PR #6, opened 2023-08-25
docs: Update README.md

laurenzlong:


README.md
```diff
@@ -23,6 +23,9 @@ const (
   OAuthClentSecret = "<client secret>"
   OAuthAccessToken = "<access token>"
   OAuthRefreshToken = "<refresh token>"
+  // If you don't know the expiry time, then set this to `time.Now().Add(-1 * time.Hour)`
+  // so the token will get refreshed right away.
+  AccessTokenExpiry = "<expiry time>"
```
eberle1080, line 28: I know this is just an example, but technically the compiler would error out if you put a time.Now() value in a const. It needs to be a var. But I won't tell if you don't :)

eberle1080 (approved):
LGTM. The syntax is a bit off, but if it clarifies the usage I'm good with it.

README.md
laurenzlong: ah good catch! updated.

PR #8, opened 2023-08-25
Expose JSONHTTPClient in the Connector interface

eberle1080:
This PR addresses a few small issues:
* Exposes JSONHTTPClient in the connector interface (avoids having to explicitly cast to get the client)
* Allow relative URLs to be passed in to JSONHTTPClient (full URLs will be auto-constructed)

### DIFF
--- a/common/json.go
+++ b/common/json.go
@@ -7,6 +7,8 @@ import (
 	"log/slog"
 	"mime"
 	"net/http"
+	"net/url"
+	"strings"
 
 	"github.com/spyzhov/ajson"
 )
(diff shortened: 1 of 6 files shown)


common/json.go
```diff
@@ -61,6 +67,18 @@ func (j *JSONHTTPClient) Get(ctx context.Context,
 	return jsonBody, nil
 }
 
+func (j *JSONHTTPClient) getURL(u string) (string, error) {
+	if strings.HasPrefix(u, "http://") || strings.HasPrefix(u, "https://") {
```
laurenzlong, line 71: aren't both of these conditions the same?

connectors.go
```diff
@@ -26,6 +26,11 @@ type Connector interface {
 	// Authentication corner cases are handled internally, but all other errors
 	// are returned to the caller.
 	Read(ctx context.Context, params ReadParams) (*ReadResult, error)
+
+	// JSONHTTPClient returns the underlying JSON HTTP client. This is useful for
+	// testing, or for calling methods that aren't exposed by the Connector
+	// interface directly. Authentication will be handled automatically.
```
laurenzlong, line 32:
```suggestion
	// interface directly. Authentication and token refreshes will be handled automatically.
```

laurenzlong (approved):

common/json.go
eberle1080: one has an `s` in it

PR #10, opened 2023-09-08
[ENG-206] feat: Handle common Salesforce errors

eberle1080:
Some Salesforce errors are structured as JSON. This PR adds the ability to introspect that JSON and return a more specific/relevant error to the caller.

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -16,6 +16,18 @@ var (
 	// ErrForbidden means the user doesn't have access to this resource.
 	ErrForbidden = errors.New("forbidden")
 
+	// ErrInvalidSessionId means the session ID is invalid.
+	ErrInvalidSessionId = errors.New("invalid session id")
+
+	// ErrUnableToLockRow means the resource couldn't be locked.
+	ErrUnableToLockRow = errors.New("unable to lock row")
+
+	// ErrInvalidGrant means the OAuth grant is invalid.
+	ErrInvalidGrant = errors.New("invalid grant")
+
+	// ErrLimitExceeded means a quota limit was exceeded.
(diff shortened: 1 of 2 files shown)


common/types.go
```diff
@@ -16,6 +16,18 @@ var (
...
 	ErrForbidden = errors.New("forbidden")
 
+	// ErrInvalidSessionId means the session ID is invalid.
+	ErrInvalidSessionId = errors.New("invalid session id")
+
+	// ErrUnableToLockRow means the resource couldn't be locked.
+	ErrUnableToLockRow = errors.New("unable to lock row")
+
+	// ErrInvalidGrant means the OAuth grant is invalid.
+	ErrInvalidGrant = errors.New("invalid grant")
```
laurenzlong, line 26:
```suggestion
	ErrInvalidGrant = errors.New("invalid OAuth grant")
```

common/types.go
```diff
@@ -16,6 +16,18 @@ var (
 	// ErrForbidden means the user doesn't have access to this resource.
 	ErrForbidden = errors.New("forbidden")
 
+	// ErrInvalidSessionId means the session ID is invalid.
+	ErrInvalidSessionId = errors.New("invalid session id")
+
+	// ErrUnableToLockRow means the resource couldn't be locked.
+	ErrUnableToLockRow = errors.New("unable to lock row")
```
laurenzlong, line 23: I wonder if this should just be `ErrRetryable` instead? Since it is easily retryable and will likely work in a few seconds after the lock is gone.

laurenzlong (approved):

PR #13, opened 2023-10-03
feat: add Salesforce write capabilities

stevenokada:
- Extends types in `common.go` to support write access pattern
- Adds support for salesforce writes + tests
- Adds dotenv package to connectors


common/json.go
```diff
@@ -100,6 +100,11 @@ func parseJSONResponse(res *http.Response, body []byte) (*ajson.Node, error) {
 	// Unmarshall the response body into JSON
 	jsonBody, err := ajson.Unmarshal(body)
 	if err != nil {
+
+		// ajson errs when trying to unmarshal empty body, even if the req was successful so we swallow error
+		if res.StatusCode == http.StatusNoContent {
```
laurenzlong, line 105: This relies on SaaS APIs providing the correct HTTP status codes, which might not always be the case, a better way to present this error is to check for the length of body before trying to unmarshal it with ajson

common/types.go
```diff
@@ -67,6 +67,19 @@ type ReadParams struct {
 	Deleted bool // optional, defaults to false
 }
 
+// WriteParams defines how we are writing data from a SaaS API.
```
laurenzlong, line 70:
```suggestion
// WriteParams defines how we are writing data to a SaaS API.
```

common/types.go
```diff
@@ -67,6 +67,19 @@ type ReadParams struct {
...
+type WriteParams struct {
+	// The name of the object we are writing, e.g. "Account"
+	ObjectName string // required
+
+	// The external id of the object instance we are updating
```
laurenzlong, line 75:
```suggestion
	// The external ID of the object instance we are updating
```

common/types.go
```diff
@@ -67,6 +67,19 @@ type ReadParams struct {
...
+	ObjectId string // optional
+
+	// ObjectData is a JSON node representing the record of data we want to insert in the case of CREATE
+	// or fields of data we want to modify in case of an update
+	ObjectData map[string]interface{} `json:"data"`
```
laurenzlong, line 80: Is there a reason why this field has a json annotation?

common/types.go
```diff
@@ -85,6 +98,18 @@ type ReadResult struct {
 	Done bool `json:"done,omitempty"`
 }
 
+// WriteResult is what's returned from writing data via the Write call.
+type WriteResult struct {
+	// ObjectId is the id of the written record.
```
laurenzlong, line 103:
```suggestion
	// ObjectId is the ID of the written record.
```

common/types.go
```diff
@@ -85,6 +98,18 @@ type ReadResult struct {
...
+	ObjectId string
+	// Errors is list of error objects returned by the API.
+	Errors []interface{}
+	// Success is true if write succeeded.
+	Success bool
+	// RecordData is a JSON node containing data about the properties that were updated.
+	RecordData map[string]interface{} `json:"data"` //optional
```
laurenzlong, line 110: Same here, why is `json:"data"` a necessary annotation?

common/types.go
```diff
@@ -67,6 +67,19 @@ type ReadParams struct {
 	Deleted bool // optional, defaults to false
 }
 
+// WriteParams defines how we are writing data from a SaaS API.
+type WriteParams struct {
+	// The name of the object we are writing, e.g. "Account"
+	ObjectName string // required
+
+	// The external id of the object instance we are updating
+	ObjectId string // optional
```
laurenzlong, line 76: Can you make it more clear that if ObjectId is present then it's an update, otherwise is a create?

test/salesforce.go
```diff
@@ -31,7 +33,12 @@ func main() {
 }
 
 func mainFn() int {
-	subdomain := flag.String("subdomain", "ampersand-dev-ed.develop", "Salesforce subdomain")
+	if err := godotenv.Load(); err != nil {
+		slog.Error("Error loading .env file", "error", err)
+		return 1
+	}
+
+	subdomain := flag.String("subdomain", "boxit2-dev-ed", "Salesforce subdomain")
```
laurenzlong, line 41: Can you make the subdomain another env variable?

salesforce/write.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	success, err := getSuccess(data)
+	if err != nil {
+		return nil, err
+	}
+	return &common.WriteResult{
+		ObjectId: createdObjectId,
+		Errors:   errors,
+		Success:  success,
```
laurenzlong, line 68: I don't actually see where you're adding `RecordData` to the WriteResult

laurenzlong (changes requested):

salesforce/write.go
stevenokada: That's because the Salesforce API doesn't return any written record, just the object id.

https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_sobject_create.htm

stevenokada: The RecordData is there for future connectors to populate

common/json.go
stevenokada: that's a great point. i'll lift the logic out of the json parsing helper

common/types.go
stevenokada: if we marshal the `WriteResult` to json to return in the future (following the same pattern as ReadResult), we'd want this tag. I think we could/should actually add the json tag to the other fields.

common/types.go
stevenokada: If we don't intend on unmarshalling a json response directly to a WriteParams, we can remove the tag here. I'll remove for now

common/json.go
laurenzlong: I see you lifted the logic out, but I think you forgot to delete this?

laurenzlong (approved):
Just one more comment, then we're good to go!

stevenokada:
should've addressed all comments!

PR #17, opened 2023-10-30
[ENG-408] Add hubspot connector with read functionality

RajatPawar:


RajatPawar:
* main
  * **PR #17** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/17" target="_blank">Graphite</a>  👈
    * **PR #18** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/18" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/amp-labs/connectors/17?utm_source=stack-comment).

hubspot/search.go
```diff
@@ -0,0 +1,61 @@
...
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return parseResult(data, getNextRecordsAfter)
+}
+
+// makeFilterBody is specifically implemented for the Since filter currently.
+func makeFilterBody(config common.ReadParams) map[string]interface{} {
```
eberle1080, line 41: I've been trying to use `any` rather than `interface{}` since the former is clearer to read. I haven't been 100% successful (old habits die hard) but still that's the intent. We should make a linter rule for this.

hubspot/read.go
```diff
@@ -0,0 +1,134 @@
...
+
+	if config.Deleted {
+		queryValues.Add("archived", "true")
+	}
+
+	return queryValues.Encode()
+}
+
+// parseResult parses the response from the Hubspot API. A 2xx return type is assumed.
+func parseResult(data *ajson.Node, paginationFunc func(*ajson.Node) (string, error)) (*common.ReadResult, error) {
```
eberle1080, line 60: I really like `pagniationFunc`, great idea

hubspot/search.go
```diff
@@ -0,0 +1,61 @@
...
+func (c *Connector) search(ctx context.Context, config common.ReadParams) (*common.ReadResult, error) {
+	var (
+		data *ajson.Node
+		err  error
+	)
+
+	// If filtering is not required, then we have to use the read endpoint.
+	if !requiresFiltering(config) {
+		return c.Read(ctx, config)
+	}
```
eberle1080, line 24: This seems unnecessary since you already do the same check in `Read`

hubspot/post.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// post writes data to Hubspot. It handles retries and access token refreshes.
+func (c *Connector) post(ctx context.Context, url string, body any) (*ajson.Node, error) {
+	node, err := c.Client.Post(ctx, url, body)
+	if err != nil {
+		switch {
+		case errors.Is(err, common.ErrAccessToken):
+			// Retryable, so just log and retry
+			slog.Warn("Access token invalid, retrying", "error", err)
```
eberle1080, line 19: It's not retrying though. It should but it's not. In the Salesforce connector I was (at one point) doing this, but it turned out that the retry library I was using had an incompatible license, so I ripped it out. Bits like this are left over from that, but you probably don't want to carry that over to the hubspot connector.

Also there's an entire conversation to be had about where retry logic should happen (in the connector or in the backend).

eberle1080 (approved):
A few small suggestions but mostly looks great

hubspot/read.go
RajatPawar: 🙌

hubspot/search.go
RajatPawar: This is to avoid any accidental calls directly to `search`. I agree it's unnecessary but should be a good guard clause to have in case.

hubspot/post.go
RajatPawar: Sure - keeping it as a todo item

hubspot/search.go
RajatPawar: Changed this, thanks

PR #20, opened 2023-11-01
[ENG-420] Add ListObjectMetadata to Salesforce & Hubspot connectors

RajatPawar:
* Adds ListObjectMetadata to describe an object's properties
* Adds ObjectMetadata to hold an object's metadata
* Adds the ListObjectMetadata function in the Connector interface

### Hubspot
Screenshot 2023-11-01 at 4 18 50 PM

### Salesforce
Screenshot 2023-11-01 at 4 23 42 PM

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -54,6 +54,9 @@ var (
 
 	// ErrNotImplemented is returned when a method is not implemented.
 	ErrNotImplemented = errors.New("not implemented")
+
+	// ErrMissingObjects is returned when no objects are provided in the request.
+	ErrMissingObjects = errors.New("no objects provided")
 )
 
 // ReadParams defines how we are reading data from a SaaS API.
--- a/common/types.go
+++ b/common/types.go
@@ -149,10 +149,11 @@ func (r HTTPStatusError) Unwrap() error {
 	return r.err
 }
 
+type ListObjectMetadataResult map[string]ObjectMetadata
+
 type ObjectMetadata struct {
 	DisplayName string
+
 	// FieldsMap is a map of field names to field display names
 	FieldsMap map[string]string
 }
-
-type ListObjectMetadataResponse map[string]ObjectMetadata
--- a/common/types.go
+++ b/common/types.go
@@ -145,3 +145,11 @@ func (r HTTPStatusError) Error() string {
 func (r HTTPStatusError) Unwrap() error {
 	return r.err
 }
+
+type ObjectMetadata struct {
+	DisplayName string
+	// FieldsMap is a map of field names to field display names
+	FieldsMap map[string]string
+}
+
+type ListObjectMetadataResponse map[string]ObjectMetadata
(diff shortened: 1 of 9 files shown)


common/types.go
```diff
@@ -145,3 +148,12 @@ func (r HTTPStatusError) Error() string {
 func (r HTTPStatusError) Unwrap() error {
 	return r.err
 }
+
+type ListObjectMetadataResult map[string]ObjectMetadata
+
+type ObjectMetadata struct {
+	DisplayName string
```
laurenzlong, line 155: Can you add a comment for DisplayName? Since the types in this file are publicly exposed, we should generally try to document all of them.

hubspot/metadata.go
```diff
@@ -0,0 +1,125 @@
...
+// describeObject returns object metadata for the given object name.
+func (c *Connector) describeObject(ctx context.Context, objectName string) (*common.ObjectMetadata, error) {
+	data, err := c.get(ctx, c.BaseURL+"/properties/"+objectName)
+	if err != nil {
+		return nil, fmt.Errorf("error fetching HubSpot fields: %w", err)
+	}
+
+	rawResponse, err := ajson.Marshal(data)
+	if err != nil {
+		return nil, fmt.Errorf("error marshalling composite response into byte array: %w", err)
```
laurenzlong, line 97:
```suggestion
		return nil, fmt.Errorf("error marshalling object metadata response into byte array: %w", err)
```

hubspot/metadata.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	resp := &describeObjectResponse{}
+
+	err = json.Unmarshal(rawResponse, resp)
+	if err != nil {
+		return nil, fmt.Errorf("error unmarshalling composite response into JSON: %w", err)
```
laurenzlong, line 104:
```suggestion
		return nil, fmt.Errorf("error unmarshalling object metadata response into JSON: %w", err)
```

laurenzlong (approved):
Feel free to do this in a follow up PR, as we move this logic into the connectors library, rather than failing the entire method if one of the object's metadata fails to be fetched, we should return a list of results and a list of errors so that the library user can decide what to do. For Ampersand's own use case, in the future, once we fully support optional objects, we wouldn't want to fail the entire request anymore. If we have trouble fetching certain objects (e.g. a consumer's Salesforce only has the license for Account and Contact, but not Opportunity and Lead), they should still be allowed to install the integration if Opportunity and Lead are optional objects.

RajatPawar:
> Feel free to do this in a follow up PR, as we move this logic into the connectors library, rather than failing the entire method if one of the object's metadata fails to be fetched, we should return a list of results and a list of errors so that the library user can decide what to do. For Ampersand's own use case, in the future, once we fully support optional objects, we wouldn't want to fail the entire request anymore. If we have trouble fetching certain objects (e.g. a consumer's Salesforce only has the license for Account and Contact, but not Opportunity and Lead), they should still be allowed to install the integration if Opportunity and Lead are optional objects in the integration.

Agreed, I ran into some issues while attempting to do this. I'll follow up with another PR for this.

PR #21, opened 2023-11-01
feat: Add a callback to save OAuth tokens

eberle1080:
Right now whenever a token is refreshed, we just discard the result. This leads to a lot of waste since we're effectively refreshing each time a workflow runs.

This PR makes it possible for us to be notified whenever a token is refreshed, so that we can persist it back to the database. This should cause fewer token refreshes over time.

### DIFF
--- a/common/oauth.go
+++ b/common/oauth.go
@@ -3,6 +3,7 @@ package common
 import (
 	"context"
 	"net/http"
+	"sync"
 
 	"golang.org/x/oauth2"
 )
@@ -40,6 +41,7 @@ type oauthClientParams struct {
 	token       *oauth2.Token
 	config      *oauth2.Config
 	tokenSource oauth2.TokenSource
+	updateToken func(tok *oauth2.Token) error
 }
 
 type OAuthOption func(params *oauthClientParams)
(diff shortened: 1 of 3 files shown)


common/oauth.go
```diff
@@ -40,6 +41,7 @@ type oauthClientParams struct {
 	token       *oauth2.Token
 	config      *oauth2.Config
 	tokenSource oauth2.TokenSource
+	updateToken func(tok *oauth2.Token) error
```
laurenzlong, line 44: Would `onTokenUpdate` be a better name for this? Since this function itself is not updating the token, it is the callback function for when the token is updated.

common/oauth.go
```diff
@@ -99,11 +110,62 @@ func newOAuthClient(ctx context.Context, params *oauthClientParams) Authenticate
...
-		return oauth2.NewClient(ctx, params.tokenSource)
-	} else {
-		return oauth2.NewClient(ctx, params.config.TokenSource(ctx, params.token))
+		return params.tokenSource
+	}
+
+	return params.config.TokenSource(ctx, params.token)
+}
+
+type writeBackTokenSource struct {
```
laurenzlong, line 135: Let's make this struct name more generic? Since persisting token update is just one use case, there might be other use cases of a callback function.

common/oauth.go
```diff
@@ -99,11 +110,62 @@ func newOAuthClient(ctx context.Context, params *oauthClientParams) Authenticate
...
+	mut         sync.Mutex
+	update      func(tok *oauth2.Token) error
```
laurenzlong, line 137: Similar idea as above, `onUpdate` is a better name for this function.

laurenzlong (approved):

common/oauth.go
eberle1080: `tokenUpdated`? The `on` prefix is something I'd usually call a parameter exposed to the caller, but this doesn't get exposed to the caller.

common/oauth.go
eberle1080: How about `observableTokenSource`?

common/oauth.go
eberle1080: `tokenUpdated`?

PR #24, opened 2023-11-07
[ENG-436] Modify ListObjectMetadata to return metadata & a list of errors

RajatPawar:
* Instead of failing completely, this returns successful results & a list of errors indexed against the object name

Once the PR is merged, I will create a PR in the server repo to update the connector dependency & work off this new structure

### DIFF
--- a/hubspot/metadata.go
+++ b/hubspot/metadata.go
@@ -10,12 +10,12 @@ import (
 	"github.com/spyzhov/ajson"
 )
 
-type ObjectMetadataResult struct {
+type objectMetadataResult struct {
 	ObjectName string
 	Response   common.ObjectMetadata
 }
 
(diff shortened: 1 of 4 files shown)


hubspot/metadata.go
```diff
@@ -10,19 +10,29 @@ import (
 	"github.com/spyzhov/ajson"
 )
 
+type ObjectMetadataResult struct {
```
laurenzlong, line 13: can you lower case this and `ObjectMetaDataError` since they are not used outside of this package?

laurenzlong (approved):

hubspot/metadata.go
RajatPawar: Fixed

PR #28, opened 2023-11-16
[ENG-442] Bulk write from csv

jlimatampersand:
Test results

UPDATED
```
jaehyunlim@Jaehyuns-MacBook-Pro bulkwrite % go run main.go
Upload complete.
{
    "state": "UploadComplete",
    "jobId": "750Do00000D12bIIAR"
}
Write Result
{
    "id": "750Do00000D12bIIAR",
    "object": "Touchpoint__c",
    "createdById": "005Do000001y7jOIAQ",
    "externalIdFieldName": "external_id__c",
    "state": "JobComplete",
    "operation": "upsert",
    "columnDelimiter": "COMMA",
    "lineEnding": "LF",
    "numberRecordsFailed": 0,
    "numberRecordsProcessed": 20,
    "apexProcessingTime": 0,
    "apiActiveProcessingTime": 57,
    "apiVersion": 58,
    "concurrencyMode": "Parallel",
    "contentType": "CSV",
    "createdDate": "2023-11-17T19:50:42.000+0000",
    "jobType": "V2Ingest",
    "retries": 0,
    "systemModstamp": "2023-11-17T19:50:43.000+0000",
    "totalProcessingTime": 118
}
```


common/http.go
```diff
@@ -37,3 +39,35 @@ func InterpretError(res *http.Response, body []byte) error {
...
+// All non-2xx responses will be passed to the error handler. If the error handler
+// returns nil, then the error is ignored and the caller is responsible for handling
+// the error. If the error handler returns an error, then that error is returned
+// to the caller, as-is. Both the response and the response body are passed
+// to the error handler as arguments.
+type ErrorHandler func(rsp *http.Response, body []byte) error
+
+// HTTPClient is an HTTP client which makes certain assumptions, such as
+// that the response body is JSON. It also handles OAuth access token refreshes.
+type HTTPClient struct {
```
jlimatampersand, line 65: Since now the HTTPClient is not just JSONHTTPClient, renamed and moved to this file from json.go

salesforce/bulkwrite.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	jobId, ok := jobCreateRes["id"] //nolint:varnamelen
+	if !ok {
+		return nil, fmt.Errorf("%w for key %s in %v", ErrKeyNotFound, "id", jobCreateRes)
+	}
+
+	jobIdString, ok := jobId.(string) //nolint:varnamelen
+	if !ok {
+		return nil, fmt.Errorf("%w. expected id to be string, gott %T", ErrInvalidType, jobId)
```
laurenzlong, line 57:
```suggestion
		return nil, fmt.Errorf("%w. expected id to be string, got %T", ErrInvalidType, jobId)
```

salesforce/bulkwrite.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	// upload csv and there is no response body other than status code
+	_, err = c.uploadCSV(ctx, jobIdString, config)
+	if err != nil {
+		return nil, fmt.Errorf("uploadCSV failed: %w", err)
+	}
+
+	//
```
laurenzlong, line 66: leftover `//` from a comment that no longer exists?

salesforce/bulkwrite.go
```diff
@@ -0,0 +1,227 @@
...
+func (c *Connector) GetJobInfo(ctx context.Context, jobId string) (*ajson.Node, error) {
+	location, err := joinURLPath(c.BaseURL, "jobs/ingest", jobId)
+	if err != nil {
+		return nil, err
+	}
+
+	return c.get(ctx, location)
+}
+
+func (c *Connector) FailedResults(ctx context.Context, jobId string) ([]byte, error) {
```
laurenzlong, line 190:
```suggestion
func (c *Connector) GetFailedResults(ctx context.Context, jobId string) ([]byte, error) {
```

salesforce/bulkwrite.go
```diff
@@ -0,0 +1,227 @@
...
+	"os"
+	"strings"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/spyzhov/ajson"
+)
+
+var (
+	ErrKeyNotFound     = errors.New("key not found")
+	ErrUnknownNodeType = errors.New("unknown node type")
```
laurenzlong, line 17:
```suggestion
	ErrUnknownNodeType = errors.New("unknown node type when parsing JSON")
```

salesforce/bulkwrite.go
```diff
@@ -0,0 +1,227 @@
...
+func (c *Connector) FailedResults(ctx context.Context, jobId string) ([]byte, error) {
+	location, err := joinURLPath(c.BaseURL, fmt.Sprintf("jobs/ingest/%s/failedResults", jobId))
+	if err != nil {
+		return nil, err
+	}
+
+	return c.getCSV(ctx, location)
+}
+
+func ParseNodeToMap(node *ajson.Node) (map[string]interface{}, error) {
```
laurenzlong, line 199: ParseAjsonNodeToMap would be more clear (since this is an exported function, we should make its usage very clear)

salesforce/bulkwrite.go
```diff
@@ -0,0 +1,227 @@
...
+
+	res, err := c.patch(ctx, location, updateLoadCompleteBody)
+	if err != nil {
+		return nil, fmt.Errorf("patch failed: %w", errors.Join(err, common.ErrRequestFailed))
+	}
+
+	return ParseNodeToMap(res)
+}
+
+func (c *Connector) GetJobResult(ctx context.Context, jobId string) ([]byte, error) {
```
laurenzlong, line 154: For this function and the other functions below this, I generally wouldn't want our library to export any functions that have `[] byte` or `*ajson.Node` as results since they are not the most user friendly and don't offer any type safety. Eventually these exported methods should all have well-typed results (e.g. a `GetJobResultResponse` struct), since later you'll need to parse the byte array or ajson.Node into structs anyhow. But it's probably hard to test these out right now, so we can go with these for now and then add the types later.

laurenzlong (changes requested):

salesforce/errors.go
```diff
@@ -70,3 +71,24 @@ func (c *Connector) interpretJSONError(res *http.Response, body []byte) error {
 	// No known errors, just do the normal error handling logic
 	return common.InterpretError(res, body)
 }
+
+func (c *Connector) HandleError(err error) error {
+	switch {
+	case errors.Is(err, common.ErrAccessToken):
+		// Retryable, so just log and retry
+		slog.Warn("Access token invalid, retrying", "error", err)
```
laurenzlong, line 79: Can we remove this warning line? since we're not actually retrying yet (it's still a TODO)

salesforce/bulkwrite.go
```diff
@@ -0,0 +1,299 @@
...
+			parsed[key] = data.MustObject()
+		default:
+			return nil, fmt.Errorf("%w: %d", ErrUnknownNodeType, data.Type())
+		}
+	}
+
+	return parsed, nil
+}
+
+type GetJobInfoResult struct {
```
laurenzlong, line 203: can you move this to the top of the file so all the types are up top?

salesforce/bulkwrite.go
```diff
@@ -0,0 +1,299 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	node, err := c.get(ctx, location)
+	if err != nil {
+		return nil, fmt.Errorf("getGetInfo failed: %w", errors.Join(err, common.ErrRequestFailed))
+	}
+
+	dataMap, err := ParseAjsonNodeToMap(node)
```
laurenzlong, line 238: Since you've already defined a struct for GetJobInfoResult, this will be a lot less lines and won't require type assertion if you do the following:
- call https://pkg.go.dev/github.com/spyzhov/ajson#Marshal to turn node into a byte array
- call [json.Unmarshall](https://pkg.go.dev/encoding/json#Unmarshal) to turn the byte array into GetInfoJobResult

salesforce/bulkwrite.go
```diff
@@ -0,0 +1,299 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("createJob failed: %w", err)
+	}
+
+	jobCreateRes, err := ParseAjsonNodeToMap(res)
+	if err != nil {
+		return nil, fmt.Errorf("parsing result of createJob failed: %w", errors.Join(err, common.ErrParseError))
+	}
+
+	state, ok := jobCreateRes["state"].(string) //nolint:varnamelen
```
laurenzlong, line 57: ajson already offers pretty good helper functions.

For example there's the [GetObject](https://pkg.go.dev/github.com/spyzhov/ajson#Node.GetObject) function which will turn res into a map of nodes (which replaces the need for the ParseAjsonNodeToMap function), and [GetString](https://pkg.go.dev/github.com/spyzhov/ajson#Node.GetString) function for getting the string value for each key.
```
nodesMap, err := res.GetObject("state")
if err != nil { ... }
state, err := s["state"].GetString()
if err != nil { ...}
```

This is safe to do and will not panic even if the keys don't exist, I tried this out ([see playground](https://play.golang.com/p/DZQAYprdsix))

salesforce/bulkwrite.go
```diff
@@ -0,0 +1,299 @@
...
+
+	res, err := c.patch(ctx, location, updateLoadCompleteBody)
+	if err != nil {
+		return nil, fmt.Errorf("patch failed: %w", errors.Join(err, common.ErrRequestFailed))
+	}
+
+	return ParseAjsonNodeToMap(res)
+}
+
+func ParseAjsonNodeToMap(node *ajson.Node) (map[string]interface{}, error) {
```
laurenzlong, line 173: if you make the other changes I suggest, this function will no longer be necessary.

laurenzlong (approved):

salesforce/bulkwrite.go
```diff
@@ -0,0 +1,299 @@
...
+// BulkWriteParams defines how we are writing data to a SaaS API.
+type BulkWriteParams struct {
+	// The name of the object we are writing, e.g. "Account"
+	ObjectName string // required
+
+	// The external ID of the object instance we are updating. Provided in the case of UPDATE, but not CREATE.
+	ExternalId string // required
+
+	// The path to the CSV file we are writing
+	FilePath string // required
```
laurenzlong, line 24: Can we add one more param for the "mode" and have it be "upsert" or "insert"? Then `createJob` can create a different kind of job depending on if it's upsert or insert?

salesforce/bulkwrite.go
```diff
@@ -0,0 +1,299 @@
...
+	"github.com/amp-labs/connectors/common"
+	"github.com/spyzhov/ajson"
+)
+
+// BulkWriteParams defines how we are writing data to a SaaS API.
+type BulkWriteParams struct {
+	// The name of the object we are writing, e.g. "Account"
+	ObjectName string // required
+
+	// The external ID of the object instance we are updating. Provided in the case of UPDATE, but not CREATE.
```
laurenzlong, line 20: Can we call this `ExternalIdField` or something? since this is a field name, not an actual ID.

laurenzlong, line 20:
```suggestion
	// The name of a field on the object which is an External ID. Provided in the case of upserts, not inserts.
```

salesforce/bulkwrite.go
laurenzlong, line 24: As we discussed, this should probably just be an in-memory CSV string or byte array rather than a filePath, so the Temporal activity does not have to write the file to disk before calling this method.

eberle1080 (changes requested):

common/csv.go
```diff
@@ -0,0 +1,69 @@
...
+		return nil, err
+	}
+
+	_, body, err := j.httpPutCSV(ctx, fullURL, headers, reqBody) // nolint:bodyclose
+	if err != nil {
+		return nil, err
+	}
+
+	return body, nil
+}
```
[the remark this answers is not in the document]
laurenzlong: Sorry, it was my suggestion to keep this as JSONHTTPClient for now since the HTTP response from the bulk write API is still in JSON format, and it's helpful to have the JSON parsing functionality still. Using the HTTP client directly means that the JSON response parsing functionality would have to be re-implemented. I do agree this is a bit confusing still, and we should come up with a better strategy here, whether that's renaming this client or creating a different client. Since you're in flight today and this PR has already been outstanding for a while and is blocking the server changes, let's merge it in, and do the refactoring in a follow up.

common/csv.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"net/http"
+)
+
+/*
+  // TODO: Below is patch work to have CSV file type to be supported
+  // Currently done on JSONHTTPClient, but might need be reimplmented on CSVHTTPClient
```
laurenzlong, line 12:
```suggestion
  // Currently done on JSONHTTPClient, but we might need a separate CSVHTTPClient instead
```

common/csv.go
```diff
@@ -0,0 +1,75 @@
...
+func addHeaders(req *http.Request, headers []Header) (*http.Request, error) {
+	// Apply any custom headers
+	for _, hdr := range headers {
+		req.Header.Add(hdr.Key, hdr.Value)
+	}
+
+	return req, nil
+}
+
+// func (j *JSONHTTPClient) GetCSV(ctx context.Context, url string, headers ...Header) ([]byte, error) {
```
laurenzlong, line 62: For later: this method should definitely not be on the JSONHTTPClient, since there's neither the request type nor the response type is JSON.

PR #32, opened 2023-12-07
chore: Sort hubspot reads by default

RajatPawar:
The hubspot `Read` method redirects a request to `Search` if `ReadParams.Since` is set, but unknown to the caller, they will get cut off at 10,000 records. This sorts the result by the id that the request can be continued via a search if required.

### Testing 
Screenshot 2023-12-07 at 10 18 50 AM

### DIFF
--- a/hubspot/read.go
+++ b/hubspot/read.go
@@ -26,7 +26,7 @@ func (c *Connector) Read(ctx context.Context, config common.ReadParams) (*common
 		searchParams := SearchParams{
 			ObjectName:   config.ObjectName,
 			FilterGroups: buildLastModifiedFilterGroup(config.Since),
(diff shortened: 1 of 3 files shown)


hubspot/read.go
```diff
@@ -26,6 +26,7 @@ func (c *Connector) Read(ctx context.Context, config common.ReadParams) (*common
 		searchParams := SearchParams{
 			ObjectName:   config.ObjectName,
 			FilterGroups: buildLastModifiedFilterGroup(config.Since),
+			SortBy:       buildSort(ObjectFieldHsObjectId, SortDirectionAsc),
```
laurenzlong, line 29: Can you add a quick comment to explain this?

PR #33, opened 2023-12-12
[ENG-470] feat: create custom object on Salesforce

jlimatampersand:
This PR is trying to createMeta to Salesforce using Metadata API. Test runs are done to create custom Object and custom Field in salesforce. 

- Unit tests are written to test xml conversions. 

Note TODO
- If we decide to fully support this feature, we might consider parsing XML responses to JSON. 

Blocker
- Didn't have chance to test run after token expiration. Should wait for a day or two to test it out. 
 

Test Run
Screenshot 2023-12-13 at 9 39 00 AM
```
jaehyunlim@Jaehyuns-MBP test % go run metadata/main.go
{
    "xmlName": "metadata",
    "attributes": [
        {
            "key": "xsi:type",
            "value": "CustomObject"
        }
    ],
    "children": [
        {
            "xmlName": "fullName",
            "attributes": null,
            "children": [
                "TestObject15__c"
            ],
            "selfClosing": false
        },
        {
            "xmlName": "label",
            "attributes": null,
            "children": [
                "Test Object 15"
            ],
            "selfClosing": false
        },
        {
            "xmlName": "pluralLabel",
            "attributes": null,
            "children": [
                "Test Objects 15"
            ],
            "selfClosing": false
        },
        {
            "xmlName": "nameField",
            "attributes": null,
            "children": [
                {
                    "xmlName": "type",
                    "attributes": null,
                    "children": [
                        "Text"
                    ],
                    "selfClosing": false
                },
                {
                    "xmlName": "label",
                    "attributes": null,
                    "children": [
                        "Test Object Name"
                    ],
                    "selfClosing": false
                }
            ],
            "selfClosing": false
        },
        {
            "xmlName": "deploymentStatus",
            "attributes": null,
            "children": [
                "Deployed"
            ],
            "selfClosing": false
        },
        {
            "xmlName": "sharingModel",
            "attributes": null,
            "children": [
                "ReadWrite"
            ],
            "selfClosing": false
        }
    ],
    "selfClosing": false
}
Field Operation Result 
<?xml version="1.0" encoding="UTF-8"?>
<soapenv:Envelope
	xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/"
	xmlns="http://soap.sforce.com/2006/04/metadata">
	<soapenv:Body>
		<createMetadataResponse>
			<result>
				<fullName>TestObject15__c</fullName>
				<success>true</success>
			</result>
			<result>
				<fullName>TestObject15__c.Comments__c</fullName>
				<success>true</success>
			</result>
		</createMetadataResponse>
	</soapenv:Body>
</soapenv:Envelope>
```


salesforce/metadataApi.go
```diff
@@ -0,0 +1,268 @@
...
+const (
+	versionPrefix             = "v"
+	openParenthesis           = "<"
+	closeParenthesis          = ">"
+	closeWithSlashParenthesis = "/>"
+)
+
+var ErrCreateMetadata = fmt.Errorf("error in CreateMetadata")
+
+type XMLSchema interface {
```
RajatPawar, line 27: We should extract the XML helpers to another file or lib & only have the main connector method logic in this file.

salesforce/metadataApi.go
```diff
@@ -0,0 +1,268 @@
...
+
+	chilren := []string{}
+	for _, child := range x.Children {
+		chilren = append(chilren, child.ToXML())
+	}
+
+	return fmt.Sprintf("%s%s%s", start, strings.Join(chilren, ""), end)
+}
+
+func (x *XMLData) startTag() string {
```
RajatPawar, line 120: Can we use [encoding/xml](https://gobyexample.com/xml) to implement this? It should provide easy & tested translation between json & xml.

RajatPawar (changes requested):

salesforce/metadataApi.go
jlimatampersand: I have done researches on that package, but the pacakge has many limitations that I found we could not use it for our use cases. With that package  I ran into problem was that we have to define struct for every object exists in Saleforce, plus there is no clear way to create custom object since we can't define struct for every custom object. 

My approach is to make the xml data structure to be more of wrapper container that can be used for any sort of object regardless of type. The way we would get definition for custom objects will be in yaml file, which need to be translated to some json. 

This data structure I created can be easily transformed from json, or even from yaml we can make it to transform easier since the underliying structure has json representation.

RajatPawar (approved):
The XML code is not very easy to comprehend at a quick glance. It would be great if you could add comments for anyone who has to extend/modify it.

jlimatampersand:
* `main`
  * **PR #33** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/33?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/amp-labs/connectors/33?utm_source=stack-comment).

common/oauth.go
```diff
@@ -164,8 +165,8 @@ func (w *observableTokenSource) HasChanged(tok *oauth2.Token) bool {
 		return true
 	}
 
-	return w.lastKnown.AccessToken == tok.AccessToken &&
+	return !(w.lastKnown.AccessToken == tok.AccessToken &&
```
jlimatampersand, line 168: fixing a bug, token was not updating because of this.

laurenzlong: Nice catch!

As an optimization, the following is equivalent and more readable (doesn't require a large expression to be put into parenthesis with a `!` in front of it.)

```
          return w.lastKnown.AccessToken != tok.AccessToken ||
		w.lastKnown.RefreshToken != tok.RefreshToken ||
		w.lastKnown.TokenType != tok.TokenType ||
		!w.lastKnown.Expiry.Equal(tok.Expiry)
```

common/xml.go
```diff
@@ -0,0 +1,133 @@
+package common
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+const (
```
laurenzlong, line 9: nit: these constant names can be a bit shorter:
- `openParen`
- `closeParen`
- `closeParenWithSlash`

salesforce/metadataApi.go
```diff
@@ -0,0 +1,184 @@
...
+		return "", err
+	}
+
+	res, body, err := c.makeRequest(ctx, req)
+	// below is a workaround to refresh token if it is expired
+	// normally oauth2 library should handle this
+	// but SOAP API does not take token in header
+	// but takes it in body
+	// So in case of 500 error and INVALID_SESSION_ID in body
+	// we know it is session expired, and automatically the client refreshes token
```
laurenzlong, line 40:
```suggestion
	// we know it is session expired, and automatically refresh the token.
```

salesforce/metadataApi.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	if res.StatusCode < 200 || res.StatusCode > 299 {
+		return "", fmt.Errorf("%w: %s", ErrCreateMetadata, string(body))
+	}
+
+	return string(body), nil
+}
+
+func (c *Connector) prepareRequest(ctx context.Context, operation *common.XMLData, tok *oauth2.Token) (*http.Request, error) {
```
laurenzlong, line 62: can you name this prepareXmlRequest or something like that? (Since it's a method on the connector, we should be more explicit in its purpose)

salesforce/metadataApi.go
```diff
@@ -0,0 +1,184 @@
...
+	data := preparePayload(operation, tok.AccessToken)
+
+	endPointURL, err := url.JoinPath(c.Client.Base, "services/Soap/m/"+removePrefix(c.APIVersion(), versionPrefix))
```
laurenzlong, line 65: can we actually have a different constant called `APIVersionSOAP` or something like that, rather than having to remove the prefix like this, and needing to expose an `APIVersion` method. It also gives us the flexibility in the future to use a different version of the SOAP API vs the REST API.

salesforce/metadataApi.go
```diff
@@ -0,0 +1,184 @@
...
+
+func removePrefix(s string, prefix string) string {
+	if !strings.HasPrefix(s, prefix) {
+		return s
+	}
+
+	return s[len(prefix):]
+}
+
+func addSOAPAPIHeaders(req *http.Request) {
```
laurenzlong, line 113: Can we name this `addSOAPHeaders`? A bit shorter, and doesn't require so many capital letters in a row, which is a bit hard to read.

salesforce/metadataApi.go
```diff
@@ -0,0 +1,184 @@
...
+	req.Header.Add("Content-Type", "text/xml")
+	req.Header.Set("SOAPAction", "''")
```
laurenzlong, line 115: I remember you said this was a gotcha, can you add a comment explaining why this is needed? otherwise others might be confused by it in the future since it looks a bit odd.

salesforce/test/metadataApi_test.go
```diff
@@ -0,0 +1,111 @@
...
+	xmlData = &common.XMLData{
+		XMLName:     "test",
+		Attributes:  []*common.XMLAttributes{{Key: "test", Value: "test"}},
+		Children:    []common.XMLSchema{common.XMLString("test")},
+		SelfClosing: true,
+	}
+
+	xmlStr = xmlData.ToXML()
+	if xmlStr != `<test test="test"/>` {
+		testing.Errorf("XMLData.ToXML() = %s; want <test test=\"test\"/>", xmlStr)
```
laurenzlong, line 36: can you use `"github.com/stretchr/testify/assert"` instead of manually doing string comparison and constructing error messages? There's a `assert.Equal` method, which we use in some tests in the server.

test/metadata/main.go
```diff
@@ -0,0 +1,306 @@
...
+	// fmt.Println(string(createObjectData))
+
+	var objectOperation *common.XMLData
+
+	if err := json.Unmarshal(createObjectData, &objectOperation); err != nil {
+		slog.Error("Error marshalling testOperation.json", "error", err)
+		os.Exit(1)
+	}
+
+	// fmt.Println(operation.ToXML())
```
laurenzlong, line 95: I think you forgot to remove these.

test/metadata/main.go
```diff
@@ -0,0 +1,306 @@
...
+		slog.Error("Error opening testOperation.json", "error", err)
+		os.Exit(1)
+	}
+
+	if string(createFieldData) == "" {
+		slog.Error("Error opening testOperation.json", "error", err)
+		os.Exit(1)
+	}
+
+	// fmt.Println(string(createFieldData))
```
laurenzlong, line 115: same here.

test/metadata/main.go
```diff
@@ -0,0 +1,306 @@
...
+		SelfClosing: false,
+	}
+
+	operation := &common.XMLData{
+		XMLName:     "createMetadata",
+		Children:    []common.XMLSchema{metadata},
+		SelfClosing: false,
+	}
+
+	// fmt.Println(operation.ToXML())
```
laurenzlong, line 199: same here

test/metadata/main.go
```diff
@@ -0,0 +1,306 @@
...
+
+func getCreateFieldOperation() *common.XMLData {
+	metadata := &common.XMLData{
+		XMLName:    "metadata",
+		Attributes: []*common.XMLAttributes{{Key: "xsi:type", Value: "CustomField"}},
+		Children: []common.XMLSchema{
+			&common.XMLData{
+				XMLName: "fullName",
+				Children: []common.XMLSchema{
+					common.XMLString("TestObject13__c.Comments__c"),
```
laurenzlong, line 219: I think having the `XMLString` type makes constructing XMLData more tedious than necessary. Is there a strong benefit you see in using it vs just a string type here?

salesforce/metadataApi.go
```diff
@@ -0,0 +1,184 @@
...
+const (
+	versionPrefix = "v"
+)
+
+var (
+	ErrCreateMetadata  = fmt.Errorf("error in CreateMetadata")
+	ErrCreatingRequest = fmt.Errorf("error in creating request")
+)
+
+func (c *Connector) CreateMetadata(ctx context.Context, operation *common.XMLData, tok *oauth2.Token) (string, error) {
```
laurenzlong, line 27: Can we name the second parameter something other than operation? Since Operation means something else in our backend and API, this might get confusing.

laurenzlong (approved):
Wow this is a lot of hard work, thank you! This is a great start.

Also, not sure if this is helpful, but it looks like there's a REST API for getting the status of the deployment rather than having to go to through the. SOAP API again. https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_rest_deploy_checkstatus.htm

test/metadata/main.go
```diff
@@ -0,0 +1,306 @@
...
+		os.Exit(1)
+	}
+
+	operation := &common.XMLData{
+		XMLName:     "createMetadata",
+		Children:    []common.XMLSchema{objectOperation, fieldOperation},
+		SelfClosing: false,
+	}
+
+	res2, err := sfc.CreateMetadata(context.Background(), operation, tok)
```
laurenzlong, line 130: what's the difference between res1 and res2? did you forget to push a commit? this file looks unfinished.

test/metadata/main.go
```diff
@@ -0,0 +1,306 @@
...
+	if err != nil {
+		slog.Debug("err", "err", err)
+	}
+
+	fmt.Println("Field Operation Result", res2)
+}
+
+func getCreateObjectOperationDefinition() *common.XMLData {
```
laurenzlong, line 138: is this being used for anything?

laurenzlong (changes requested):
Rest looks good, but I think the test file needs work.

test/metadata/main.go
jlimatampersand: Children is array of XMLSchema which is an interface that requires ToXML() for parsing purpose of children. 
Children can be string or another XMLData, and when parsing through, parse operates by `child.ToXML()` and appends to an array.

Otherwise I need to make `Children` to be interface{}, the in side the parser, check switch case to check the child component which I thought was uneccessary complexicty.

test/metadata/main.go
jlimatampersand: removed in stacked PR.

jlimatampersand:
@laurenzlong  Comments are resolved to this PR, but there are some updates to Prototyping Create Metadata on json unmarshalling and validation. These concerns were addressed in separate PR and will merge only after both PRs are approved.

test/metadata/main.go
```diff
@@ -0,0 +1,275 @@
...
+		ClientSecret: clientSecret,
+		Endpoint: oauth2.Endpoint{
+			AuthURL:   fmt.Sprintf("https://%s.my.salesforce.com/services/oauth2/authorize", salesforceSubdomain),
+			TokenURL:  fmt.Sprintf("https://%s.my.salesforce.com/services/oauth2/token", salesforceSubdomain),
+			AuthStyle: oauth2.AuthStyleInParams,
+		},
+	}
+
+	tok := &oauth2.Token{
+		AccessToken:  accessToken,
```
laurenzlong, line 51: Since expiry time is set to `time.Now().Add(-1 * time.Hour)` below, the access token will get refreshed right away, so you can actually just pass a dummy value to accessToken, there's then no need to load it from the credential file.

test/metadata/main.go
```diff
@@ -0,0 +1,275 @@
...
+		os.Exit(1)
+	}
+
+	operation := &common.XMLData{
+		XMLName:     "createMetadata",
+		Children:    []common.XMLSchema{objectOperation, fieldOperation},
+		SelfClosing: false,
+	}
+
+	res, err := sfc.CreateMetadata(context.Background(), operation, tok)
```
laurenzlong, line 117: you can use `ctx` that was created on line 57 instead of creating a new context here.

laurenzlong (approved):
Seems like `testCreateCustomField.xml` and `testCreateCustomObject.xml` are not used anywhere, can those files be deleted? Also, looks like there are some linter errors.

The https://pkg.go.dev/github.com/subchen/go-xmldom library looks like it could replace a lot of this code, but feel free to file a follow up Linear ticket for it instead, there's no urgency in switching to it yet if this is working.

jlimatampersand:
> Seems like `testCreateCustomField.xml` and `testCreateCustomObject.xml` are not used anywhere, can those files be deleted? Also, looks like there are some linter errors.
> 
> The https://pkg.go.dev/github.com/subchen/go-xmldom library looks like it could replace a lot of this code, but feel free to file a follow up Linear ticket for it instead, there's no urgency in switching to it yet if this is working.

I left those as an example so that next person not familiar with SOAP API can take a look and find out how the SOAP API body looks like before going into documentation. I can delete if you insist.

test/metadata/main.go
```diff
@@ -0,0 +1,275 @@
...
+func main() {
+	creds, err := os.ReadFile("../creds.json")
+	if err != nil {
+		slog.Error("Error opening creds.json", "error", err)
+		return
+	}
+
+	var credsMap map[string]interface{}
+
+	if err := json.Unmarshal(creds, &credsMap); err != nil {
```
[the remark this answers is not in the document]
jlimatampersand: Chris and I discussed on this. I think it can be better addressed in separate PR to consolidate ways to get credentials. 

Created task https://linear.app/ampersand/issue/ENG-522/consolidate-reading-credentials-from-testing-runs-in-connectors

PR #35, opened 2023-12-15
chore: Update hubspot error handling

RajatPawar:


hubspot/errors.go
```diff
@@ -50,7 +51,7 @@ type ErrLinks struct {
 }
 
 func (c *Connector) interpretError(res *http.Response, body []byte) error {
-	if res.Header.Get("Content-Type") == "application/json" {
+	if res.Header.Get("Content-Type") == "application/json;charset=utf-8" {
```
eberle1080, line 54: If you're having to deal with this level of differentiation, I'd suggest using [mime.ParseMediaType](https://pkg.go.dev/mime#ParseMediaType). This will give you `application/json` for the first return value (even if the charset is set). It would behave properly regardless of whether or not the charset is added on. More robust, in other words.

hubspot/errors.go
```diff
@@ -74,6 +75,9 @@ func (c *Connector) interpretJSONError(res *http.Response, body []byte) error {
 	}
 
 	switch res.StatusCode {
+	// Hubspot sends us a 400 when the search endpoint returns over 10K records.
+	case http.StatusBadRequest:
+		return createError(ErrSearchLimitExceeded, apiError)
```
eberle1080, line 80: I'm sure I know the answer to this, but is there really no other hint being included in the response? Headers with extra details? `400` (alone) is such an inappropriate marker for this scenario. I know you didn't design it, I'm just baffled that there's no way to tell the difference between "too many records found" vs. "hey user, you sent us garbage".

eberle1080 (approved):

hubspot/errors.go
RajatPawar: This is great, thank you!

hubspot/errors.go
RajatPawar: Nope :( It's annoying that there's not even one way to tell if it's a legit 400 or a 400 due to the search limit. I inspected the raw error that we get and here it is -
Screenshot 2023-12-14 at 5 22 01 PM
Screenshot 2023-12-14 at 5 21 48 PM

PR #40, opened 2023-12-20
[ENG-493] feat: Improve the JSON response interface to be more flexible

eberle1080:
I was writing this for another PR (not yet published) when I realized the changes here are substantial enough to warrant a dedicated PR of its own.

Basically I needed access to the HTTP status code. This led me down a rabbit hole to eventually re-working the JSON return type. The caller can still get the ajson.Node body, but they can also access the response code, the response body -- and crucially then can even unmarshal the response in to a structure if it makes sense to do so. This should cut down on boilerplate code in other places.

### DIFF
--- a/common/json.go
+++ b/common/json.go
@@ -29,6 +29,24 @@ type JSONHTTPClient struct {
 	ErrorHandler ErrorHandler            // optional error handler. If not set, then the default error handler is used.
 }
 
+type JSONHTTPResponse struct {
+	bodyBytes []byte
+
+	Code    int
+	Headers http.Header
+	Body    *ajson.Node
+}
+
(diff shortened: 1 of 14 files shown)


common/json.go
```diff
@@ -29,6 +29,24 @@ type JSONHTTPClient struct {
 	ErrorHandler ErrorHandler            // optional error handler. If not set, then the default error handler is used.
 }
 
+type JSONHTTPResponse struct {
+	bodyBytes []byte
```
laurenzlong, line 33: Can you add a comment that Body is the same data as bodyBytes, but unmarshalled into JSON?

laurenzlong (approved):

PR #41, opened 2023-12-20
[ENG-493] feat: add Hubspot write functionality

eberle1080:
This PR adds write functionality for the hubspot connector. This relies on functionality introduced in #40 

Related:
* https://github.com/amp-labs/connectors/pull/40

Example run:

```
$> go run ./main.go
{
  "success": true,
  "objectid": "29651",
  "data": {
    "company": "Compendia Bioscience Life Technologies",
    "createdate": "2023-12-21T21:51:13.394Z",
    "email": "nathanjacobson@kuhic.io",
    "firstname": "Jarod",
    "hs_all_contact_vids": "29651",
    "hs_email_domain": "kuhic.io",
    "hs_is_contact": "true",
    "hs_is_unworked": "true",
    "hs_lifecyclestage_lead_date": "2023-12-21T21:51:13.394Z",
    "hs_object_id": "29651",
    "hs_object_source": "INTEGRATION",
    "hs_object_source_id": "2317233",
    "hs_object_source_label": "INTEGRATION",
    "hs_pipeline": "contacts-lifecycle-pipeline",
    "hs_searchable_calculated_phone_number": "1554714476",
    "lastmodifieddate": "2023-12-21T21:51:13.394Z",
    "lastname": "Quigley",
    "lifecyclestage": "lead",
    "phone": "1554714476",
    "website": "http://www.dynamicinteractive.io/virtual/optimize/bricks-and-clicks"
  }
}
```

PR #44, opened 2024-01-03
[ENG-459] feat: Add hubspot write tests

eberle1080:
I forgot to add these tests to the write PR. These tests actually validate that hubspot writes work.

See https://github.com/amp-labs/connectors/pull/41 for an example run.

### DIFF
--- a/test/hubspot/main.go
+++ b/test/hubspot/main.go
@@ -17,6 +17,7 @@ import (
 	"golang.org/x/oauth2"
 )
 
+// Contact is a basic Hubspot contact.
 type Contact struct {
 	Email     string `json:"email"`
 	Phone     string `json:"phone"`
@@ -27,14 +28,18 @@ type Contact struct {
 }
 
 func main() {
+	// Handle Ctrl-C gracefully.
 	ctx, done := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
 	defer done()
 
+	// Set up slog logging.
 	utils.SetupLogging()
 
+	// Get the Hubspot connector.
 	hsConn := getConnector(ctx)
 	defer utils.Close(hsConn)
 
+	// Write an artificial contact to Hubspot.
 	result, err := hsConn.Write(ctx, common.WriteParams{
 		ObjectName: "contacts",
 		ObjectId:   "",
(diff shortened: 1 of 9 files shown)


test/hubspot/main.go
```diff
@@ -0,0 +1,139 @@
...
+		utils.Fail("CLIENT_ID not found in creds")
+	}
+
+	clientSecret, found := config["CLIENT_SECRET"]
+
+	if !found {
+		utils.Fail("CLIENT_SECRET not found in creds")
+	}
+
+	accessToken, found := config["ACCESS_TOKEN"]
```
laurenzlong, line 99: access token is not needed from the file if it's going to be refreshed anyways.

test/utils/creds.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"os"
+
+	"github.com/spyzhov/ajson"
+)
+
+// Credentials returns the credentials from the creds.json file.
+func Credentials() (*ajson.Node, error) {
+	fileName := "creds.json"
```
laurenzlong, line 11: Can this be a parameter to the function instead? There will be more than one creds file for the different tests, within the Hubspot tests, can pass in `hubspot-creds.json` as the parameter, and within the Salesforce tests, can pass in `salesforce-creds.json`, etc.

test/hubspot/main.go
```diff
@@ -0,0 +1,139 @@
...
+type Contact struct {
+	Email     string `json:"email"`
+	Phone     string `json:"phone"`
+	Company   string `json:"company"`
+	Website   string `json:"website"`
+	Lastname  string `json:"lastname"`
+	Firstname string `json:"firstname"`
+}
+
+func main() {
```
laurenzlong, line 30: Can you please add comments on how to run this test, and also create a sample creds file in OnePassword so others on the team can easily get up and running with this test?

test/hubspot/main.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil {
+		utils.Fail("error creating hubspot connector", "error", err)
+	}
+
+	return conn
+}
+
+// getOAuthInfo returns the OAuth2 config and token from the creds.json file.
+func getOAuthInfo(creds *ajson.Node) (*oauth2.Config, *oauth2.Token) {
+	config := creds.MustObject()
```
laurenzlong, line 86: Can you put lines 86 to 109 into a helper function that can be used by all the different provider tests not just the Hubspot test?

laurenzlong (approved):
Looks good, I mostly have feedback on how to make this reusable for other tests, filed https://linear.app/ampersand/issue/ENG-525/align-saleforce-connector-tests-to-use-same-helper-functions-as to align Salesforce tests with these tests.

test/hubspot/main.go
eberle1080: Fair point, I'll just make it optional then

PR #45, opened 2024-01-04
chore: Update contacts lastmodifieddate field

RajatPawar:
* The `contacts` object silently fails to return objects according to criteria even if we use `hs_lastmodifieddate`. 
* According to https://community.hubspot.com/t5/APIs-Integrations/CRM-V3-API-Search-issue-with-Contacts-when-using-Filters/m-p/324617 (and testing), the right field to use is `lastmodifieddate` for contacts and `hs_lastmodifieddate` for other objects.

Also shifts a function that gets the last ID of a result into the connector, so that the server has minimal knowledge of the internal structure of a connector's response

## Concerns
* The check for which `lastmodifiedfield` to use is a direct string comparison, so we need to be careful (https://linear.app/ampersand/issue/ENG-520/how-do-we-handle-inconsistent-plural-object-name)

### DIFF
--- a/hubspot/search.go
+++ b/hubspot/search.go
@@ -35,11 +35,24 @@ func (c *Connector) Search(ctx context.Context, config SearchParams) (*common.Re
 }
 
 // BuildLastModifiedFilterGroup filters records modified since the given time.
-func BuildLastModifiedFilterGroup(since time.Time) Filter {
+// If the time is zero, it returns an empty filter. For contacts, it uses the
+// lastmodifieddate field. For other objects, it uses the hs_lastmodifieddate.
+// Read more: https://community.hubspot.com/t5/APIs-Integrations/CRM-V3-API-Search-issue-with-Contacts-when-using-Filters/m-p/324617
+func BuildLastModifiedFilterGroup(params *common.ReadParams) Filter {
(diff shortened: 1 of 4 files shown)


hubspot/search.go
```diff
@@ -35,11 +35,24 @@ func (c *Connector) Search(ctx context.Context, config SearchParams) (*common.Re
 }
 
 // BuildLastModifiedFilterGroup filters records modified since the given time.
-func BuildLastModifiedFilterGroup(since time.Time) Filter {
+// If the time is zero, it returns an empty filter. For contacts, it uses the
+// lastmodifieddate field. For other objects, it uses the hs_lastmodifieddate.
+// Read more: https://community.hubspot.com/t5/APIs-Integrations/CRM-V3-API-Search-issue-with-Contacts-when-using-Filters/m-p/324617
```
jlimatampersand, line 40: Please add //nolint:lll to fix linter error

jlimatampersand (approved):
LGTM

hubspot/search.go
RajatPawar: Local & github action linters both aren't complaining about this, where did you see this? Merging it for now, will follow up with this if there's an error anywhere

PR #47, opened 2024-01-05
fix: Make the data format for hubspot more consistent

eberle1080:
no more wrapping everything in 'properties'. This will ensure a consistent experience for our users.

### DIFF
--- a/hubspot/write.go
+++ b/hubspot/write.go
@@ -31,7 +31,10 @@ func (c *Connector) Write(ctx context.Context, config common.WriteParams) (*comm
 		write = c.Client.Post
 	}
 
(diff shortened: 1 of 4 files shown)


hubspot/write.go
```diff
@@ -31,7 +31,10 @@ func (c *Connector) Write(ctx context.Context, config common.WriteParams) (*comm
 		write = c.Client.Post
 	}
 
-	json, err := write(ctx, url, config.ObjectData)
+	data := make(map[string]interface{})
+	data["properties"] = config.ObjectData
```
laurenzlong, line 35: Can you add a comment that Hubspot requires all fields to be nested within "properties"?

laurenzlong (approved):

PR #54, opened 2024-01-12
fix: check header before adding

RajatPawar:


common/http.go
```diff
@@ -316,11 +313,20 @@ func getURL(baseURL string, urlString string) (string, error) {
...
 	for _, hdr := range headers {
 		req.Header.Add(hdr.Key, hdr.Value)
 	}
 
-	return req, nil
+	return req
+}
+
+// addContentTypeIfNotPresent adds the Content-Type header if it is not already present.
+func addContentTypeIfNotPresent(req *http.Request) *http.Request {
```
laurenzlong, line 326: Can you rename to `addJSONContentTypeIfNotPresent` or `ensureJSONHeader`? (Basically add JSON to the name so it's obvious what the behavior is.)

laurenzlong (approved):

PR #60, opened 2024-01-30
[ENG-620] chore: Add empty module to enable hubspot proxying

RajatPawar:
* Our current setup always requires a module to be input which is a problem for proxy requests since a CRM specific prefix is automatically added to it. 
* This PR adds an empty module which we can use in the server repo for proxy request building

### Testing
- [x] Read
- [x] Write
- [x] ListObjectMetadata
- [x] Search

### DIFF
--- a/hubspot/modules.go
+++ b/hubspot/modules.go
@@ -9,9 +9,16 @@ type APIModule struct {
 	Version string // e.g. "v3"
 }
 
+// ModuleCRM is the module used for accessing standard CRM objects
 var ModuleCRM = APIModule{ // nolint: gochecknoglobals
 	Label:   "crm",
 	Version: "v3",
+}
+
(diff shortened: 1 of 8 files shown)


hubspot/modules.go
```diff
@@ -9,11 +9,18 @@ type APIModule struct {
 	Version string // e.g. "v3"
 }
 
+// ModuleCRM is the module used for accessing standard CRM objects.
 var ModuleCRM = APIModule{ // nolint: gochecknoglobals
 	Label:   "crm",
 	Version: "v3",
 }
 
+// ModuleEmpty is Used for proxying requests through.
```
laurenzlong, line 18:
```suggestion
// ModuleEmpty is used for proxying requests through.
```

laurenzlong (approved):

PR #61, opened 2024-01-30
[ENG-586] Introduce providers.yaml 

RajatPawar:
* Adds `providers` package to house all providers
* Adds a `providers.ReadConfig` which uses the caller's file loc to find the `providers.yaml` file which stores all the config regarding providers. Also adds a test for it 
* Refactors connectors to use these providers instead of variables

### Follow up 
* Actually consume the config to make requests, etc
* Update server code to use `connector/provider` types


providers/providers.yaml
```diff
@@ -0,0 +1,28 @@
...
+# 1) Connector specific, which define properties specific to the go connector itself (e.g. connector_type)
+# 2) Provider specific, which define configuration related to the actual provider API (e.g. provider_base_url)
+# Each field is prefixed accordingly as a convenience and convention.
+
+# This YAML file allows for {{.variables}} (note the period) in the configuration, which can be replaced at runtime
+# by the connector. This syntax (and more) comes from the https://pkg.go.dev/html/template package.
+
+providers:
+  salesforce:
+    connector_type: full
```
laurenzlong, line 13: we need much more granular tracking of supported features than "full" and "basic", https://github.com/amp-labs/server/blob/f38444bef62e823e721e738ddbda5554fd02ad76/shared/common/provider.go#L19 can probably be moved to here.

providers/providers.yaml
```diff
@@ -0,0 +1,28 @@
...
+    connector_version: 1.0.0
+    provider_auth_type: oauth2
+    provider_base_url: https://{{.subdomain}}.salesforce.com
```
laurenzlong, line 16: Can we use `workspace` instead of `subdomain` as the term here? Workspace is a more generic term, whereas subdomain is very Salesforce-specific. We also use `workspace` in our data schemas already (e.g. the Connections table has a `providerWorkspaceRef`)

providers/providers.yaml
```diff
@@ -0,0 +1,28 @@
...
+# 2) Provider specific, which define configuration related to the actual provider API (e.g. provider_base_url)
+# Each field is prefixed accordingly as a convenience and convention.
+
+# This YAML file allows for {{.variables}} (note the period) in the configuration, which can be replaced at runtime
+# by the connector. This syntax (and more) comes from the https://pkg.go.dev/html/template package.
+
+providers:
+  salesforce:
+    connector_type: full
+    connector_version: 1.0.0
```
laurenzlong, line 14: Not sure I understand the usefulness of tracking connector versions, since we are not distributing each connectors as its own package, therefore there's no way to revert to a previous version of a particular connector.

hubspot/provider.go
```diff
@@ -0,0 +1,10 @@
+package hubspot
```
laurenzlong, line 1: We are getting an over-proliferation of files inside of each connector directory, which increases the overhead of creating new connectors. Can this live in `string.go`? It might also make sense to rename that file to `strings.go`  and consolidate `name.go`, `url.go` to all live in that file.

hubspot/string.go
```diff
@@ -4,5 +4,5 @@ import "fmt"
 
 // String returns a string representation of the connector, which is useful for logging / debugging.
 func (c *Connector) String() string {
-	return fmt.Sprintf("hubspot.Connector[%s]", c.Module)
```
laurenzlong, line 7: I just noticed this `String` method is not a part of any the interfaces yet, it should probably be a part of BasicConnector

basic.go
```diff
@@ -18,4 +19,7 @@ type BasicConnector interface {
 
 	// HTTPClient returns the underlying HTTP client. This is useful for proxy requests.
 	HTTPClient() *common.HTTPClient
+
+	// Provider returns the connector provider.
+	Provider() providers.Provider
```
laurenzlong, line 24: It's a bit redundant to have both the `Name` method and the `Provider` method (and it increases the overhead of creating new connectors. I think we should add Provider and remove `Name`, since `Provider().String()` is effectively equivalent to `Name()`

providers/config.go
```diff
@@ -0,0 +1,100 @@
...
+	"gopkg.in/yaml.v3"
+)
+
+const (
+	// configFileLoc is the name of the config file.
+	configFileRelativeLoc = "providers.yaml"
+)
+
+var (
+	ErrProviderConfigNotFound = errors.New("provider config not found")
```
laurenzlong, line 21:
```suggestion
	ErrProviderConfigNotFound = errors.New("provider config file not found")
```

providers/config.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+// Config is the entire configuration for all providers.
+type Config struct {
```
laurenzlong, line 25: Can we name this something other than Config is it doesn't get confused with the `Config` in our backend (which means the configuration of an installation). Maybe `Catalog`?

providers/config.go
```diff
@@ -0,0 +1,100 @@
...
+	Providers map[Provider]map[string]string `yaml:"providers"`
+}
+
+// ReadConfig reads the configuration from the config file for
+// a specific provider. It also performs string substitution
+// on the values in the config that are surrounded by {{}}.
+// The provider YAML has more details on how it works.
+func ReadConfig(provider Provider, substitutions map[string]string) (map[string]string, error) {
```
laurenzlong, line 33: It would be good if this actually returned a Go type rather than a map of strings, then the calling code can work with the returned result more easily.

test/providers/config_test.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"errors"
+	"testing"
+
+	"github.com/amp-labs/connectors/providers"
+)
+
+func TestReadConfig(t *testing.T) {
+	// Define test cases
+	testCases := []struct {
```
laurenzlong, line 12: Can you add a field to this struct called `description` so each test case has a quick description and if any of them fail, the logged error also contains the description?

laurenzlong (changes requested):
Super excited about the direction this is taking us!

hubspot/provider.go
RajatPawar: Moved Name() & Provider() inside `strings`, all HTTP method files into one `methods.go` file and `Close` into `client.go`

hubspot/string.go
RajatPawar: `BasicConnector` already has `fmt.Stringer`

providers/providers.yaml
RajatPawar: We don't, but since this is open source, it might be nice to version it.

providers/providers.yaml
RajatPawar: This is meant to be salesforce specific. So when the salesforce client will read config, it will look like - 
```
params = &salesforceParams{
  subdomain: ..,
}

configMap, err := config.Read(providers.Salesforce, map[string]string{
  "subdomain": params.subdomain,
})

params.base_url = configMap["provider_base_url"]
```

The substitutions are done per provider, so we should use the same term that we are using in the params and inputs for ease of use / understanding

providers/providers.yaml
RajatPawar: Sure, this is a non-exhaustive list of the configuration, just as a POC. Let's catch up to discuss more

providers/config.go
RajatPawar: I played around with this, but configuration for providers will diverge, leaving us with two options - 
* Define config structs for each provider, which is a lot of work 
* Use a map (optionally define keys as constants with which to look the map up)

Or we could do something in between where we define some static configuration that is a struct, and some per-provider config that is a map.

I went with a map because the person adding a new config key will be the one using it so it should be okay as long as we are careful & we keep extending the testcases. Secondly, this is something that will not change often.

providers/config.go
RajatPawar: Renamed this type, but let me know if you want me to avoid the word 'Config' even in other symbols like ReadConfig / ProviderConfig, etc.

providers/config.go
```diff
@@ -0,0 +1,122 @@
...
+	err = tmpl.Execute(&result, substitutions)
+	if err != nil {
+		return "", err
+	}
+
+	return result.String(), nil
+}
+
+// read reads the entire configuration from the config file.
+func read() (*Catalog, error) {
```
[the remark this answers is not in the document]
laurenzlong: maybe also renamed to `getCatalog` or something to be more clear?

providers/types.go
```diff
@@ -0,0 +1,35 @@
...
+
+// Catalog is the top-level structure of the configuration file.
+type Catalog struct {
+	Providers map[Provider]ProviderConfig `yaml:"providers"`
+}
+
+// ProviderConfig is the configuration for a specific provider.  We use reflection to substitute any variables
+// in the configuration. The substitution is only done on string fields. If you want to use pointers in the struct,
+// you might have to update the code to handle it.
+type ProviderConfig struct {
```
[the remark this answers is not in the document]
RajatPawar: Updated

PR #65, opened 2024-02-10
[ENG-522] Create credential getter. 

jlimatampersand:
This is Part 1 for consolidation of credential getting mechanism. 

Created Credential getting struct and methods that support multi-source value getting ( currently from env var , and json file). 

Part 2 will do clean-up of current code base test examples to use these methods and consolidate and reduce redundant code.

### DIFF
--- a/utils/credentials.go
+++ b/utils/credentials.go
@@ -1,0 +1,402 @@
+// nolint: ireturn
+package utils
+
+import (
+	"errors"
+	"fmt"
+	"log"
+	"log/slog"
+	"os"
+
+	"github.com/spyzhov/ajson"
+)
+
+var (
+	ErrWrongType            = errors.New("wrong type")
+	ErrEnvVarNotSet         = errors.New("environment variable not set")
+	ErrEmptyStringEnvVar    = errors.New("empty value for environment variable")
+	ErrEmptyFilePathPathVar = errors.New("empty value for file path")
+	ErrJSONPathNotFound     = errors.New("empty value at json path")
+	ErrReaderNotFound       = errors.New("Reader not found")
+	ErrCredentialNotFound   = errors.New("credential not found")
+)
+
+type CredentialsRegistry map[string]Reader
+
+type Reader interface {
+	Value() (any, error)
+	Key() (string, error)
+}
+
+type EnvReaderParam struct {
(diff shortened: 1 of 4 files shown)


jlimatampersand:
* `main`
  * **PR #65** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/65?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/amp-labs/connectors/65?utm_source=stack-comment).

utils/credentials.go
```diff
@@ -0,0 +1,608 @@
...
+	return list[0].Value()
+}
+
+func (c CredentialsOptions) AddLocator(locator Locator) error {
+	key, err := locator.GetKey()
+	if err != nil {
+		return err
+	}
+
+	c[key] = locator
```
eberle1080, line 100: Only thing about this is that you can have 1 and only 1 locator per key. Not a huge deal, but it might be nice if the locators could chain (i.e. try a first locator, if not found try a second, if not found try a third, etc until the value is found). But that might be more complexity than its worth.

utils/credentials.go
```diff
@@ -0,0 +1,608 @@
...
+		value, err := locator.Locate()
+		if err != nil {
+			return nil, err
+		}
+
+		creds[key] = value
+	}
+
+	return creds, nil
+}
```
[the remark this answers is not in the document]
jlimatampersand: Couldn't find good argument for that lol 

Thought of it as just a nice to have, but that registry and the credentials are almost identical. So I removed it :)

utils/credentials.go
jlimatampersand: I like the idea. Working on it.

jlimatampersand: Actually, this can be achieved by defining multi chained Reader. CredentialRegistry accepts a Reader interface, so how you defined the reader can make the difference.

jlimatampersand: For this PR, I will leave it as is just with basic 1 to 1 reader. We can add more readers to it.

PR #66, opened 2024-02-12
[ENG-668] Add generic basic connector creation

RajatPawar:
* This PR allows a caller to construct a connector given there is some basic configuration for the request provider inside the provider catalog. 
* Updates readme
* Disallows creation of a basic connector for salesforce/hubspot using these methods as they currently don't use the catalog to set their internal state


basic.go
RajatPawar, line 1: Thought it's just easier to read BasicConnector & Connector interface together

README.md
```diff
@@ -93,3 +98,31 @@ client, err := connectors.Salesforce.New(
 ```go
 client, err := connectors.New("salesforce", map[string]any{"workspace": "salesforce-instance-name"})
 ```
+
+## Basic connectors
```
eberle1080, line 102: Could we name this something like "proxy connectors"? Basic makes it sounds like something barebones which should be copied more or less verbatim, or derived from. Not sure we want to give that impression.

basic/connector.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+// Connector is a Hubspot connector.
+type Connector struct {
+	ProviderInfo *providers.ProviderInfo
+	Client       *common.JSONHTTPClient
+	provider     providers.Provider
+}
+
+// NewConnector returns a new Hubspot connector.
```
eberle1080, line 15: might want to remove the reference to hubspot

eberle1080 (changes requested):
Overall looks good. I have concerns about the name though. `BasicConnector` doesn't say anything about what it does. I could see `ProxyConnector` or `ShallowConnector` or `PrimitiveConnector` or something else, but as is I think this is a recipe for confusion.

README.md
RajatPawar: Agreed - the terminology isn't great. This concerns product terminology so probably best to wait for @laurenzlong

eberle1080 (approved):

README.md
RajatPawar: Converting the naming decision into an async call which @laurenzlong will get back to us with

PR #67, opened 2024-02-14
[ENG-684] Update hubspot & SF connectors to consume catalog

RajatPawar:
* Reads values from catalog to set internal connector state
* This PR intentionally doesn't try to replace URL constructions, etc - we need to find a better way to do that in cases of Salesforce, etc. Sometimes we need base parts of the URL and sometimes we need the suffix of a URL, etc.

## Next steps
* Figure out what is the maximum amount of information we can transfer from code into the YAML without breaking anything or having too many overlapping fields in the YAML. For example, in metadata requests, we need only `services/data/v59.0/...` but for a read request we need `https://......./services...` - how do you store the least amount of information to get the most out of this?
* Refactor these connectors to get rid of struct members like 'BaseURL' or 'Domain' and work off a `ProviderInfo` member?

### DIFF
--- a/providers/providers.yaml
+++ b/providers/providers.yaml
@@ -10,24 +10,27 @@
 
 providers:
   salesforce:
-    baseUrl: https://{{.subdomain}}.salesforce.com
+    baseUrl: "https://{{.subdomain}}.salesforce.com"
     authType: oauth2
     oauthOpts:
-      authUrl: https://{{.subdomain}}.my.salesforce.com/services/oauth2/authorize
-      tokenUrl: https://{{.subdomain}}.my.salesforce.com/services/oauth2/token
+      authUrl: "https://{{.subdomain}}.my.salesforce.com/services/oauth2/authorize"
+      tokenUrl: "https://{{.subdomain}}.my.salesforce.com/services/oauth2/token"
     support:
       bulkWrite: true
       proxy: true
       read: true
       subscribe: false
       write: true
+    providerOptions:
+      restApiUrl: "https://{{.subdomain}}.my.salesforce.com/services/data/v59.0"
+      domain: "{{.subdomain}}.my.salesforce.com"
 
   hubspot:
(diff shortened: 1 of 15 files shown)


providers/types.go
```diff
@@ -33,3 +34,13 @@ type AuthType string
 const (
 	AuthTypeOAuth2 AuthType = "oauth2"
 )
+
+func (i *ProviderInfo) GetOption(key string) (string, bool) {
+	if i.ProviderOptions == nil {
```
laurenzlong, line 39: This if statement isn't needed, ` i.ProviderOptions[key]` will already return `"", false` if key is not found. See https://go.dev/play/p/fGhFnVJoZzy

providers/providers.yaml
```diff
@@ -10,24 +10,27 @@
...
-      tokenUrl: https://{{.subdomain}}.my.salesforce.com/services/oauth2/token
+      authUrl: "https://{{.subdomain}}.my.salesforce.com/services/oauth2/authorize"
+      tokenUrl: "https://{{.subdomain}}.my.salesforce.com/services/oauth2/token"
     support:
       bulkWrite: true
       proxy: true
       read: true
       subscribe: false
       write: true
+    providerOptions:
```
laurenzlong, line 24: Can you add comments to each of these options to explain what they do?

laurenzlong (approved):

providers/providers.yaml
```diff
@@ -10,24 +10,27 @@
 
 providers:
   salesforce:
-    baseUrl: https://{{.subdomain}}.salesforce.com
+    baseUrl: "https://{{.subdomain}}.salesforce.com"
```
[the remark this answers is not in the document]
RajatPawar: Updated, can you please review the PR again?

providers/types.go
RajatPawar: This is to check if the map even exists (it won't for hubspot which doesn't have any options)

PR #70, opened 2024-02-21
[ENG-729] feat: Add helper scripts for running proxy & obtaining oauth token

RajatPawar:
The `oauth.go` script creates the token as a file, which the `proxy.go` script then uses to run a server that forwards requests to the provider.

### DIFF
--- a/scripts/oauth/token.go
+++ b/scripts/oauth/token.go
@@ -1,0 +1,371 @@
+package main
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"log"
+	"log/slog"
+	"net/http"
+	"os"
+	"os/exec"
+	"runtime"
+	"strconv"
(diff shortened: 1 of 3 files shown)


scripts/oauth/token.go
```diff
@@ -0,0 +1,374 @@
+package main
```
eberle1080, line 1: Uh there's already a script in this directory. Having 2 main methods side-by-side like this is gonna be a problem. Can you rename? Also the script currently in oauth is special-made to obtain oauth tokens. This seems like duplicated functionality.

RajatPawar: I created the `scripts` directory, you might be thinking of the server?

RajatPawar: (Also, I copied most of the functionality here from the server script)

eberle1080 (approved):

PR #71, opened 2024-02-21
feat: Use OpenAPI spec to hold ProviderInfo types 

RajatPawar:
* Handling YAML / file changes is trickier than expected. This PR defines an OpenAPI spec that holds the type of the provider information. 
* The Go types generated using this spec are used to store the catalog.
* This will allow us to also generate types for the FE and use the provider information 

**Note** that this is technically an 'invalid' or 'incomplete' spec that linters will give warnings for, because we are missing `paths` and some other attributes of the spec.

### Go
Works as expected, can be seen in `types.gen.go`

### Typescript
Using the command - 
```
openapi-generator-cli generate -i ../connectors/providers/types.yaml -g typescript-fetch -o generated-sources/providers --skip-validate-spec --additional-properties=npmName=amp-labs-generated-provider-sdk,supportsES6=true,withInterfaces=true
```
gives us - 

Screenshot 2024-02-22 at 2 54 07 PM

### Remote reference works
This spec can be remotely referred to using the full URL - 
Screenshot 2024-02-22 at 2 51 27 PM

## readme.io docs work
Screenshot 2024-02-22 at 3 06 57 PM


salesforce/events.go
```diff
@@ -258,8 +258,6 @@ func (c *Connector) postToSFAPI(ctx context.Context, body any, path string, enti
 		return nil, err
 	}
 
-	fmt.Println("location: ", location)
```
laurenzlong, line 261: did yo accidentally commit this?

providers/utils.go
```diff
@@ -0,0 +1,107 @@
...
+	"reflect"
+	"strings"
+
+	"github.com/go-playground/validator"
+)
+
+var ErrProviderCatalogNotFound = errors.New("provider or provider catalog not found")
+
+// ReadConfig reads the configuration from the config file for a specific provider. It also performs string substitution
+// on the values in the config that are surrounded by {{}}. The provider YAML has more details on how it works.
```
laurenzlong, line 15: outdated comment since it's not a provider YAML anymore

laurenzlong (approved):

salesforce/events.go
RajatPawar: No, this was a linter warning, unless we need this?

PR #78, opened 2024-03-07
[ENG-788] feat: Add SalesLoft connector

Jkarage:
## Catalog variables
No required catalog variables

## Notes 
The proxy and API responses differ when the response from the API is Bad Request(400). The proxy returns an empty body. The API has some details.

### GET
Screenshot 2024-03-07 at 13 34 40

### POST
Screenshot 2024-03-07 at 08 42 34

### PUT
Screenshot 2024-03-07 at 11 29 50

### DELETE
Screenshot 2024-03-07 at 11 26 06

## Pagination
Showing pagination sample. The response data will have _next_page_ in the metadata.pagination object.
If the next_page value is not "null", There are more pages.
Screenshot 2024-03-07 at 12 45 58

Then we can access the next page, by adding the page query parameter in the next response.
Screenshot 2024-03-07 at 12 46 29

### DIFF
--- a/providers/catalog.go
+++ b/providers/catalog.go
@@ -8,6 +8,7 @@ const (
 	Salesforce Provider = "salesforce"
 	Hubspot    Provider = "hubspot"
 	LinkedIn   Provider = "linkedIn"
+	Salesloft  Provider = "salesloft"
 )
 
 // ================================================================================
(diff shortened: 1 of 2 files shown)


providers/catalog.go
```diff
@@ -8,6 +8,7 @@ const (
 	Salesforce Provider = "salesforce"
 	Hubspot    Provider = "hubspot"
 	LinkedIn   Provider = "linkedIn"
+	SalesLoft  Provider = "salesLoft"
```
RajatPawar, line 11: Can we name this as `Salesloft`? From their website, seems like they refer to themselves as a single word

RajatPawar (changes requested):

RajatPawar (approved):

providers/catalog.go
RajatPawar: Approved - feel free to 'Squash and merge' once you resolve this.

PR #81, opened 2024-03-11
[ENG-796] feat: Add Keap connector

Listener430:
## Catalog variables
No required variables.

### GET 
 localhost:4444/contacts/model
proxy_get

### POST
localhost:4444/companies 
proxy_post

### PUT
localhost:4444/account/profile
proxy_put

### PATCH
doesn't work via proxy localhost:4444/contacts?contactId=60 (returns 405 error - method not allowed)
patch-proxy-doesn't work
while if hit directly at https://api.infusionsoft.com/crm/rest/v1/contacts/contactId=60 everything works fine
patch-direct-works

### DEL
localhost:4444/contacts?contactId=60 via proxy returns 405 
del-via-proxy-doesn't work

if run directly works correctly - https://api.infusionsoft.com/crm/rest/v1/contacts/contactId=60 
del-directly-works

## Pagination
Pagination via proxy returns all contracts and there are two links at the bottom for next and prev page. 

pagination-proxy

if run directly, the same output is shown. I.e. seems they have a bug as it should have returned only fraction of contacts, not all of them. At the same time, the pagination links are correct. 

pagination_next_page

### DIFF
--- a/providers/catalog.go
+++ b/providers/catalog.go
@@ -10,6 +10,7 @@ const (
 	LinkedIn   Provider = "linkedIn"
 	Salesloft  Provider = "salesloft"
 	Outreach   Provider = "outreach"
+	Keap       Provider = "keap"
 )
 
 // ================================================================================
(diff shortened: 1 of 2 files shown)


providers/catalog.go
```diff
@@ -115,4 +116,24 @@ var catalog = CatalogType{ // nolint:gochecknoglobals
 			Write:     false,
 		},
 	},
+
+	Keap: {
+		AuthType: Oauth2,
+		BaseURL:  "https://api.infusionsoft.com/crm/rest/v1",
```
RajatPawar, line 122: We want to allow users to call endpoints that may not be crm / rest / v1, so we should probably just have this as `https://api.infusionsoft.com`

RajatPawar (changes requested):

RajatPawar (approved):

PR #88, opened 2024-03-13
[ENG-825] feat: Add Dropbox connector

sanjay-a01:
## Catalog variables
Dropbox    Provider = "dropbox"

## Notes
Supports batching requests and pagination as well

### GET 
URL: http://localhost:4444/file_requests/get
GET

### POST
URL: http://localhost:4444/file_requests/create
POST

### PUT
URL: http://localhost:4444/file_requests/update
PUT

### DIFF
--- a/providers/catalog.go
+++ b/providers/catalog.go
@@ -12,6 +12,7 @@ const (
 	Outreach   Provider = "outreach"
 	Keap       Provider = "keap"
 	Asana      Provider = "asana"
+	Dropbox    Provider = "dropbox"
 )
 
 // ================================================================================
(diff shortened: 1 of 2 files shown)


providers/catalog.go
```diff
@@ -156,4 +157,23 @@ var catalog = CatalogType{ // nolint:gochecknoglobals
 			Write:     false,
 		},
 	},
+	
+	// Dropbox configuration
+    Dropbox: {
+        AuthType: Oauth2,
+        BaseURL:  "https://api.dropboxapi.com/2/",
```
RajatPawar, line 164: Users should probably have the freedom to call the v1 of the api, so I would remove `2` from the base URL - what do you think?

RajatPawar: Sure - please add this information to the dropbox slab if you haven't already. Thanks!

RajatPawar (approved):

PR #95, opened 2024-03-15
[ENG-758] Add metadata fields to catalog

RajatPawar:
# Context
* OAuth tokens obtained from providers can come with extra metadata that contain information about -
  * The account/workspace that the token is associated with (referred to as the `workspaceRef` here on)
  * The user that the token is associated with (referred to as the `consumerRef` here on)
  * The scopes related to the token

* For example, a salesforce token looks like - 
```
{
        "access_token": "1234",
        "refresh_token": "4321",
        "scope": "refresh_token api id full",
        "instance_url": "https://ampersand-dev-ed.my.salesforce.com",
        "id": "https://login.salesforce.com/id/00D5x0000019AWSGA1/0053x000000orG1ABI",
        "token_type": "Bearer",
        "issued_at": "1680295102278"
    }
```

* and a notion token looks like - 
```
{
        "access_token": "1234",
        "refresh_token": "4321",
        "workspace_id": "65b72915-fddc-4107-b76c-8575109ea65a",
        "owner": {
          "user": {
             "id": "some-user-id"
         }
        "token_type": "Bearer",
        "issued_at": "1680295102278"
    }
```

# Requirement
* In each oauth token obtained from a provider, if available, we are interested in capturing these fields. For Salesforce, `instance_url` is the `workspaceRef`, `id` is the the `consumerRef`, and the `scope` field is as is.
* Since the names of these payload keys can change from provider to provider, we want to store the names of these 3 fields in the catalog.
* In Notion's case, the `workspaceRef` is found under the `workspace_id` field, the `consumerRef` is found under `owner.user.id` field (note that nesting is denoted by a dot) and there is no scope field.

# Changes
This PR introduces a struct called `TokenMetadataFields` under `OAuthOpts` which stores the `WorkspaceRefField`, the `ConsumerRefField` and the `ScopesField`. 

For Salesforce, the `WorkspaceRefField` would be `instance_url`, while for Notion it would be `workspace_id`, and so on. 

This will allow us to be provider-agnostic when extracting token metadata.

### DIFF
--- a/providers/catalog.go
+++ b/providers/catalog.go
@@ -134,7 +134,6 @@ var catalog = CatalogType{ // nolint:gochecknoglobals
 		AuthType: Oauth2,
 		BaseURL:  "https://api.close.com/api",
 		OauthOpts: OauthOpts{
-
 			AuthURL:                   "https://app.close.com/oauth2/authorize",
 			TokenURL:                  "https://api.close.com/oauth2/token",
 			ExplicitScopesRequired:    false,
@@ -154,7 +153,6 @@ var catalog = CatalogType{ // nolint:gochecknoglobals
 		BaseURL:  "https://api.infusionsoft.com",
 
 		OauthOpts: OauthOpts{
-
 			AuthURL:                   "https://accounts.infusionsoft.com/app/oauth/authorize",
 			TokenURL:                  "https://api.infusionsoft.com/token",
 			ExplicitScopesRequired:    false,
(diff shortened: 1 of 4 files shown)


laurenzlong (approved):
One thing I realized that we should also have in the catalog is a display name for the provider (e.g. "LinkedIn", "DocuSign", "Close CRM") this will be useful for both the UI library and Management Console. Filed https://linear.app/ampersand/issue/ENG-847/add-display-name-to-providers-catalog

PR #98, opened 2024-03-15
[ENG-786] feat: Add Gong connector

Jkarage:
## Checklist
- [x] Ran Linter
- [x] Catalog tests passing

## Catalog variables
workspace variable is required. It's given by Gong. Sample is us-49467 in https://us-49467.api.gong.io/

## Notes
It's currently private, The only OAuth user who can access the app is the Creator. Publishing it requires filling a form.
Also the token endpoint requires Basic Authorization header whose values is base64 encoded `client_id:client_secret`

### GET
Screenshot 2024-03-15 at 12 14 22

### POST
Screenshot 2024-03-15 at 13 03 09

### PUT
Screenshot 2024-03-15 at 13 10 26

### DELETE
Screenshot 2024-03-15 at 13 26 22

## Pagination
For pagination, we would have the cursor value in the records object below: We would in turn use that value as the query parameter in the next request. 
Screenshot 2024-03-15 at 12 12 41

### DIFF
--- a/providers/catalog.go
+++ b/providers/catalog.go
@@ -15,6 +15,7 @@ const (
 	Asana      Provider = "asana"
 	Dropbox    Provider = "dropbox"
 	Notion     Provider = "notion"
+	Gong       Provider = "gong"
 )
 
 // ================================================================================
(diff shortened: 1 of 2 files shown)


Jkarage:
@RajatPawar 
I have also changed the test values for Notion, as the changes for proxy support were not made in the tests as well.

RajatPawar (approved):

PR #102, opened 2024-03-18
[ENG-724] [draft] Refactor connector interfaces & consume catalog information

RajatPawar:


connectors.go
```diff
@@ -4,30 +4,36 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io"
 
 	"github.com/amp-labs/connectors/common"
 	"github.com/amp-labs/connectors/hubspot"
 	"github.com/amp-labs/connectors/providers"
+	"github.com/amp-labs/connectors/proxy"
```
eberle1080, line 11: Seems like this is going to cause an import cycle, you might want to move some files around to avoid this.

connectors.go
```diff
@@ -97,74 +86,66 @@ var (
...
 
-// newHubspot returns a new Hubspot Connector, by unwrapping the options and passing them to the Hubspot API.
-func newHubspot(opts map[string]any) (Connector, error) { //nolint:ireturn
-	var options []hubspot.Option
-
-	c, valid := getParam[common.AuthenticatedHTTPClient](opts, "client")
-	if valid {
-		options = append(options, hubspot.WithAuthenticatedClient(c))
+// NewReadConnector returns a new read connector for the given provider with the given build options.
+func NewReadConnector(
```
eberle1080, line 98: The use of "proxy.Option" is kind of confusing here. If I'm creating a reader, I'm not expecting to deal with proxy things. I'd say if you're going down this route, just move proxy.Option in to the providers package (`providers.Option`) which would seem more intuitive.

connectors.go
```diff
@@ -97,74 +86,66 @@ var (
...
+	if err != nil {
+		return nil, err
 	}
 
-	w, valid := getParam[hubspot.APIModule](opts, "module")
-	if valid {
-		options = append(options, hubspot.WithModule(w))
+	switch provider {
+	case providers.Hubspot:
+		conn, outErr = hubspot.NewConnector(hubspot.WithProxyConnector(proxyConn))
```
eberle1080, line 114: we don't want the proxy connector though right? We just want the HTTP client. Why pass in the whole connector? Seems like a terminology thing here. I still like "proxy connector" IF the intended use is for acting as a proxy. But if all we're doing is using it to make calls -- to other places within the same repo, I think something like "ProviderClient" might make the distinction more clear. Not sure if that's the best name but hopefully you get my meaning.

hubspot/connector.go
```diff
@@ -1,57 +1,45 @@
...
+// and extends it with Hubspot-specific functionality (read, write, etc).
 type Connector struct {
-	Module  string
-	BaseURL string
-	Client  *common.JSONHTTPClient
+	*proxy.Connector
 }
 
-// NewConnector returns a new Hubspot connector.
+// NewConnector returns a new hubspot connector.
```
eberle1080, line 18: nit: why downcase this?

laurenzlong (commented):
This is a good direction, a few high level thoughts:
- we should also have an interface for BulkWriteConnector, and a MetadataConnector (which supports  `ListObjectMetadata`)
- There can be a lot more code reuse between `NewReadConnector`, `NewWriteConnector` (and `NewBulkWriteConnector` and `NewMetadataConnector` also). Instead of each of these functions keeping a giant switch statement, they can call a helper function called `newConcreteConnector` or something after checking with the catalog that the provider supports that action. This means that when we add new deep connectors, we only have to update the `newConcreteConnector` function. 

Something like this:

```
func NewReadConnector(..) {
   info, err := providers.ReadInfo(provider, nil)
   if !info.Support.Read {
      return NotSuportedError
   }
   return newConcreteProvider(provider, ...)
}

func newConcreteProvider(provider, ...) { 
   proxyConn, err := NewProxyConnector(provider, opts...)
   switch provider {
    case provider.Hubspot:
         conn, outErr = hubspot.NewConnector(hubspot.WithProxyConnector(proxyConn))
    ....
   }
  }
```

connectors.go
eberle1080, line 98: I like the interfaces approach. But I don't like New<interface_name>.

Generally speaking (not always) interfaces are provided for those who *use* the interface, not for those who *construct* the object. For example:

Bad:

```go
type MyInterface interface {
    Foo()
}

type MyImpl struct {}

func (m *MyImpl) Foo() { }

func (m *MyImpl) Bar() { } // CAN'T use this

func NewMyInterface() MyInterface {
    return &MyImpl{}
}

func UseMyInterface(i MyInterface) {
    i.Foo()
}

func main() {
    inst := NewMyInterface()
    UseMyInterface(inst)

    inst.Bar() // won't work
}
```

Good:

```go
type MyInterface interface {
    Foo()
}

type MyImpl struct {}

func (m *MyImpl) Foo() { }

func (m *MyImpl) Bar() { } // CAN use this

func UseMyInterface(i MyInterface) {
    i.Foo()
}

func main() {
    inst := &MyImpl{}
    UseMyInterface(inst)

    inst.Bar() // works
}
```

The distinction may seem academic but it's actually pretty important. A single object (er, struct) can implement multiple interfaces. And the interfaces may or may not have overlap with one another. So if you use the concrete type (for construction), the object is allowed to bring the full set of capabilities with it, rather than being whittled down to a singular purpose.

If you're still confused I'd recommend this article which explains it better than I did: https://duncanleung.com/go-idiom-accept-interfaces-return-types/

Note that I only loosely follow this idiom when I code. However in this case, because this is so central to the connectors repo, I think it's important.

connectors.go
eberle1080, line 114: If all of the connectors are just going to accept a single param (the proxy connector) it might be easier to just have a single param rather than the "With<Feature>" variety. Partly for simplicity but also because then it's impossible to forget to include it.

hubspot/connector.go
```diff
@@ -1,57 +1,45 @@
...
-	}
+	return
+}
 
-	conn.Client.HTTPClient.ErrorHandler = conn.interpretError
+func WithProxyConnector(conn connectors.ProxyConnector) Option {
+	return func(connector *Connector) {
+		proxyConn, ok := conn.(*proxy.Connector)
+		if !ok {
+			return
```
eberle1080, line 40: Honestly, I'd panic here. There's code (the recover block) which will catch this panic and do the right thing. You don't want to silently fail.

hubspot/vars.go
```diff
@@ -0,0 +1,14 @@
...
+	// This section holds all the variables present in the hubspot catalog info.
+	// These can be referred to in the code to avoid hardcoding values.
+
+	// PlaceholderModule is the variable name used in the hubspot catalog info.
+	PlaceholderModule = "module"
+)
+
+const (
+	// DefaultPageSize is the default page size for paginated requests.
+	DefaultPageSize = "100"
```
eberle1080, line 13: Why isn't this an integer?

proxy/connector.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/amp-labs/connectors/providers"
+)
+
+// Connector is a Hubspot connector.
+type Connector struct {
+	Client *common.JSONHTTPClient
+
+	provider      providers.Provider
+	providerInfo  *providers.ProviderInfo
+	substitutions *map[string]string
```
eberle1080, line 14: This is only needed at construction time, right? Seems wasteful to keep it alive in memory if it'll only ever get used once.

connectors.go
eberle1080: To put it in more ampersand-specific terms, many of our deep connectors will implement at least ProxyConnector, ReadConnector, and WriteConnector (probably others too). I don't think it makes sense that I have to construct 3 separate objects (using the 3 New functions) when a single object (using its original concrete type) can still do all these jobs.

RajatPawar:
https://ampersand.slab.com/posts/connectors-interface-v-2-yle1ac4g

PR #115, opened 2024-03-20
[ENG-872] feat: Add Microsoft Dynamics 365 Sales connector

Cobalt0s:
## Checklist
- [x] Ran Linter

> WARN [runner/source_code] Failed to get line 391 for file utils/credentials.go: invalid file line index0 (390) >= len(fc) (92) 
WARN [runner/source_code] Failed to get line 402 for file utils/credentials.go: invalid file line index0 (401) >= len(fc) (92) 
::error file=utils/credentials.go,line=391,col=1::getFromReader returns generic interface (T) (ireturn)
::error file=utils/credentials.go,line=402,col=1::get returns generic interface (T) (ireturn)
make: *** [Makefile:14: fix] Error 1

- [x] Catalog tests passing

## Catalog variables
Workspace: `org5bd08fdd`
Scopes: `https://org5bd08fdd.crm.dynamics.com/user_impersonation,offline_access`
As of now substitution works for BaseURL. Here user impersonation scope itself contains workspace variable.

## Notes
Acquires AccessToken from Microsoft Entra ID that has preconfigured permissioning to access Sales App APIs making it a Sales Connector.

## Testing 
All Postaman requests are compared to the Web Portals View of Dynamics365Sales to demonstrate that both of them access the same data.

### GET 
URL: http://localhost:4444/
Top level discovery endpoint. Lists all entry sets for Sales CRM.
image

URL: http://localhost:4444/contacts?$select=fullname

List Contacts fullnames.
image
image

### POST
URL: http://localhost:4444/leads
Create new Lead. To verify that it was created lets view this information in CRM.
image
image

### GET
After lead creation filter GET response by firstname to get entry id, this will be used in update and delete step below. URL: http://localhost:4444/leads?$count=true&$filter=contains(firstname, 'Bob')
Now request `Single Entry`.
URL: http://localhost:4444/leads(dd2f7870-3fe8-ee11-a204-0022481f9e3c)?$select=firstname,lastname,subject
image

### PATCH
URL: http://localhost:4444/leads(dd2f7870-3fe8-ee11-a204-0022481f9e3c)
Update `Subject Name` of the lead.
image
image

### DELETE
URL: http://localhost:4444/leads(dd2f7870-3fe8-ee11-a204-0022481f9e3c)
Removes lead. After which entry is not found.
image
image

### GET
URL: http://localhost:4444/phonecalls?$count=true
List phone calls counting the number of entries.
image
image

### First page
URL: http://localhost:4444/phonecalls?$count=true&$select=subject
List phone calls, limit to max page size of 2 via `Prefer` header.
image

### Next page
URL: http://localhost:4444/phonecalls?$count=true&$select=subject&$skiptoken=%3Ccookie%20pagenumber=%222%22%20pagingcookie=%22%253ccookie%2520page%253d%25221%2522%253e%253cactivityid%2520last%253d%2522%257b3D1D5314-ED24-EB11-A814-000D3A30F257%257d%2522%2520first%253d%2522%257b1D1D5314-ED24-EB11-A814-000D3A30F257%257d%2522%2520%252f%253e%253c%252fcookie%253e%22%20istracking=%22False%22%20/%3E
Follow next page. We specify the same `maxpagesize` header. The link cannot be constructed or altered(as of MS docs), we are treating it as opaque string.
image

### DIFF
--- a/providers/catalog.go
+++ b/providers/catalog.go
@@ -5,29 +5,30 @@ package providers
 // ================================================================================
 
 const (
-	Salesforce        Provider = "salesforce"
-	Hubspot           Provider = "hubspot"
-	LinkedIn          Provider = "linkedIn"
-	Salesloft         Provider = "salesloft"
-	Outreach          Provider = "outreach"
-	Pipedrive         Provider = "pipedrive"
-	Copper            Provider = "copper"
-	ZohoCRM           Provider = "zohoCRM"
-	Sellsy            Provider = "sellsy"
-	Attio             Provider = "attio"
-	Close             Provider = "close"
-	Keap              Provider = "keap"
-	Asana             Provider = "asana"
-	Dropbox           Provider = "dropbox"
-	Notion            Provider = "notion"
-	Gong              Provider = "gong"
-	Zoom              Provider = "zoom"
-	Intercom          Provider = "intercom"
-	DocuSign          Provider = "docuSign"
-	DocuSignDeveloper Provider = "docuSignDeveloper"
-	Calendly          Provider = "calendly"
-	AWeber            Provider = "aWeber"
-	GetResponse       Provider = "getResponse"
+	Salesforce                Provider = "salesforce"
+	Hubspot                   Provider = "hubspot"
+	LinkedIn                  Provider = "linkedIn"
+	Salesloft                 Provider = "salesloft"
+	Outreach                  Provider = "outreach"
+	Pipedrive                 Provider = "pipedrive"
+	Copper                    Provider = "copper"
+	ZohoCRM                   Provider = "zohoCRM"
+	Sellsy                    Provider = "sellsy"
+	Attio                     Provider = "attio"
+	Close                     Provider = "close"
+	Keap                      Provider = "keap"
+	Asana                     Provider = "asana"
+	Dropbox                   Provider = "dropbox"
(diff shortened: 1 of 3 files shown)


Cobalt0s:
@RajatPawar it is ready for review :smiley:

providers/catalog.go
```diff
@@ -467,4 +468,23 @@ var catalog = CatalogType{ // nolint:gochecknoglobals
 			Write:     false,
 		},
 	},
+
+	// MS Sales configuration
+	MicrosoftDynamics365Sales: {
+		AuthType: Oauth2,
+		BaseURL:  "https://{{.workspace}}.api.crm.dynamics.com/api/data/v9.2",
```
RajatPawar, line 475: Can you please remove the suffix in this? We want users of this to be able to access all versions. I would keep the base URL as just `https://{{.workspace}}.api.crm.dynamics.com` - there might be other sections of the API that might be accessible via this

RajatPawar (changes requested):

providers/catalog.go
Cobalt0s: That makes sense. So when going over proxy the url would look like `http://localhost:444/api/data/v9.2` giving end user all control. I believe v9.2 uri part only changes in their API

RajatPawar (approved):

PR #116, opened 2024-03-20
fix: Use text/template rather than html/template

eberle1080:
We're using `html/template` for substitutions. This probably won't work as intended, since part of html templating includes making the contents HTML-safe, which (for things like raw URLs) we probably don't want.

This PR just switches over to `text/template` so that we don't get this unintended side-effect.

### DIFF
--- a/providers/utils.go
+++ b/providers/utils.go
@@ -6,7 +6,7 @@ import (
 	"errors"
 	"reflect"
 	"strings"
-	"text/template"
+	"text/template" // nosemgrep: go.lang.security.audit.xss.import-text-template.import-text-template
 
 	"github.com/go-playground/validator"
 )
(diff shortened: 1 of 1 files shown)


eberle1080:
> From https://semgrep.dev/docs/cheat-sheets/go-xss/#1a-using-the-texttemplate-package, I see the following which seems to be already the case?
> 
> > **Mitigation**
> > Ban text/template. Alternatively, use html/template, or a stricter alternative such as safehtml.

I don't follow. Semgrep had a problem because it assumed we were using the output for HTML. We're not. So I turned off that recommendation. It's good advice if you're dealing with HTML, but it's not like there's no use case for text/template. Their use of the word "ban" is pretty heavy-handed, and overly broad. Unless there's some deeper security problem than "it might output raw HTML directly to a browser client" (which it won't), I don't see a problem.

RajatPawar (approved):

PR #120, opened 2024-03-21
fix silent bug

jlimatampersand:
There was if statement that will never be executed because the condition is always false. It did not affect anything and didn't do any harm, but need to be cleaned up.

### DIFF
--- a/salesforce/bulkwrite.go
+++ b/salesforce/bulkwrite.go
@@ -374,10 +374,6 @@ func (c *Connector) getPartialFailureDetails(ctx context.Context, jobInfo *GetJo
 			failureMap = failInfo.FailedCreates
 		}
 
-		if failureMap == nil {
-			failureMap[errMsg] = []string{}
-		}
-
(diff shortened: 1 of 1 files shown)


salesforce/bulkwrite.go
```diff
@@ -374,10 +374,6 @@ func (c *Connector) getPartialFailureDetails(ctx context.Context, jobInfo *GetJo
 			failureMap = failInfo.FailedCreates
 		}
 
-		if failureMap == nil {
```
jlimatampersand, line 377: This `failureMap` is never `nil` because we set it up in line 367 as `failInfo.FailedUpdates` and `failInfo.FailedUpdates` is not `nil` according to assignemtn in line 330. 

`failure[errMsg] == nil` check is not necessary since `failureMap[errMsg] = append(failureMap[errMsg], externalId)` will append without resulting in nil pointer exception for `failureMap[errMsg]` is typed `nil`

laurenzlong (approved):
Good catch!

PR #121, opened 2024-03-22
[ENG-800] feat: Add ZohoCRM connector

Listener430:
## Catalog variables
No required catalog variables

### GET 
URL: <localhost:4444/crm/v6/Contacts?fields=Full_Name,Account_Name> 
Zoho - get proxy ok

### POST
URL: <localhost:4444/api/v2/unattended_computer/group> - fails if run via proxy
Zoho - post proxy fails
In case it is run directly, works fine - 
Zoho post direct works fine

### PUT
URL: <localhost:4444/api/v2/unattended_computer/group>
Zoho put proxy ok

### DELETE
URL: <localhost:4444/api/v2/unattended_computer/group> - fails
Zoho - del proxy fails
But if we hit the endpoint directly - everything works fine 
Zoho - del direct is ok

## Pagination
Zoho pagination sample

### DIFF
--- a/providers/catalog.go
+++ b/providers/catalog.go
@@ -189,11 +189,11 @@ var catalog = CatalogType{ // nolint:gochecknoglobals
 			ExplicitWorkspaceRequired: false,
 		},
 		Support: Support{
-			BulkWrite: true,
+			BulkWrite: false,
 			Proxy:     false,
-			Read:      true,
-			Subscribe: true,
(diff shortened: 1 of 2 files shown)


Listener430:
@RajatPawar hi, could you please kindly take a look when avaialable?

RajatPawar (approved):

PR #129, opened 2024-03-26
[ENG-878] feat: bulk delete

jlimatampersand:
Tested as below

```
jaehyunlim@Jaehyuns-MBP connectors % go run test/salesforce/bulkdelete/main.go
time=2024-03-26T16:03:05.571-07:00 level=INFO msg="Bulk write job created" res="&{State:UploadComplete JobId:750Dp0000090pptIAA}"
.time=2024-03-26T16:03:07.906-07:00 level=INFO msg="Bulk write job done"
time=2024-03-26T16:03:08.798-07:00 level=INFO msg="Bulk delete job created" res="&{State:UploadComplete JobId:750Dp0000090pcEIAQ}"
.time=2024-03-26T16:03:11.137-07:00 level=INFO msg="Bulk delete job done"
{
  "jobId": "750Dp0000090pcEIAQ",
  "state": "JobComplete",
  "jobInfo": {
    "id": "750Dp0000090pcEIAQ",
    "object": "Touchpoint__c",
    "createdById": "005Dp000003Cd1SIAS",
    "externalIdFieldName": "",
    "state": "JobComplete",
    "operation": "delete",
    "columnDelimiter": "COMMA",
    "lineEnding": "LF",
    "numberRecordsFailed": 0,
    "numberRecordsProcessed": 10,
    "apexProcessingTime": 0,
    "apiActiveProcessingTime": 178,
    "apiVersion": 59,
    "concurrencyMode": "Parallel",
    "contentType": "CSV",
    "createdDate": "2024-03-26T23:03:08.000+0000",
    "jobType": "V2Ingest",
    "retries": 0,
    "systemModstamp": "2024-03-26T23:03:10.000+0000",
    "totalProcessingTime": 239
  }
} <nil>
```

### DIFF
--- a/salesforce/bulk.go
+++ b/salesforce/bulk.go
@@ -67,25 +67,26 @@ type GetJobInfoResult struct {
 	Id                     string  `json:"id"`
 	Object                 string  `json:"object"`
 	CreatedById            string  `json:"createdById"`
-	ExternalIdFieldName    string  `json:"externalIdFieldName"`
+	ExternalIdFieldName    string  `json:"externalIdFieldName,omitempty"`
 	State                  string  `json:"state"`
 	Operation              string  `json:"operation"`
 	ColumnDelimiter        string  `json:"columnDelimiter"`
 	LineEnding             string  `json:"lineEnding"`
 	NumberRecordsFailed    float64 `json:"numberRecordsFailed"`
 	NumberRecordsProcessed float64 `json:"numberRecordsProcessed"`
-	ErrorMessage           string  `json:"errorMessage,omitempty"`
+	ErrorMessage           string  `json:"errorMessage"`
 
-	ApexProcessingTime      float64 `json:"apexProcessingTime"`
-	ApiActiveProcessingTime float64 `json:"apiActiveProcessingTime"`
-	ApiVersion              float64 `json:"apiVersion"`
-	ConcurrencyMode         string  `json:"concurrencyMode"`
-	ContentType             string  `json:"contentType"`
-	CreatedDate             string  `json:"createdDate"`
-	JobType                 string  `json:"jobType"`
-	Retries                 float64 `json:"retries"`
-	SystemModstamp          string  `json:"systemModstamp"`
-	TotalProcessingTime     float64 `json:"totalProcessingTime"`
+	ApexProcessingTime      float64 `json:"apexProcessingTime,omitempty"`
+	ApiActiveProcessingTime float64 `json:"apiActiveProcessingTime,omitempty"`
+	ApiVersion              float64 `json:"apiVersion,omitempty"`
+	ConcurrencyMode         string  `json:"concurrencyMode,omitempty"`
+	ContentType             string  `json:"contentType,omitempty"`
+	CreatedDate             string  `json:"createdDate,omitempty"`
+	JobType                 string  `json:"jobType,omitempty"`
+	Retries                 float64 `json:"retries,omitempty"`
+	SystemModstamp          string  `json:"systemModstamp,omitempty"`
+	TotalProcessingTime     float64 `json:"totalProcessingTime,omitempty"`
+	IsPkChunkingSupported   bool    `json:"isPkChunkingSupported,omitempty"`
 }
 
 type FailInfo struct {
@@ -490,4 +491,71 @@ func (c *Connector) bulkOperation(
 		JobId: updatedJobId,
 		State: completeState,
 	}, nil
+}
+
+func (c *Connector) BulkQuery(
+	ctx context.Context,
+	query string,
+) (*GetJobInfoResult, error) {
+	location, err := joinURLPath(c.BaseURL, "jobs/query")
+	if err != nil {
+		return nil, err
+	}
+
+	jobBody := map[string]any{
+		"operation": "query",
+		"query":     query,
+	}
+
+	res, err := c.post(ctx, location, jobBody)
+	if err != nil {
+		return nil, fmt.Errorf("bulk query failed: %w", err)
+	}
+
+	return common.UnmarshalJSON[GetJobInfoResult](res)
+}
+
+func (c *Connector) GetBulkQueryInfo(
+	ctx context.Context,
+	jobId string,
+) (*GetJobInfoResult, error) {
+	location, err := joinURLPath(c.BaseURL, "jobs/query", jobId)
+	if err != nil {
+		return nil, err
+	}
+
+	res, err := c.get(ctx, location)
+	if err != nil {
+		return nil, fmt.Errorf(
+			"failed to get bulk query info for job '%s': %w",
+			jobId,
+			errors.Join(err, common.ErrRequestFailed),
+		)
+	}
+
+	return common.UnmarshalJSON[GetJobInfoResult](res)
+}
+
+func (c *Connector) GetBulkQueryResults(
+	ctx context.Context,
+	jobId string,
+) (*http.Response, error) {
+	location, err := joinURLPath(c.BaseURL, fmt.Sprintf("jobs/query/%s/results", jobId))
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := common.MakeJSONGetRequest(ctx, location, []common.Header{
+		{
+			Key:   "Accept",
+			Value: "text/csv",
+		},
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get results for bulk query %s: %w", jobId, err)
+	}
+
+	// Get the connector's JSONHTTPClient, which is a special HTTPClient that handles JSON responses,
+	// and use it's underlying http.Client to make the request.
+	return c.Client.HTTPClient.Client.Do(req)
 }
--- a/salesforce/bulk.go
+++ b/salesforce/bulk.go
@@ -53,12 +53,7 @@ type BulkOperationParams struct {
 	Mode BulkWriteMode
 }
 
-type BulkWriteParams BulkOperationParams
-
 type BulkWriteMode string
-
-// BulkWriteResult is what's returned from writing data via the BulkWrite call.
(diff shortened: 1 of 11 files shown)


salesforce/bulkwrite.go
```diff
@@ -52,10 +52,15 @@ type BulkWriteParams struct {
 	Mode BulkWriteMode
 }
 
+type BulkWriteParams BulkJobParams
```
jlimatampersand, line 55: bulk write endpoint and bulk delete endpoints are the same, so I renamed the type struct. But for backward compatibility, we keep `BulkWriteParams` to not break existing usages.

salesforce/bulkwrite.go
```diff
@@ -52,10 +52,15 @@ type BulkWriteParams struct {
...
+
 type BulkWriteMode string
 
 // BulkWriteResult is what's returned from writing data via the BulkWrite call.
-type BulkWriteResult struct {
+type BulkWriteResult BulkJobResult
```
jlimatampersand, line 60: same as above, return result is the same for bulk write and bulk delete

salesforce/bulkwrite.go
```diff
@@ -118,87 +123,20 @@ func (c *Connector) BulkWrite( //nolint:funlen,cyclop
 		return nil, fmt.Errorf("%w: %s", ErrUnsupportedMode, config.Mode)
 	}
 
-	// cretes batch upload job, returns json with id and other info
```
jlimatampersand, line 121: The below code is moved to `BulkJob` so that we can use it in both bulk write and bulk delete with a few modifications.

jlimatampersand:
* **#129** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/129?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @jlimatampersand and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

salesforce/bulkwrite.go
```diff
@@ -430,3 +399,100 @@ func (c *Connector) getIncompleteJobResults(jobInfo *GetJobInfoResult) *JobResul
...
+	jobBody := map[string]any{
+		"operation": "delete",
+		"object":    params.ObjectName,
+	}
+
+	return c.BulkJob(ctx, params, jobBody)
+}
+
+//nolint:funlen,cyclop
+func (c *Connector) BulkJob(
```
jlimatampersand, line 413: rewritten from the previous `BulkWrite()` method because below can be reused for `BulkDelete` and any bulk job.

test/salesforce/bulkwrite/main.go
```diff
@@ -30,25 +27,9 @@ func main() { //nolint:funlen
 	logger := slog.New(handler)
 	slog.SetDefault(logger)
 
-	// assumes that this code is being run from the root of the project
```
jlimatampersand, line 33: coped to test/salesforce/utils/connector.go to reuse.

salesforce/bulkwrite.go
laurenzlong: You can keep it as BulkWrite, delete is technically a form of write. `BulkJob` is too ambiguous.

salesforce/bulkwrite.go
laurenzlong: same as above, let's keep it as BulkWriteResult

salesforce/bulkwrite.go
```diff
@@ -348,17 +295,27 @@ func (c *Connector) getPartialFailureDetails(ctx context.Context, jobInfo *GetJo
 			return nil, err
 		}
 
+		fieldNames := []string{sfIdFieldName, sfErrorFieldName}
+
+		if jobInfo.ExternalIdFieldName != "" {
```
laurenzlong, line 300: if ` jobInfo.ExternalIdFieldName` is an empty string, should we error out? When would that happen?

test/salesforce/bulkdelete/main.go
```diff
@@ -0,0 +1,211 @@
...
+		slog.Error("Error creating Salesforce connector", "error", err)
+
+		return
+	}
+
+	defer func() {
+		_ = sfc.Close()
+	}()
+
+	// create recrods to delete
```
laurenzlong, line 41:
```suggestion
	// create records to delete
```

salesforce/bulkwrite.go
jlimatampersand: In case of delete job, it is going to be empty. for that reason in line 335, I am using switch to get correct rows according to the operation.

jlimatampersand: But I think it will be more clear for me to add another if to check if it is delete operation

salesforce/bulkwrite.go
jlimatampersand: I think keeping it BulkWriteXX will confuse others. I think we can use BulkOperationXX. Salesforce distinguishes those jobs by field "operation"

jlimatampersand: like operation==delete, operation==upsert, etc.

salesforce/bulk.go
```diff
@@ -52,10 +53,15 @@ type BulkWriteParams struct {
 	Mode BulkWriteMode
 }
 
+type BulkWriteParams BulkOperationParams
```
jlimatampersand, line 56: @laurenzlong  I think using BulkWriteXX to all different operation is going to be confusing, ex) bulk query might also use the same parameter struct. I think we can call it `BulkOperationXX` instead. what do you think?

jlimatampersand: operations: delete, upsert, query, ...

salesforce/bulk.go
```diff
@@ -118,87 +124,20 @@ func (c *Connector) BulkWrite( //nolint:funlen,cyclop
 		return nil, fmt.Errorf("%w: %s", ErrUnsupportedMode, config.Mode)
 	}
 
-	// cretes batch upload job, returns json with id and other info
-	res, err := c.createJob(ctx, config)
```
jlimatampersand, line 122: this portion is reused for many operations, so moved to `BulkOperation` function

salesforce/bulkwrite.go
laurenzlong: yep that sounds good

salesforce/bulk.go
```diff
@@ -430,3 +399,100 @@ func (c *Connector) getIncompleteJobResults(jobInfo *GetJobInfoResult) *JobResul
...
+	jobBody := map[string]any{
+		"operation": "delete",
+		"object":    params.ObjectName,
+	}
+
+	return c.BulkOperation(ctx, params, jobBody)
+}
+
+//nolint:funlen,cyclop
+func (c *Connector) BulkOperation(
```
laurenzlong, line 413: This should not be exported, this should just be an internal method (not capitalized)

test/salesforce/bulkdelete/main.go
```diff
@@ -0,0 +1,240 @@
...
+		slog.Error("Error creating Salesforce connector", "error", err)
+
+		return
+	}
+
+	defer func() {
+		_ = sfc.Close()
+	}()
+
+	// Prepare objects and write to Salesforce to be deleted
```
laurenzlong, line 41: Can you revise this section of comments to:
"We first create objects in Salesforce, and then we generate an in-memory CSV of the Salesforce IDs of the newly created objects, so that we can bulk-delete them."

test/salesforce/bulkdelete/main.go
```diff
@@ -0,0 +1,240 @@
...
+
+	defer func() {
+		if fileToWrite != nil {
+			if closeErr := fileToWrite.Close(); closeErr != nil {
+				slog.Warn("unable to close file", "error", closeErr)
+			}
+		}
+	}()
+
+	// Write the file to Salesforce
```
laurenzlong, line 144:
```suggestion
	// Write the records to Salesforce, so that we can delete them later.
```

test/salesforce/bulkdelete/main.go
```diff
@@ -0,0 +1,240 @@
...
+
+	slog.Info("Preparing objects to delete", "res", writeRes)
+
+	// wait for the job to complete
+	_, err = getResultInLoop(ctx, sfc, writeRes.JobId)
+	if err != nil {
+		return nil, fmt.Errorf("error getting bulk write job results: %w", err)
+	}
+
+	slog.Info("Preparation finished")
```
laurenzlong, line 163:
```suggestion
	slog.Info("Records created, now deleting them.")
```

test/salesforce/bulkdelete/main.go
```diff
@@ -0,0 +1,240 @@
...
+			}
+		}
+	}()
+
+	headers, rows, err := csvBytesToSlice(successBody)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing CSV: %w", err)
+	}
+
+	// remove all columns except the id	column
```
laurenzlong, line 193:
```suggestion
	// remove all columns except the id column
```

salesforce/bulk.go
laurenzlong: name sounds good, no need to alias it as discussed.

salesforce/bulk.go
```diff
@@ -52,10 +53,15 @@ type BulkWriteParams struct {
 	Mode BulkWriteMode
 }
 
+type BulkWriteParams BulkOperationParams
+
 type BulkWriteMode string
 
 // BulkWriteResult is what's returned from writing data via the BulkWrite call.
-type BulkWriteResult struct {
+type BulkWriteResult BulkOperationResult
```
laurenzlong, line 61: Same here, no need to alias it, we can just start using BulkOperationResult

laurenzlong (approved):

PR #138, opened 2024-03-29
[ENG-878] Add delete mode in bulk operation

jlimatampersand:
adding "delete" mode and some linting and clean up

### DIFF
--- a/salesforce/bulk.go
+++ b/salesforce/bulk.go
@@ -17,6 +17,7 @@ const (
 	Insert BulkWriteMode = "insert"
 	Upsert BulkWriteMode = "upsert"
 	Update BulkWriteMode = "update"
+	Delete BulkWriteMode = "delete"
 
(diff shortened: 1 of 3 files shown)


salesforce/bulk.go
```diff
@@ -109,7 +110,7 @@ func (c *Connector) BulkWrite( //nolint:funlen,cyclop
 	config BulkOperationParams,
 ) (*BulkOperationResult, error) {
 	// Only support upsert for now
-	switch config.Mode {
+	switch config.Mode { //nolint:exhaustive
```
laurenzlong, line 113: Instead of a switch statement, this would be better written as an if statement:
```
if config.Mode != Upsert {
 return nil, fmt.Errorf("%w: %s", ErrUnsupportedMode, config.Mode)
}
```

laurenzlong (approved):

eberle1080 (approved):

PR #157, opened 2024-04-03
feat: Deep connector MS Dynamics365 Sales (Read)

Cobalt0s:
Closes https://github.com/amp-labs/connectors/issues/186


common/interpreter/error.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+func (h ErrorHandler) Handle(res *http.Response, body []byte) error {
+	mediaType, _, err := mime.ParseMediaType(res.Header.Get("Content-Type"))
+	if err != nil {
+		return fmt.Errorf("mime.ParseMediaType failed: %w", err)
+	}
+
+	if h.JSON != nil && mediaType == "application/json" {
+		return h.JSON(res, body)
```
Cobalt0s, line 30: Depending on a ContentType a connector can parse error response in JSON, XML, HTML whatever impl we want to provide.
Every connector understands JSON structure in their own way. Other media types could be added.

common/json.go
```diff
@@ -76,16 +71,24 @@ func (j *JSONHTTPClient) Patch(ctx context.Context,
...
 
 	return parseJSONResponse(res, body)
 }
 
 // parseJSONResponse parses the given HTTP response and returns a JSONHTTPResponse.
 func parseJSONResponse(res *http.Response, body []byte) (*JSONHTTPResponse, error) {
+	// empty response body should not be parsed as JSON since it will cause ajson to err
+	if len(body) == 0 {
+		return nil, nil //nolint:nilnil
+	}
```
Cobalt0s, line 91: it looks like every CRUD operation does this check (except GET, could be a bug), pulled it to this method since all of them call it at the end.

common/paramsbuilder/params.go
```diff
@@ -0,0 +1,104 @@
...
+	"net/http"
+
+	"github.com/amp-labs/connectors/common"
+	"golang.org/x/oauth2"
+)
+
+// ParamAssurance checks that param data is valid
+// Every param instance must implement it.
+type ParamAssurance interface {
+	ValidateParams() error
```
Cobalt0s, line 16: There are several parameters with single responsibilities: Client/Workspace/Module. There could be more of these shared parameters.
Every parameter must determine if its value is required/optional by implementing `ParamAssurance`.

common/parse.go
```diff
@@ -73,3 +75,100 @@ func ExtractLowercaseFieldsFromRaw(fields []string, record map[string]interface{
...
+var (
+	ErrNotArray   = errors.New("results is not an array")
+	ErrNotObject  = errors.New("result is not an object")
+	ErrNotString  = errors.New("link is not a string")
+	ErrNotNumeric = errors.New("value is not numeric")
+	ErrNotInteger = errors.New("value is not integer")
+
+	// JSONManager is a helpful wrapper of ajson library that adds errors when querying JSON payload
+	// and provides common conversion methods.
+	JSONManager = jsonManager{}
```
Cobalt0s, line 88: Querying JSON content deals with finding "keys" and validating expected data type of the "value".
ErrNotArray, ErrNotObject errors can be universally shared for all JSON parsers

common/reqrepeater/standard.go
```diff
@@ -0,0 +1,50 @@
...
+	"errors"
+	"time"
+)
+
+var ErrRetry = errors.New("try again later")
+
+// Strategy is configuration of how a connector handles faulty requests
+// connector will create Retry based on this config every time it attempts to make API call.
+type Strategy interface {
+	Start() Retry
```
Cobalt0s, line 13: Connector can be in future configured on how request retries are performed.
For now there is only one `UniformRetryStrategy` and is hard coded for MS Sales with 3 retries every 1 sec.

common/types.go
```diff
@@ -82,6 +82,8 @@ type ReadParams struct {
 	Fields []string // required, at least one field needed
 	// NextPage is an opaque token that can be used to get the next page of results.
 	NextPage string // optional, only set this if you want to read the next page of results
+	// PageSize number of elements per page
+	PageSize int64 // optional, desired page size
```
Cobalt0s, line 86: I think client should have control over page size.

msdsales/errors.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+type EnhancedSalesError struct {
+	HelpLink     string `json:"@Microsoft.PowerApps.CDS.HelpLink"`
+	TraceText    string `json:"@Microsoft.PowerApps.CDS.TraceText"`
+	InnerMessage string `json:"@Microsoft.PowerApps.CDS.InnerError.Message"`
+}
+
+func (s SalesResponseError) Error() string {
+	return s.Err.Message
```
Cobalt0s, line 63: Error response has so much more. In existing connectors we return a tiny bit. Same strategy used here.
Should we care about other fields? Ex: HelpLink

msdsales/methods.go
```diff
@@ -0,0 +1,129 @@
...
+	"context"
+	"errors"
+	"log/slog"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/reqrepeater"
+)
+
+// FIXME these methods look repetitive
+// FIXME arguments: Retry Strategy + HTTP Client.
```
Cobalt0s, line 13: Every CRUD method is identical to other connectors. This could be refactored/generated.

msdsales/params.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+// Option is a function which mutates the connector configuration.
+type Option func(params *parameters)
+
+// parameters MS Sales supports auth client, workspace, etc. by delegation.
+type parameters struct {
+	paramsbuilder.Client
+	paramsbuilder.Workspace
+	paramsbuilder.Module
```
Cobalt0s, line 25: MS Sales allows configuration via Options Pattern just like other connectors.
Other connectors would compose paramteres like this by delegation.
Reason: If underlying Client parameter changes it will be applied to all parameters that delegate to it.

msdsales/params.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func (p parameters) FromOptions(opts ...Option) (*parameters, error) {
+	params := &p
+	for _, opt := range opts {
+		opt(params)
+	}
+
+	return params, params.ValidateParams()
```
Cobalt0s, line 34: This is a construcotr of connector parameters from `Option functions` with required field validation

msdsales/parse.go
```diff
@@ -0,0 +1,40 @@
...
+func getNextRecordsURL(node *ajson.Node) (string, error) {
+	return common.JSONManager.GetString(node, "@odata.nextLink")
+}
+
+// FIXME we must differentiate between GET and LIST (it is LIST now)
+// sometimes the fields that user requests are either
+// * singular record
+// * list of records
+// * hybrid, list of records with extra fields describing list.
+func getMarshaledData(records []map[string]interface{}, fields []string) ([]common.ReadResultRow, error) {
```
Cobalt0s, line 30: Can we marshal singular record in case of 
`GET domain/cats/<catID>` vs `GET domain/cats`
Also LIST of cats could return additional parameters stored outside of array, How those fields are returned?

msdsales/read.go
```diff
@@ -0,0 +1,95 @@
...
+// $select = ReadParams{fields} [COMPLETED]
+// $expand = nested response
+// $orderby = list of fields with asc/desc keyword
+// TODO batch
+// $apply = batch
+// TODO search
+// $filter = query functions, comparisons
+// TODO pagination
+// $top = <int> of entries to return (ignored if header <Prefer: odata.maxpagesize>)
+// $count = counts all existing rows (@odata.count)
```
Cobalt0s, line 24: Microsoft's Read is very flexible, see [options](https://learn.microsoft.com/en-us/power-apps/developer/data-platform/webapi/query-data-web-api#odata-query-options). Should we support all query capabilities?

msdsales/read.go
```diff
@@ -0,0 +1,95 @@
...
+	if len(config.NextPage) == 0 {
+		// First page
+		relativeURL := config.ObjectName + makeQueryValues(config)
+		fullURL = c.getURL(relativeURL)
+	} else {
+		// Next page
+		fullURL = config.NextPage
+	}
+	// TODO given that one of the fields is annotation we can automatically add annotation header (how the hell the end user gonna know about the names of those fields)
+	rsp, err := c.get(ctx, fullURL, newPaginationHeader(resolvePageSize(config)), annotationsHeader)
```
Cobalt0s, line 50: Response can be more descriptive if annotation header is provided.
Right now it is always passed. How much control will we give? It relates to `$expand` query for descriptive responses.

test/msdsales/paginate/main.go
```diff
@@ -0,0 +1,70 @@
...
+		utils.Fail("there was no second page cursor")
+	}
+
+	if res.Rows != customPageSize {
+		utils.Fail(
+			fmt.Sprintf("first page has record count mismatch, given %v, expected %v", res.Rows, customPageSize))
+	}
+
+	res, err = conn.Read(ctx, common.ReadParams{
+		NextPage: string(res.NextPage), // TODO should Read params have NextPageToken type?
```
Cobalt0s, line 58: ReadParams.NextPage(string) `<- doesn't match type with ->` ReadResult.NextPage(NextPageToken)

Cobalt0s (commented):
@RajatPawar, here is an initial explanation and questions to existing/future code changes

common/reqrepeater/standard.go
Cobalt0s: https://www.loom.com/share/39f1d4ea8dc446148eeff83a3f1c8681

msdsales/params.go
Cobalt0s: https://www.loom.com/share/97b52ea5eeab449cb0dd00abc2842785

msdsales/read.go
Cobalt0s: https://www.loom.com/share/5b109408daf9416aa07ac9427e13e228

Cobalt0s: https://www.loom.com/share/5b109408daf9416aa07ac9427e13e228

common/paramsbuilder/params.go
laurenzlong: This is a very cool package, thanks for introducing it!

msdsales/errors_test.go
```diff
@@ -0,0 +1,91 @@
...
+import (
+	"errors"
+	"net/http"
+	"testing"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/interpreter"
+)
+
+func Test_interpretJSONError(t *testing.T) {
```
laurenzlong, line 12: We generally name tests without underscores, so this would be: TestInterpretJSONError

msdsales/methods.go
laurenzlong: Can you replace `FIXME` with `TODO`? then it's easier for us to do to a search for TODOs across the codebase at a later point.

msdsales/read.go
Cobalt0s: Yes, bottom line is I have described these options in docs.

msdsales/read.go
```diff
@@ -0,0 +1,100 @@
...
+	if len(config.NextPage) == 0 {
+		// First page
+		relativeURL := config.ObjectName + makeQueryValues(config)
+		fullURL = c.getURL(relativeURL)
+	} else {
+		// Next page
+		fullURL = config.NextPage
+	}
+	// TODO given that one of the fields is annotation we can automatically add annotation header
+	// (how the hell the end user gonna know about the names of those fields)
```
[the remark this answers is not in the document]
Cobalt0s: This was relevant in the context of annotations and quering them to be included under `fields` not just `raw`.
It is not relevant anymore.

msdsales/read_test.go
```diff
@@ -0,0 +1,292 @@
+package msdsales
```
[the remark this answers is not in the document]
Cobalt0s: Glad you like it. I tried to use realistic data for these kind of tests. More of them are in other PRs for other MS Sales methods.

msdsales/read_test.go
```diff
@@ -0,0 +1,292 @@
...
+		{
+			name: "Incorrect key in payload",
+			server: httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				w.Header().Set("Content-Type", "application/json")
+				w.WriteHeader(http.StatusOK)
+				writeBody(w, `{
+					"garbage": {}
+				}`)
+			})),
+			expectedErrs: []error{errors.New("wrong request: wrong key 'value'")},
```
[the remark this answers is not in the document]
Cobalt0s: Yes, this is a smell code. There was another test where I used type comparison. To be more general I added reflection check. For this test checking that error comes from `ajson` lib should be enough.
```
expectedErrs - used when you know exact error
expectedErrTypes - used less frequently and relies on type rather than content
```

test/msdsales/paginate/main.go
Cobalt0s: Changed.
In places where it was expected to use as string instead of direct conversion I added a method. Let me know if it is not ok.
```
type NextPageToken string

func (t NextPageToken) String() string {
	return string(t)
}
```

test/msdsales/read/main.go
```diff
@@ -0,0 +1,48 @@
...
+		filePath = "./ms-sales-creds.json"
+	}
+
+	conn := msTest.GetMSDynamics365SalesConnector(ctx, filePath)
+	defer utils.Close(conn)
+
+	res, err := conn.Read(ctx, common.ReadParams{
+		ObjectName: "contacts",
+		Fields: []string{
+			"fullname", "emailaddress1", "fax", "familystatuscode",
```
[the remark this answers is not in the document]
Cobalt0s: I have removed it. 
But I am not sure what do you mean that it is not a real field? MS Sales returns it. Is it because of the enum data type?

msdsales/read.go
```diff
@@ -3,38 +3,19 @@ package msdsales
...
-// $orderby = list of fields with asc/desc keyword
-// TODO batch
-// $apply = batch
-// TODO search
-// $filter = query functions, comparisons
-// TODO pagination
-// $top = <int> of entries to return (ignored if header <Prefer: odata.maxpagesize>)
-// $count = counts all existing rows (@odata.count)
-
 var annotationsHeader = common.Header{
```
laurenzlong: Can you add a comment that adding this annotations header to the request will cause MS to return all available odata annotations in the response? (Just so someone reading this code in the future will understand the purpose of this header)

laurenzlong (approved):
Thanks so much!

PR #159, opened 2024-04-03
feat: Add oauth grant types to catalog

RajatPawar:
We currently support only OAuth 2.0 Authorization code flow (& refresh token flow), but there's [multiple types of grants possible with OAuth 2.0](https://oauth.net/2/grant-types/). 

Some providers that we are working on are using these flows, so it might be a good idea to specify the grant types.


eberle1080 (approved):

laurenzlong (approved):
Great idea!

PR #160, opened 2024-04-03
feat: Make bulk write support more granular

eberle1080:
Certain providers do support bulk writes, but not for all modes. This makes validation difficult from the backend. This PR breaks out the modes in to their own fields so that we can more accurately model what is and isn't allowed.


laurenzlong (approved):
Can you also add `Delete` as a key to `BulkWriteSupport`? Also @RajatPawar what do you think of this?

PR #173, opened 2024-04-05
draft: Representing provider groups

RajatPawar:


providers/catalog.go
```diff
@@ -888,4 +867,36 @@ var catalog = CatalogType{ // nolint:gochecknoglobals
 			Write:     false,
 		},
 	},
+
```
RajatPawar, line 870: We could create a separate catalog for provider groups too, but then we'd have to think about provider <=> providerGroup mapping, and how we would be able to make out what is a provider and what is a providerGroup

PR #179, opened 2024-04-08
feat: Deep connector MS Dynamics365 CRM (ListMetadata) 

Cobalt0s:
Closes https://github.com/amp-labs/connectors/issues/196


microsoftdynamicscrm/metadata_test.go
```diff
@@ -0,0 +1,217 @@
...
+				Errors: nil,
+			},
+			expectedErrs: nil,
+		},
+		{
+			// In total phonecall will have 65 fields, where
+			// phonecall 		(has 7 fields) and inherits from
+			// activitypointer 	(has 58 fields), which in turn inherits from
+			// crmbaseentity 	(has 0 fields)
+			name:  "Correctly list metadata for phone calls including inherited fields",
```
Cobalt0s, line 136: Real schemas are huge. Even the small one selected for the test will make response check cumbersome. For this case we just check that total number of fields are indeed inherited and match expectation.

test/microsoftdynamicscrm/metadata/README.md
```diff
@@ -0,0 +1,14 @@
...
+Listing object metadata for MS Dynamics 365 Sales will give huge XML document describing schema for entire service.
+A portion is propagated based on the list of desired Objects. This script is concerned about "account" structure.
+In order to validate that metadata is extracted as expected we compare against the response from connector's READ operation.
+
+The test ensures the following rules:
+* Every field returned by `GET ~/accounts` must be in schema description (except annotations)
+* No schema property is omitted
+
+If you would like to validate for other objects, here where you can find more information.
+Under this repo `sortedStructNamesFromSchema.txt` file lists all objects that are available.
```
Cobalt0s, line 12: This file has entity names for the whole CRM covering:
* Microsoft Dynamics 365 Sales
* Microsoft Dynamics 365 Customer Service

test/microsoftdynamicscrm/metadata/main.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+var (
+	objectName       = "account"
+	objectNamePlural = "accounts"
+)
+
+// we want to compare fields returned by read and schema properties provided by metadata methods
+// they must match for all such objects
+func main() {
```
Cobalt0s, line 25: Algorithm:
* Select object (ex: account)
* Ask for Metadata
* Query endpoint via READ
* Response filed must match schema acquired from Metadata

providers/catalog.go
```diff
@@ -712,8 +712,8 @@ var catalog = CatalogType{ // nolint:gochecknoglobals
 				Upsert: false,
 				Delete: false,
 			},
-			Proxy:     false,
-			Read:      false,
+			Proxy:     true,
```
laurenzlong, line 715: can we revert `proxy` and `read` back to `false` until we've had a chance to do more thorough integration testing with the server?

Cobalt0s: Removed. I thought that switching those on was a required task but something I had missed for read PR

common/xml.go
```diff
@@ -18,11 +23,93 @@ const (
 var (
 	//nolint:gochecknoglobals
 	validate          = validator.New()
+	ErrNoXMLRoot      = errors.New("xml document has no root")
 	ErrNotXMLChildren = errors.New("children must be of type 'XMLData' or 'XMLString'")
 	ErrNoSelfClosing  = errors.New("selfClosing cannot be true if children are not present")
 	ErrNoParens       = errors.New("value cannot contain < or >")
 )
 
+// XMLHTTPClient speaks from http client in XML.
```
laurenzlong, line 32: This comment line is a bit awkward sounding, do you mean something like "XMLHTTPClient is an HTTP client that can parse XML responses"?

microsoftdynamicscrm/metadata.go
```diff
@@ -2,13 +2,125 @@ package microsoftdynamicscrm
...
+	root, err := rsp.GetRoot()
+	if err != nil {
+		return nil, err
+	}
+
+	entities, err := extractEntities(root)
+	if err != nil {
+		return nil, err
+	}
+	// FIXME a call to the metadata endpoint is costly, once queried and processed we should Cache entities
```
laurenzlong, line 42: Can you remove this FIXME? this is a concern of the server, this library doesn't do any caching.

microsoftdynamicscrm/metadata_test.go
```diff
@@ -0,0 +1,217 @@
...
+				w.Header().Set("Content-Type", "application/xml")
+				w.WriteHeader(http.StatusOK)
+				_, _ = w.Write(fakeServerResp)
+			})),
+			expected: &common.ListObjectMetadataResult{
+				Result: map[string]common.ObjectMetadata{
+					"accountleads": {
+						DisplayName: "accountleads",
+						FieldsMap: map[string]string{
+							"accountid":                 "accountid",
```
laurenzlong, line 106: So Microsoft Dynamics doesn't return the display name for objects or fields? What I mean is: for the object `accountleads`, does the API not return something like `Account Leads`?

Cobalt0s: Yes, indeed I couldn't find the human readable forms in either Endpoint response nor in Schema response.
https://github.com/Cobalt0s/connectors/blob/1cbb441a9b93a82ba1b7280a471b895e49d1151b/microsoftdynamicscrm/test/metadata.xml#L14-L37
Some screenshot I took earlier. All responses look quite "unnatural" for display.
image

laurenzlong: that's unfortunate 😢

laurenzlong (approved):

PR #193, opened 2024-04-09
Add Outreach read connector 

Jkarage:
This PR adds the Outreach read capability.

- Updated the jsonResponse to allow application/vnd+.json. All providers implementing JSONAPISpecification will have this as the Content-Type in the response.

Written test to read an Object from Outreach api. The test reads users object from the outreach API.

### DIFF
--- a/common/json.go
+++ b/common/json.go
@@ -97,6 +97,7 @@ func parseJSONResponse(res *http.Response, body []byte) (*JSONHTTPResponse, erro
 			return nil, fmt.Errorf("failed to parse content type: %w", err)
 		}
 
+		// Providers implementing JSONAPISpeicifcations returns application/vnd.api+json
 		if mimeType != "application/json" && mimeType != "application/vnd.api+json" {
 			return nil, fmt.Errorf("%w: expected content type to be application/json or application/vnd.api+json, got %s",
 				ErrNotJSON, mimeType,
--- a/common/json.go
+++ b/common/json.go
@@ -97,8 +97,10 @@ func parseJSONResponse(res *http.Response, body []byte) (*JSONHTTPResponse, erro
 			return nil, fmt.Errorf("failed to parse content type: %w", err)
 		}
 
-		if mimeType != "application/json" {
-			return nil, fmt.Errorf("%w: expected content type to be application/json, got %s", ErrNotJSON, mimeType)
+		if mimeType != "application/json" && mimeType != "application/vnd.api+json" {
+			return nil, fmt.Errorf("%w: expected content type to be application/json or application/vnd.api+json, got %s",
+				ErrNotJSON, mimeType,
+			)
 		}
 	}
 
--- a/outreach/connector.go
+++ b/outreach/connector.go
@@ -1,0 +1,61 @@
+package outreach
+
+import (
+	"fmt"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/providers"
+)
+
+const (
+	providerOptionRestApiURL = "restAPIURL"
+)
+
+type Connector struct {
+	BaseURL string
+	Client  *common.JSONHTTPClient
+}
+
+func NewConnector(opts ...Option) (conn *Connector, outErr error) {
+	defer func() {
+		if re := recover(); re != nil {
+			tmp, ok := re.(error)
+			if !ok {
+				panic(re)
+			}
+
+			outErr = tmp
+			conn = nil
+		}
+	}()
+
+	params := &outreachParams{}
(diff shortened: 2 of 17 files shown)


outreach/metadata.go
```diff
@@ -0,0 +1,16 @@
+package outreach
```
RajatPawar, line 1: This can be removed because of https://github.com/amp-labs/connectors/pull/195

outreach/string.go
```diff
@@ -0,0 +1,7 @@
+package outreach
+
+import "fmt"
+
+func (c *Connector) String() string {
```
RajatPawar, line 5: As the linter says, this can be replaced with string concat

outreach/write.go
```diff
@@ -0,0 +1,13 @@
+package outreach
```
RajatPawar, line 1: Same as the metadata function, can be removed because of https://github.com/amp-labs/connectors/pull/195

common/types.go
```diff
@@ -86,8 +86,34 @@ type ReadParams struct {
 	Since time.Time // optional, omit this to fetch all records
 	// Deleted is true if we want to read deleted records instead of active records.
 	Deleted bool // optional, defaults to false
+	// The size of the returned records per page
```
RajatPawar, line 89: Still thinking about this as we need to be cognizant of other providers while doing this cc (filters look different from provider to provider, for example) @laurenzlong @eberle1080

common/json.go
```diff
@@ -94,7 +94,8 @@ func parseJSONResponse(res *http.Response, body []byte) (*JSONHTTPResponse, erro
 			return nil, fmt.Errorf("failed to parse content type: %w", err)
 		}
 
-		if mimeType != "application/json" {
+		// Providers implementing JSONAPISpeicifcations returns application/vnd.api+json
+		if mimeType != "application/json" && mimeType != "application/vnd.api+json" {
 			return nil, fmt.Errorf("%w: expected content type to be application/json, got %s", ErrNotJSON, mimeType)
```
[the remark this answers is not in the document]
Jkarage: Ok.

outreach/connector.go
```diff
@@ -0,0 +1,58 @@
...
+	"strings"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/providers"
+)
+
+const (
+	// we need to change the BaseURL
+	// for this to be v2.
+	apiVersion = "api/v2"
```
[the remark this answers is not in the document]
Jkarage: Noted.

outreach/metadata.go
Jkarage: Ok

common/types.go
laurenzlong: While these are really good ideas, generally we don't introduce new capabilities to connectors unless we are confident we can support it across **all** connectors now and in the future. While Outreach has support for custom page sizes, filters and sorting, a lot of other APIs won't, and it would be really unexpected behavior if these fields of ReadParams sometimes works and sometimes does nothing. While there are things we can do to "fake it" for providers that don't support these capabilities natively, it adds complexity and we would be committing to taking on that work for future connectors. So can you please remove this PageSize, FilterBy and SortBy, and remove the relevant code? Sorry to make you take out code that you've already written but we just have to be really thoughtful about future maintenance of this project.

Jkarage: Noted Lauren

outreach/errors.go
```diff
@@ -0,0 +1,31 @@
...
+	case errors.Is(err, common.ErrRetryable):
+		return err
+	case errors.Is(err, common.ErrApiDisabled):
+		fallthrough
+	case errors.Is(err, common.ErrForbidden):
+		fallthrough
+	default:
+		// Anything else is a permanent error
+		return err
+	}
```
eberle1080, line 30: This could be collapsed down to `return err`

test/outreach/read/read.go
```diff
@@ -0,0 +1,85 @@
...
+	)
+	if err != nil {
+		testUtils.Fail("error creating outreach connector", "error", err)
+	}
+
+	return conn
+}
+
+func main() {
+	outreach := GetOutreachConnector(context.TODO(), DefaultCredsFile)
```
eberle1080, line 66: given that this is in the main function, you can just use context.Background. Same with line 72.

laurenzlong (approved):
Once you resolve the merge conflicts, you can "squash and merge". Thanks!

Jkarage:
Moving this to #232, rebased and resolved the conflicts.

PR #216, opened 2024-04-15
feat: Add a mock connector

eberle1080:
Certain tests are difficult to run against real providers. A prime example of that is a rate-limiting test, since we simply don't have enough actual data (in our test instance of salesforce, for example) to reliably trigger a 429 code. Another example is a load test, since we want to test the load tolerance of *our* system, not the load tolerance of the actual providers.

This PR introduces a mock connector, whose behavior can be dynamically configured via callback functions. It strives to be minimal, and leave the mock details up to the consumer.

This would be useful both in unit tests, as well as end-to-end tests.

### DIFF
--- a/providers/catalog.go
+++ b/providers/catalog.go
@@ -52,7 +52,7 @@ const (
 var catalog = CatalogType{ // nolint:gochecknoglobals
 	Mock: {
 		AuthType: None,
-		BaseURL:  "https://mock.com",
+		BaseURL:  "https://not-a-real-domain.mock",
 		Support: Support{
 			BulkWrite: BulkWriteSupport{
 				Insert: true,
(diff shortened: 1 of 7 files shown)


providers/catalog.go
```diff
@@ -49,6 +50,26 @@ const (
 // ================================================================================
 
 var catalog = CatalogType{ // nolint:gochecknoglobals
+	Mock: {
+		AuthType: None,
+		BaseURL:  "https://not-a-real-domain.mock",
+		Support: Support{
+			BulkWrite: BulkWriteSupport{
```
laurenzlong, line 57: doesn't look like you've actually implemented the BulkWrite and BulkDelete methods for the mock connector, so can you set these to `false` instead? (Lines 58 to 61)

laurenzlong (approved):
Great idea!

providers/catalog.go
eberle1080: ah right, I didn't see a dedicated interface for that so I wasn't sure how to approach it. Yeah I'll set to false.

PR #231, opened 2024-04-17
fix: Update catalog entry for DocuSign to be more accurate

laurenzlong:


RajatPawar (approved):
The catalog test is failing since you've changed the config, so please update that file before merging it in. It might also be good to just comment docusign out given the server/region collection issue. Also, docusign requires different oauth apps for these. Well not really, but you promote a dev app into prod, which changes the client ID

PR #232, opened 2024-04-18
feat: Add outreach read connector

Jkarage:
This PR adds the Outreach read capability.

- Updated the jsonResponse to allow application/vnd+.json. All providers implementing JSONAPISpecification will have this as the Content-Type in the response.

Written test to read an Object from Outreach api. The test reads users object from the outreach API.

draft PR for #187


RajatPawar (approved):
Approving this as it is a duplicate of #193

PR #246, opened 2024-04-23
feat: Add Typeform auth connector

### LINKED ISSUE #242: Add Typeform auth connector

Cobalt0s:
Closes #242 

## Checklist
- [x] Ran Linter
- [x] Catalog tests passing

## Token Response
```
{
  "access_token":"....",
  "expires_in":604800,
  "refresh_token":"...",
  "state":"",
  "token_type":"bearer"
}
```

## Catalog variables
We are using default data center by omitting it. We couldv'e specfied `eu.`
```
  "substitutions": {
    "dataCenter" : ""
  }
```

### GET
URL: http://localhost:4444/forms
List forms.
image

### POST
URL: http://localhost:4444/forms
image

### PATCH
URL: http://localhost:4444/forms/BvXy6fa1
Changed form tittle. We are giving an array of changes to apply.
image

### PUT
URL: http://localhost:4444/forms/BvXy6fa1
Same as Patch but the whole object is published.
image

### DELETE
URL: http://localhost:4444/forms/BvXy6fa1
Remove form.
image

### First page
URL: http://localhost:4444/forms?page=1&page_size=3
image

### Second Page
URL: http://localhost:4444/forms?page=2&page_size=3
image

### DIFF
--- a/providers/catalog.go
+++ b/providers/catalog.go
@@ -55,6 +55,7 @@ const (
 	GoogleMail                          Provider = "googleMail"
 	Monday                              Provider = "monday"
 	Miro                                Provider = "miro"
+	Typeform                            Provider = "typeform"
 )
 
 // ================================================================================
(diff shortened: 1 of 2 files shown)


providers/catalog.go
```diff
@@ -1215,4 +1216,28 @@
 			Write:     false,
 		},
 	},
+
+	Typeform: {
+		AuthType: Oauth2,
+		BaseURL:  "https://api.{{.dataCenter}}typeform.com",
```
RajatPawar, line 1222: Can you please update this to `api.typeform.com`, and add a separate `TypeformEU` connector?

RajatPawar (changes requested):
This is region based, so splitting up into two connectors cc @laurenzlong

providers/catalog.go
Cobalt0s: Added

laurenzlong:
@Cobalt0s @RajatPawar Sorry can we actually remove the EU connector and just support US for now (no substitution variables, just api.typeform.com). I read up on this more and it seems like:
- EU data hosting[ is only available on the Enterprise plan for Typeform](https://www.typeform.com/help/a/what-happens-to-my-data-360029581691/).  (which means not a lot of companies will have this, because enterprise plans are expensive)
- we'll eventually want to support both US and EU endpoints within the same connector (because the same OAuth app is used). Our platform doesn't support this yet. I don't want to introduce a EU connector and then have to remove it later and worry about backwards compatibility, migrating customers, etc.

RajatPawar (approved):

PR #258, opened 2024-04-26
fix: update errors values

Jkarage:
Updates the error variables being used, Updates outreach connector to use the exported errors in common package instead of creating it's own variables. Would also prefer if they were all in error.go. All providers may need this update

### DIFF
--- a/outreach/errors.go
+++ b/outreach/errors.go
@@ -1,16 +1,5 @@
 package outreach
 
-import (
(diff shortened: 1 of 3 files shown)


Jkarage:
Just closing this as I believe the type system should be within the package, Moving to the common package might cause issues later on, Any changes in common package would break another package.

PR #264, opened 2024-04-29
Add Adobe Sign Auth Connector

### LINKED ISSUE #181: Add Adobe Acrobat Sign auth connector

Listener430:
+ Tests
+ Linter
+ Branch

## GET 
localhost:4444/api/rest/v6/users?pageSize=1
adobe - get proxy ok

## POST
localhost:4444/api/rest/v6/users
adoby - post proxy ok

## Pagination

Cursor based pagination
pagination_cursor

Next page
pagination_next_page

Closes #181

### DIFF
--- a/providers/catalog.go
+++ b/providers/catalog.go
@@ -206,12 +206,12 @@ var catalog = CatalogType{ // nolint:gochecknoglobals
 	// Adobe Sign configuration
 	AdobeSign: {
 		AuthType: Oauth2,
-		BaseURL:  "https://api.na3.adobesign.com",
+		BaseURL:  "https://api.{{.shard}}.adobesign.com",
 		OauthOpts: OauthOpts{
-			AuthURL:                   "https://secure.echosign.com/public/oauth/v2",
-			TokenURL:                  "https://secure.adobesign.com/oauth/v2/token",
+			AuthURL:                   "https://secure.{{.shard}}.adobesign.com/public/oauth/v2",
+			TokenURL:                  "https://api.{{.shard}}.echosign.com/oauth/v2/token",
 			ExplicitScopesRequired:    true,
-			ExplicitWorkspaceRequired: false,
+			ExplicitWorkspaceRequired: true,
 			TokenMetadataFields: TokenMetadataFields{
 				ConsumerRefField:  "api_access_point",
 				WorkspaceRefField: "web_access_point",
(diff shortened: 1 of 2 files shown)


providers/catalog.go
```diff
@@ -202,6 +203,34 @@ var catalog = CatalogType{ // nolint:gochecknoglobals
 		},
 	},
 
+	// Adobe Sign configuration
+	AdobeSign: {
+		AuthType: Oauth2,
+		BaseURL:  "https://api.na3.adobesign.com",
```
RajatPawar, line 209: Is this also a shard or would this work globally?

Listener430: @RajatPawar, I guess for the partner apps that we now use only the auth and token endpoints have no shards. 
Other endpoints (non auth/token) are shard based, and partners app seem to have no restrictions on which ones they can hit. So in this case, we can put na1 or eu or any other shard and it will work fine.

RajatPawar: From [here](https://opensource.adobe.com/acrobat-sign/embedpartner/gstarted.html), seems like we have to collect the endpoint that we need to use from a token's metadata. If you can confirm this, we will not be able to support it for now, and can update the issue as such.

Listener430: @RajatPawar have just tested once again, partners can access any shards. So the partner access seems to be  shard agnostic, let me know if you want to stick with the approach from the documentation or to the reality on the ground.

Listener430: Reading the document, I guess they say that once you have the token, you know the shard your client data resides from the token meta data. And they suggest to work with that shared. It isn’t enforced though. 

We can meet them halfway, I.e. we take out shards from auth/token endpoints and keep it in the BaseURL. Will it be something that might make more sense?

Listener430: Or there could be another alternative, we add setBaseURL method to the connector, so once they got the token, they can override whatever default shard is there. Just thinking out loud.

RajatPawar: The document does say that we might get empty responses from shards that are not associated with the partner app. Let's support this with surety later. Can you please - 
1. Add all of this information (links, docs, summary, etc) into the slab
2. Close the PR
3. Close the issue as cancelled

Thanks!

Listener430: @RajatPawar have doublechecked, seems that this PR is the only one that has the code for Adobe auth connector. 
So instead of cancelling, I have added the shards backs. Also left a comment in slab on parnters connectors. 
Kindly asking to review this one.

PR #269, opened 2024-04-30
feat: Deep connector Salesloft (Write)

### LINKED ISSUE #266: Add Salesloft Write

Cobalt0s:
Closes #266 

## Checklist
- [x] Ran Linter
- [x] All tests passing

## Test script output
Location: `test/salesloft/write-delete/main.go`

image


common/parse.go
```diff
@@ -114,66 +122,175 @@ func (jsonManager) ArrToMap(arr []*ajson.Node) ([]map[string]any, error) {
...
+			return nil, createKeyNotFoundErr(message)
+		}
+
+		node, err = node.GetKey(key)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if !node.IsObject() {
```
laurenzlong, line 257: This check is undesirable, even if the nested node is a string or another type, it is not a problem and we should return it.

common/parse.go
```diff
@@ -114,66 +122,175 @@ func (jsonManager) ArrToMap(arr []*ajson.Node) ([]map[string]any, error) {
...
 
-	return txt, nil
+	return formatProblematicKeyError(key, ErrNullJSON)
+}
+
+func formatProblematicKeyError(key string, baseErr error) error {
+	return fmt.Errorf("problematic key: %v %w", key, baseErr)
+}
+
+func DotZoom(keys string) []string {
```
laurenzlong, line 290: This function appears to only be used in the test, and once you implement my suggestion of revising test input to be an array instead of a string, this function will no longer be necessary and can be removed.

common/jsonManager_test.go
```diff
@@ -0,0 +1,315 @@
...
+
+	tests := []struct {
+		name        string
+		input       string
+		expectedErr error
+		withErr     bool
+	}{
+		{
+			name:    "Missing nested object",
+			input:   "payload.random",
```
laurenzlong, line 238: Why don't you make the input an array, just like what `GetNestedNode` expects? Then you don't have to parse it into an array on line 263.

common/types.go
```diff
@@ -43,6 +43,9 @@ var (
 	// ErrNotJSON is returned when a response is not JSON.
 	ErrNotJSON = errors.New("response is not JSON")
 
+	// ErrNullJSON is returned when a JSON key is null.
+	ErrNullJSON = errors.New("value of JSON key is null")
```
laurenzlong, line 47: this isn't used anywhere

salesloft/parse.go
```diff
@@ -0,0 +1,85 @@
...
+
+	return common.JSONManager.ArrToMap(arr)
+}
+
+func makeNextRecordsURL(reqLink *linkutils.URL) common.NextPageFunc {
+	return func(node *ajson.Node) (string, error) {
+		nested, err := common.JSONManager.GetNestedNode(node, []string{"metadata", "paging"})
+		if err != nil {
+			if errors.Is(err, common.ErrKeyNotFound) {
+				// list resource doesn't support pagination, hence no next page
```
laurenzlong, line 50: What do you mean by "list resource doesn't support pagination"? Are you referring to a particular endpoint?

salesloft/read.go
```diff
@@ -0,0 +1,39 @@
...
+	"context"
+	"strconv"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/linkutils"
+)
+
+// nolint:lll
+// Microsoft API supports other capabilities like filtering, grouping, and sorting which we can potentially tap into later.
+// See https://learn.microsoft.com/en-us/power-apps/developer/data-platform/webapi/query-data-web-api#odata-query-options
```
laurenzlong, line 13: This comment seems to be for microsoft, not salesloft

laurenzlong (changes requested):

common/parse.go
```diff
@@ -114,66 +122,175 @@ func (jsonManager) ArrToMap(arr []*ajson.Node) ([]map[string]any, error) {
...
-		return "", ErrNotString
+		return nil, ErrNotString
+	}
+
+	return &txt, nil
+}
+
+// GetNestedNode reaches into the JSON node by zooming via dot separated keys
+// Ex: keys = item.shipping.address => item has object with shipping key and so on.
+func (jsonManager) GetNestedNode(node *ajson.Node, keys []string) (*ajson.Node, error) {
```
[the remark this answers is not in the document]
Cobalt0s: Yes, I found out that some APIs have dot in the key itself.
I cannot make assumptions like this. So explicit slice is better. I will update the comment.

common/parse.go
Cobalt0s: I will rename the method to GetNestedObject. That what I meant.
In addition I could introduce a combination of GetNestedObject with [GetInteger, GetString...] so it will be GetNestedString.
The number of methods would drasticly increase. I could make query builder struct. Ex:
```
q := common.JSONQuery(node *ajson.Node, keys ...String)
val, err := q.GetString()
q.GetStringWithDefault()
q.GetObject()
q.GetInteger()
```
This would make querying more explicit. Node, query path via keys and then a finishing method to enforce data type.
What do you think? @laurenzlong

common/types.go
Cobalt0s: It is used here. Which in turn is used by `GetString, GetInteger` methods
https://github.com/amp-labs/connectors/blob/d6481dd91feba64de85faaab0105eeb917b3f564/common/parse.go#L278-L284

salesloft/parse.go
Cobalt0s: Metadata sometimes doesn't have `paging` object. There are tests for these. Sample response that I am refering to is below.

**Reading users:**
https://github.com/amp-labs/connectors/blob/d6481dd91feba64de85faaab0105eeb917b3f564/salesloft/test/read-list-users.json#L2-L8
**Reading people:**
https://github.com/amp-labs/connectors/blob/d6481dd91feba64de85faaab0105eeb917b3f564/salesloft/test/read-list-people.json#L2-L14

common/linkutils/link.go
```diff
@@ -0,0 +1,109 @@
...
+	// some special symbols are allowed
+	result := u.queryParams.Encode()
+	if len(result) != 0 {
+		for before, after := range map[string]string{
+			"%40": "@",
+			"%24": "$",
+			"%2C": ",",
+		} {
+			result = strings.ReplaceAll(result, before, after)
+		}
```
eberle1080, line 87: Not sure how I feel about this. On the one hand you're probably right that they're technically allowed. But there's "allowed by the spec" and then there's how common implementations treat them, as well as practical implications. Take the `$` symbol - in the wrong place (e.g. in the middle of a script) it might try to do string interpolation. Is it wrong? No. Is it hazardous? Maybe. It's also a regex anchor symbol, which - again, in the wrong context - might impact the correctness of the output.

Is there a specific reason to do this? Like does some API dislike these escape codes? Unless there's a good reason to do this, my inclination is to say that we should abide by the internal escape logic.

common/linkutils/link.go
```diff
@@ -0,0 +1,109 @@
...
+	"strings"
+)
+
+var ErrInvalidURL = errors.New("URL format is incorrect")
+
+// URL focuses on query params and treats the rest as immutable string.
+// You can use any URL string to construct this object.
+// Its primary goal is to "Expose Query Manipulation".
+// Under the hood it uses url.URL library.
+type URL struct {
```
eberle1080, line 15: This is good. I like the basic idea here. However the specific implementation is problematic - reason being that URLs have fairly esoteric escape logic. If memory serves, there are like 6 distinct parts of a URL, and each part has its own escape rules. Trying to roll your own is going to fall short (not that you can't do it, just that you'll eventually end up re-inventing the wheel).

One idea is to look at something like [RFC 6570](https://datatracker.ietf.org/doc/html/rfc6570). There's a good library which already implements this: https://github.com/jtacoma/uritemplates . It handles the escape logic nicely. You can still have this wrapper layer to simplify things, but I think string concatenation alone will eventually cause problems.

Another idea is to take these URL fields and use it to construct (on the fly) a `url.URL` instance (with the scheme, host, path, query params, fragment, etc all set explicitly). Then you can rely on the URL to properly escape things.

common/parse.go
eberle1080: Or maybe a `GetNestedNode` to do this generic logic, and then you can have `GetNested<Type>` to unwrap the "zoomed-in" value as the desired type.

common/parse.go
```diff
@@ -114,66 +118,171 @@ func (jsonManager) ArrToMap(arr []*ajson.Node) ([]map[string]any, error) {
...
+		return nil, err
+	}
+
+	result := make(map[string]any)
+	for k, v := range data {
+		result[k] = v
 	}
 
-	return txt, nil
+	return result, nil
```
eberle1080, line 271: Use `node.Unpack` instead -- that will actually turn them in to go native types. Assuming that's your goal.

common/parse.go
Cobalt0s: I haven't considered generics. I have expiremented right now with it and I get into some weird type conversions. Under the hood ajson unwraps float64 converting this to int is easier relying on convential ajson methods. Otherwise would need to have some switch on generic type to determine post processing.
The idea is intriguing to me (especially to use generics in golang as I used it only in Java). I will do the composition style.

common/linkutils/link.go
Cobalt0s: Microsoft Dynamics CRM uses special query syntax. Nothing close to "traditional" URL encoding rules.
I think it is better that tests for this are moved back and live with Microsoft. This package may allow exceptions to be registred. I won't impose such limitations for all connectors, indeed thats bad and I don't like it.
Those query parameters if not sanitised could lead to bad outcomes if they haven't put precautions in place. I will treat those as unique "language" per connector.

common/parse.go
Cobalt0s: This is great. It makes response better for the end user of this library.
I previously had some issues with tests because of it. With this change tests will be better and common.WriteResult will have golang types which is a heavy plus.

laurenzlong (approved):

laurenzlong:
changes look good to me, @eberle1080 if you're happy with it, feel free to approve and merge

eberle1080 (approved):

PR #307, opened 2024-05-09
feat: Implement Outreach ListMetadata

Jkarage:
Implemented the ListMetdata function for the Outreach connector. Loops through the objects one by one and from the response we generate the metadata field values or errors.
- Include tests for the success and failures.

Running the tests:
Screenshot 2024-05-09 at 17 56 04

### DIFF
--- a/outreach/metadata.go
+++ b/outreach/metadata.go
@@ -41,6 +41,8 @@ func (c *Connector) ListObjectMetadata(ctx context.Context,
 
 		// Check nil response body, to avoid panic.
 		if res == nil || res.Body == nil {
+			objMetadata.Errors[obj] = ErrEmptyResponse
+
 			continue
 		}
 
--- a/outreach/metadata.go
+++ b/outreach/metadata.go
@@ -1,0 +1,77 @@
+package outreach
+
+import (
+	"context"
+	"encoding/json"
+	"net/url"
+
+	"github.com/amp-labs/connectors/common"
+)
+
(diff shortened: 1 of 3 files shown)


outreach/metadata.go
```diff
@@ -0,0 +1,78 @@
...
+		// Constructing the  request url.
+		objURL, err := url.JoinPath(c.BaseURL, obj)
+		if err != nil {
+			return nil, err
+		}
+
+		res, err := c.get(ctx, objURL)
+		if err != nil {
+			objMetadata.Errors[obj] = err
+			continue
```
eberle1080, line 38:
```suggestion

			continue
```

outreach/metadata.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+
+		// Check nil response body, to avoid panic.
+		if res == nil || res.Body == nil {
+			objMetadata.Errors[obj] = ErrEmptyResponse
+			continue
```
eberle1080, line 44:
```suggestion

			continue
```

eberle1080 (approved):

outreach/metadata.go
laurenzlong: @Jkarage the suggestion wasn't to remove `objMetadata.Errors[obj] = ErrEmptyResponse`, it is to simply add a blank line between it and `continue` can you please add the line back?

Jkarage: Noted!!

PR #318, opened 2024-05-13
Update provider catalog with fixes

RajatPawar:
* This PR fixes some provider options that were set incorrectly.
* It also sorts the providers alphabetically.
* Renames `GoogleMail` to `Gmail`

PR #319, opened 2024-05-13
feat: Add userInfoURL to docusign and add a connector

RajatPawar:
## Catalog changes
* Adds a `PostAuthInfoNeeded` provider option which can be set to true to indicate that further metadata collection might be required post auth flow.
* Adds an `AuthMetadataConnector` which implements a `GetPostAuthInfo` method 
* Adds docusign connector that implements `AuthMetadataConnector` and gets metadata from [userInfo endpoint](https://developers.docusign.com/platform/auth/reference/user-info/)
* Deletes `catalog_test.go`


RajatPawar:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/319?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#319** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/319?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#318** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/318?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

 Join @RajatPawar and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

providers/catalog.go
```diff
@@ -673,30 +673,31 @@ var catalog = CatalogType{ // nolint:gochecknoglobals
...
+				Upsert: false,
+				Delete: false,
+			},
+			Proxy:     false,
+			Read:      false,
+			Subscribe: false,
+			Write:     false,
+		},
+		ProviderOpts: ProviderOpts{
+			"userInfoURL": "https://account.docusign.com/oauth/userinfo",
```
laurenzlong, line 698: Instead of this being in `ProviderOpts`, can we make this a top-level field to track all URLs that we need to call in an additional step after oauth flow (e.g. Hubspot's [GetAccountDetails](https://legacydocs.hubspot.com/docs/methods/get-account-details), Mailchimp's OAuth metadata endpoint). I'm imagining something like this:

```
Docusign: {
		AuthType: Oauth2,
		BaseURL:  "https://{{.server}}.docusign.net",
                UserInfoURL: "https://account.docusign.com/oauth/userinfo",
```

Alternative names for this field:
- AccountInfoURL
- ConnectionInfoURL
- ConnectionMetadataURL

laurenzlong (approved):
Looks good once the field is moved to a top-level field in Catalog

providers/catalog.go
RajatPawar: What do you think of storing a reference to the method instead? So it could look like - 
```
         Docusign: {
		AuthType: Oauth2,
		BaseURL:  "https://{{.server}}.docusign.net",
                FetchConnectionMetadata: docusign.GetUserInfo,
```

laurenzlong: I'm not a big fan of this approach since it's optimizing for having different method names for different connectors. (i.e. `docusign.GetUserInfo`, `mailchimp.GetPrefix`, etc.), which I don't think is needed. For consistency, I'd much rather always use the same method name for all connectors' "post-auth metadata collection" step, just like how all connectors that support read have a `Read` function. The method name can be something more generic like `getPostAuthInfo` or `getConnectionMetadata`. 

Then the catalog can either have:
- a boolean value like `postAuthInfoNeeded`, which tells the server whether or not to invoke the `getPostAuthInfo` method
- a string value like `postAuthInfoURL` or `userInfoURL`, and the presence of the string value tells the server whether or not to invoke the `getPostAuthInfo` method.

RajatPawar: Going with `postAuthInfoNeeded`, which will assume that connectors will implement `GetPostAuthInfo`

laurenzlong (approved):

laurenzlong:
FYI @RajatPawar this PR is included in the connectors update in https://github.com/amp-labs/server/pull/963

PR #321, opened 2024-05-13
feat: introduce `displayName` to catalog

laurenzlong:
## Checklist
- [x] Ran Linter
- [x] Catalog tests passing

This PR introduces a `displayName` field to ProviderInfo and adds display names for Zendesk Chat and Zendesk Support

### DIFF
--- a/providers/catalog.go
+++ b/providers/catalog.go
@@ -930,6 +930,7 @@ var catalog = CatalogType{ // nolint:gochecknoglobals
 
 	// Zendesk Support configuration
 	ZendeskSupport: {
+		DisplayName: "Zendesk Support",
 		AuthType: Oauth2,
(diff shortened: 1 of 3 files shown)


providers/catalog.go
```diff
@@ -66,6 +66,14 @@ const (
 	Zuora                               Provider = "zuora"
 )
 
+func optionalString(s string) *string {
```
RajatPawar, line 69: Instead of this, can we use `x-go-type-skip-optional-pointer`? ([ref](https://github.com/deepmap/oapi-codegen?tab=readme-ov-file#openapi-extensions))

RajatPawar (approved):

providers/catalog.go
laurenzlong: oh neat! did not know about that.

PR #346, opened 2024-05-15
feat: Deep connector Intercom (Read)

### LINKED ISSUE #347: Intercom connector: Read

Cobalt0s:
Closes #347 
- [x] Linter ok
- [x] Tests passing
- [x] Manual testing

## Manual `read`

Sample of getting notes on a contact (temporal request, not part of code):
image

From `test/intercom/read/main.go` reading conversations:
image

### DIFF
--- a/intercom/errors.go
+++ b/intercom/errors.go
@@ -4,17 +4,60 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/amp-labs/connectors/common/interpreter"
 )
 
+/*
+Response example:
+
+	{
+	  "type": "error.list",
+	  "request_id": "00066ltgfertncb684rg",
+	  "errors": [
+		{
+		  "code": "...",
+		  "message": "...",
+		  "field": "..."
+		}
+	  ]
+	}
+*/
+
 func (*Connector) interpretJSONError(res *http.Response, body []byte) error { //nolint:cyclop
-	payload := make(map[string]any)
+	var payload ResponseError
 	if err := json.Unmarshal(body, &payload); err != nil {
 		return fmt.Errorf("interpretJSONError general: %w %w", interpreter.ErrUnmarshal, err)
 	}
 
-	// TODO handle error
+	return payload.combineErr(interpreter.DefaultStatusCodeMappingToErr(res, body))
+}
 
-	return interpreter.DefaultStatusCodeMappingToErr(res, body)
+type ResponseError struct {
+	Type      string             `json:"type"`
+	RequestId *string            `json:"request_id"` // nolint:tagliatelle
+	Errors    []DescriptiveError `json:"errors"`
+}
+
+type DescriptiveError struct {
+	Code    string  `json:"code"`
+	Message *string `json:"message"`
+	Field   *string `json:"field"`
+}
+
+func (r ResponseError) combineErr(base error) error {
+	messages := make([]string, len(r.Errors))
+
+	for i, descr := range r.Errors {
+		var message string
+
+		if descr.Message != nil {
+			message = fmt.Sprintf("[%v]", *descr.Message)
+		}
+
+		messages[i] = descr.Code + message
+	}
+
+	return fmt.Errorf("%w: %s", base, strings.Join(messages, ","))
 }
--- a/intercom/errors.go
+++ b/intercom/errors.go
@@ -2,10 +2,12 @@ package intercom
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"strings"
 
+	"github.com/amp-labs/connectors/common"
 	"github.com/amp-labs/connectors/common/interpreter"
 )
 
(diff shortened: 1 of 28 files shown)


intercom/errors.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+type ResponseError struct {
+	Type      string             `json:"type"`
+	RequestId *string            `json:"request_id"` // nolint:tagliatelle
+	Errors    []DescriptiveError `json:"errors"`
+}
+
+type DescriptiveError struct {
+	Code    string  `json:"code"`
```
Cobalt0s, line 65: In all APIs the `code` field is required when responding with an error

intercom/params.go
```diff
@@ -0,0 +1,61 @@
...
+	"net/http"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/paramsbuilder"
+	"golang.org/x/oauth2"
+)
+
+const (
+	// DefaultPageSize is number of elements per page.
+	DefaultPageSize = 50
```
Cobalt0s, line 15: However, the docs say it is 150. In practice some endpoints have a limit of 60. Exceeding it returns bad request instead of defaulting to their limit :man_shrugging: .

intercom/parse.go
```diff
@@ -0,0 +1,155 @@
...
+
+	arr, err := jsonquery.New(node).Array(arrKey, false)
+	if err != nil {
+		return nil, err
+	}
+
+	return jsonquery.Convertor.ArrayToMap(arr)
+}
+
+func makeNextRecordsURL(reqLink *urlbuilder.URL) common.NextPageFunc {
```
Cobalt0s, line 57: `page.next` can be object or string, hence 2 ways of handling parsing.

intercom/parse.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	// usually when we have a pure `list` type it means array is stored at `data` field
+	return defaultFieldName, nil
+}
+
+func applyPluralForm(word string) string {
+	// The requirement for Intercom's object pluralisation is low and simple
+	// There is no need to consider all English orthography cases
+	return word + "s"
```
Cobalt0s, line 154: Adding single `s` is significant for all existing endpoints under Intercom. Maybe later we would need to handle `es`, or words ending with `y`, some sort of library would help.

intercom/read_test.go
```diff
@@ -0,0 +1,286 @@
...
+func TestRead(t *testing.T) { //nolint:funlen,gocognit,cyclop,maintidx
+	t.Parallel()
+
+	responseErrorFormat := mockutils.DataFromFile(t, "page-req-too-large.json")
+	responseContactsFirstPage := mockutils.DataFromFile(t, "read-contacts-1-first-page.json")
+	responseContactsSecondPage := mockutils.DataFromFile(t, "read-contacts-2-second-page.json")
+	responseContactsThirdPage := mockutils.DataFromFile(t, "read-contacts-3-last-page.json")
+	responseReadConversations := mockutils.DataFromFile(t, "read-conversations.json")
+	responseNotesFirstPage := mockutils.DataFromFile(t, "read-notes-1-first-page.json")
+	responseNotesSecondPage := mockutils.DataFromFile(t, "read-notes-2-last-page.json")
```
Cobalt0s, line 28: Last page of contacts and notes slightly difffers in page object.
Array is stored under different field names for conversations.

intercom/url.go
```diff
@@ -0,0 +1,19 @@
+package intercom
+
+import "github.com/amp-labs/connectors/common/urlbuilder"
+
+// Intercom pagination cursor sometimes ends with `=`.
+var intercomQueryEncodingExceptions = map[string]string{ //nolint:gochecknoglobals
+	"%3D": "=",
```
Cobalt0s, line 7: Example: 
```
    "next": {
      "page": 2,
      "starting_after": "WzE3MTU2OTU2NzkwMDAsIjY2NDM3MDNmZmFlNzgzNGQxNzkyZmQzMCIsMl0="
    },
```
`starting_after` will be used as query parameter and it has `=` symbol. Sometimes it doesn't.

test/salesloft/read/main.go
```diff
@@ -43,7 +43,7 @@ func main() {
 	fmt.Println("Reading people..")
 	utils.DumpJSON(res, os.Stdout)
 
-	if res.Rows > microsoftdynamicscrm.DefaultPageSize {
-		utils.Fail(fmt.Sprintf("expected max %v rows", microsoftdynamicscrm.DefaultPageSize))
+	if res.Rows > salesloft.DefaultPageSize {
```
Cobalt0s, line 46: Wrong pacakage but the same value used for previous connector Salesloft.

intercom/errors.go
```diff
@@ -0,0 +1,89 @@
...
+		var message string
+
+		if descr.Message != nil {
+			message = fmt.Sprintf("[%v]", *descr.Message)
+		}
+
+		messages[i] = descr.Code + message
+	}
+
+	data := strings.Join(messages, ",")
```
eberle1080, line 83:
```suggestion
	data := strings.Join(messages, ", ")
```

intercom/params.go
eberle1080: Why 50 instead of 60 then?

intercom/parse.go
eberle1080: If it's easier you could use something like this: https://github.com/gertd/go-pluralize

Generally speaking, dependencies on 3rd party libraries are ok, but the license has to be closed-source friendly (MIT, BSD, Apache, etc).

intercom/url.go
eberle1080: looks like base64 data. Decoding it shows a value of `[1715695679000,"6643703ffae7834d1792fd30",2]`. Guessing it's ok to treat this as an opaque value though?

eberle1080 (approved):

intercom/params.go
Cobalt0s: 50 is large enough and under 60 to be safe. Also this number looks better.
The arguments on my side are not solid at all :sweat_smile:

intercom/url.go
Cobalt0s: Yes, it is.
I implemented cursor based pagination before. The implementation requires a cursor to be opaque string with the following properties:
* Ordered (this acts as a pointer in history. Before/after is robust against rapid modifications like news feed, avoiding stuttering, slipage when switching pages in offset style)
* Totally unqiue (this breaks ties if something happend at exact moment)
Not all tables/documents have a field that falls under both categories. First number is ordered, second is unique, third is unclear what it means - Intercom does composite approach.

intercom/params.go
Cobalt0s: Changes: https://github.com/amp-labs/connectors/pull/346/commits/9a19a0c4e50cd7829cc1622e64dc61bcd1033041

intercom/parse.go
Cobalt0s: Cool, I added this library. https://github.com/amp-labs/connectors/pull/346/commits/62e51e2860e7e16ac5079bbd38fe15764b212f5b
Tests pass.

PR #352, opened 2024-05-15
feat: Deep connector Intercom (Write)

### LINKED ISSUE #348: Intercom connector: Write

Cobalt0s:
Closes #348

# Test Result
image


intercom/write.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"context"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/jsonquery"
+	"github.com/spyzhov/ajson"
+)
+
+type writeMethod func(context.Context, string, any) (*common.JSONHTTPResponse, error)
```
laurenzlong, line 11: Can you actually add `writeMethod` to `common/types/go` now that is used in several connectors now.

laurenzlong (approved):
Can you please update the PR description with the test output from `write-delete/main.go`?

intercom/write.go
Cobalt0s: As a tidy task I could apply some common methods or those coming from packages like jsonquery, urlbuilder, paramsbuilder. But that kind of refactoring can only happen after adding mock tests for all connectors. Let me know if I could do such task.

PR #375, opened 2024-05-20
fix: avoid panics incase nil response

Jkarage:
We are allowing returning nil in parseJSONResponse, this will cause panics in parseResult function. This fixes that.

### DIFF
--- a/common/parse.go
+++ b/common/parse.go
@@ -24,7 +24,7 @@ func ParseResult(
 	fields []string,
 ) (*ReadResult, error) {
 	if resp == nil {
-		return &ReadResult{}, nil
+		return nil, ErrEmptyResponse
 	}
 
 	totalSize, err := sizeFunc(resp.Body)
--- a/common/parse.go
+++ b/common/parse.go
@@ -24,7 +24,7 @@ func ParseResult(
 	fields []string,
 ) (*ReadResult, error) {
 	if resp == nil {
-		return nil, ErrEmptyResponse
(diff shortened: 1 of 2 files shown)


common/parse.go
```diff
@@ -23,6 +23,10 @@ func ParseResult(
 	marshalFunc func([]map[string]any, []string) ([]ReadResultRow, error),
 	fields []string,
 ) (*ReadResult, error) {
+	if resp == nil {
+		return &ReadResult{}, nil
```
RajatPawar, line 27: Nice catch! This should probably be an error, because even if the HTTP response body itself is empty, we'd expect it to still be wrapped around by `JSONHTTPResponse`

common/types.go
```diff
@@ -78,6 +78,9 @@ var (
 
 	// ErrBadRequest is returned when we get a 400 response from the provider.
 	ErrBadRequest = errors.New("bad request")
+
+	// ErrEmptyResponse is returned when the jsonResponse is nil
+	ErrEmptyResponse = errors.New("empty http response error")
```
RajatPawar, line 83: Can you name this as `ErrEmptyJSONHTTPResponse` and change the error message to `empty json http response`? You will also need to fix linting errors.

RajatPawar (changes requested):

laurenzlong (approved):
@Jkarage can you please fix the merge conflict? Thanks!

RajatPawar (approved):

PR #391, opened 2024-05-20
feat: Connector Template Generator

### LINKED ISSUE #392: Template Connector Starter

Cobalt0s:
Closes #392 

# Run

Run `make connector-gen` to create executable under `./bin`.
Then execute the following:
```
./bin/cgen base -o microsoftdynamics-example -p msdcrm -n MicrosoftDynamicsCRM
./bin/cgen read contact -o microsoftdynamics-example -p msdcrm -n MicrosoftDynamicsCRM
./bin/cgen write lead -o microsoftdynamics-example -p msdcrm -n MicrosoftDynamicsCRM
./bin/cgen delete lead -o microsoftdynamics-example -p msdcrm -n MicrosoftDynamicsCRM
./bin/cgen metadata admin -o microsoftdynamics-example -p msdcrm -n MicrosoftDynamicsCRM
```
or shorter version 
```
make connector-gen 

./bin/cgen base -p xero
./bin/cgen read contact -p xero
./bin/cgen write lead -p xero
./bin/cgen delete lead -p xero
./bin/cgen metadata account -p xero
```

# Review Output

If you want to see **output example** for xero see this `commit`: https://github.com/Cobalt0s/connectors/commit/8b27bfb50aec0090232fc16b52dc803a5c4cd43b

image


common/error.go
```diff
@@ -42,3 +44,37 @@ func PanicRecovery(wrapup func(cause error)) {
...
+
+		fallthrough
+	case errors.Is(err, ErrRetryable):
+		fallthrough
+	case errors.Is(err, ErrApiDisabled):
+		fallthrough
+	case errors.Is(err, ErrForbidden):
+		fallthrough
+	default:
+		if p.Process != nil {
```
Cobalt0s, line 74: No connector does anything special with handling errors returned from JSON HTTP Client.
Only salesforce cares about post processing an error.
Every connector has `methods.go` removed making them obsolete.

Cobalt0s (commented):
Removing `methods.go` for all connectors.

common/connectorgen/cmd/commands.go
```diff
@@ -0,0 +1,107 @@
...
+	data := &ManualTestParams{
+		Recipe:     recipe,
+		ObjectName: objectName,
+	}
+	applyTemplatesFromDirectory(filepath.Join("test", directory), data,
+		filepath.Join(recipe.Output, "test", recipe.Package, directory),
+	)
+}
+
+func init() {
```
Cobalt0s, line 101: Base command generates connector struct and Constructor.

Sample implementation, mock tests and manual tests are generated for:
* Read
* Write
* Delete
* Metadata

Cobalt0s (commented):
Templates use Provider name and Golang Package name for generation.
Manual tests also need objectName to make Read request, etc.

image

eberle1080 (changes requested):
I really, really like this. I've thought of doing something like this but haven't had the cycles to devote. This could be a big timesaver for everyone. I have 2 asks:

* Move the connectorgen stuff out of common. Being in common implies that this functionality is meant for general connector runtime, which isn't the case.
* Wherever reasonable, move re-usable functions from template functions in to non-template files. For example, `boolPtr` doesn't need to be copy-pasted in each connector, it would be better to have it in a central location so that future maintenance can be centralized. These re-usable bits can and should live somewhere in common, since those will be used at runtime.

Again, awesome idea, really happy to see this.

Cobalt0s:
> I really, really like this. I've thought of doing something like this but haven't had the cycles to devote. This could be a big timesaver for everyone. I have 2 asks:
> 
> * Move the connectorgen stuff out of common. Being in common implies that this functionality is meant for general connector runtime, which isn't the case.
> * Wherever reasonable, move re-usable functions from template functions in to non-template files. For example, `boolPtr` doesn't need to be copy-pasted in each connector, it would be better to have it in a central location so that future maintenance can be centralized. These re-usable bits can and should live somewhere in common, since those will be used at runtime.
> 
> Again, awesome idea, really happy to see this.

I am glad you like it! I moved this package under tools directory. (I think scrapper should move there too. I will create a separate PR for that). I cleaned up those utility methods in another PR https://github.com/amp-labs/connectors/pull/443 as you asked. This is the first step for template generator (which I think should be as simple as possible), it should evolve as more requirements will be coming.

I changed command to imply output directory and provider name (they can be overriden if needed):
```
make connector-gen 

./bin/cgen base -p xero
./bin/cgen read contact -p xero
./bin/cgen write lead -p xero
./bin/cgen delete lead -p xero
./bin/cgen metadata account -p xero
```

laurenzlong (approved):
This is a good start! Let's get this merged in and then iterate. I'd like to try this out locally but I don't have access to your forked repo. So once this is merged I'll try it out locally, and will likely have more feedback then.

One thing to keep in mind: we've now added other forms of auth in https://github.com/amp-labs/connectors/pull/453, so this will eventually need to be updated to accommodate the other auth types also. But no action required at the moment since we want to build a few more connectors using the new auth types first to make sure we are happy with the flow and the interfaces.

RajatPawar (commented):
This is very cool, @Cobalt0s! I don't think we can perfect this in this PR, would be good to start using it and refine it over time.

Makefile
```diff
@@ -33,3 +33,8 @@ test:
 .PHONY: pr-template
 pr-template:
 	. ./scripts/bash/pr_options.sh; pr_template
+
+# Compiles connector generator CLI. For more information locate README file near main.go.
+.PHONY: connector-gen
+connector-gen:
+	go build -o ./bin/cgen ./scripts/connectorgen/main.go && echo "now run command: ./bin/cgen"
```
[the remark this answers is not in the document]
Cobalt0s: It only `builds` an executable. It doesn't run it, only tells where it is located and how it is called to run later.
As a modification it can be installed so that the command will be available under `$PATH`.

Cobalt0s:
I was trying to remove `bin/` locally but wasn't sure why git was still indexing it.... and the reason is that it is present on the main branch. It was commited by mistake by another PR because of different `.gitignore` files, sorry about that.

laurenzlong (approved):
The semgrep Github Action is failing, and can be fixed by replacing `text/template` with `html/template`, once you do that and fix the merge conflict, you can "squash and merge" (and I believe you should have the appropriate permissions now)

eberle1080:
> The semgrep Github Action is failing, and can be fixed by replacing `text/template` with `html/template`, once you do that and fix the merge conflict, you can "squash and merge" (and I believe you should have the appropriate permissions now)

Eh, that semgrep rule is a bit aggressive. `text/template` is only insecure IF you're using it to render actual HTML. Since that's not the case here, I think it's fine. Probably more appropriate to `nolint` that particular rule.

Cobalt0s:
> > The semgrep Github Action is failing, and can be fixed by replacing `text/template` with `html/template`, once you do that and fix the merge conflict, you can "squash and merge" (and I believe you should have the appropriate permissions now)
> 
> Eh, that semgrep rule is a bit aggressive. `text/template` is only insecure IF you're using it to render actual HTML. Since that's not the case here, I think it's fine. Probably more appropriate to `nolint` that particular rule.

Indeed, we are not genereting HTML. We don't want any escaping rules applied that are relevant for it. I found how to ignore semgrep see commit: https://github.com/amp-labs/connectors/pull/391/commits/58ec895de61fe36aebd55357665c6680d8cef6a0. Rebased and there are no more conficts.
If you have some ideas on improving this template generator let me know and I can address those.

eberle1080 (approved):

PR #411, opened 2024-05-22
Remove Modules from Intercom connector

laurenzlong:
We only want to have modules for products like Hubspot where there are several APIs for different product areas, and the APIs are versioned differently or have different ways of constructing URLs from object names. For Intercom, there's only a single REST API and there aren't sub-products. Requiring the connector to be initialized with a Module parameter (`intercom.WithModule(intercom.DefaultModule)`) adds unnecessary complexity and decreases usability. 

cc @Cobalt0s

### DIFF
--- a/common/paramsbuilder/params.go
+++ b/common/paramsbuilder/params.go
@@ -79,6 +79,8 @@ func (p *Workspace) WithWorkspace(workspaceRef string) {
 }
 
 // Module params adds suffix to URL controlling API versions.
+// This is relevant where there are several APIs for different product areas or sub-products, and the APIs
+// are versioned differently or have different ways of constructing URLs from object names.
 type Module struct {
 	Suffix string
 }
--- a/intercom/connector.go
+++ b/intercom/connector.go
@@ -6,20 +6,14 @@ import (
 
 	"github.com/amp-labs/connectors/common"
 	"github.com/amp-labs/connectors/common/interpreter"
-	"github.com/amp-labs/connectors/common/paramsbuilder"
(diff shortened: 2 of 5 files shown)


RajatPawar (changes requested):
There's some leftover module functionality in `intercom/params.go` that should be removed as well.

Cobalt0s:
> We only want to have modules for products like Hubspot where there are several APIs for different product areas, and the APIs are versioned differently or have different ways of constructing URLs from object names. For Intercom, there's only a single REST API and there aren't sub-products. Requiring the connector to be initialized with a Module parameter (`intercom.WithModule(intercom.DefaultModule)`) adds unnecessary complexity and decreases usability.
> 
> cc @Cobalt0s

Here is a PR https://github.com/amp-labs/connectors/pull/417 to merge into your branch that does the cleaning and adds your comment to Module param. I think it should be documented in code to know its purpose.

laurenzlong:
> > We only want to have modules for products like Hubspot where there are several APIs for different product areas, and the APIs are versioned differently or have different ways of constructing URLs from object names. For Intercom, there's only a single REST API and there aren't sub-products. Requiring the connector to be initialized with a Module parameter (`intercom.WithModule(intercom.DefaultModule)`) adds unnecessary complexity and decreases usability.
> > cc @Cobalt0s
> 
> Here is a PR #417 to merge into your branch that does the cleaning and adds your comment to Module param. I think it should be documented in code to know its purpose.

Thank you!!

eberle1080 (approved):

Issue #366, opened 2024-05-17
Previously created auth Connectors Integration testing


Listener430:
Integration testing for the previously written connectors - tracker - https://docs.google.com/spreadsheets/d/17aM_5wNwXImK7DWElxskpryNZNENCyZUpU1i-b1awrI/edit#gid=0

Listener430:
There are more connectors in pipe for integration testings

Listener430:
Reopening as there are more conntectors left for integration testing / refer to the tracker in the issue description

Listener430:
Reopening as there are more conntectors left for integration testing / refer to the tracker in the issue description

Listener430:
Reopening the  issue as more integration tests PRs are added to this issue

Listener430:
Reopening the issue as more integration tests PRs are added to this issue

Listener430:
Reopening the issue as more integration tests PRs are added to this issue

PR #414, opened 2024-05-23
Add Adobe Sign Auth Connector

### LINKED ISSUE #181: Add Adobe Acrobat Sign auth connector

Listener430:
Replacement of previous PR that wasn't merged due to shards addition and then removal. This version includes the shards as it was planned originally. 

## GET 
localhost:4444/api/rest/v6/users?pageSize=1
adobe - get proxy ok

## POST
localhost:4444/api/rest/v6/users
adoby - post proxy ok

## Pagination

Cursor based pagination
pagination_cursor

Next page
pagination_next_page

fixes #366

### DIFF
--- a/providers/catalog.go
+++ b/providers/catalog.go
@@ -5,6 +5,7 @@ package providers
 // ================================================================================
 
 const (
+	AdobeSign                           Provider = "adobeSign"
 	Airtable                            Provider = "airtable"
 	AWeber                              Provider = "aWeber"
 	Asana                               Provider = "asana"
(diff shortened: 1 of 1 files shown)


laurenzlong:
@Listener430 Can you please add a screenshot of the token being printed out by the token script?

providers/catalog.go
```diff
@@ -79,6 +80,36 @@ const (
 // ================================================================================
 
 var catalog = CatalogType{ // nolint:gochecknoglobals
+
+	// Adobe Sign configuration with substitution
+	AdobeSign: {
+		AuthType: Oauth2,
+		BaseURL:  "https://api.{{.shard}}.adobesign.com",
```
RajatPawar, line 87: Can you also add the endpoint to call for fetching the shard, and a sample response?

Listener430: @RajatPawar, I guess the flow has the two branches for customers and partners respectively (the branch/app type is set upon the app is created). 

- If you use the customer app you would need to specify shard in your Auth URL: https://secure.{{shard}}.echosign.com/public/oauth/v2 
- If you use the partner app you don't need to write a shard in Auth URL: https://secure.echosign.com/public/oauth/v2

Customer and partner apps both share the same  access token URL: https://secure.adobesign.com/oauth/v2/token 

Attaching a sceenshot of a token for customer as the code we have is for customer. Please let me know if you want additional code/connector to handle the partner use case. 
shard

Listener430:
> @Listener430 Can you please add a screenshot of the token being printed out by the token script?

@laurenzlong hi, this provider requires an https callback URI, so can't use localhost/share a raw token data from the console. 
Code_received
And in postman the token looks like this 
shard
Please let me know if more inputs are required.

Listener430:
Closing as per discussion with @RajatPawar

PR #423, opened 2024-05-23
feat: Add a client which automatically appends fixed header values

eberle1080:
This PR adds a new kind of client which automatically appends fixed header values. Meant to be used with API keys or basic auth.


common/auth.go
```diff
@@ -0,0 +1,12 @@
+package common
+
+import "net/http"
+
+// AuthenticatedHTTPClient is an interface for an http client which can automatically
+// authenticate itself. This is useful for OAuth authentication, where the access token
+// needs to be refreshed automatically. The signatures are a subset of http.Client,
+// so it can be used as a (mostly) drop-in replacement.
+type AuthenticatedHTTPClient interface {
```
eberle1080, line 9: There's no changes here, I just moved it from another file

PR #453, opened 2024-05-29
feat: Add new auth types

eberle1080:
This PR adds support for:

* OAuth 2.0 (Client Credentials Flow -- a.k.a. 2-legged OAuth)
* API keys
* Basic authentication

This specifically targets the proxy script so that contractors can begin work on new auth types. This won't fully enable these auth types for the server, that will require additional work.

See https://ampersand.slab.com/posts/adding-a-new-proxy-connector-mu7822bu for intended usage


eberle1080:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/453?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#453** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/453?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#423** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/423?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

 Join @eberle1080 and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

scripts/proxy/proxy.go
```diff
@@ -241,32 +413,126 @@ func configureOAuth(clientId, clientSecret string, scopes []string, providerInfo
...
+	return conn.HTTPClient().Client
+}
+
+func basicAuth(username, password string) string {
+	auth := username + ":" + password
+	return base64.StdEncoding.EncodeToString([]byte(auth))
+}
+
+func setupBasicAuthHttpClient(provider string, user, pass string) common.AuthenticatedHTTPClient {
+	client, err := common.NewHeaderAuthHTTPClient(context.Background(),
```
laurenzlong, line 462: Right now, the heavy lifting is done by the user of the connector, not the connector itself. Ideally it should be the other way around. The code here for constructing the headers, doing base64 encoding, etc. now has to be repeated inside the server (and any future codebases that uses the connectors library)

I'm imagining that instead of `common.NewHeaderAuthHTTPClient`, we have:
- `common.NewBasicAuthHttpClient`
- `common.NewApiKeyAuthHttpClient`

Usage would be something like:
```
client, err := common.NewBasicAuthHttpClient(context.Background(), provider, common.WithDebug(common.DefaultDebug)) // see my other comment about a default debug function
```
The implementation of NewBasicAuthHttpClient would include looking at ProviderInfo and constructing the appropriate headers.

laurenzlong (changes requested):

scripts/proxy/proxy.go
eberle1080: Got it. I've made it much more centralized in this version. Everyone can just use `ProviderInfo.NewClient` which will handle all of the construction logic. I also implemented your suggestion for dedicated basic and api key clients.

common/basic.go
```diff
@@ -0,0 +1,27 @@
...
+		Key:   "Authorization",
+		Value: "Basic " + basicAuth(user, pass),
+	}))...)
+}
+
+// shamelessly stolen from https://pkg.go.dev/net/http#Request.SetBasicAuth
+func basicAuth(username, password string) string {
+	auth := username + ":" + password
+
+	return base64.StdEncoding.EncodeToString([]byte(auth))
```
RajatPawar, line 26: Can we turn this into an option? It is possible that we see nonstandard requirements from some providers on how they expect this to be.

common/debug.go
```diff
@@ -0,0 +1,52 @@
...
+	} else {
+		rspBody, err := httputil.DumpResponse(rsp, true)
+		if err != nil {
+			return
+		}
+
+		rspStr = string(rspBody)
+	}
+
+	_, _ = fmt.Fprintln(writer, ">>>>>>>> Request >>>>>>>>")
```
RajatPawar, line 44: This is great, would be nice to add a `DEBUG` prefix to this

common/basic.go
eberle1080: no, basic auth is extremely standard. It will always look like this. Even the go standard http library does it this way. And on the off-chance some provider really does have some novel way of doing this (which wouldn't technically be basic auth anymore), we can just construct a custom header-client with the custom-computed header values.

laurenzlong (approved):
Looks good! Can you please make a follow up PR to update the README.md for this repo to explain the new capabilities and different types of supported clients?

common/debug.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/httputil"
+	"os"
+)
+
+func DebugRequest(req *http.Request, rsp *http.Response) {
```
laurenzlong, line 11: This might be more appropriately named as `PrintRequestAndResponse` or something like that, since that's a more literal representation of what it's doing.

common/debug.go
```diff
@@ -0,0 +1,52 @@
...
+	DumpRequest(os.Stdout, req, rsp)
+}
+
+func DumpRequest(writer io.Writer, req *http.Request, rsp *http.Response) {
```
laurenzlong, line 15: does this need to be exported? should this be `dumpRequest`? It cannot be directly used by the caller of the library as a `debug` function anyhow since it doesn't match the expected function signature of `func(req *http.Request, rsp *http.Response)`

common/header.go
```diff
@@ -31,6 +31,13 @@ func WithHeaderClient(client *http.Client) HeaderAuthClientOption {
 	}
 }
 
+// WithHeaderDebug sets a debug function to be called on every request and response.
```
laurenzlong, line 34:
```suggestion
// WithHeaderDebug sets a debug function to be called on every request and response,
// after the response has been received from the downstream API. 
```

common/oauth.go
```diff
@@ -60,6 +61,13 @@ func WithOAuthConfig(config *oauth2.Config) OAuthOption {
 	}
 }
 
+// WithOAuthDebug sets a debug function to be called on every request and response.
```
laurenzlong, line 64:
```suggestion
// WithOAuthDebug sets a debug function to be called on every request and response,
// after the response has been received from the downstream API. 
```

providers/types.yaml
```diff
@@ -109,6 +109,20 @@ components:
...
+      type: object
+      required:
+        - headerName
+      properties:
+        headerName:
+          type: string
+          example: X-Api-Key
+          x-oapi-codegen-extra-tags:
+            validate: required
+        valuePrefix:
```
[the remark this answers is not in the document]
eberle1080: got it, done

common/debug.go
eberle1080: I went with yes, but it's like a very light yes. I figure, you could use this to dump to a file if you wanted (you'd have to pass in a custom callback function but it could invoke this function with a file handle). But I'm basically indifferent, so I can un-export it if you have even modestly strong feelings about it.

common/debug.go
eberle1080: agreed, done

common/debug.go
eberle1080: I un-exported it. If we need it later, it's simple enough to re-export.

PR #473, opened 2024-06-04
fix: make the authURL parameter optional

eberle1080:
For client credentials, the authURL parameter will be empty. Right now it's not allowed to be empty. This PR allows it to be empty.

### DIFF
--- a/providers/types.yaml
+++ b/providers/types.yaml
@@ -81,7 +81,6 @@ components:
         validate: required
       required:
           - grantType
-          - authURL
           - tokenURL
           - explicitScopesRequired
           - explicitWorkspaceRequired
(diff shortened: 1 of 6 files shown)


providers/types.yaml
```diff
@@ -93,8 +92,6 @@ components:
         authURL:
           type: string
           example: https://login.salesforce.com/services/oauth2/authorize
-          x-oapi-codegen-extra-tags:
```
laurenzlong, line 96: Can you add:
```
x-go-type-skip-optional-pointer: true
```
Then you don't need the `pointerFor` method and can revert the changes in catalog.go

laurenzlong (approved):

PR #494, opened 2024-06-07
feature: add Salesloft Read.Since

Cobalt0s:
# Background

Ampersand schedules `read` without time interval by default.

# Salesloft API

About 30% of Read APIs use `updated_at` query parameter. It is **15** out of **48** total objects as of June 10th, 2024.

# Action

Salesloft read method is now capable of returning a subset of data since specified time focusing on recent updates.

### DIFF
--- a/salesloft/read.go
+++ b/salesloft/read.go
@@ -45,8 +45,10 @@ func (c *Connector) buildReadURL(config common.ReadParams) (*urlbuilder.URL, err
 	link.WithQueryParam("per_page", strconv.Itoa(DefaultPageSize))
 
 	if !config.Since.IsZero() {
-		// documentation states ISO8601, while server accepts different formats
-		// but for consistency sticking to one format
+		// Documentation states ISO8601, while server accepts different formats
+		// but for consistency we are sticking to one format to be sent.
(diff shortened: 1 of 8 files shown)


salesloft/read.go
```diff
@@ -43,5 +44,12 @@ func (c *Connector) buildReadURL(config common.ReadParams) (*urlbuilder.URL, err
 
 	link.WithQueryParam("per_page", strconv.Itoa(DefaultPageSize))
 
+	if !config.Since.IsZero() {
+		// documentation states ISO8601, while server accepts different formats
```
laurenzlong, line 48: can you link to the documentation page?

laurenzlong (approved):
Can you please add a few sentences to the description of this PR? Even if we've already talked about something over slack, we still want PRs to contain context about what the change was, so that someone who's looking at it in the future will understand what it does. You can merge after that.

salesloft/read.go
Cobalt0s: Updated comment, see:
https://github.com/amp-labs/connectors/blob/2a37a22d11404e3eefe95f591738d89dd07f0c98/salesloft/read.go#L48-L53

PR #527, opened 2024-06-12
fix: Add space between prefix and key

Jkarage:
For Providers having a prefix in the API Key scheme, we need to add a space between the prefix and the token.


Jkarage:
Closing this as it seems the prefix value should be followed by a space.

PR #529, opened 2024-06-12
feat: add support for api-key authentication using HTTP query params

eberle1080:
...instead of headers.

There's apparently a requirement for this for a new connector.

This PR also renames OAuthOpts to OAuth2Opts since... that's what they are.


RajatPawar (approved):
Looks great!

PR #539, opened 2024-06-14
chore: Refactor catalog so that each provider has own file

eberle1080:
Just to keep things maintainable.

Note that none of the connectors have been modified, they were just split in to their own files.


providers/utils.go
```diff
@@ -71,6 +71,10 @@ func ReadCatalog(opts ...CatalogOption) (CatalogType, error) {
 // provider information. As a general rule, don't ever call this function
 // in production code unless you have a compelling reason to do so.
 func SetInfo(provider Provider, info ProviderInfo) {
```
laurenzlong, line 73: The comments above this function are now outdated - since you are now definitely calling this function in production code.

providers/aWeber.go
```diff
@@ -0,0 +1,30 @@
+package providers
+
+const AWeber Provider = "aWeber"
+
+func init() {
```
laurenzlong, line 5: Maybe I'm missing something but when are these init functions called?

laurenzlong (approved):

providers/utils.go
eberle1080: Lol fair. Yes I backtracked on that pretty quickly. I'll update the comments.

PR #554, opened 2024-06-16
Add Recurly Connector

### LINKED ISSUE #520: Add Recurly connector

Beadko:
Closes #520 

## Checklist
- [x] Ran Linter

## Notes
include version in the headers `Accept: application/vnd.recurly.v2021-02-25`
Not possible to print out the token as it is generated in platform.

### GET 
URL: <localhost:4444/accounts> 
Screenshot 2567-06-16 at 17 13 15

### POST
URL: <localhost:4444/accounts> 
Screenshot 2567-06-16 at 17 16 40

### PUT
URL: <localhost:4444/accounts/code-2> 
Screenshot 2567-06-16 at 17 21 03

### DELETE
URL: <localhost:4444/accounts/code-2> 
Screenshot 2567-06-16 at 17 21 27

## Pagination
**Before pagination**
Screenshot 2567-06-16 at 17 18 05
**After pagination with limit and cursor params included**
Screenshot 2567-06-16 at 17 19 27

### DIFF
--- a/providers/recurly.go
+++ b/providers/recurly.go
@@ -18,6 +18,6 @@ func init() {
 			Subscribe: false,
 			Write:     false,
 		},
-		PostAuthInfoNeeded: true,
+		PostAuthInfoNeeded: false,
 	})
 }
--- a/providers/recurly.go
+++ b/providers/recurly.go
@@ -1,0 +1,23 @@
+package providers
+
+const Recurly Provider = "recurly"
+
+func init() {
+	SetInfo(Recurly, ProviderInfo{
+		AuthType: Basic,
+		BaseURL:  "https://v3.recurly.com",
+		Support: Support{
+			BulkWrite: BulkWriteSupport{
+				Insert: false,
+				Update: false,
+				Upsert: false,
+				Delete: false,
+			},
+			Proxy:     false,
+			Read:      false,
+			Subscribe: false,
+			Write:     false,
+		},
+		PostAuthInfoNeeded: true,
+	})
+}


providers/catalog.go
```diff
@@ -2147,4 +2148,23 @@ var catalog = CatalogType{ // nolint:gochecknoglobals
 			Write:     false,
 		},
 	},
+
+	// Recurly Basic authentication
+	Recurly: {
+		AuthType: Basic,
+		BaseURL:  "https://v3.recurly.com",
```
RajatPawar, line 2155: @laurenzlong Interesting position of the API version. We might have to just hardcode this one.

laurenzlong: yeah agreed

providers/recurly.go
```diff
@@ -0,0 +1,23 @@
...
+				Update: false,
+				Upsert: false,
+				Delete: false,
+			},
+			Proxy:     false,
+			Read:      false,
+			Subscribe: false,
+			Write:     false,
+		},
+		PostAuthInfoNeeded: true,
```
laurenzlong, line 21: @Beadko `PostAuthInfoNeeded` is used to indicate whether Ampersand needs to make an API call to an additional endpoint after getting a token to retrieve info such as `workspaceRef` or server region, etc that we need to know to make subsequent API calls. I don't see any mention of an endpoint like this in the Slab doc, and this kind of an endpoint like this.

RajatPawar (approved):

PR #574, opened 2024-06-20
feat: Pre-allocate the catalog to avoid too many allocs during initialization

eberle1080:
The way the current catalog gets initialized tends to thrash memory a bit. This is easy to fix, all we have to do is pre-allocate the memory. The only part that's a little tricky is figuring out the number of providers, which we can just generate via script. And since we already have a script which reads the catalog and outputs something, this can just piggy-back on that.


internal/provider_count.go
```diff
@@ -0,0 +1,5 @@
+package internal
+
+// This file will be updated automatically, do not edit it manually.
```
jlimatampersand, line 3: Can we rename the folder or file to include "generated" to make it little more obvious?

scripts/catalog/json.go
```diff
@@ -10,6 +11,13 @@ import (
 
 const writePerm = 0o644
 
+const countFileContents = `package internal
+
+// This file will be updated automatically, do not edit it manually.
```
jlimatampersand, line 16:
```suggestion
// This file is generated automatically, do not edit it manually.
```

scripts/catalog/json.go
```diff
@@ -28,5 +36,13 @@ func main() {
 		log.Fatal(err)
 	}
 
+	countFile := "internal/provider_count.go"
```
jlimatampersand, line 39: As commented in the file, make it to include `generated` to make it easier to understand when importing.

jlimatampersand (approved):

scripts/catalog/json.go
```diff
@@ -28,5 +36,13 @@ func main() {
 		log.Fatal(err)
 	}
 
+	countFile := "internal/generated/provider_count.go"
+	str := fmt.Sprintf(countFileContents, len(catalog))
```
RajatPawar, line 40: cool idea!

PR #597, opened 2024-06-24
[ENG-1231] feat: More examples, update README

eberle1080:
Made a number of new examples, and moved all of it outside of the readme in to a dedicated `examples` directory. The updated README now renders like this:

Screenshot 2024-06-25 at 4 20 37 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,72 +8,11 @@ It can be either be used as a standalone library, or as a part of the [Ampersand
 - Handling API quotas from SaaS APIs
 - A dashboard for observability and troubleshooting
 
-Sample usage:
+## Examples
 
-```go
-import (
-  "context"
-  "fmt"
-  "net/http"
-  "time"
+See the [examples directory](https://github.com/amp-labs/connectors/tree/main/examples) for examples of how to use the library.
 
-  "github.com/amp-labs/connectors"
-  "github.com/amp-labs/connectors/salesforce"
-  "golang.org/x/oauth2"
-)
-
-const (
-  // Replace these with your own values.
-  Workspace = "<workspace>"
-  OAuthClientId = "<client id>"
-  OAuthClentSecret = "<client secret>"
-  OAuthAccessToken = "<access token>"
-  OAuthRefreshToken = "<refresh token>"
-
-)
-
-// Replace with when the access token will expire,
-// or leave as-is to have the token be refreshed right away.
-var AccessTokenExpiry = time.Now().Add(-1 * time.Hour)
-
-func main() {
-  // Set up the OAuth2 config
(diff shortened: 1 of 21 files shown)


examples/auth_connectors/adobe/adobe.go
```diff
@@ -0,0 +1,80 @@
...
+	// Print out the names of the data sets
+	for _, node := range nodes {
+		fmt.Printf("DataSet.Name: %s\n", node.MustString())
+	}
+
+	return nil
+}
+
+// Create an auth connector with the Adobe provider.
+func createAuthConnector(ctx context.Context) *connector.Connector {
```
jlimatampersand, line 59: It is just examples, so maybe not a big deal. But since this func is defined 5 times with slightly different details, maybe we can define this once and pass options in?

scripts/oauth/token.go
```diff
@@ -132,8 +132,8 @@ func (a *OAuthApp) ServeHTTP(writer http.ResponseWriter, request *http.Request)
 	case request.URL.Path == "/" && request.Method == "GET":
 		// Redirect to the OAuth provider.
 		encState := base64.URLEncoding.EncodeToString([]byte(a.State))
-		url := a.Config.AuthCodeURL(encState, a.Options...)
-		writer.Header().Set("Location", url)
+		u := a.Config.AuthCodeURL(encState, a.Options...)
```
jlimatampersand, line 135: can we use url instead unless there is a need for it? I sometimes like shortcuts, but I think it would be easier to read for others if it is spelled out. I think the linter would also complain about short parameter names.

jlimatampersand (approved):
Couple comments left. You can either resolve or respond, then feel free to merge after.

README.md
```diff
@@ -8,112 +8,38 @@ It can be either be used as a standalone library, or as a part of the [Ampersand
...
 
-There are 3 ways to initialize a Connector:
-
-1. Initializing a provider-specific Connector (returns a concrete type). This method of initialization will allow you to use methods that only exist for that provider.
+## How to initialize a Connector
 
 ```go
-client, err := connectors.Salesforce(
+// Example for Salesforce
+client, err := salesforce.NewConnector(
```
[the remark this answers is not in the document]
eberle1080: it was recently broken. Rajat removed certain functions, but the docs referred to one of the old ways of initializing.

scripts/oauth/token.go
eberle1080: I changed the name because the var name `url` collides with the import. I was trying to reduce the chance of a future bug. But it's not a big deal and I can change it back if it's confusing.

PR #601, opened 2024-06-25
feature: Add Oauth2 Password grant type

Cobalt0s:
# Scripts

## Token

Generating tokens for SugarCRM using Oauth2 password grant type:
image

## Proxy

SugarCRM uses `Access Token` that we acquired using previous script and successfuly proxies to emails resource:
image


providers/utils.go
```diff
@@ -239,6 +244,9 @@ type NewClientParams struct {
 	// If the provider uses auth code, this field must be set.
 	OAuth2AuthCodeCreds *OAuth2AuthCodeParams
 
+	// TODO comment
```
laurenzlong, line 247: I think you forgot to update the comment

providers/sugarcrm.go
```diff
@@ -0,0 +1,29 @@
+package providers
```
laurenzlong, line 1: can you put this file in a different PR? so have one PR for adding password grant and one PR for adding sugarCRM?

providers/sugarcrm.go
```diff
@@ -0,0 +1,29 @@
...
+const SugarCRM Provider = "sugarCRM"
+
+func init() {
+	// 2-legged auth
+	SetInfo(SugarCRM, ProviderInfo{
+		AuthType: Oauth2,
+		BaseURL:  "{{.workspace}}",
+		Oauth2Opts: &Oauth2Opts{
+			GrantType:                 Password,
+			TokenURL:                  "{{.workspace}}/rest/{{.restVersion}}/oauth2/token",
```
laurenzlong, line 12: let's hard code in a restVersion right now, we don't support a way to get the restVersion from the builder currently, so this won't work with our existing system.

providers/utils.go
```diff
@@ -387,6 +397,31 @@ func createOAuth2ClientCredentialsHTTPClient( //nolint:ireturn
...
 
+func createOAuth2PasswordHTTPClient(ctx context.Context, client *http.Client,
+	dbg bool, params *Oauth2PasswordParams,
+) (common.AuthenticatedHTTPClient, error) {
+	options := []common.OAuthOption{
+		common.WithOAuthClient(getClient(client)),
+		common.WithTokenSourceProvider(func() (*oauth2.Token, error) {
+			// we could use refresh token, but it works a bit differently when refresh token was
+			// acquired via oauth2 password grant type.
+			// TODO this is called every time a request is made. Store AT/RT
```
laurenzlong, line 408: what's AT/RT? Can you update the comment to make it more clear?

providers/types.yaml
```diff
@@ -86,7 +86,7 @@ components:
       properties:
         grantType:
           type: string
-          enum: [authorizationCode, clientCredentials, PKCE]
+          enum: [authorizationCode, clientCredentials, password, PKCE]
```
laurenzlong, line 89: we're starting to migrate all of our openapi definition to its own repo, can you make the same change in this file also? https://github.com/amp-labs/openapi/blob/main/catalog/catalog.yaml

providers/utils.go
Cobalt0s: sorry I use AT/RT a short version of access token and refresh token

providers/sugarcrm.go
Cobalt0s: Dedicated PR: https://github.com/amp-labs/connectors/pull/602

providers/sugarcrm.go
Cobalt0s: I updated the base URL to match the version and restrict to rest usage. https://github.com/amp-labs/connectors/pull/602/commits/c8a11f40ac59c98aa88fcf2442aabf895e27a1e4

providers/types.yaml
Cobalt0s: PR created: https://github.com/amp-labs/openapi/pull/4

eberle1080 (approved):
Please resolve the issues pointed out by @laurenzlong but otherwise this looks good. Thanks! This is very helpful. We won't support this immediately (in the server and frontend) but we can start planning for it. And it does unlock some additional providers.

PR #603, opened 2024-06-25
[ENG-1257] feat: Add WithWorkspace, connectors delegate to paramsbuilder

Cobalt0s:
# Changes

* paramsbuilder.Workspace is used by relevant connectors, method returns substitutions map making usage concise
* Based on Hubstpot formulated behaviour of paramsbuilder.Module with fallback module
* all connectors delegate to paramsbuilder, shorter format

### DIFF
--- a/common/paramsbuilder/params.go
+++ b/common/paramsbuilder/params.go
@@ -78,6 +78,10 @@ func (p *Workspace) WithWorkspace(workspaceRef string) {
 	p.Name = workspaceRef
 }
 
+func (p *Workspace) Substitution() map[string]string {
+	return map[string]string{"workspace": p.Name}
+}
+
 // Module params adds suffix to URL controlling API versions.
 // This is relevant where there are several APIs for different product areas or sub-products, and the APIs
 // are versioned differently or have different ways of constructing URLs from object names.
--- a/common/paramsbuilder/params.go
+++ b/common/paramsbuilder/params.go
@@ -17,8 +17,9 @@ type ParamAssurance interface {
 }
 
 var (
-	ErrMissingClient    = errors.New("http client not set")
-	ErrMissingWorkspace = errors.New("missing workspace name")
+	ErrMissingClient     = errors.New("http client not set")
+	ErrMissingWorkspace  = errors.New("missing workspace name")
+	ErrNoSupportedModule = errors.New("no supported module was chosen")
 )
 
 // Client params sets up authenticated proxy HTTP client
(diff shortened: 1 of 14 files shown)


laurenzlong:
Thanks for this @Cobalt0s! Next time can you please:
- request reviewer (if you don't have a dedicated reviewer for a particular connector, then type `amp-labs/go` into the reviewer section, and it'll auto-pick someone from our team)
- make PRs isolated to one feature (e.g. Module related changes could have been in a different PR)

Cobalt0s:
@laurenzlong yes, I felt it is not good that when I started refactoring I ended up with more than needed. I will try keep the changes better organised for multiple PRs.
Speaking of reviewer and linear, will do!

hubspot/modules.go
```diff
@@ -1,40 +1,21 @@
...
-func supportsModule(module string) bool {
-	for _, mod := range supportedModules {
-		if module == mod.String() {
-			return true
-		}
-	}
-
-	return false
-}
+var supportedModules = []paramsbuilder.APIModule{ModuleCRM} // nolint: gochecknoglobals
```
RajatPawar, line 21: Would this mean that an empty module initialization of Hubspot would error out?

common/paramsbuilder/params.go
```diff
@@ -78,20 +79,54 @@ func (p *Workspace) WithWorkspace(workspaceRef string) {
...
+}
+
 // Module params adds suffix to URL controlling API versions.
 // This is relevant where there are several APIs for different product areas or sub-products, and the APIs
 // are versioned differently or have different ways of constructing URLs from object names.
 type Module struct {
-	Suffix string
+	Suffix    string
+	supported []APIModule
+	fallback  *APIModule
```
RajatPawar, line 92: Does a fallback module have to be in the supported list?

RajatPawar (approved):
Have a couple of comments, please reply / resolve those before merging it

RajatPawar (commented):
You might need to update the README as well.

common/paramsbuilder/params.go
```diff
@@ -78,20 +79,54 @@ func (p *Workspace) WithWorkspace(workspaceRef string) {
 	p.Name = workspaceRef
 }
 
+func (p *Workspace) Substitution() map[string]string {
```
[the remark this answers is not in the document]
Cobalt0s: Of course, that would be much better. Also the pointer to the map kind of bothered my OCD ahaha. I will think about your suggestion and be back

hubspot/modules.go
Cobalt0s: Thats the understanding I got from the way it is implemented right now. It is refactored in such a way where any connector that will have support of module option must be one of the listed, where we enforce the default.

Empty Module is not used in connectors and I have just checked that there is one usage in `server`. So what happens there is it will fallback to ModuleCRM.

common/paramsbuilder/params.go
Cobalt0s: Yes I validate the fallback too to be in the supported list. 
https://github.com/amp-labs/connectors/blob/730046a36c2e81e96b83b8b9292e516e84151bfe/common/paramsbuilder/params.go#L98-L110

For hubspot it was hard coded deep inside but now this intention will be explicit. Note: supported/fallback are internal fields that should be set by us.
https://github.com/amp-labs/connectors/blob/730046a36c2e81e96b83b8b9292e516e84151bfe/hubspot/params.go#L59-L64

common/paramsbuilder/params.go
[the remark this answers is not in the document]
Cobalt0s: I will create a dedicated PR for this!

Cobalt0s:
> You might need to update the README as well.

Main readme file already has `WithWorkspace` example.

common/paramsbuilder/params.go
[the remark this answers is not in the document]
Cobalt0s: Here is a PR https://github.com/amp-labs/connectors/pull/615

hubspot/modules.go
RajatPawar: We need it to be able to initialize with an empty module so that it can work as a proxy connector. Can you please add a fix for this? The fallback should be NoModule.

Cobalt0s: The PR has a commit with this fix: https://github.com/amp-labs/connectors/pull/615/commits/1e631f2632e1873f2e11a89085ddb6b897118708

PR #640, opened 2024-07-01
tidy: remove restApiUrl provider option

Cobalt0s:
# Changes

Some connectors have unnecessary Provider Options, which in fact can be derived from core variables of catalog.

Salesforce options equivalence:
* restApiUrl = BaseURL + `/services/data/v59.0`
* domain = BaseURL

There was ambiguity and several paths to the same values (API "sections"). All URLs can be formed via connector.getURL (for rest api, which is most common usage) and connector.getDomainURL (for special cases). 

Outreach options equivalence:
* restApiUrl = BaseURL + `api/v2`

Also for convenience `urlbuilder.URL` offers managment of query params and paths.

# Refactoring

Every connector tries to handle uri parts over and over while the utility method already exists for URL.
Adding tests which dictate how URL parts should be handled created some changes to AddPath and constructor.

# Testing

All tests are passing. Running read operation on Salesloft produces successful output.
image

# Notes

There are no longer any ProviderOpts.


outreach/connector.go
```diff
@@ -33,19 +32,22 @@ func NewConnector(opts ...Option) (conn *Connector, outErr error) {
...
-		BaseURL: restApi,
 	}
 
-	conn.Client.HTTPClient.Base = providerInfo.BaseURL
+	conn.setBaseURL(providerInfo.BaseURL)
 
 	return conn, nil
 }
+
+func (c *Connector) getURL(arg string) (*urlbuilder.URL, error) {
```
RajatPawar, line 46: I think this might be a little unintuitive for someone new - we should rename this to be a little more explicit (`getApiUrl` or something like that)

salesforce/connector.go
```diff
@@ -4,22 +4,18 @@ import (
...
 
-const (
-	providerOptionRestApiURL = "restApiUrl"
-	providerOptionDomain     = "domain"
-)
-
 const (
 	apiVersion    = "59.0"
 	versionPrefix = "v"
+	restAPISuffix = "/services/data/v59.0"
```
RajatPawar, line 14: We should construct this using `apiVersion`

salesforce/connector.go
```diff
@@ -50,27 +46,42 @@ func NewConnector(opts ...Option) (conn *Connector, outErr error) {
...
+func (c *Connector) Provider() providers.Provider {
+	return providers.Salesforce
+}
+
+// String returns a string representation of the connector, which is useful for logging / debugging.
+func (c *Connector) String() string {
+	return fmt.Sprintf("%s.Connector", c.Provider())
+}
+
+func (c *Connector) getURL(paths ...string) (*urlbuilder.URL, error) {
```
RajatPawar, line 72: I'd rename this to be more explicit about the fact that it is joining against the REST API url, because there are other kinds of suffixes being used in this connector too

salesforce/events.go
```diff
@@ -189,6 +189,7 @@ func (c *Connector) CreateEventRelayConfig(
 // nolint: lll
 // https://developer.salesforce.com/docs/atlas.en-us.api_tooling.meta/api_tooling/tooling_api_objects_eventrelayconfig.htm?q=EventRelayConfig
 func (c *Connector) RunEventRelay(ctx context.Context, cfg *EventRelayConfig) error {
+	// TODO should there be a BaseURL?
 	location, err := joinURLPath("tooling/sobjects/EventRelayConfig", cfg.Id)
```
RajatPawar, line 193: This could be a constant and be updated to use getURL / getDomainURL

salesforce/provider.go
RajatPawar, line 1: Good one :)

RajatPawar (approved):
Looks good, just a couple of minor naming changes. Please test salesforce, outreach, intercom, dynamicscrm and gong connectors before you merge this in. URL related changes frequently break things very subtly, and I want us to be very sure that this isn't doing that.

salesforce/events.go
Cobalt0s: I changed URL methods such that:

getRestApiURL -> rest prefix
getDomainURL -> url starting from base domain
getURIPartEventRelayConfig -> returns relative path to `tooling/sobjects/EventRelayConfig`

RajatPawar (approved):

PR #646, opened 2024-07-02
test: mock Salesforce Read method

Cobalt0s:
# Mock Tests

Added mock tests for salesforce Read operation.
Used common packages to simplify code.
By making server requests collected its output and stored under `salesforce/test/*.json`.

There was no way to make server return `nextRecordsUrl`. This requires over 2000 items. Such scenario was forged to match [documentation.](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_query.htm)


salesforce/read_test.go
```diff
@@ -0,0 +1,214 @@
...
+		input        common.ReadParams
+		server       *httptest.Server
+		connector    Connector
+		comparator   func(serverURL string, actual, expected *common.ReadResult) bool // custom comparison
+		expected     *common.ReadResult
+		expectedErrs []error
+	}{
+		{
+			name: "At least one field must be provided",
+			server: httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
laurenzlong, line 37: to make this test more readable, `input` should explicitly written out

salesforce/read_test.go
```diff
@@ -0,0 +1,214 @@
...
+func TestRead(t *testing.T) { //nolint:funlen,gocognit,cyclop,maintidx
+	t.Parallel()
+
+	responseUnknownObject := mockutils.DataFromFile(t, "unknown-object.json")
+	responseLeadsFirstPage := mockutils.DataFromFile(t, "read-list-leads.json")
+	responseListContacts := mockutils.DataFromFile(t, "read-list-contacts.json")
+
+	tests := []struct {
+		name         string
+		input        common.ReadParams
```
laurenzlong, line 28: I'm realizing that we're not doing any validation that ReadParams has a `ObjectName` field, which should be a required field. Once that validation exists, these test cases should be updated to include an ObjectName in the input, and there should be one more test case added to ensure ObjectName is required.

common/types.go
```diff
@@ -65,12 +65,12 @@ var (
 	// ErrMissingRecordID is returned when resource id is missing in the request.
 	ErrMissingRecordID = errors.New("no object ID provided")
 
+	// ErrMissingFields is returned when no fields are provided for reading.
+	ErrMissingFields = errors.New("no fields to select")
```
laurenzlong, line 69:
```suggestion
	ErrMissingFields = errors.New("no fields provided in ReadParams")
```

laurenzlong (approved):

salesforce/read_test.go
Cobalt0s: This happened before. I will be creating test refactoring PR. As a chore I will apply this to all tests then.

salesforce/read_test.go
Cobalt0s: Yes, this was not enforced for any connector Read operation unlike the Write method.

PR #659, opened 2024-07-05
Fix: nil reference error in Salesforce CreateMetadata when there's invalid token

laurenzlong:
When there's an invalid access token used, `res` can actually be empty, so `res.StatusCode` results in a nil dereference error. 

Screenshot 2024-07-04 at 6 49 21 PM

I revised the code to first check if `err` contains the string "invalid_grant" (see screenshot below for the printed `error` string)
Screenshot 2024-07-04 at 6 16 53 PM

Then if err is nil and there's a non-200 response check the body. 

I also refactored the code to be a recursive function.


laurenzlong:
likely obsoleted by https://github.com/amp-labs/connectors/pull/656

PR #668, opened 2024-07-08
feat: Infer credentials for Provider from JSON/ENV

Cobalt0s:
# Goal

When constructing connector field readers should be automatically infered from the provider information. Catalog has all information that we need.

# Changes

Any credential registry files moved under `credsregistry` package.
Select provider from the catalog that matches provider name from ENV or JSON file. Then create list of fields that are required. Example: oauth2 then for sure clientID+clientSecret. If basic auth then username+password.

The field names in JSON(or ENV vars) will be consistent and hence meet expactations when dealing with any connector. (as of now only DynamicsCRM is adapted).

# Example DynamicsCRM

```
	filePath := credsregistry.LoadPath(providers.DynamicsCRM)
	reader := utils.MustCreateProvCredJSON(filePath, true) // all field readers are automatically added

	conn, err := dynamicscrm.NewConnector(
                // construction of oauth token is no longer repetative
		dynamicscrm.WithClient(ctx, http.DefaultClient, getConfig(reader), reader.GetOauthToken()),
		dynamicscrm.WithWorkspace(reader.Get(credsregistry.Fields.Workspace)),
	)
```


common/credsregistry/fields.go
```diff
@@ -0,0 +1,157 @@
...
+
+	"github.com/amp-labs/connectors/common/handy"
+	"github.com/amp-labs/connectors/providers"
+	"github.com/iancoleman/strcase"
+)
+
+// Fields is a grouping of constant values that dictate the keys that
+// can be present inside *creds.json file.
+var Fields = struct { // nolint:gochecknoglobals
+	Provider Field
```
Cobalt0s, line 15: These are constant values that define consistent naming of ENV variables and JSON fields

common/credsregistry/fields.go
```diff
@@ -0,0 +1,157 @@
...
+	case providers.None:
+	case providers.Oauth2:
+		lists.Add(requiredType, Fields.ClientId, Fields.ClientSecret)
+	default:
+		return nil, ErrProviderInfo
+	}
+
+	options := info.Oauth2Opts
+	if options != nil {
+		// TODO workspace can happen not only to Oauth2
```
Cobalt0s, line 136: As of now we can conclude that workspace is needed for Oauth2. Other auth types dont have this flag. Here is a PR for changing models: https://github.com/amp-labs/openapi/pull/15

common/credsregistry/path.go
```diff
@@ -0,0 +1,24 @@
...
+	"os"
+
+	"github.com/iancoleman/strcase"
+)
+
+// LoadPath will give path to creds.json.
+// For provider called `dynamicsCRM` the file location will either be
+// * value of DYNAMICS_CRM_CRED_FILE env var, or
+// * ./dynamics-crm-creds.json.
+func LoadPath(providerName string) string {
```
Cobalt0s, line 14: This defines how creds file should be named for a provider

common/credsregistry/provider.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	registry := NewCredentialsRegistry()
+	readers := handy.Lists[Reader]{}
+
+	// add readers for every field
+	for kind, fieldList := range fields {
+		for _, field := range fieldList {
+			reader := selectReader(field, filePath, providerName)
+			readers.Add(kind, reader)
```
Cobalt0s, line 69: Readers are grouped as optional and required. Different validation applies.

common/credsregistry/provider.go
```diff
@@ -0,0 +1,146 @@
...
+	registry := NewCredentialsRegistry()
+
+	reader := Fields.Provider.GetJSONReader(filePath)
+
+	err := registry.AddReader(reader)
+	if err != nil {
+		return ""
+	}
+
+	name, _ := registry.GetString(Fields.Provider.Name)
```
Cobalt0s, line 143: This registry can only read provider name from a file. Then lookup the catalog and imply the rest of fields.

common/credsregistry/tokens.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"time"
+
+	"golang.org/x/oauth2"
+)
+
+// GetOauthToken constructs Token group from creds file.
+// Some connectors may implement Refresh tokens, when it happens expiry must be provided alongside.
+// Library shouldn't attempt to refresh tokens if API doesn't support `refresh_token` grant type.
+func (r ProviderCredentials) GetOauthToken() *oauth2.Token {
```
Cobalt0s, line 12: This procedure was unique to proxy script. I moved it here and every connector that needs to load AT/RT tokens will do it using this method. Should be a common thing.

common/handy/lists.go
```diff
@@ -0,0 +1,11 @@
+package handy
+
+type Lists[T any] map[string][]T
```
Cobalt0s, line 3: General data structure.
In this PR we use this `handy.Lists` to conceptually split fields into required and optional kinds. Adding elements is convenient. Looping over values allows to apply different operations based on the list kind.

test/dynamicscrm/connector.go
```diff
@@ -2,61 +2,46 @@ package dynamicscrm
...
-	"github.com/amp-labs/connectors/utils"
+	"github.com/amp-labs/connectors/providers"
+	"github.com/amp-labs/connectors/test/utils"
+	"golang.org/x/oauth2"
 )
 
-func GetMSDynamics365CRMConnector(ctx context.Context, filePath string) *dynamicscrm.Connector {
-	registry := utils.NewCredentialsRegistry()
-
-	readers := []utils.Reader{
```
Cobalt0s, line 15: These readers are implied from catalog.

jlimatampersand:
@Cobalt0s  I was waiting on upstream PR https://github.com/amp-labs/openapi/pull/15/files which seems to be closed after @laurenzlong  reviewed this. Can you modify this PR to reflect that as well? Feel free to ping me on slack after you modify the PR.

jlimatampersand:
I will remove myself from this reviewer list. You can add me back after modification.

Cobalt0s:
@jlimatampersand, I replaced a `todo` with a comment. See commit: https://github.com/amp-labs/connectors/pull/668/commits/c6ef634660eff80533dda819ee77bc59feb6b245
I think when this change will happen a person removing `ExplicitWorkspaceRequired` will end up reading of what is needed from that part of code.

Cobalt0s:
* **#668** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/668?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

common/credsregistry/registry.go
```diff
@@ -1,5 +1,5 @@
 // nolint: ireturn
-package utils
+package credsregistry
```
jlimatampersand, line 2: This registry is not just for credentials. It has general purpose we can read any values from any file or env that are not necessarily credentials. Also that is why most of the variables/methods don't really mention much about credentials, secrets, tokens, etc in the package. That is why those are called `Readers` because they read anything. 

There is probably one or two variables, but those are to be modified so that it won't refer to anything auth related. 

It just happens to be the case that Auth is currently the only example we have that is using this package for. Please change revert these changes. 

You are welcome to create a credsregistry package as wrapper package of these readers with an isolated purpose.

jlimatampersand (changes requested):
Overall direction looks good, but the readers should remain unchanged. Please refactor and revert some. Feel free to create registry package as wrapper of readers. Reason stated in another comment.

common/credsregistry/registry.go
Cobalt0s: Makes sense. I wasn't sure about purpose of other methods.

I have cleaned up the code such that:
* Moved registry from `utils` to `scanning` package which is preserving multipurpose reading intent.
* CredentialsRegistry is now `scanning.Registry`
* Readers interface and 3 impls are moved to a sibling file `readers.go`
* Field rename: Readers.CredKey -> Readers.KeyName
* Error: "credential not found" -> "value not found"

As a result `scanning` has no trace of `creds` words. Everything else is preserved.

Cobalt0s: Now wrapper is called `credscanning` implying relationship to `scanning` package.

Important methods:
* NewJSONProviderCredentials
* NewENVProviderCredentials
* ProviderCredentials.Get - utilises MustString
* ProviderCredentials.GetOauthToken - returns AcessToken and optional Refresh
* LoadPath - JSON file path to put creds based on provider name

Maybe when other requirement will arise we could add wrappers under `scanning` directory.

Cobalt0s: All new changes are here: https://github.com/amp-labs/connectors/pull/668/commits/5be7a2267de0e7e06d363089e4e948f5ccdd20d6

Example running Zendesk Support - Read, Write.
image
image

Cobalt0s: All new changes are here: https://github.com/amp-labs/connectors/pull/668/commits/5be7a2267de0e7e06d363089e4e948f5ccdd20d6

Example running Zendesk Support - Read, Write.
image
image

Cobalt0s:
Running lint locally and clearing cache by running `golangci-lint cache clean` still succeeds :man_shrugging:

laurenzlong:
@jlimatampersand can you please re-review?

jlimatampersand (approved):

jlimatampersand:
Feel free to merge after fixing the line errors.

PR #677, opened 2024-07-10
fix: update json catalog package and action

RajatPawar:
## Why hasn't this action worked?
We have github branch protection set up to protect `main` against accidental deletes or force pushes. We've also set it up to ensure that every commit goes through a PR. 

It has an option to let certain users bypass it, which I assumed would be enough to let our `ampersand-ops` user commit directly. 

Unfortunately, life isn't simple - if you have also set up required status checks, then you actually can't truly bypass these restrictions. Even if the user has a bypass branch protection permission 🙃 ([ref](https://github.com/orgs/community/discussions/43460))

### Other options
The way that Github offers out of this is to use something called as `Rulesets`. These are just sets of rules you can apply against branches. 

For our usecase, we need to 'layer' two rulesets - one basic anti-deletion ruleset without a bypass, and then a ruleset that has the required PRs/status check related stuff, with the bypass you need.

So finally, the way I fixed this was - 
1. Deleted our `main` branch protection rules
2. Created a base `main` ruleset which protects against deletion and force pushes.
3. Created another `main` ruleset on top of that which requires approvals, status checks, etc, and added a bypass for our custom `devops` role which is just a `Write` role + the `bypass branch protection` permission.
  - Actually added another nicety, which is a requirement to resolve all comments before merging 
  - It also has some cool options to enforce branch naming conventions.
4. Give our `ampersand-ops` user the required role (this is currently not possible, context ahead)

### Other fixes
1. Seems like the action also doesn't work without a couple of write permissions.
2. The script was generating `internal/generated/provider_count.go`, but this action wasn't committing it. Adds it in the automated commit.
3. Tweaks the commit message to be better.
4. You also need to checkout a repo with the PAT you intend to use.

### Other blockers
- Our enterprise (for some unknown reason) got downgraded to a free plan (ticket raised about this), so we don't have access to custom roles anymore. Once this is resolved, I will assign the `ampersand-ops` user its deserved custom `devops` role. Until then, the action will not work. 

## Testing
Temporarily, I let the `Write` role bypass the ruleset and assigned it to `ampersand-ops`, and it was able to finally auto-commit things -


RajatPawar:
* **#677** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/677?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @RajatPawar and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

eberle1080 (approved):

laurenzlong:
I gave `ampersand-ops` admin role on this repo, and now the Github Action is working again: https://github.com/amp-labs/connectors/commit/e25f29893d44ad6e0f132ad72c5b19d407e8aebb

RajatPawar:
> I gave `ampersand-ops` admin role on this repo, and now the Github Action is working again: https://github.com/amp-labs/connectors/commit/e25f29893d44ad6e0f132ad72c5b19d407e8aebb. Custom roles are no longer available due to the (intentional, albeit poorly communicated on my part 😛 ) downgrade from Github Enterprise to Github Team. 
Thanks! I did freak out for a second when I saw that amp-labs was removed from Ampersand haha

PR #704, opened 2024-07-18
fix: DynamicsCRM ListObjectMetadata readable fields only, add DisplayNames

Cobalt0s:
# Problem

Not all fields returned by `ListObjectMetadata` can be used by `Read`. There are multiple endpoints to get metadata each has its differences.
Its funny how some fields are not allowed to be used for query parameters on Collection resource but on Single resource would ignore and not produce an error. We have to deal with it.

# Solution

I found a total replacment of XML schema request ($metadata endpoint) by using EntityDefinitions.Attributes endpoint with certian parameters.
After investigation I developed steps to return fields just like before, but taking into account the problem described above!

Here is a short pseudo algorithm. For every object name:
* make request to `EntityDefinitions` to get DisplayName of the object
* make request to `EntityDefinitions.Attributes` such that
    * IsValidODataAttribute is true - field can be found in Read response
    * IsValidForRead is true - is allowed to be used in query fields

Algorithm for infering display name of an Attribute:
* Pick Localized Display name if available, otherwise
* Pick SchemaName, otherwise
* Default: display name will be same as field name.

## Edge Case
There is one more extra edge case. Addressing it opened a possiblity of a **full switch to the new endpoint**.
Some fields in Read payload are in the form of `_<some-name>_value`. Contacts has 25 underscore fields. There are 25 attributes with `Targets` field. Turns out this is the patern among entities. In conclusion, we manually format the field, while harnessing human readable display names.

Ex: Attribute name is `createdby`, since it has targets, the field name is `_createdby_value` with `Created By` display.

# Tests

The script of comparying set of fields from Read and ListObjectMetadata is no longer applicable. Reason is ListObjectMetadata can be only a subset of Read because of `IsValidForRead` rule.
A new test asks Metadata first. Then uses output to query fields by Read method. This test also fulfills the direct purpose of ListObjectMetadata.

Mock tests imitate 2 API calls triggered by ListObjectMetadata.


common/json.go
```diff
@@ -104,7 +104,12 @@ func (j *JSONHTTPClient) Delete(ctx context.Context, url string, headers ...Head
 func parseJSONResponse(res *http.Response, body []byte) (*JSONHTTPResponse, error) {
 	// empty response body should not be parsed as JSON since it will cause ajson to err
 	if len(body) == 0 {
-		return nil, nil //nolint:nilnil
+		return &JSONHTTPResponse{
```
Cobalt0s, line 107: It caused a panic during testing. It can happen in prod. This was something preserved during refactoring but not addressed.
At least the caller should check if Body is nil. Sometimes GET can return nothing, shouldn't make assumptions.

test/dynamicscrm/metadata/main.go
```diff
@@ -5,18 +5,16 @@ import (
...
 	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/handy"
 	connTest "github.com/amp-labs/connectors/test/dynamicscrm"
 	"github.com/amp-labs/connectors/test/utils"
-	"github.com/amp-labs/connectors/test/utils/mockutils"
 )
 
 var (
-	objectName = "accounts"
+	objectName = "contacts"
```
Cobalt0s, line 17: Contacts is problematic it has `subscriptionid` which is not valid for read.

IsValidForRead = False

common/handy/maps.go
```diff
@@ -0,0 +1,45 @@
+package handy
```
RajatPawar, line 1: Could you put these in utils or rename this as utils?

common/handy/maps.go
```diff
@@ -0,0 +1,45 @@
...
+
+type Map[K comparable, V any] map[K]V
```
RajatPawar, line 3: Let's use something battle tested like https://github.com/deckarep/golang-set? Could be pretty handy later on

dynamicscrm/connector.go
```diff
@@ -64,6 +61,20 @@ func (c *Connector) getURL(arg string) (*urlbuilder.URL, error) {
 	return constructURL(c.BaseURL, apiVersion, arg)
 }
 
+func (c *Connector) getEntityDefinitionURL(arg naming.SingularString) (*urlbuilder.URL, error) {
```
RajatPawar, line 64: Can you please add comments above `getEntityDefinitionURL` just like you have for `getEntityAttributesURL` describing what it does?

dynamicscrm/metadata.go
```diff
@@ -34,109 +34,181 @@ func (c *Connector) listObjectMetadata(
 		return nil, common.ErrMissingObjects
 	}
 
-	url, err := c.getURL("$metadata")
+	// Collect display name for the Object and its fields (name+display)
+	result := map[string]common.ObjectMetadata{}
+
+	for _, objectName := range objectNames {
+		objectDisplayName, err := c.metadataGetObjectDisplayName(ctx, objectName)
```
RajatPawar, line 41: We could skip the `metadata` prefix, `getObjectDisplayName` is verbose enough

dynamicscrm/metadata.go
```diff
@@ -34,109 +34,181 @@ func (c *Connector) listObjectMetadata(
...
+		if err != nil {
+			return nil, err
+		}
+
+		fields, err := c.metadataGetFields(ctx, objectName)
```
RajatPawar, line 46: Likewise, we could skip the `metadata` prefix and call this `getFieldsForObject` or something instead

dynamicscrm/metadata.go
```diff
@@ -34,109 +34,181 @@ func (c *Connector) listObjectMetadata(
...
+		if err != nil {
+			return nil, err
+		}
+
+		result[objectName.Plural().String()] = common.ObjectMetadata{
```
RajatPawar, line 51: Can you add a callout on why the plural() is being done here?

dynamicscrm/metadata.go
```diff
@@ -34,109 +34,181 @@ func (c *Connector) listObjectMetadata(
...
 		return nil, err
 	}
 
-	result, err := convertEntitySetToMetadataSet(objectNames, entities)
+	// Our goal is to get properties that can be found in Read response.
+	// Not all Attributes are real fields.
+	// IsValidODataAttribute == true -> filters fields returned in the Read payload.
+	// Now we have all good fields that certainly will be in payload.
+	// But if they are returned doesn't mean you can request them!
+	// IsValidForRead == true -> filters fields that can be requested in GET query params.
```
RajatPawar, line 101:
```suggestion
// Filter attributes to ensure they are:
// 1. Present in Read responses (IsValidODataAttribute == true)
// 2. Can be queried in GET requests (IsValidForRead == true)
// This ensures we only work with fields that are both 
// returned in the payload and can be used in query parameters.
```

dynamicscrm/metadata.go
```diff
@@ -34,109 +34,181 @@ func (c *Connector) listObjectMetadata(
...
-	for _, abstractEntity := range root.FindMany(queryListAbstractEntities) {
-		if !abstractEntity.HasChildren() {
-			// these entities were not included by previous query as they have no properties
-			// we programmatically find these special types, which are "primary values" but for structs
-			// Ex: crmbaseentity, crmmodelbaseentity,
-			entityName := abstractEntity.Attr("Name")
-			// effectively only create
-			_ = entities.GetOrCreate(entityName, "")
+// Given Collection of attributes from endpoint response, select those applicable to Read response.
+func extractFields(attributes *ajson.Node) (map[string]string, error) {
```
RajatPawar, line 120: `selectValidReadFields` or something similar would read better

dynamicscrm/metadata.go
```diff
@@ -34,109 +34,181 @@ func (c *Connector) listObjectMetadata(
...
 	}
 
-	// link every child with parent completing hierarchy
-	schemaAlias := schema.Attr("Alias")
-	if err := entities.MatchParentsWithChildren(schemaAlias); err != nil {
-		return nil, errors.Join(ErrMetadataProcessing, err)
+	return fieldsMap, nil
+}
+
+func getObjectDisplayName(item *ajson.Node, objectName naming.SingularString) (string, error) {
```
RajatPawar, line 145: There's a lot of stuff going on here, and it might make sense to separate out these calls into another file - `entityDefinitions.go` or something - where you could summarize the steps we are going through to make metadata collection happen (the more examples, the better)

dynamicscrm/metadata_test.go
```diff
@@ -40,119 +42,96 @@ func TestListObjectMetadata(t *testing.T) { // nolint:funlen,gocognit,cyclop
...
+			input: []string{"accounts"},
 			server: httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 				w.WriteHeader(http.StatusTeapot)
 			})),
 			expectedErrs: []error{interpreter.ErrMissingContentType},
 		},
 		{
-			name:  "Missing XML response on status OK",
-			input: []string{"msfp_surveyinvite"},
+			name:  "Schema endpoint is not available for object",
```
RajatPawar, line 52: Can you mention in the metadata code how we currently handle objects that don't respond to schema endpoints?

RajatPawar (approved):

RajatPawar:
Left a couple of comments, please make the changes wherever applicable (or drop a comment if you think they aren't applicable) before merging.

common/handy/maps.go
Cobalt0s: There are too many utils I wanted to limit increase of that package.
I wasn't sure about adding the library for data structures. But more or less thats what it is. I will incorporate it.

common/handy/maps.go
Cobalt0s: It doesn't use generics but rather `interface{}`. Type checking is much needed.
Lets keep this implementation for now. Later we could replace it with alternative.

dynamicscrm/metadata.go
Cobalt0s: Changed the comment it was a bit misleading. Function is `extractFieldsFromJSON`. The input is parsed, there is no selection going on. API request covered the filtering part.

dynamicscrm/metadata_test.go
Cobalt0s: Added a comment at the top of `ListObjectMetadata` method.

dynamicscrm/metadata.go
Cobalt0s: This is a great idea. I have split methods into 2 groups. EntityDefinition vs EntityAttributes.
Metadata file uses both, with 1 main method. This should spook less people if they will open it.

Cobalt0s:
* **#704** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/704?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #710, opened 2024-07-22
feat: Deep connector Jira (Read)

### LINKED ISSUE #709: Add Jira deep connector
- GetPostAuthInfo
- ListObjectMetadata
- Read
- Write

Cobalt0s:
Closes #709 

# Description

Read operation uses search API endpoint to return Jira issues.
The data returned from `issue endpoint` is not returned untouched.

Fields property contains data of interest, the gist of it.
Id property is manually attached. This is the only relevant property outside of fields.

# Testing

Manual script to retrieve list of issues.
image

### DIFF
--- a/atlassian/connector.go
+++ b/atlassian/connector.go
@@ -1,6 +1,7 @@
 package atlassian
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/amp-labs/connectors/common"
@@ -9,6 +10,9 @@ import (
 	"github.com/amp-labs/connectors/common/urlbuilder"
 	"github.com/amp-labs/connectors/providers"
 )
+
+// ErrMissingCloudId happens when cloud id was not provided via WithMetadata.
+var ErrMissingCloudId = errors.New("connector missing cloud id")
 
 type Connector struct {
 	Client  *common.JSONHTTPClient
@@ -66,6 +70,17 @@ func (c *Connector) String() string {
 	return fmt.Sprintf("%s.Connector[%s]", c.Provider(), c.Module)
 }
 
+// URL format follows structure applicable to Oauth2 Atlassian apps.
+// https://developer.atlassian.com/cloud/jira/platform/rest/v2/intro/#other-integrations
+func (c *Connector) getJiraRestApiURL(arg string) (*urlbuilder.URL, error) {
+	cloudId, err := c.getCloudId()
+	if err != nil {
+		return nil, err
+	}
+
+	return constructURL(c.BaseURL, "ex/jira", cloudId, c.Module, arg)
+}
+
 // URL allows to get list of sites associated with auth token.
 // https://developer.atlassian.com/cloud/confluence/oauth-2-3lo-apps/#3-1-get-the-cloudid-for-your-site
 func (c *Connector) getAccessibleSitesURL() (*urlbuilder.URL, error) {
@@ -75,4 +90,12 @@ func (c *Connector) getAccessibleSitesURL() (*urlbuilder.URL, error) {
 func (c *Connector) setBaseURL(newURL string) {
 	c.BaseURL = newURL
 	c.Client.HTTPClient.Base = newURL
+}
+
+func (c *Connector) getCloudId() (string, error) {
+	if len(c.cloudId) == 0 {
+		return "", ErrMissingCloudId
+	}
+
+	return c.cloudId, nil
 }
--- a/atlassian/parse.go
+++ b/atlassian/parse.go
@@ -1,0 +1,109 @@
+package atlassian
+
+import (
+	"errors"
+	"strconv"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/jsonquery"
+	"github.com/spyzhov/ajson"
+)
+
+func getTotalSize(node *ajson.Node) (int64, error) {
+	return jsonquery.New(node).ArraySize("issues")
+}
+
+/*
+Records cannot be returned as is from the API. Extra processing is described below.
+ 1. First of all, main properties are located under "fields" key.
+ 2. Secondly, item id is not mirrored under "fields" property, which means
+    unwrapping is not enough, we must attach "id" property.
+
+Visual example of what will happen to each property:
+
+	{
+		"id": "", 					=> preserved
+		"expand": "",				=> removed
+		"self": "",					=> removed
+		"key": "",					=> removed
+		"fields": {					=> unwrapped/flattened
+			"project": "",			=> moved outside
+			"lastViewed": ""		=> moved outside
+			...						...
+			...						=> moved outside
+		}
+	}
+*/
+func getRecords(node *ajson.Node) ([]map[string]any, error) {
+	arr, err := jsonquery.New(node).Array("issues", false)
+	if err != nil {
+		return nil, err
+	}
+
+	list := make([]map[string]any, len(arr))
+
+	for index, item := range arr {
+		fieldsObject, err := jsonquery.New(item).Object("fields", false)
+		if err != nil {
+			return nil, errors.Join(common.ErrParseError, err)
+		}
+
+		fields, err := jsonquery.Convertor.ObjectToMap(fieldsObject)
+		if err != nil {
+			return nil, errors.Join(common.ErrParseError, err)
+		}
+
+		id, err := jsonquery.New(item).Str("id", false)
+		if err != nil {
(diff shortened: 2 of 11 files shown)


atlassian/test/cloud2.json
```diff
@@ -0,0 +1,30 @@
+[
+  {
+    "id": "9e1477fd-54ef-41fe-b747-bc9e6a11a925",
```
Cobalt0s, line 3: This is what token introspeciton returns.
There are options of what container to choose. I employed the strategy of selecting container by workspace name. 
Id means CloudID.

atlassian/modules.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"github.com/amp-labs/connectors/common/paramsbuilder"
+)
+
+// Atlassian's modules can include CloudID variable.
+// Token discovery API will be called to substitute CloudID.
+
+// ModuleJira is the module used for accessing standard CRM objects.
+var ModuleJira = paramsbuilder.APIModule{ // nolint: gochecknoglobals
+	Label:   "ex/jira/{{.cloudID}}/rest/api",
```
Cobalt0s, line 12: @RajatPawar do you think the having a module with variables is a good approach?

atlassian/parse.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/jsonquery"
+	"github.com/spyzhov/ajson"
+	"strconv"
+)
+
+func getTotalSize(node *ajson.Node) (int64, error) {
+	return jsonquery.New(node).ArraySize("issues")
```
Cobalt0s, line 11: I am not sure about general solution for response parsing.
There should be a dependency between Atlassian module and response format.

For now we make only 1 read call for "issues".

atlassian/connector.go
```diff
@@ -0,0 +1,85 @@
...
+	}.Handle
+
+	// CloudID is dynamic data and cannot be supplied via ProviderInfo.
+	// This information will be filled by making discovery API calls.
+	cloudID, err := conn.getCloudID(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	conn.Module, err = params.Module.ResolveSuffix(map[string]string{
```
Cobalt0s, line 56: Module is a template awaiting for cloud id variable.
Connector hold this state to construct URLs. See methods below in this file.

Cobalt0s (commented):
@RajatPawar ignore test files they are autogenerated.
Current code allows reading Jira issues via Search api call. 
(need to address lots of TODOs)

atlassian/modules.go
RajatPawar: Can this be a part of the base URL? I'd like us to do modules a little more comprehensively ideally. In this particular case, if it becomes a part of the base URL, it just is a workspace var and can behave like docusign.

Cobalt0s: I am struggling with balancing 2 things. Jira APIs and whatever should come in like Confluence. I would honestly introduce modules only after working on the second deep connector.

I was thinking that BaseURLs are shared among 2 while each has its own structure one could be `ex/jira` the other something else. Checking [Confluence docs](https://developer.atlassian.com/cloud/confluence/rest/v2/intro/#auth) it seems that base URL differs entierly.

I could keep `rest/api/3` as module for now.

Cobalt0s: It is a bit confusing to have `WithModule(.)` if there is only 1 or raw aka empty. Honestly, all connectors fall under this category. Only some will have `{ empty, defaultVersion, option[1...many] }.`

Cobalt0s:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/710?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#710** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/710?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#712** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/712?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#713](https://github.com/amp-labs/connectors/pull/713) <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/713?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#711** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/711?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

common/jsonquery/converters.go
```diff
@@ -66,3 +66,14 @@ func (convertor) ObjectToMap(node *ajson.Node) (map[string]any, error) {
 
 	return result, nil
 }
+
+// NodeToArray treats current node as an array of objects.
+// Sometimes JSON response from the server right of the bet is an array. This method performs the necessary conversion.
+func (convertor) NodeToArray(node *ajson.Node) ([]*ajson.Node, error) {
```
Cobalt0s, line 72: This is a small utility function. 
When getting `{baseURL}/oauth/token/accessible-resources` the response is array `[]` instead of `{}`.
Thats the primary reason for this method. Existing methods to get array expect the key, while in this case it is "root", "dot".

providers/utils.go
```diff
@@ -156,14 +157,17 @@ func (i *ProviderInfo) SubstituteWith(vars []paramsbuilder.CatalogVariable) erro
 	return paramsbuilder.NewCatalogSubstitutionRegistry(vars).Apply(i)
 }
 
-func (i *ProviderInfo) GetOption(key string) (string, bool) {
+func (i *ProviderInfo) GetOption(key string) (string, error) {
```
Cobalt0s, line 160: This method was not used. I repurposed it and it came in handy.
I verified that server repo doesn't use it either.

atlassian/connector.go
```diff
@@ -66,6 +66,10 @@ func (c *Connector) String() string {
 	return fmt.Sprintf("%s.Connector[%s]", c.Provider(), c.Module)
 }
 
+func (c *Connector) getJiraRestApiURL(arg string) (*urlbuilder.URL, error) {
```
jlimatampersand, line 69: In comment, can we have link to jira api doc where it explains the endpoint?

atlassian/parse.go
```diff
@@ -0,0 +1,109 @@
...
+
+/*
+Records cannot be returned as is from the API. Extra processing is described below.
+ 1. First of all, main properties are located under "fields" key.
+ 2. Secondly, item id is not mirrored under "fields" property, which means
+    unwrapping is not enough, we must attach "id" property.
+
+Visual example of what will happen to each property:
+
+	{
```
jlimatampersand, line 24: thanks for example. This is super helpful!

atlassian/parse.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/amp-labs/connectors/common/jsonquery"
+	"github.com/spyzhov/ajson"
+)
+
+func getTotalSize(node *ajson.Node) (int64, error) {
+	return jsonquery.New(node).ArraySize("issues")
+}
+
+/*
+Records cannot be returned as is from the API. Extra processing is described below.
```
jlimatampersand, line 17: What do you mean records cannot be returned as is? Why is that?
Can you add the link to the doc as well?

jlimatampersand: If that is what the API is returning, then we should return as is since that is what the users will be expecting. @laurenzlong Please correct me if I am wrong.

jlimatampersand (commented):
Overall looks good, but not sure about changing object schema. I think we shouldn't unless there is some requirements that I am not aware of. Please let me know if there is any.

All looks good, but I will leave the object transformation part to @laurenzlong.

jlimatampersand (approved):
I just had chat about flattening the data with @laurenzlong This is good.

jlimatampersand:
But please do provide links to docs in the comments before merging this. Thanks.

atlassian/parse.go
Cobalt0s: I remember some connector long ago had alike problem, hence I went with such approach.

atlassian/parse.go
Cobalt0s: :pray:

PR #724, opened 2024-07-26
fix: Outreach read to be able to filter by time

Jkarage:
### MailMonkey Cases:

Set-Up:
Screenshot 2024-07-26 at 09 52 23

Authentication:
Screenshot 2024-07-26 at 09 52 43

Installations:
Screenshot 2024-07-26 at 09 54 35

Screenshot 2024-07-26 at 09 54 46

Screenshot 2024-07-26 at 09 55 01

Screenshot 2024-07-26 at 09 55 31

## The Ampersand Console:
Screenshot 2024-07-26 at 09 41 24

## The Webhook Response:
Screenshot 2024-07-26 at 09 39 57

Screenshot 2024-07-26 at 09 40 17

Screenshot 2024-07-26 at 09 40 32

### Accounts 
Screenshot 2024-07-26 at 21 43 37


laurenzlong (commented):
As mentioned on Slack, please add `optionalFieldsAuto: true ` to at least one of the objects and then make a screenshot of that tab of the UI library. I also don't see a svix screenshot for `accounts`, did you test that?

Jkarage:
@laurenzlong  Added the `accounts` webhook response. On the `optionalFieldsAuto: all` I did add that in the `sequence` object.

RajatPawar (dismissed):

laurenzlong (changes requested):
2 things I noticed:
- for the sequence object, I would have expected all the fields in https://developers.outreach.io/api/reference/tag/Sequence/#tag/Sequence/section/Sequence-Resource-Metadata to be shown in the list, but only `type` and `links` are shown, this is unexpected. 
- the `Since` field of ReadParams is not being used at all. When `Since` is provided, the `Read` function should only return records that have been updated **after** that timestamp. (See the Salesforce connector for an example). In rare cases where the API does not support filtering by timestamp, but from my quick research it looks like that the Outreach API does support it.

There is a [filter query param](https://developers.outreach.io/api/making-requests/#filter-by-exact-attribute) that you can attach to GET requests, and I see that objects have an `updatedAt` field you can filter & sort by. Can you please:
1) figure out why ListObjectMetadata is not returning the full list of fields (which is what would have caused the "Sequence" tab in the UI library to not show all the available fields
2) revise the `Read` function to filter for only records that have been updated after the `Since` field.

Jkarage:
I have worked on the since parameter as you suggested. Regarding the metadata, all top level-fields are captured, Unless we want to filter inner fields as well.I have added more details on slack.

outreach/read.go
```diff
@@ -27,11 +29,24 @@ func (c *Connector) Read(ctx context.Context, config common.ReadParams) (*common
...
 	}
 
-	return c.getApiURL(config.ObjectName)
+	url, err := c.getApiURL(config.ObjectName)
+	if err != nil {
+		return nil, err
+	}
+
+	// If Since is not set, then we're doing a backfill. We read all rows (in pages)
+	// Making sure the time passed, is in the format outreach expects.
```
laurenzlong, line 44:
```suggestion
	// Making that Since is in the format the Outreach API expects, and that the time is in the past.
```

test/outreach/connector.go
```diff
@@ -0,0 +1,56 @@
+package outreach_test
```
laurenzlong, line 1: this package should be called `outreach` to be consistent with the other packages in this `test` directory.

test/outreach/metadata/responses/metadata.json
```diff
@@ -0,0 +1,35 @@
+{
```
laurenzlong, line 1: We actually only want to return `id` plus the subfields within `attributes`, since those are the "real" fields of the object that hold a value. so the expected response is:

```
  "Result": {
    "mailings": {
      "DisplayName": "mailings",
      "FieldsMap": {
        "id": "id",
        "bodyHtml": "bodyHtml",
        "bodyText: "bodyText",
        "bouncedAt": "bouncedAt",
       // All the other fields in https://developers.outreach.io/api/reference/tag/Mailing/
      }
    },
   // ...
}
```

test/outreach/read/read.go
```diff
@@ -0,0 +1,70 @@
...
+		log.Fatal(err)
+	}
+}
+
+func testReadConnector(ctx context.Context, conn connectors.ReadConnector) error {
+	config := connectors.ReadParams{
+		ObjectName: "sequences",
+		Since:      time.Now().Add(-720 * time.Hour), // Goes back a Month.
+		Fields:     []string{"type", "id", "relationships"},
+		// NextPage:   "https://api.outreach.io/api/v2/users?page%5Blimit%5D=1\u0026page%5Boffset%5D=2",
```
laurenzlong, line 52: Please remove this line

test/outreach/read/sample_responses/sequence-using-since.json
```diff
@@ -0,0 +1,611 @@
+{
+  "rows": 5,
+  "data": [
+    {
+      "fields": {},
```
laurenzlong, line 5: `field` should not be empty, it should contain the values of the fields that were provided to ReadParams. 

For example, if ReadParams was:

```
connectors.ReadParams{
		ObjectName: "sequences",
		Fields:     []string{"id", "maxActivations", "createdAt"},
```
Then the output should be:
```
{
   "rows": 5,
   "data": [
      {
         "fields": {
                  "id": 1,
                  "maxActivations": 150,
                  "createdAt": "2024-07-26T06:36:16.000Z"
         }
        "raw": {
           // same as what you already have
        }
}
```

test/outreach/metadata/responses/metadata.json
laurenzlong, line 1: Can you also name this folder `sample_responses`?

test/outreach/read/sample_responses/sequnce-uses-both-since-fields.json
```diff
@@ -0,0 +1,946 @@
+{
+  "rows": 5,
```
laurenzlong, line 2: we don't need so many sample response files, just `sequences.json` and `sequences-using-since.json` is good enough. Also What does "sequences using both since fields" actually mean? What was the input to this file?

laurenzlong (changes requested):

outreach/read.go
```diff
@@ -27,11 +34,24 @@ func (c *Connector) Read(ctx context.Context, config common.ReadParams) (*common
...
 	}
 
-	return c.getApiURL(config.ObjectName)
+	url, err := c.getApiURL(config.ObjectName)
+	if err != nil {
+		return nil, err
+	}
+
+	// If Since is not set, then we're doing a backfill. We read all rows (in pages)
+	// Making that Since is in the format the Outreach API expects, and that the time is in the past.
```
laurenzlong, line 49:
```suggestion
	// If Since is present, we turn it into the format the Outreach API expects
```

test/outreach/metadata/sample_responses/metadata.json
```diff
@@ -0,0 +1,35 @@
+{
```
laurenzlong, line 1: This sample response appears outdated.

laurenzlong (approved):

PR #729, opened 2024-07-27
add: Logo - Zendesk Support

Cobalt0s:
# Notes

Each mode shares the same logo, icon.
**Reason:** Icon looks better than logo (has no name).

**zendeskSupport** and **zendeskSupport** use identical Media. 2 providers though 1 company.

Selection_648

### DIFF
--- a/providers/zendesk.go
+++ b/providers/zendesk.go
@@ -1,11 +1,11 @@
 package providers
 
 const (
-	ZendeskChat    Provider = "zendeskChat"
+	ZendeskChat    Provider = "zendeskSupport"
 	ZendeskSupport Provider = "zendeskSupport"
 )
 
-func init() {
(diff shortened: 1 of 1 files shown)


providers/zendesk.go
```diff
@@ -1,11 +1,11 @@
 package providers
 
 const (
-	ZendeskChat    Provider = "zendeskChat"
+	ZendeskChat    Provider = "zendeskSupport"
```
eberle1080, line 4: This seems like a bug. `ZendeskChat` and `ZendeskChat` both map to the same value.

Cobalt0s: Likely when I was copy pasting I did this crime :hot_face:

eberle1080 (approved):

PR #820, opened 2024-08-01
feat: rename Module.Suffix to Module.Name

laurenzlong:
Sometimes sub-product APIs might have different prefixes, not necessarily suffixes. Also the goal of this field is represents the name of the product that a caller of the library puts into the `WithModule` function, so name is more appropriate.


Cobalt0s (approved):
This makes sense.
Parameters are used by composition in dedicated connector, where I use full path to values anyway, so it is safe, no collisions.
Ex:
```
parameters.Workspace.Name
parameters.Module.Name
```

PR #837, opened 2024-08-02
fix: update writing (outreach)

Jkarage:
This updates the Outreach Write connector. 

Changes:
- All the attributes data are expected at the top level field of the requestData. 
- The `relationships` field as well.
- Add checks for the `type` field, includes it if not available.
- Adds `id` to the request body for updating.


test/outreach/write/sample_response/prospects.json
```diff
@@ -0,0 +1,443 @@
+{
+  "success": true,
+  "recordId": "9",
+  "data": {
+    "attributes": {
```
Jkarage, line 5: @laurenzlong should I update the response to include the `attributes` fields as top-level fields in the `data` response?

outreach/errors.go
```diff
@@ -10,4 +10,6 @@ var (
 	ErrNotObject     = errors.New("record is not an object")
 	ErrNotString     = errors.New("next is not a string")
 	ErrEmptyResponse = errors.New("empty response body")
+	ErrIdMustInt     = errors.New("provided id must be convertable to int")
+	ErrMustJSON      = errors.New("request data couldn't be decoded to json")
```
Cobalt0s, line 14: There is `common.ErrNotJSON` but it is about response. This error could belong in `common`.
I think it is possible for other connectors to care about request being not json, should they parse it.

Cobalt0s (commented):
Hi, I came across your PR. Some minor suggestions.

outreach/write.go
```diff
@@ -18,25 +18,36 @@ var JSONAPIContentTypeHeader = common.Header{ //nolint:gochecknoglobals
...
 		return nil, err
 	}
 
+	// prepares the updating data request.
 	if len(config.RecordId) > 0 {
 		url.AddPath(config.RecordId)
 
 		write = c.Client.Patch
+
+		data, err = parseData(config.RecordData, config.ObjectName, config.RecordId)
```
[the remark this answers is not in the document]
Jkarage: Thanks, I also did not like slice of ids... What do you say if I just change the signature of this function to be `parseData(config)`. Since am using all the writeParam details anyway.  And this will remove the slice of ids concern.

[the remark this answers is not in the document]
Cobalt0s: You know that's even better.

test/outreach/write/sample_response/prospects.json
laurenzlong: this is good as-is! we want the result of the Write to be the raw response from the API

laurenzlong:
> Hi, I came across your PR. Some minor suggestions.

laurenzlong (approved):
Thanks for the helpful comments @Cobalt0s and thanks for the PR @Jkarage ! I've added a commit that cleans things up a bit and will go ahead to merge this PR so since there's a customer waiting on it. @Jkarage if you can create a follow up PR to address[ the comment on the test](https://github.com/amp-labs/connectors/pull/837#discussion_r1701812498) that would be great!

PR #843, opened 2024-08-05
fix: Outreach Read to include Id

Jkarage:
This updates the read connector, 
- On the previous PR, The output did not include the `id` field as recommended.  This adds it.
- Adds more tests.


laurenzlong (approved):
Good catch!

PR #888, opened 2024-08-09
feat: Handle empty HTTP response

Cobalt0s:
# Problem
Any API call may return empty response. We are not explicit about it. And caller may forget to do nil check.

# Changes
I hid `common.JSONHTTPResponse.Body` field. It enforces access via `GetBody` which returns a flag indicating if response body was empty. This makes error handling explicit. Implementor of a deep connector must decide what empty response body signifies and how to address it.

# Result
The content of response body can be accessed 
* as *ajson.Node by calling `GetBody` (with flag check)
* or loaded into struct via `common.UnmarshalJSON` (empty body means empty struct, tests illustrate that)

# Review
Commits are grouped into related changes. Easier to review commit by commit.

### DIFF
--- a/common/json.go
+++ b/common/json.go
@@ -3,6 +3,7 @@ package common
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"mime"
 	"net/http"
@@ -18,6 +19,8 @@ type JSONHTTPClient struct {
 }
 
 // JSONHTTPResponse is a JSON response from an HTTP request.
+// Consider using Body to operate on ajson.Node or
+// Unmarshal into the struct of your choosing via UnmarshalJSON.
 type JSONHTTPResponse struct {
 	// bodyBytes is the raw response body. It's not JSON-unmarshalled.
 	// We keep it around so that we can unmarshal it into a struct later,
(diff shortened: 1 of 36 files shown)


providers/salesforce/write_test.go
```diff
@@ -62,21 +62,6 @@ func TestWrite(t *testing.T) { // nolint:funlen,cyclop
 				errors.New("Unable to create/update fields: MasterRecordId"), // nolint:goerr113
 			},
 		},
-		{
-			Name:  "Write must act as a Create",
```
Cobalt0s, line 66: A bit duplicate to another test at the end

common/json.go
```diff
@@ -30,9 +33,20 @@ type JSONHTTPResponse struct {
...
+	// body is the JSON-unmarshalled response body. Aside from the fact
 	// that it's JSON-unmarshalled, it's identical to bodyBytes.
-	Body *ajson.Node
+	// If there were no bytes this will be nil.
+	body *ajson.Node
+}
+
+// GetBody returns JSON node. If it is empty the flag will indicate so.
+// Empty response body is a special case and should be handled explicitly.
+func (j *JSONHTTPResponse) GetBody() (*ajson.Node, bool) {
```
RajatPawar, line 44: Let's just call this Body() (https://go.dev/doc/effective_go#Getters)

RajatPawar (approved):
LGTM, minor comments to resolve before merging

common/json.go
Cobalt0s: I like this!

PR #920, opened 2024-08-15
feat: Add clientcredentials OAuth2 option

Jkarage:
This PR adds 2 legged as an option in creating the Authenticated client using OAuth2. This will allow adding deep connectors implementing Client Credentials Grant Type.

### DIFF
--- a/common/clientcredentials.go
+++ b/common/clientcredentials.go
@@ -1,0 +1,27 @@
+package common
+
+import (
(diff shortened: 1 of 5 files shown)


eberle1080:
This isn't necessary, as long as it can accept an `AuthenticatedClient` instance. The client creation logic is already implemented.

PR #921, opened 2024-08-15
feat: Add Read (Marketo deep connector)

### LINKED ISSUE #908: Add deep connector for Marketo
Include:
- ListObjectMetadata
- Read
- Write

Jkarage:
This adds Read capability for the Marketo deep connector. 
Note:
- Marketo API responds with 200 Ok, for most cases, even when having errors.  Created a mapping function for mapping native marketo error codes to http status codes. So then the server can determine what to do with the response.

-doc link: https://experienceleague.adobe.com/en/docs/marketo-developer/marketo/rest/error-codes 

Sample test when the response from Marketo is 200, But the custom error handler returns a 404.
Screenshot 2024-08-15 at 18 53 40

Closes #908

### DIFF
--- a/common/http.go
+++ b/common/http.go
@@ -262,7 +262,7 @@ func makeDeleteRequest(ctx context.Context, url string, headers []Header) (*http
 }
 
 // sendRequest sends the given request and returns the response & response body.
-func (h *HTTPClient) sendRequest(req *http.Request) (*http.Response, []byte, error) {
+func (h *HTTPClient) sendRequest(req *http.Request) (*http.Response, []byte, error) { //nolint:cyclop
 	// Send the request
 	res, err := h.Client.Do(req)
 	if err != nil {
--- a/common/http.go
+++ b/common/http.go
@@ -31,6 +31,7 @@ type HTTPClient struct {
 	Base         string                  // optional base URL. If not set, then all URLs must be absolute.
 	Client       AuthenticatedHTTPClient // underlying HTTP client. Required.
 	ErrorHandler ErrorHandler            // optional error handler. If not set, then the default error handler is used.
+	OKStatusErr  bool
 }
 
 // getURL returns the base prefixed URL.
@@ -283,11 +284,19 @@ func (h *HTTPClient) sendRequest(req *http.Request) (*http.Response, []byte, err
 		return nil, nil, fmt.Errorf("error reading response body: %w", err)
 	}
 
-	if h.ErrorHandler != nil {
+	if h.OKStatusErr && (res.StatusCode >= 200 && res.StatusCode <= 299) {
 		return res, body, h.ErrorHandler(res, body)
 	}
 
-	return res, body, InterpretError(res, body)
+	if res.StatusCode < 200 || res.StatusCode > 299 {
+		if h.ErrorHandler != nil {
+			return res, body, h.ErrorHandler(res, body)
+		}
+
+		return res, body, InterpretError(res, body)
+	}
+
+	return res, body, nil
 }
 
 // getURL returns the given URL if it is an absolute URL, or the given URL joined with the base URL.
--- a/common/http.go
+++ b/common/http.go
@@ -26,12 +26,14 @@ type Header struct {
 // to the error handler as arguments.
 type ErrorHandler func(rsp *http.Response, body []byte) error
 
+type ResponseHandler func(rsp *http.Response) (*http.Response, error)
+
 // HTTPClient is an HTTP client that handles OAuth access token refreshes.
 type HTTPClient struct {
-	Base         string                  // optional base URL. If not set, then all URLs must be absolute.
-	Client       AuthenticatedHTTPClient // underlying HTTP client. Required.
-	ErrorHandler ErrorHandler            // optional error handler. If not set, then the default error handler is used.
-	OKStatusErr  bool                    // optional, If set true indicates 200OK http response may include errors.
+	Base            string                  // optional base URL. If not set, then all URLs must be absolute.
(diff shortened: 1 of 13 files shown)


common/http.go
```diff
@@ -283,16 +283,11 @@ func (h *HTTPClient) sendRequest(req *http.Request) (*http.Response, []byte, err
 		return nil, nil, fmt.Errorf("error reading response body: %w", err)
 	}
 
-	// Check the response status code
-	if res.StatusCode < 200 || res.StatusCode > 299 {
```
Jkarage, line 287: Moved this logic to the error handler func as I want, to check for errors, even when status code is 200.

laurenzlong:
@Jkarage looks like tests are not passing, can you please fix?

laurenzlong:
@Jkarage build still failing, please re-request review from Rajat once you fix it.

Jkarage:
Closing this and reopening a new one(test works in the new one), I can't seem to find the reason why the test is failing in this one.

marketo/errors.go
```diff
@@ -0,0 +1,154 @@
...
+
+	return common.NewHTTPStatusError(res.StatusCode, fmt.Errorf("%w: %s", common.ErrUnknown, string(body)))
+}
+
+// statusCodeMap maps the erroneous response from marketo, with a valid http status code.
+// The response body can be sent as is.
+// https://experienceleague.adobe.com/en/docs/marketo-developer/marketo/rest/error-codes
+func statusCodeMap(code int) int { //nolint:funlen,cyclop
+	switch code {
+	case 502: //nolint:gomnd
```
RajatPawar, line 93: This must have been painful, thanks for the hard work! :) A quick tip would be to use a function level no lint (just add it one line before the function) to disable the check for the entire function, instead of adding it per line. 

ref - https://github.com/golangci/golangci-lint/discussions/1895

common/http.go
```diff
@@ -31,6 +31,7 @@ type HTTPClient struct {
 	Base         string                  // optional base URL. If not set, then all URLs must be absolute.
 	Client       AuthenticatedHTTPClient // underlying HTTP client. Required.
 	ErrorHandler ErrorHandler            // optional error handler. If not set, then the default error handler is used.
+	OKStatusErr  bool                    // optional, If set true indicates 200OK http response may include errors.
```
RajatPawar, line 34: I see why you added this, but how about having a `ResponseHandler` instead? We can add that in place of this. You would shift the response handling in `sendRequest` to this `ResponseHandler` - i.e. 
```
type ResponseHandler func(rsp *http.Response) (*http.Response, error)

// sendRequest sends the given request and returns the response & response body.
func (h *HTTPClient) sendRequest(req *http.Request) (*http.Response, error) { //nolint:cyclop
	// Send the request
	res, err := h.Client.Do(req)
	if err != nil {
		return nil, err
	}

	// Apply the ResponseHandler if provided
	if h.ResponseHandler != nil {
		res, err = h.ResponseHandler(res)
                 ......
```

Since it's optional, if there's no ResponseHandler, there will be no change. But for Marketo, you can set the ResponseHandler to parse it and convert it into a http Response with the error code in the body set to the http.Response.Code

RajatPawar (changes requested):

marketo/url.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+var restAPIPrefix string = "rest" //nolint:gochecknoglobals
+
+func (c *Connector) getURL(params common.ReadParams) (*urlbuilder.URL, error) {
+	// make sure the object is in lowercase format.
+	objName := strings.ToLower(params.ObjectName)
+
+	bURL := strings.Join([]string{c.BaseURL, restAPIPrefix, c.Module, objName}, "/")
+	bURL += ".json"
```
RajatPawar, line 19: Can you add a check to make sure that the base URL already doesn't have a `.json` suffix? It's possible that a connector user might consider the entire name (`leads.json`) as the object name, so we should plan for that.

marketo/module.go
```diff
@@ -0,0 +1,22 @@
...
+
+// Assets is the module/API used for accessing assets objects.
+var Assets = paramsbuilder.APIModule{ // nolint: gochecknoglobals
+	Label:   "asset",
+	Version: "v1",
+}
+
+// Leads is the module/API used for accessing leads objects.
+var Leads = paramsbuilder.APIModule{ //nolint: gochecknoglobals
+	Label:   "",
```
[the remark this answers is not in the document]
Jkarage: They don't say that explicitly, But with the looks of leads API url structures, it should be it.

marketo/parse.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+// getNextRecordsURL returns the URL for the next page of results.
+func getNextRecordsURL(node *ajson.Node) (string, error) {
+	return jsonquery.New(node).StrWithDefault("nextPageToken", "")
+}
+
+// getRecords returns the records from the response.
+func getRecords(node *ajson.Node) ([]map[string]any, error) {
+	result, err := jsonquery.New(node).Array("result", true)
```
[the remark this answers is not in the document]
Jkarage: Well, thanks for this. I believe this should be false as even when the data is empty the key will just respond with an empty list.

marketo/errors.go
```diff
@@ -0,0 +1,137 @@
...
+	"net/http"
+	"strconv"
+)
+
+type Error struct {
+	Code    string `json:"code"`
+	Message string `json:"message"`
+}
+
+type Respose struct {
```
RajatPawar, line 16: Typo

RajatPawar (approved):

PR #922, opened 2024-08-15
add: Get URL query param

Cobalt0s:
Add capability to return query parameter from our custom URL. 

Note: URL primarily manages path and query params.

PR #915, opened 2024-08-14
feat: Enable proxy: Instantly

Cobalt0s:
Closes #892

### GET
```
curl --location --request GET 'https://proxy.withampersand.com/v1/campaign/list' \
--header 'x-amp-project-id: 8b257c62-6b89-4b9b-9271-6fe3f6b700f1' \
--header 'x-amp-proxy-version: 1' \
--header 'x-api-key: <API_KEY>' \
--header 'x-amp-installation-id: cebcc727-9724-4032-bb79-273856c0c567'
```
image

### POST
```
curl --location --request POST 'https://proxy.withampersand.com/v1/custom-tag' \
--header 'x-amp-project-id: 8b257c62-6b89-4b9b-9271-6fe3f6b700f1' \
--header 'x-amp-proxy-version: 1' \
--header 'x-api-key: <API_KEY>' \
--header 'x-amp-installation-id: cebcc727-9724-4032-bb79-273856c0c567' \
--header 'Content-Type: application/json' \
--data-raw '{
  "label": "Tag label",
  "description": "Some description goes here"
}'
```
image

### PATCH
```
curl --location --request PATCH 'https://proxy.withampersand.com/v1/custom-tag/28bad393-3205-42dc-981a-2ae3d88a12a2' \
--header 'x-amp-project-id: 8b257c62-6b89-4b9b-9271-6fe3f6b700f1' \
--header 'x-amp-proxy-version: 1' \
--header 'x-api-key: <API_KEY>' \
--header 'x-amp-installation-id: cebcc727-9724-4032-bb79-273856c0c567' \
--header 'Content-Type: application/json' \
--data-raw '{
    "label": "Super High Delivery"
}'
```
image

### DELETE
```
curl --location --request DELETE 'https://proxy.withampersand.com/v1/custom-tag/28bad393-3205-42dc-981a-2ae3d88a12a2' \
--header 'x-amp-project-id: 8b257c62-6b89-4b9b-9271-6fe3f6b700f1' \
--header 'x-amp-proxy-version: 1' \
--header 'x-api-key: <API_KEY>' \
--header 'x-amp-installation-id: cebcc727-9724-4032-bb79-273856c0c567'
```
image

### Invalid requests
Wrong verb applied.
image

Invalid path.
image

### First Page
image

### Second Page
image

## Successful console operation (operation & events)
image

## Failed console operation (operation & events)
image

PR #914, opened 2024-08-14
feat: Add Instantly connector

Cobalt0s:
Closes #892

# Configuration
None.

### GET
List campagins.
URL: http://localhost:4444/v1/campaign/list
image

### POST
Create tag.
URL: http://localhost:4444/v1/custom-tag
image

### PATCH
Update tag.
URL: http://localhost:4444/v1/custom-tag/f560a82c-1fba-4b4d-8661-8c0242742f50
image

### DELETE
Remove tag.
URL: http://localhost:4444/v1/custom-tag/f560a82c-1fba-4b4d-8661-8c0242742f50
image
Attempt to remove one more time.
image

### First Page
List tags, 2 per page.
URL: http://localhost:4444/v1/custom-tag?limit=2
image

### Second Page
List tags, next 2.
URL: http://localhost:4444/v1/custom-tag?limit=2&skip=2
image

# Logo & Icons
image

PR #923, opened 2024-08-15
feat: Add Deep connector Instantly (Read)

### LINKED ISSUE #912: Add Instantly deep connector

Cobalt0s:
Closes #912 

# Description

There are few endpoints that perform listing. Those that do have varying endpoint paths.
Therefore, there is a hard coded mapping between object name (in plural) to the appropriate URL path.
Pagination is manually calculated as response is not a reliable source as it differes among endpoints.

# Testing
image

### DIFF
--- a/providers/instantly/parse.go
+++ b/providers/instantly/parse.go
@@ -1,0 +1,31 @@
+package instantly
+
+import (
+	"strconv"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+	"github.com/spyzhov/ajson"
+)
+
+func makeNextRecordsURL(reqLink *urlbuilder.URL) common.NextPageFunc {
+	return func(node *ajson.Node) (string, error) {
+		previousStart := 0
+
+		skipQP, ok := reqLink.GetFirstQueryParam("skip")
+		if ok {
+			// Try to use previous "skip" parameter to determine the next skip.
+			skipNum, err := strconv.Atoi(skipQP)
+			if err == nil {
+				previousStart = skipNum
+			}
+		}
+
+		nextStart := previousStart + DefaultPageSize
+
+		reqLink.WithQueryParam("limit", strconv.Itoa(DefaultPageSize))
+		reqLink.WithQueryParam("skip", strconv.Itoa(nextStart))
+
+		return reqLink.String(), nil
+	}
+}
--- a/providers/instantly/client.go
+++ b/providers/instantly/client.go
@@ -1,0 +1,16 @@
+package instantly
+
+import "github.com/amp-labs/connectors/common"
+
+// JSONHTTPClient returns the underlying JSON HTTP client.
+func (c *Connector) JSONHTTPClient() *common.JSONHTTPClient {
+	return c.Client
+}
+
+func (c *Connector) HTTPClient() *common.HTTPClient {
+	return c.Client.HTTPClient
+}
+
+func (c *Connector) Close() error {
+	return nil
+}
--- a/providers/instantly/connector.go
+++ b/providers/instantly/connector.go
@@ -1,0 +1,73 @@
+package instantly
+
+import (
+	"fmt"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/interpreter"
+	"github.com/amp-labs/connectors/common/paramsbuilder"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+	"github.com/amp-labs/connectors/providers"
+)
+
+const apiVersion = "v1"
+
+type Connector struct {
+	BaseURL string
+	Client  *common.JSONHTTPClient
+}
+
+func NewConnector(opts ...Option) (conn *Connector, outErr error) {
+	defer common.PanicRecovery(func(cause error) {
+		outErr = cause
+		conn = nil
+	})
+
+	params, err := paramsbuilder.Apply(parameters{}, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	httpClient := params.Client.Caller
+	conn = &Connector{
+		Client: &common.JSONHTTPClient{
+			HTTPClient: httpClient,
+		},
+	}
+
+	providerInfo, err := providers.ReadInfo(conn.Provider())
+	if err != nil {
+		return nil, err
+	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	// connector and its client must mirror base url and provide its own error parser
+	conn.setBaseURL(providerInfo.BaseURL)
+	conn.Client.HTTPClient.ErrorHandler = interpreter.ErrorHandler{
+		JSON: interpreter.NewFaultyResponder(errorFormats, nil),
+	}.Handle
+
+	return conn, nil
+}
+
+func (c *Connector) Provider() providers.Provider {
+	return providers.Instantly
+}
+
+func (c *Connector) String() string {
+	return fmt.Sprintf("%s.Connector", c.Provider())
+}
+
+func (c *Connector) getURL(parts ...string) (*urlbuilder.URL, error) {
+	return urlbuilder.New(c.BaseURL, append([]string{
+		apiVersion,
+	}, parts...)...)
+}
+
+func (c *Connector) setBaseURL(newURL string) {
+	c.BaseURL = newURL
+	c.Client.HTTPClient.Base = newURL
+}
--- a/providers/instantly/errors.go
+++ b/providers/instantly/errors.go
@@ -1,0 +1,29 @@
+package instantly
+
+import (
+	"fmt"
+
+	"github.com/amp-labs/connectors/common/interpreter"
+)
+
+var errorFormats = interpreter.NewFormatSwitch( // nolint:gochecknoglobals
+	[]interpreter.FormatTemplate{
+		{
+			MustKeys: nil,
+			Template: func() interpreter.ErrorDescriptor { return &ResponseError{} },
+		},
+	}...,
+)
+
+type ResponseError struct {
+	// Message usually has API Key query parameter. This shouldn't be returned.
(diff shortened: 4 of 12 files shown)


Cobalt0s:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/923?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#923** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/923?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#922** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/922?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#914** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/914?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#915](https://github.com/amp-labs/connectors/pull/915) <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/915?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`

<h2></h2>

instantly/url.go
```diff
@@ -0,0 +1,18 @@
+package instantly
+
+import "github.com/amp-labs/connectors/common/urlbuilder"
+
+var queryEncodingExceptions = map[string]string{ //nolint:gochecknoglobals
+	// none
+}
+
+func constructURL(base string, path ...string) (*urlbuilder.URL, error) {
```
jlimatampersand, line 9: Same as the `WithClient` can we also remove the duplicate definition for `constructURL` and define in `common` package and reuse it?

instantly/parse.go
```diff
@@ -0,0 +1,72 @@
...
+		nextStart := previousStart + DefaultPageSize
+
+		reqLink.WithQueryParam("limit", strconv.Itoa(DefaultPageSize))
+		reqLink.WithQueryParam("skip", strconv.Itoa(nextStart))
+
+		return reqLink.String(), nil
+	}
+}
+
+func getMarshaledData(records []map[string]interface{}, fields []string) ([]common.ReadResultRow, error) {
```
jlimatampersand, line 61: same as other functionss, define in `common` and remove duplicate definitions.

jlimatampersand (changes requested):
For many part of the code seem to be redundant. I commented on some of them, but wherever applicable, lets move to common package and reduce duplicate code. 

You can do it just for this this connector for the moment, and apply to other connectors in follow up PR to separate the scopes.

instantly/params.go
```diff
@@ -0,0 +1,65 @@
...
+	if p.setupError != nil {
+		return p.setupError
+	}
+
+	return errors.Join(
+		p.Client.ValidateParams(),
+	)
+}
+
+func WithClient(ctx context.Context, client *http.Client,
```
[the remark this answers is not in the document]
Cobalt0s: This method so far looked identical, but there are more non-oauth2 connectors added which would change the arguments and the impl of the function.

We are using `paramsbuilder.Client` as a delegate. So each connector then selects one of `WithOauthClient` / `WithBasicClient` / `WithApiKeyHeaderClient` / `WithApiKeyQueryParamClient`.

Note: `paramsbuilder` has a dependency on `common` package. In case we need to move such methods somewhere it will be new package to avoid circular dependency. (I tried some refactoring while working on this PR and thats what I had run into).

I used delegation to have some consistency in the way parameters are defined. Since we are using Options pattern the repitition will be there.

instantly/read.go
```diff
@@ -0,0 +1,102 @@
...
+		return "unibox/emails", "data", nil
+	case objectNameTags:
+		// https://developer.instantly.ai/tags/list-tags
+		return "custom-tag", "data", nil
+	default:
+		return "", "", common.ErrOperationNotSupportedForObject
+	}
+}
+
+func (c *Connector) buildReadURL(config common.ReadParams) (*urlbuilder.URL, string, error) {
```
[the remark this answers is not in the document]
Cobalt0s: Yes, dataLocation could be related to URL. I didn't know how to use it.
With link I try to call it url, from older code I kind of introduced link word... anyway I see that there is more semantic dfiference then I give to it.

I will change those, and pull the method out.

instantly/url.go
Cobalt0s: See: https://github.com/amp-labs/connectors/pull/972

instantly/parse.go
Cobalt0s: Please, review: https://github.com/amp-labs/connectors/pull/971

instantly/read.go
```diff
@@ -0,0 +1,102 @@
...
+		link, err := urlbuilder.New(config.NextPage.String())
+		if err != nil {
+			return nil, err
+		}
+
+		return link, nil
+	}
+
+	// First page
+	link, err := c.getURL(urlPath)
```
jlimatampersand, line 93: let's use `url` instead of `link`

instantly/params.go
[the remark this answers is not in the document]
jlimatampersand: I looked into the code, I think we can define those functions once in. Looking at how much lines we are replicating for each connector, I think this design is wrong. We always need to think in scale. We will end up having 1000+ connectors. which means this will end up being hundreds of thousands of lines. Which is simply not scalable. I am not sure how we came up this design from the beginning, but if this is current way, we should stop and redesign this. 

We shouldn't declare `type Option = func(*parameters)` type for every connector. 

Option should be in a separate package so that those can be imported and reused, not redefined. 

Let's hold off on this until I can think through and redesign this.

instantly/parse.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"strconv"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/jsonquery"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+	"github.com/spyzhov/ajson"
+)
+
+func makeGetTotalSize(dataLocation string) common.ListSizeFunc {
```
jlimatampersand, line 12: again, this should not be defined for every connector and same for all other functions in this file. Maybe except for the `makeNextRecordsURL` since url are specific to the  provider.

jlimatampersand (changes requested):

instantly/read.go
Cobalt0s: Sorry, missed that one.
I did rename in other files too. This naming will stop.

instantly/parse.go
Cobalt0s: I went further. You are right that there is repitition, in this case this method shouldn't even exist. Here is a PR: https://github.com/amp-labs/connectors/pull/991

instantly/read.go
Cobalt0s: It started to get messy for this PR, I split it: https://github.com/amp-labs/connectors/pull/993

instantly/params.go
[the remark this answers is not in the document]
Cobalt0s: First of all by design we require setting up connectors using  `With<MethodName>`. This means every connector must define this method. Also IDE completion makes it useful when say instantly.With... and it autocompletes potential arguments.

`Option`s golang alias is necessary for generic constructor. It provides typed parameter. Every connector must define its set of parameters. So there is nothing that can be refactored here. In other words type `Option = func(*connectorName.parameters)` is not common or shared in any way, only the visual is.

And about methods that live under `parse.go`. This file is responsible to understand response format of a provider. This again cannot be known universally and every connector stores array of items under different field name, everybody does pagination in their unique way. Maybe as more patterns become apparent we could add utility functions to shorten it.

The patterns that could evolve could be more configuration/data driven. But thats the information that makes the identity of that connector. I agree that adding connectors can be confusing especially to outsiders. But slowly these things are polished. I think what we did with `errors.go` can be done here (config driven).

jlimatampersand (approved):
Approving this PR for the time being. There were some comments that I had about having repetitive code for each connectors. I found refactoring this will take time and certainly this PR does not need to wait for it. But please do try to minimize the repetition for PRs later on. 

Currently we are adding 1000 lines per connectors, which will quickly become 1 million lines of code if we get to 1000 connectors. 

I will work on refactoring in next couple weeks to come up with better solution. It seems like this pattern has ben inherited from prior to your works. Thanks for working on this and other optimizations you are doing!

providers/instantly/parse.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/spyzhov/ajson"
+)
+
+func makeGetTotalSize(dataLocation string) common.ListSizeFunc {
+	return func(node *ajson.Node) (int64, error) {
+		return jsonquery.New(node).ArraySize(dataLocation)
+	}
+}
+
+func makeGetRecords(dataLocation string) common.RecordsFunc {
```
jlimatampersand, line 18: This and above function should be in common pacakge or utils package and imported to be used. Also `return jsonquery.New(node).ArraySize(dataLocation)` is being used twice for no reason and drawing performance back. 

Please resolve these before merging

jlimatampersand (changes requested):
Please take out `makeGetRecords` to common package and reuse. Please create a follow up PR to implement this to other connectors as well.

providers/instantly/parse.go
jlimatampersand: I initially approved, but found there are some code `makeGetRecords` that is redefined for each connector and the function has nothing specific to the connector. Please move to common or utils and import to use.

Cobalt0s: I will create common utilities for this method.

Cobalt0s: Dependent PR: https://github.com/amp-labs/connectors/pull/1004

jlimatampersand (approved):

PR #925, opened 2024-08-15
feat: Deep connector Instantly (Metadata)

Cobalt0s:
Closes #912 

# Description

Metadata is taken from documentation and saved as static files under `schema.json`.

# Testing
image

PR #933, opened 2024-08-16
feat: Deep connector Instantly (Write)

### LINKED ISSUE #912: Add Instantly deep connector

Cobalt0s:
Closes #912

# Description
Write endpoints are hard coded to the list of object names associated with them.

There is a second commit with minor tidy things. They are not significant to be in the independent PR but they are not related to current.

# Testing
image

### DIFF
--- a/providers/instantly/delete.go
+++ b/providers/instantly/delete.go
@@ -1,0 +1,39 @@
+package instantly
+
+import (
+	"context"
+
+	"github.com/amp-labs/connectors/common"
+)
+
+// Delete removes object. As of now only removal of Tags are allowed.
+func (c *Connector) Delete(ctx context.Context, config common.DeleteParams) (*common.DeleteResult, error) {
+	if len(config.ObjectName) == 0 {
+		return nil, common.ErrMissingObjects
+	}
+
+	if len(config.RecordId) == 0 {
+		return nil, common.ErrMissingRecordID
+	}
+
+	if !supportedObjectsByDelete.Has(config.ObjectName) {
+		// Removing tags is the only to be supported at this time.
+		// https://developer.instantly.ai/tags/delete-a-tag
+		return nil, common.ErrOperationNotSupportedForObject
+	}
+
+	url, err := c.getURL("custom-tag", config.RecordId)
+	if err != nil {
+		return nil, err
+	}
+
+	// 200 OK is expected
+	_, err = c.Client.Delete(ctx, url.String())
+	if err != nil {
+		return nil, err
+	}
+
+	return &common.DeleteResult{
+		Success: true,
+	}, nil
+}
--- a/providers/instantly/delete.go
+++ b/providers/instantly/delete.go
@@ -6,7 +6,9 @@ import (
 	"github.com/amp-labs/connectors/common"
 )
 
-// Delete removes object. As of now only removal of Tags are allowed.
+// Delete removes object. As of now only removal of Tags are allowed because
+// deletion of other object types require a request payload to be added
+// c.Client.Delete does not yet support this.
 func (c *Connector) Delete(ctx context.Context, config common.DeleteParams) (*common.DeleteResult, error) {
 	if len(config.ObjectName) == 0 {
 		return nil, common.ErrMissingObjects
--- a/providers/instantly/write.go
+++ b/providers/instantly/write.go
@@ -1,0 +1,121 @@
+package instantly
+
+import (
+	"context"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/handy"
+	"github.com/amp-labs/connectors/common/jsonquery"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+	"github.com/spyzhov/ajson"
+)
+
+// Write method allows to
+// * create campaigns
+// * create/update email-accounts
+// * create client
+// Documentation links can be found within constructURLPath function.
+func (c *Connector) Write(
+	ctx context.Context, config common.WriteParams,
+) (*common.WriteResult, error) {
+	if len(config.ObjectName) == 0 {
+		return nil, common.ErrMissingObjects
+	}
+
+	if !supportedObjectsByWrite.Has(config.ObjectName) {
+		return nil, common.ErrOperationNotSupportedForObject
+	}
+
+	// Object name will be used to complete further URL construction below.
+	url, err := c.getURL()
+	if err != nil {
+		return nil, err
+	}
+
+	var write common.WriteMethod
+	if len(config.RecordId) == 0 {
+		write = c.Client.Post
+		constructURLPathCreate(config, url) // nolint:wsl
+	} else {
+		write = c.Client.Patch
+		constructURLPathUpdate(config, url)
+	}
+
+	res, err := write(ctx, url.String(), config.RecordData)
+	if err != nil {
+		return nil, err
+	}
+
+	if res == nil || res.Body == nil {
+		// it is unlikely to have no payload
+		return &common.WriteResult{
+			Success: true,
+		}, nil
+	}
+
+	recordIdNodePath := recordIdPaths[config.ObjectName]
+
+	// write response was with payload
+	return constructWriteResult(res.Body, recordIdNodePath)
+}
+
+var recordIdPaths = map[string]*string{ // nolint:gochecknoglobals
+	objectNameLeads:            nil, // ID is not returned for Leads.
+	objectNameBlocklistEntries: handy.Pointers.Str("blocklist_id"),
+	objectNameUniboxReplies:    handy.Pointers.Str("message_id"),
+	objectNameTags:             handy.Pointers.Str("id"),
+}
+
+func constructURLPathCreate(config common.WriteParams, url *urlbuilder.URL) {
+	switch config.ObjectName {
+	case objectNameLeads:
+		// Add lead to campaign.
+		// https://developer.instantly.ai/campaign/add-leads-to-a-campaign
+		url.AddPath("lead/add")
+	case objectNameBlocklistEntries:
+		// Add blocklist entry.
+		// https://developer.instantly.ai/blocklist/add-entries-to-blocklist
+		url.AddPath("blocklist/add/entries")
+	case objectNameUniboxReplies:
+		// Create message - unibox reply.
+		// https://developer.instantly.ai/unibox/send-reply
+		url.AddPath("unibox/emails/reply")
+	case objectNameTags:
+		// Create tag.
+		// https://developer.instantly.ai/tags/create-a-new-tag
+		url.AddPath("custom-tag")
+	}
+}
+
+func constructURLPathUpdate(config common.WriteParams, url *urlbuilder.URL) {
+	if config.ObjectName == objectNameTags {
+		// Update account.
+		// https://developer.instantly.ai/tags/update-tag
+		url.AddPath("custom-tag", config.RecordId)
+	}
+}
+
+func constructWriteResult(body *ajson.Node, recordIdLocation *string) (*common.WriteResult, error) {
+	if recordIdLocation == nil {
+		// ID is not present inside response. Therefore, empty.
+		return &common.WriteResult{
+			Success:  true,
+			RecordId: "",
+			Errors:   nil,
+			Data:     nil,
+		}, nil
+	}
+
+	// ID is integer that is always stored under different field name.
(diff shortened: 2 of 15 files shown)


Cobalt0s:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/933?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#933** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/933?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#923** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/923?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#925](https://github.com/amp-labs/connectors/pull/925) <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/925?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#922** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/922?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#914** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/914?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#915](https://github.com/amp-labs/connectors/pull/915) <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/915?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`

<h2></h2>

instantly/read.go
```diff
@@ -14,6 +14,10 @@ const (
 	objectNameAccounts  = "accounts"
 	objectNameEmails    = "emails"
 	objectNameTags      = "tags"
+
+	objectNameLeads            = "leads"
+	objectNameBlocklistEntries = "blocklist-entries"
+	objectNameUniboxReplies    = "unibox-replies"
```
jlimatampersand, line 20: Since these constants are also used in delete and write, we should move to a new file that can be shared across actions.

jlimatampersand (changes requested):
Please resolve comments and re-request review.

instantly/write.go
```diff
@@ -0,0 +1,149 @@
...
+// Documentation links can be found within completeURLPath function.
+func (c *Connector) Write(
+	ctx context.Context, config common.WriteParams,
+) (*common.WriteResult, error) {
+	if len(config.ObjectName) == 0 {
+		return nil, common.ErrMissingObjects
+	}
+
+	// Object name is fully replace
+	url, err := c.getURL()
```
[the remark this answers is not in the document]
Cobalt0s: 100%

instantly/write.go
```diff
@@ -0,0 +1,149 @@
...
+		}, nil
+	}
+
+	// write response was with payload
+	return constructWriteResult(res.Body, recordIdLocation)
+}
+
+// Maps common.WriteParams to URL format for that object.
+// In addition, returns field name where ID will be stored on successful response.
+func completeURLPath(
```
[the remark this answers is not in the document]
Cobalt0s: I decided to create a map to hold json path to recordIDs.
The same thing I applied to Smartlead.
Instantly: https://github.com/amp-labs/connectors/pull/933/commits/3dc08bba7db8511baea9c9886f56923e62790781
Smartlead: https://github.com/amp-labs/connectors/pull/933/commits/43a9b2e49c4e90376a1b9e7ae2a5eab9e459f90c

PS: Its funny how when I work on multiple deep connectors both of them have shared unique code. Those connectors come in pairs ahaha.

providers/instantly/write.go
```diff
@@ -0,0 +1,121 @@
...
+		// Create tag.
+		// https://developer.instantly.ai/tags/create-a-new-tag
+		url.AddPath("custom-tag")
+	}
+}
+
+func constructURLPathUpdate(config common.WriteParams, url *urlbuilder.URL) {
+	if config.ObjectName == objectNameTags {
+		// Update account.
+		// https://developer.instantly.ai/tags/update-tag
```
jlimatampersand, line 93: nice, thanks for having all these doc url. These are super helpful.

providers/smartlead/objectNames.go
```diff
@@ -0,0 +1,19 @@
...
+
+import "github.com/amp-labs/connectors/common/handy"
+
+const (
+	objectNameCampaign     = "campaigns"
+	objectNameEmailAccount = "email-accounts"
+	objectNameClient       = "client"
+)
+
+var supportedObjectsByWrite = handy.NewSet([]string{ //nolint:gochecknoglobals
```
jlimatampersand, line 11: were these only requirements? not lead? I don't have knowledge on the requirements myself.

providers/instantly/write.go
```diff
@@ -0,0 +1,121 @@
...
+		url.AddPath("custom-tag", config.RecordId)
+	}
+}
+
+func constructWriteResult(body *ajson.Node, recordIdLocation *string) (*common.WriteResult, error) {
+	if recordIdLocation == nil {
+		// ID is not present inside response. Therefore, empty.
```
jlimatampersand, line 100: how do we then return result for lead? we should read lead as well. For lead, email is the unique key they use to identify the object.

providers/instantly/write.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	recordIdNodePath := recordIdPaths[config.ObjectName]
+
+	// write response was with payload
+	return constructWriteResult(res.Body, recordIdNodePath)
+}
+
+var recordIdPaths = map[string]*string{ // nolint:gochecknoglobals
+	objectNameLeads:            nil, // ID is not returned for Leads.
```
jlimatampersand, line 63: however, emails are used to identify a lead along with campaign id. we should handle this.

jlimatampersand (changes requested):
Overall looks okay except that handling lead object seems to be missing some pieces. Please fill in the gaps or leave some comments on limitations on the lead objects. 

cc @laurenzlong I am not sure about the requirements for this connector, so please comment if missing lead is not an issue for us.

providers/instantly/delete.go
```diff
@@ -0,0 +1,39 @@
...
+		return nil, common.ErrMissingRecordID
+	}
+
+	if !supportedObjectsByDelete.Has(config.ObjectName) {
+		// Removing campaign is the only to be supported at this time.
+		// https://api.smartlead.ai/reference/delete-campaign
+		return nil, common.ErrOperationNotSupportedForObject
+	}
+
+	url, err := c.getURL("custom-tag", config.RecordId)
```
jlimatampersand, line 25: we shouldn't hard code this. I don't think you can delete campaign with custom-tag either. Campaign has id.

providers/instantly/delete.go
```diff
@@ -0,0 +1,39 @@
...
+	if len(config.ObjectName) == 0 {
+		return nil, common.ErrMissingObjects
+	}
+
+	if len(config.RecordId) == 0 {
+		return nil, common.ErrMissingRecordID
+	}
+
+	if !supportedObjectsByDelete.Has(config.ObjectName) {
+		// Removing campaign is the only to be supported at this time.
```
jlimatampersand, line 20: Is this business decision? @laurenzlong? instantly API seems to supprot delete for all objects.

providers/smartlead/objectNames.go
```diff
@@ -0,0 +1,19 @@
+package smartlead
```
jlimatampersand, line 1: this file should be in separate PR.

jlimatampersand (changes requested):

providers/instantly/delete.go
laurenzlong: no, not a business decision, Constantin can you elaborate on the reason for this?

providers/instantly/write.go
Cobalt0s: As mentioned above I couldn't explore leads.
I can make `Read leads` request and there is no ID. There is no single read. Only search by email (funny enough only then ID can be returned according to docs if we can believe it, but we cannot use it in any way).

Lead REST API doesn't follow conventions. We could return empty recordID but create custom payload instead to mentione email `common.WriteResult.Data`. But then we would need to document that we do special case. But if we go back whoever makes `/api/v1/lead/add` request will provide `email` for each lead so it defeats the purpose. Hence, write method only returns ACK.

providers/instantly/write.go
Cobalt0s: I will close this comment as we are discussing this in the thread above.

providers/instantly/delete.go
Cobalt0s: I messed up the method documentation. I didn't update it.
The delete only works on `tags` which they call `custom-tag`.

It has nothing to do with campaigns. Speaking of campaigns they cannot be removed or created programatically.

providers/instantly/delete.go
Cobalt0s: :sweat: the comment line says campaign while it should say `tags`. Thats a typo.

I created a list of supported objects for Read, Write, Delete bound to current design of inpurt params.
Speaking of Delete I created this connector interface to match others. 

I am not sure about importance of usage of Delete. So let me know if I need to spend time for the following upgrade:
* add `common.DeleteResult.Data` field to store user payload
* modify HTTP client's Delete method

Supporting Delete payload will enable the following Instantly operations:
* Delete Account
* Delete Customer Tracking Domain (this is not a List resource)
* Remove entries from blocklist
* Delete leads from a campaign

PS: my main focus was on Read/Write. This is the first connector to challenge simple OOB Delete. Usually it is a good REST API practice to not have payload during Delete. But APIs are different some may do this via POST, etc.

providers/smartlead/objectNames.go
jlimatampersand: I got confused with instantly because this is instantly PR, so I thought there should be `lead`.

jlimatampersand (approved):

providers/instantly/delete.go
jlimatampersand: @laurenzlong so I think main challenge is that delete does not have single id or key reference that can delete objects. It requires campaign id and and email to delete leads. So basically they use composite key like referencing, so we need to have multiple parameters to delete a lead. I think, if delete is not part of requirement we can skip for now. Otherwise, we need to support multi-primary key parameters. 

@Cobalt0s other parts of code look good, but Lauren should make decision on how we want to support delete. I will delegate PR approval to Lauren.

providers/smartlead/objectNames.go
Cobalt0s: Smartlead and Instantly have related set of Objects but with some differences.

providers/instantly/delete.go
Cobalt0s: Yes, `multi-primary key parameters` is the general rule that I was using to categorize objects for Read, Write, and by inertia I did it ignore Delete.
Luckily, those operations that I listed above can be managed with single ID or no IDs at all. This information can be passed via `Request Body`, but that would require implementation changes I had mentioned. It depends on the business requirements.

Having delete is nice to roll back create/update which was used as a type of cleanup. People using connectors as a library may find it useful. As of now I don't see usage of `Delete` in server.

Lets wait for the roadmap from Lauren.

providers/smartlead/objectNames.go
Cobalt0s: Moved here https://github.com/amp-labs/connectors/pull/1006

providers/instantly/delete.go
```diff
@@ -0,0 +1,39 @@
+package instantly
+
+import (
+	"context"
+
+	"github.com/amp-labs/connectors/common"
+)
+
+// Delete removes object. As of now only removal of Tags are allowed.
```
laurenzlong, line 9:
```suggestion
// Delete removes object. As of now only removal of Tags are allowed because
// deletion of other object types require a request payload to be added, c.Client.Delete does not yet support this.
```

laurenzlong (approved):

providers/instantly/delete.go
laurenzlong: We don't yet expose the ability to delete to our customers, so only supporting tags in this connector for now is fine - @Cobalt0s can you take a look at my suggestions on the comment line? Hopefully I captured the information correctly.

providers/instantly/delete.go
Cobalt0s: Slight difference: it uses DELETE with payload. Not POST.

PR #964, opened 2024-08-21
feat: Add write  Marketo connector

### LINKED ISSUE #908: Add deep connector for Marketo

Jkarage:
Currently we write to the Leads API.

Note: We currently don't handle record level errors(These are individual record errors). 

Test Response:
Screenshot 2024-08-21 at 12 14 47

Closes #908

### DIFF
--- a/marketo/url.go
+++ b/marketo/url.go
@@ -45,8 +45,13 @@ func constructURL(base string, path ...string) (*urlbuilder.URL, error) {
 
 func updateURLWithID(url *urlbuilder.URL, id string) (*urlbuilder.URL, error) {
 	s := removeJSONSuffix(url.String())
-	s += id
-	s = addJSONSuffix(s)
+
+	url, err := constructURL(s, id)
+	if err != nil {
+		return nil, err
+	}
+
+	s = addJSONSuffix(url.String())
 
 	return constructURL(s)
 }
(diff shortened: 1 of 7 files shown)


marketo/url.go
```diff
@@ -41,3 +42,10 @@ func constructURL(base string, path ...string) (*urlbuilder.URL, error) {
 
 	return link, nil
 }
+
+func updateURLWithID(url *urlbuilder.URL, id string) (*urlbuilder.URL, error) {
```
RajatPawar, line 46: Can you move the code which adds or removes `.json` to separate functions so that we can reuse it?

marketo/write.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	resp, err := common.UnmarshalJSON[writeResponse](json)
+	if err != nil {
+		return nil, err
+	}
+
+	return &common.WriteResult{
+		Success:  resp.Success,
+		RecordId: fmt.Sprint(resp.Result[0]["id"]),
```
RajatPawar, line 50: This has chances of failing, so you should add a check to see if `[0]` exists, and then if `id` exists on it

test/marketo/write/write.go
RajatPawar, line 1: Please add the output of the test results in the description of the PR

RajatPawar (changes requested):

RajatPawar (approved):

PR #985, opened 2024-08-23
feat: Update Read Marketo connector

Jkarage:
This updates the Marketo Read connector to support Since filtering with the following Objects:

- Emails
- Programs
- SmartCampaigns
- SmartLists

Removed the `SinceDateTime` since all the objects it supports are currently not supported using the Write Ampersand API.
The `sinceDateTime` supports  `lead activities` object which requires a client to send the `activityTypeIds`  as a query parameter during the request.

Screenshot 2024-08-23 at 17 13 50

Example responses:
Reading `campaign` does not use Since, even if provided
Screenshot 2024-08-26 at 14 56 08

Reading `channels` does not use Since, even if provided.
Screenshot 2024-08-26 at 14 57 02

Reading `smartCampaigns` uses provided Since value.
Screenshot 2024-08-26 at 14 56 44

### DIFF
--- a/providers/marketo/README.md
+++ b/providers/marketo/README.md
@@ -2,7 +2,7 @@
 
 Currently the Write connector supports Leads API only. 
 
-- The objects that we can write to currently are:
+The objects that we can write to currently are:
  - Companies
  - Leads
  - Named Account Lists
--- a/providers/marketo/parse.go
+++ b/providers/marketo/parse.go
@@ -21,5 +21,8 @@ func getRecords(node *ajson.Node) ([]map[string]any, error) {
 }
 
 func getTotalSize(node *ajson.Node) (int64, error) {
-	return jsonquery.New(node).ArraySize("result")
+	// When there is no result value, it indicates [], we should ot error it.
+	size, _ := jsonquery.New(node).ArraySize("result")
+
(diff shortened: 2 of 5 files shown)


marketo/url.go
```diff
@@ -12,32 +12,31 @@ import (
...
-		time := params.Since.Format(time.RFC3339)
-		fmtTime := fmt.Sprintf("%v", time)
-		link.WithQueryParam("sinceDatetime", fmtTime)
+	// If NextPage is set, then we're reading the next page of results.
+	if len(params.NextPage) > 0 {
+		link.WithQueryParam("nextPageToken", params.NextPage.String())
+	}
+
+	// The only objects in Assets API supporting this are: Emails, Programs, SmartCampaigns,SmartLists
+	if !params.Since.IsZero() && c.Module == ModuleAssets.String() {
```
RajatPawar, line 26: Would be nice if we can do something like this to make it even more readable - 
```
if !params.Since.IsZero() {
  switch c.Module {
    case ModuleAssets.String()
    case ...
  }
}
```

RajatPawar (approved):
Please run tests for all modules again & add screenshots (if you haven't) to rule out any regression

PR #990, opened 2024-08-26
chore: Move deep connectors inside providers

RajatPawar:
As discussed previously, it makes sense to organise deep connector packages inside the providers package. I've done this using my IDE which takes care of refactoring the linkages when you move a package, and our linter will also make sure of them.


laurenzlong (approved):
Please update the README's code snippets

PR #991, opened 2024-08-26
tidy: Remove RecordsSize method

Cobalt0s:
In general the implemenation of `getTotalSize` is directly connected to `getRecords`. Since we already have this array, why not get the length from it.
Sometimes, some connector do have `size` field in their response but thats not going to give any performence boost but only complicate the code.

### DIFF
--- a/common/parse.go
+++ b/common/parse.go
@@ -7,13 +7,11 @@ import (
 )
 
 type (
-	ListSizeFunc func(*ajson.Node) (int64, error)
 	NextPageFunc func(*ajson.Node) (string, error)
 	RecordsFunc  func(*ajson.Node) ([]map[string]any, error)
 )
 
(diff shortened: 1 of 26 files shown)


Cobalt0s:
* **#991** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/991?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

jlimatampersand (approved):

RajatPawar (approved):

RajatPawar (approved):
I don't think it should happen, but this would error out if `marshaledData` is ever nil, so please verify that

Cobalt0s:
@RajatPawar 
len(nil) returns 0
So it is safe just like looping over nil slice.

PR #994, opened 2024-08-26
[ENG-1423] fix: Enable rate limiting for oauth token refresher

eberle1080:
Passing in a custom HTTP client to the OAuth token refresher actually requires a context value to pull off. Without this, the default client gets used, which right now is resulting in HTTP throttle errors (since our rate-limited client is being ignored for these calls).

### DIFF
--- a/common/oauth.go
+++ b/common/oauth.go
@@ -211,6 +211,12 @@ func getTokenSource(ctx context.Context, params *oauthClientParams) oauth2.Token
 		return params.tokenSource
 	}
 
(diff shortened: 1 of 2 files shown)


RajatPawar (approved):
I am not aware of the exact details, but since our proxying uses a connector's HTTP client, would this cause any side-effects there?

eberle1080:
> Since our proxying uses a connector's HTTP client, would this cause any side-effects there?

No, this only matters when rate limiting is explicitly enabled (there's a param). In the proxy, this value is always set to false, so this code path won't even matter.

PR #996, opened 2024-08-26
chore: Fix formatting

eberle1080:
I just ran `make format`


Cobalt0s:
The changes make sense but it is a bit confusing why `make lint` doesn't catch these?
I was only using `make fix`. Just checked the makefile and it seems `make fix` is the same as `make format`.
I wonder why we get different results for `make format`, do you know why?

jlimatampersand (approved):

PR #998, opened 2024-08-27
[ENG-1498] feat: Make ReadParams.Fields unique

Cobalt0s:
# Behaviour Change
Read operation before:
image
After the change:
image

# Changes
All read access to `common.ReadParams.Fields` is replaced with `common.ReadParams.GetUniqueFields()`.

# Considirations
This implementation does not affect `go` struct inits. Creating a dedicated constructor method felt very restrictive.
Changing HTTP Client was to late and much control is lost at this level (some connectors pass fields in special ways, syntax, so no assumptions can be made here). Calling a utility method makes it very disconnected and doesn't send a right message. Hence the member method.

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"fmt"
 	"time"
+
+	"github.com/amp-labs/connectors/common/handy"
 )
 
 var (
@@ -92,7 +94,7 @@ type ReadParams struct {
 	// The name of the object we are reading, e.g. "Account"
 	ObjectName string // required
 	// The fields we are reading from the object, e.g. ["Id", "Name", "BillingCity"]
-	Fields []string // required, at least one field needed
+	Fields handy.StringSet // required, at least one field needed
 	// NextPage is an opaque token that can be used to get the next page of results.
 	NextPage NextPageToken // optional, only set this if you want to read the next page of results
 	// Since is a timestamp that can be used to get only records that have changed since that time.
--- a/common/handy/maps.go
+++ b/common/handy/maps.go
@@ -14,7 +14,7 @@ func (m Map[K, V]) Keys() []K {
 }
 
 func (m Map[K, V]) KeySet() Set[K] {
-	return NewSet(m.Keys())
+	return NewSetFromList(m.Keys())
 }
 
 func (m Map[K, V]) Has(key K) bool {
--- a/common/handy/set.go
+++ b/common/handy/set.go
@@ -1,11 +1,23 @@
 package handy
 
+// NewStringSet and StringSet are Aliases.
+var NewStringSet = NewSet[string] // nolint:gochecknoglobals
+type StringSet = Set[string]
+
 // Set data structure that can hold any type of data.
 type Set[T comparable] map[T]struct{}
 
-// NewSet creates a set from slice.
-func NewSet[V comparable](values []V) Set[V] {
-	result := make(Set[V])
+// NewSet creates a set from multiple values.
+func NewSet[V comparable](values ...V) Set[V] {
+	result := make(Set[V], len(values))
+	result.Add(values)
+
+	return result
+}
+
+// NewSetFromList creates a set from slice.
+func NewSetFromList[V comparable](values []V) Set[V] {
+	result := make(Set[V], len(values))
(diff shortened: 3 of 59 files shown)


Cobalt0s:
* **#998** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/998?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

common/types.go
```diff
@@ -101,6 +103,7 @@ type ReadParams struct {
 	// The name of the object we are reading, e.g. "Account"
 	ObjectName string // required
 	// The fields we are reading from the object, e.g. ["Id", "Name", "BillingCity"]
+	// NOTE: to access this value use ReadParams.GetUniqueFields.
 	Fields []string // required, at least one field needed
```
RajatPawar, line 107: Can we make this a string set? That way, we don't have to remember the unique function on top of it

Cobalt0s: I prefer that too. I wasn't sure about usage of `handy.Set` outside for those who will use this function.

RajatPawar (commented):
Thanks! 

* Can you specifically alias handy.Set[string] as a type - maybe `FieldSet` or `StringSet`?
* Can you re-export it in `connectors` and use that instead? Similar to what we did with ReadParams (`ReadParams = common.ReadParams`). I don't want users to directly have to dig into `common`, and `connectors.StringSet` or `connectors.FieldSet` is much easier. 
* `handy` is too broad/casual of a word, let's call it `utils` which sounds much more appropriate. 

So the usage for people outside would be like - 
```
ReadParams{
   Fields: connectors.FieldSet/StringSet{...}
}
```

Cobalt0s:
@RajatPawar 

I like this idea. This aliasing will make easier usage for sure. I am not a fan of using handy outside as I mentioned. More of internal utility.
image

However, speaking about utils naming. One utils repo I removed by doing several PRs making such obsolete. It is usually suggested not to name packages with very general names. With growing codebase it may go out of hand when utils is everywhere. But I cannot come up with a better naming. We have multiple utils but that utility is for different context so it has a different flavour or color, the way I like to think about it. Some current utils I called `mockutils`, `testutils`, `testroutines`, `naming`, `jsonquery`, etc. All of them where utils originally but then evolved separately.
Mostly `handy` deals with data structures or universal golang utilities like pointer creation or struct cloning.

We could rename handy but maybe in a different PR. Maybe better split it. I would prefer some data structure package. What would be the name? 
* `dst`  - data structure
* `container`
* `housing` - there will be no naming collisions ahah. (utils doens't tell much, but housing too, but more memorable once you know). The house because data has a place to live: as a map or as a list, etc.
* `pile` - data pile, pile.DefaultMap
* `holder`
* other

laurenzlong:
Hey @Cobalt0s what happens if the caller puts in duplicate fields?

```
result, err := conn.Read(ctx, connectors.ReadParams{
  ObjectName: "Contact",
  Fields: connectors.Fields("FirstName", "FirstName") // ideally this should not error out
})
```

Can you please start a test file inside `handy` and add a test case for this?

Can you also create a PR in the server repo to update how we are calling the Read method?

Cobalt0s:
* **#998** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/998?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Cobalt0s:
> Hey @Cobalt0s what happens if the caller puts in duplicate fields?
> 
> ```
> result, err := conn.Read(ctx, connectors.ReadParams{
>   ObjectName: "Contact",
>   Fields: connectors.Fields("FirstName", "FirstName") // ideally this should not error out
> })
> ```
> 
> Can you please start a test file inside `handy` and add a test case for this?
> 
> Can you also create a PR in the server repo to update how we are calling the Read method?

@laurenzlong 

Added test suite: https://github.com/amp-labs/connectors/blob/5ef9f57e75297b5c6525527590f3939d5c9df98f/common/handy/set_test.go.
Created draft PR which will be updated once this one is merged: https://github.com/amp-labs/server/pull/1708

laurenzlong (approved):

PR #1000, opened 2024-08-27
feat: Sync Read capability with OpenAPI

Cobalt0s:
# Changes

* Every connector that serves Metadata from `schema.json` has additional check during **Read** that prunes unsupported objects (not part of metadata)
* OpenAPI did some special mapping between `ObjectName` and `JsonPath` holding that object. This logically belongs to the connector and is moved as `DefaultMap` called `ObjectNameToResponseField`. 

# Benefits
`ObjectNameToResponseField` is used by `ZendeskSupport` and `Intercom`.

Because of the move it is not only beneficial for OpenAPI scripts to extract schemas but now it offers correct JSON resolution!
**Zendesk Support Read:**
image

# Summary
This removes the gap between ListObjectMetadata and Read methods. It syncs both, making Read cover **no more** and **no less** than what ListObjetMetadata returns.

### DIFF
--- a/providers/intercom/objectNames.go
+++ b/providers/intercom/objectNames.go
@@ -1,0 +1,26 @@
+package intercom
+
+import (
+	"github.com/amp-labs/connectors/common/handy"
+	"github.com/amp-labs/connectors/providers/intercom/metadata"
+)
+
+// Supported object names can be found under schemas.json.
+var supportedObjectsByRead = handy.NewSet( //nolint:gochecknoglobals
+	metadata.Schemas.GetObjectNames(),
+)
+
+// ObjectNameToResponseField maps ObjectName to the response field name which contains that object.
+var ObjectNameToResponseField = handy.NewDefaultMap(map[string]string{ //nolint:gochecknoglobals
+	"admins":        "admins",
+	"teams":         "teams",
+	"ticket_types":  "ticket_types",
+	"events":        "events",
+	"segments":      "segments",
+	"activity_logs": "activity_logs",
+},
+	func(key string) string {
+		// Other objects are mapped to `data`.
+		return "data"
+	},
+)
--- a/scripts/openapi/README.md
+++ b/scripts/openapi/README.md
@@ -1,0 +1,29 @@
+# Description
+
+This folder contains **scripts that process OpenAPI spec to produce Object Schemas**.
+These schemas are later will be served via ListObjectMetadata.
+
+# Structure
+
+Scripts will be located under `scripts/openapi/<PROVIDER_NAME>/metadata/main.go`.
+
+OpenAPI files that it loads can be found under `providers/<PROVIDER_NAME>/openapi/<FILE_NAME>.yaml|json`.
+The output will be saved under `providers/<PROVIDER_NAME>/metadata/schemas.json`.
+
+# Running instructions
+
+Update OpenAPI file to the latest version, then execute the following command in the project root directory:
+
+```
+go run ./scripts/openapi/intercom/metadata
+```
+Check `providers/intercom/metadata/schemas.json` for any side effects. Please monitor the log output,
+and if there are any errors, manually review the OpenAPI spec. Based on your review,
+decide whether the endpoint should be integrated or ignored.
+
+# Capability
+
+These scripts offer:
+* Control over which parts of the OpenAPI spec are relevant for processing.
+* Formatting options for display names.
+* Establishing relationship between Resource/Object Name and JSON field name, containing said object.
--- a/tools/fileconv/api3/parameters.go
+++ b/tools/fileconv/api3/parameters.go
@@ -1,6 +1,7 @@
 package api3
 
 import (
+	"github.com/amp-labs/connectors/common/handy"
 	"github.com/amp-labs/connectors/common/naming"
 	"github.com/iancoleman/strcase"
 )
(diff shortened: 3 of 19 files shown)


Cobalt0s:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1000?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1001** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1001?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1000** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1000?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#999** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/999?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

RajatPawar:
> # Changes
> The following only apply to **connectors** that use static `schema.json` when serving ListObjectMetadata:
> 
> 1. ObjectName must be passed to Read method.
> 2. Read will error out `common.ErrOperationNotSupportedForObject` for ObjectNames that are not found under schemas.json.
> 3. Parsing `schema.json` is done on package import. JSON parsing error is ignored, we assume the file is valid. If at least one test exists this will be caught.
> 4. Added mapping ObjectName to Response Field for **ZendeskSupport** and **Intercom**.
> 5. Fix ZendeskSupport to locate correct JSON path even for special objects by relying on `ObjectNameToResponseField` map described in the step 4. Note: Intercom returns special `type` field that tells where the array is located so it already covers all ObjectName space.
> 6. Log error when any OpenAPI schema cannot be processed. If we update OpenAPI file we will be notified about unexpected changes.
> 
> # PS
> Let me know if this PR needs splitting.

Please split this up and request a review

Cobalt0s:
> > # Changes
> > The following only apply to **connectors** that use static `schema.json` when serving ListObjectMetadata:
> > 
> > 1. ObjectName must be passed to Read method.
> > 2. Read will error out `common.ErrOperationNotSupportedForObject` for ObjectNames that are not found under schemas.json.
> > 3. Parsing `schema.json` is done on package import. JSON parsing error is ignored, we assume the file is valid. If at least one test exists this will be caught.
> > 4. Added mapping ObjectName to Response Field for **ZendeskSupport** and **Intercom**.
> > 5. Fix ZendeskSupport to locate correct JSON path even for special objects by relying on `ObjectNameToResponseField` map described in the step 4. Note: Intercom returns special `type` field that tells where the array is located so it already covers all ObjectName space.
> > 6. Log error when any OpenAPI schema cannot be processed. If we update OpenAPI file we will be notified about unexpected changes.
> > 
> > # PS
> > Let me know if this PR needs splitting.
> 
> Please split this up and request a review

Ok, lets start with parameter validation. I started with enforcing ObjectName to be required for Read method.
Generalized it more. https://github.com/amp-labs/connectors/pull/1013 The PR is bigger but it addresses 1 issue at the large scale. Lauren asked a while ago about ObjectName being required. 

I will be splitting this PR bit by bit by concept (this helped me generalize the problem).

Cobalt0s:
@RajatPawar I repurpopsed this PR and it was split into chunks. Check graphite for the full tree structure. It is ready for review.

scripts/openapi/README.md
RajatPawar, line 1: Cool!

providers/intercom/objectNames.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/amp-labs/connectors/providers/intercom/metadata"
+)
+
+// Supported object names can be found under schemas.json.
+var supportedObjectsByRead = handy.NewSet( //nolint:gochecknoglobals
+	metadata.Schemas.GetObjectNames(),
+)
+
+// ObjectNameToResponseField maps ObjectName to the response field name which contains that object.
+var ObjectNameToResponseField = handy.NewDefaultMap(map[string]string{ //nolint:gochecknoglobals
```
RajatPawar, line 14: Shouldn't we use this in the `intercom/read` like we are doing for zendesk?

tools/fileconv/api3/parameters.go
```diff
@@ -21,6 +22,16 @@ func DataObjectCheck(objectName, fieldName string) bool {
 	return fieldName == "data"
 }
 
+// CustomMappingObjectCheck builds ObjectCheck using mapping,
+// which knows exceptions and patterns to determine response field name.
+func CustomMappingObjectCheck(dict *handy.DefaultMap[string, string]) ObjectCheck {
```
RajatPawar, line 27: Can you add an example as a comment? This is unclear by itself

RajatPawar (changes requested):

providers/intercom/objectNames.go
Cobalt0s: That was my initial intent. Any future connector that will have unique mapping between 2 concepts would do just like Zendesk handles it. But :smile: intercom always returns `type` field. Which tells us where the array is stored. This is convinent and I prefer programatic solution over hard coded as this requires no maintances.

But we could remove this code https://github.com/amp-labs/connectors/blob/3c4deeb3cf1f01e97b9f836b90bdb7d96c7c255e/providers/intercom/parse.go#L89 and fallback to `ObjectNameToResponseField` resolution. Even if provider API gets new endpoint that this programatic `type` inference will be able to catch our `ListObjectMetadata` would still not know about this. Since both Read and Metadata are so connected we can't harvest much benefit from programatic `type` inference, because ListObjectMetadata holds us back. 

@RajatPawar what is your opinion on this? Which approach do you prefer?

tools/fileconv/api3/parameters.go
Cobalt0s: New method documentation header:
https://github.com/amp-labs/connectors/blob/13184db44673c35ef70bc145c42ee70c86014272/tools/fileconv/api3/parameters.go#L25-L38

providers/intercom/objectNames.go
RajatPawar: Sure, if we can do it programmatically, that's always easier on the soul :) Can you just add a comment here that tells the reader about the context you added in here? Something like `Intercom returns a type field which tells us where the response array is located, which is why we do not need to hardcode the mapping. If we need to override this at any point, we can add the mapping here.` (please add additional context if missing)

RajatPawar (approved):
LGTM - please resolve the comment before merging

PR #1002, opened 2024-08-28
[ENG-1511] chore: Add filter for salesforce

RajatPawar:
This adds a filter field to `ReadParams`, which is supported only for Salesforce. This should be removed in favour of a better filter concept later on.

## With filters
Screenshot 2024-08-28 at 7 19 39 PM
Screenshot 2024-08-28 at 7 17 34 PM
Screenshot 2024-08-28 at 7 17 57 PM

### DIFF
--- a/providers/salesforce/read.go
+++ b/providers/salesforce/read.go
@@ -67,5 +67,9 @@ func makeSOQL(config common.ReadParams) string {
 		soql.Where("IsDeleted = true")
 	}
 
+	if config.Filter != "" {
+		soql.Where(config.Filter)
+	}
+
 	return soql.String()
 }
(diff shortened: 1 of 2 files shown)


providers/salesforce/read.go
```diff
@@ -67,5 +67,9 @@ func makeSOQL(config common.ReadParams) string {
 		soql.Where("IsDeleted = true")
 	}
 
+	if config.Filter != "" {
+		soql.Where(config.Filter)
+	}
```
eberle1080, line 72: Does SOQL have the equivalent of SQL injection? This seems like a serious possibility here. I'm gonna approve this since Ampersand controls the contents of this field but this is a concern for me.

eberle1080 (approved):

providers/salesforce/read.go
RajatPawar: They have a section on this (https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/pages_security_tips_soql_injection.htm) and their advice is basically to stick to static queries. I am going to add a comment here to remind us to be aware of this.

PR #1003, opened 2024-08-28
feat: Add Read Apollo connector

Jkarage:


RajatPawar (changes requested):
From the docs, seems like the Apollo API basically has two different ways of reading - listing, and searching. 

To support as many objects as we can, let's route the Read request to either a Search method or a List method, based on the object name. That way, we can support most of the objects that exist in the API.

RajatPawar:
After merging this PR in, please remember to create a PR against the docs (https://github.com/amp-labs/docs) to add the supported objects and limitations in there. You can use a deep provider guide (Instantly, Hubspot, Salesforce, etc) as a reference.

PR #1008, opened 2024-08-29
feat: Add Apollo ListMetadata connector

Jkarage:
Uses reading of the object to generate the metadata. 

Screenshot 2024-08-29 at 08 54 21

### DIFF
--- a/providers/apollo/metadata.go
+++ b/providers/apollo/metadata.go
@@ -1,0 +1,90 @@
+package apollo
+
+import (
+	"context"
+	"errors"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/jsonquery"
+	"github.com/spyzhov/ajson"
+)
+
+// ListObjectMetadata creates metadata of object via reading objects using Apollo API.
+func (c *Connector) ListObjectMetadata(ctx context.Context,
+	objectNames []string,
+) (*common.ListObjectMetadataResult, error) {
+	// Ensure that objectNames is not empty
+	if len(objectNames) == 0 {
+		return nil, common.ErrMissingObjects
+	}
+
+	metadataResult := common.ListObjectMetadataResult{
+		Result: make(map[string]common.ObjectMetadata),
+		Errors: make(map[string]error),
+	}
+
+	for _, obj := range objectNames {
+		url, err := c.getAPIURL(obj)
+		if err != nil {
+			return nil, err
+		}
+
+		resp, err := c.Client.Get(ctx, url.String())
+		if err != nil {
+			metadataResult.Errors[obj] = err
+
+			continue
+		}
+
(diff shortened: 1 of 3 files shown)


providers/apollo/metadata.go
```diff
@@ -0,0 +1,90 @@
...
+		Errors: make(map[string]error),
+	}
+
+	for _, obj := range objectNames {
+		url, err := c.getAPIURL(obj)
+		if err != nil {
+			return nil, err
+		}
+
+		resp, err := c.Client.Get(ctx, url.String())
```
RajatPawar, line 32: Is there a way to get just one record? That way, we don't end up fetching hundreds of thousands of records that we do not need

providers/apollo/metadata.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		return common.ObjectMetadata{}, err
+	}
+
+	resultArr, err := jsonquery.New(root).Array(obj, true)
+	if err != nil {
+		return common.ObjectMetadata{}, err
+	}
+
+	objectResponnse := resultArr[0].MustObject()
```
RajatPawar, line 82: Are there any rate limits / quota considerations we need to take into account? Because this will count against it. We could very well just have a static metadata file like we have for some other connectors

providers/apollo/metadata.go
```diff
@@ -0,0 +1,90 @@
...
+		metadata, err := parseMetadataFromResponse(resp, obj)
+		if err != nil {
+			if errors.Is(err, common.ErrMetadataLoadFailure) {
+				metadataResult.Errors[obj] = common.ErrEmptyResponse
+			} else {
+				return nil, err
+			}
+		}
+
+		metadata.DisplayName = obj
```
RajatPawar, line 55: We should make this object name more human readable

RajatPawar (changes requested):

providers/apollo/metadata.go
Jkarage: Yes, there is (depending on plan) the minimum(free trial) is 50 calls/min, 200/hour and 600/day. So this will have an effect on it. Was looking into static metadata file as well. But couldn't find  specs files. If you agree, Let me create a script that does this for all objects and create the static file for it. So that we wouldn't be doing this for every integration.

providers/apollo/metadata.go
Jkarage: No, They are ids are random, 24 characters.

Jkarage: For some objects, we can ensure that we are only returning a single page of records.

Jkarage: @RajatPawar This is a possibility. Using pagination details(single page and limiting the records in a page), we can only return a single record. This would solve not returning so much data for the interest of one.

RajatPawar (approved):

PR #1010, opened 2024-08-29
feat: Bulk Read Salesforce

Cobalt0s:
# Changes
There is file relocation from folder `bulk<operation>` to nested `bulk/<operation>`.
I noticed some imports aliased salesforce as `salesforce2`. This should be that unintended IDE changes slipped into main.
One big `bulk.go` is split into dedicated files and every public bulk method has documentation and documentation link.
https://github.com/amp-labs/connectors/pull/1010/commits/cee45550ab30b2a8ffb8a3392548e78db3de2b6c

Read itself is a modification of existing BulkQuery: https://github.com/amp-labs/connectors/pull/1010/commits/6d8756fe7bec6fd86dbd593c0d7e3664c94f61c9.

# Breaking
Renamed BulkQuery to LaunchBulkQuery same for LaunchBulkWrite and LaunchBulkDelete.
This tells that it creates side effects of launching async job and further method calls must be made to collect result.

# Review
I have split commits to related changes but they are not clear cut. If this should be split into PRs, where themes are:
* Moved files
* Split bulk file
* Added docs

Let me know.

# Test Results.
Everything works as before.

## Bulk Read
image

## Bulk Query
image

## Bulk Write
image
image

## Bulk Delete
image


providers/salesforce/bulk-info.go
```diff
@@ -0,0 +1,129 @@
...
+func (c *Connector) GetBulkQueryResults(ctx context.Context, jobId string) (*http.Response, error) {
+	location, err := c.getRestApiURL("jobs/query/", jobId, "/results")
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := common.MakeJSONGetRequest(ctx, location.String(), []common.Header{
+		{
+			Key:   "Accept",
+			Value: "text/csv",
```
RajatPawar, line 44: From the docs, seems like we can also get JSON results, can we allow for that? That'll be useful for reads.

providers/salesforce/bulk-write.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// LaunchBulkWrite starts Bulk Job to upsert records.
+// https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/create_job.htm
+//
+// After creation inspect newly launched Bulk Job via:
+// * GetJobInfo
+// * GetJobResults
+// * GetSuccessfulJobResults.
+func (c *Connector) LaunchBulkWrite( //nolint:funlen,cyclop
```
RajatPawar, line 15: I understand the rationale behind calling it `Launch` since it's a job, I would still want to call this as `BulkWrite`, and the same applies for other methods. It needs to align with similar methods that we will want to introduce for other providers in the future.

Cobalt0s: I thought that it could be the case. I will do that.
However, the job info, job results or job successful results are confusing. I mention those in the method documentation to explain how things are related, it is not apparent from the code.

providers/salesforce/bulk-info.go
Cobalt0s: Sadly, no. I was wondering that too. According to the documentation it says the only supported option is CSV. The option is there but with one choice ahaha, is it even an option, never understood such motive.
CreateJob: https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/query_create_job.htm
image

providers/salesforce/bulk-write.go
Cobalt0s: Changed: https://github.com/amp-labs/connectors/pull/1010/commits/f2c83f52a9b267202ce9586e71fe522d56452250

providers/salesforce/bulk.go
```diff
@@ -357,35 +289,18 @@ func getColumnIndice(record []string, columnNames []string) (map[string]int, err
...
 }
 
-func (c *Connector) getIncompleteJobResults(jobInfo *GetJobInfoResult) *JobResults {
-	jobResult := &JobResults{
-		JobId:   jobInfo.Id,
-		State:   jobInfo.State,
-		JobInfo: jobInfo,
-	}
-
+func getIncompleteJobMessage(jobInfo *GetJobInfoResult) string {
```
jlimatampersand, line 292: What was the reason behind this refactor? This jobInfo state can stay as the connector method since. Though I see that there is no reference to `c`. Seems more cosmetic I guess. no big deal but just curious.

providers/salesforce/read.go
```diff
@@ -48,13 +48,13 @@ func (c *Connector) buildReadURL(config common.ReadParams) (*urlbuilder.URL, err
...
 
-	url.WithQueryParam("q", makeSOQL(config))
+	url.WithQueryParam("q", makeSOQL(config).String())
 
 	return url, nil
 }
 
 // makeSOQL returns the SOQL query for the desired read operation.
-func makeSOQL(config common.ReadParams) string {
+func makeSOQL(config common.ReadParams) *soqlBuilder {
```
jlimatampersand, line 57: nice.

providers/salesforce/bulk-query.go
```diff
@@ -0,0 +1,49 @@
...
+// After creation inspect newly launched Bulk Job via:
+// * GetBulkQueryInfo
+// * GetBulkQueryResults.
+func (c *Connector) BulkQuery(
+	ctx context.Context,
+	query string,
+	includeDeleted bool,
+) (*GetJobInfoResult, error) {
+	operation := "query"
+	if includeDeleted {
```
jlimatampersand, line 27: nice upgrade.

jlimatampersand (approved):
Looks good. But make sure to test results that are successful before merging this. You will need to add new test data that are standard objects. Currently they are custom objects which should be replace.

Cobalt0s (commented):
@jlimatampersand left some comments. Let me know what do you think about it

providers/salesforce/bulk.go
Cobalt0s: Yes that is cosmetic. I was hesitant. Because naming suggested that it had related sibling method but using `c`, while this one wasn't.
I decided that it is not a connector method. Connector already has a lot of methods.

PR #1011, opened 2024-08-30
update Marketo ListMetadata with OpenAPI fallback

Jkarage:
Context:
Using the current implementation, If the object being read has no data(0 records returned), we would fail to generate the metadata for that particular object. This update resolves this by using the marketo's swagger v2 specification files. 

- mapi.json (Leads)
- assets.json (Assets)

These files are obtained here:  https://developers.marketo.com/rest-api/endpoint-reference/download-swagger-definition/

Created a  script that generates the schemas.json file (Added in `providers/marketo/metadata` dir). We can now use this to generate the metadatas, instead of making calls and thus resolving the issue above.

Running the tests:
Screenshot 2024-08-31 at 11 54 37

After Adding the fallback: Initially we were failing to retrieve the `files` metadata, returns 0 records. 
Now we can.
Screenshot 2024-09-04 at 06 41 44


scripts/openapi/marketo/metadata/metadata.go
```diff
@@ -0,0 +1,152 @@
...
+)
+
+func main() {
+	//  read the definitions in the specificatios files.
+	def, docA, err := constructDefinitions("./scripts/openapi/marketo/metadata/asset.json", 5) //nolint:gomnd
+	if err != nil {
+		panic(err)
+	}
+
+	ldef, docL, err := constructDefinitions("./scripts/openapi/marketo/metadata/mapi.json", 4) //nolint:gomnd
```
RajatPawar, line 22: What do the `4` or `5` mean in `constructDefinitions`? Please add a comment for context

scripts/openapi/marketo/metadata/metadata.go
```diff
@@ -0,0 +1,152 @@
...
+	return s[len(s)-1]
+}
+
+func generateMetadata(objDefs map[string]string,
+	doc openapi2.T, objectMetadata map[string]scrapper.ObjectMetadata,
+) (map[string]scrapper.ObjectMetadata, error) {
+	for obj, dfn := range objDefs {
+		schemas := doc.Definitions[dfn].Value.Properties
+
+		// This is attached to the marketo
```
RajatPawar, line 123: incomplete comment

scripts/openapi/marketo/metadata/metadata.go
```diff
@@ -0,0 +1,152 @@
...
+		result, err := schemas["result"].Value.JSONLookup("items")
+		if err != nil {
+			return nil, err
+		}
+
+		r, ok := result.(*openapi3.Ref)
+		if !ok {
+			return nil, fmt.Errorf("failed the assertion, the response data type is not expected") //nolint:goerr113
```
RajatPawar, line 131:
```suggestion
			return nil, fmt.Errorf("failed to convert the response type into an openapi3 type") //nolint:goerr113
```

RajatPawar (changes requested):
This is very helpful, thanks! 

A potential concern is that specs don't exhaustively list out all the fields that actually exist on an object. What I recommend is that we retain the existing logic, and in case we are not able to retrieve a record for the object (either due to an error, or because there were 0 records returned), then we fallback to returning the metadata using the schema file.

If we can only request for 1 record instead of listing all records for that object, let's do that. And also, please add comments to explain the fallback logic.

scripts/openapi/marketo/metadata/metadata.go
Jkarage: The `paths` we support  would generate 4 and 5 substrings, when splitted using `/`, for Leads and Assets API respectively.

scripts/openapi/marketo/metadata/metadata.go
```diff
@@ -0,0 +1,152 @@
...
+	"strings"
+
+	"github.com/amp-labs/connectors/tools/scrapper"
+	"github.com/getkin/kin-openapi/openapi2"
+	"github.com/getkin/kin-openapi/openapi3"
+)
+
+func main() {
+	//  read the definitions in the specificatios files.
+	def, docA, err := constructDefinitions("./scripts/openapi/marketo/metadata/asset.json", 5) //nolint:gomnd
```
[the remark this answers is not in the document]
Cobalt0s: There is no need to specify Path strings as you can embed file byte data as in Pipeliner: https://github.com/amp-labs/connectors/blob/main/providers/pipeliner/openapi/file.go

Since this connector needs 2 files it will look something like this:
```
//go:embed asset.json
apiFileAsset []byte
//go:embed mapi.json
apiFileMapi []byte

FileManagerAsset = api3.NewOpenapiFileManager(apiFileAsset) // nolint:gochecknoglobals
FileManagerMapi  = api3.NewOpenapiFileManager(apiFileMapi)  // nolint:gochecknoglobals
```
Then these `FileManagers` can be called to extract schemas.

scripts/openapi/marketo/metadata/metadata.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	// Create a `schemas.json` file and Adds the metadata details to the file.
+	if err = writefile(mb); err != nil {
+		panic(err)
+	}
+}
+
+func writefile(b []byte) error {
+	f, err := os.Create("schemas.json")
```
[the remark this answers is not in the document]
Cobalt0s: The file manager from scrapper package which you have already declared (maybe package should be called differently as it no longer makes sense, before it was scrapper only).
```
FileManager = scrapper.NewMetadataFileManager(schemas, fileconv.NewSiblingFileLocator())
```
This filemanager has utility methods like `SaveSchemas()`. And it will save it to the neighbouring schema.json file.
This method accepts `scrapper.NewObjectMetadataResult()` data structure, which has convenient `Add` method this can be used to add schemas from one file then from the other and finally save to `schemas.json`.

@Jkarage If you have any questions I can help you, dont hesitate to ask!

Cobalt0s (commented):
Hi, you should've tagged me on OpenAPI PR.
Left some comments.
Check this README https://github.com/amp-labs/connectors/tree/main/tools/fileconv/api3

scripts/openapi/marketo/metadata/metadata.go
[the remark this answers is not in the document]
Jkarage: @Cobalt0s yeah, I saw the tool as well. The reason I decided to create the script is as the wrapper `api3` uses the OpenAPI v3 parser. And the marketo files needs v2. We can look more into it if they are full compatible and update this script. Or Update the wrapper to also include the v2 parser.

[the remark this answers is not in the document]
Cobalt0s: Well, that is really unfortunate... I guess we can later think about having api3 and api2 with some common core code. I wonder how many more connectors will have v2.
I guess only `scrapper.NewMetadataFileManager` could be of use.

providers/marketo/metadata.go
```diff
@@ -37,22 +38,57 @@ func (c *Connector) ListObjectMetadata(ctx context.Context,
 
 		resp, err := c.Client.Get(ctx, url.String())
 		if err != nil {
-			metadataResult.Errors[obj] = err
+			// Try fallback function
```
RajatPawar, line 41: * Since this fallback appears multiple times, you can put it in a helper and call that
* You can combine the checks as `err != nil || resp == nil...`

RajatPawar (approved):

PR #1017, opened 2024-09-03
tidy: Consistent test connector setup

Cobalt0s:
# Changes

Removed repo `connectors/utils` because
* Oauth2 credentials now live alongside the connector constructor.
* Required credential fields are automatically implied using previously added uility `MustCreateProvCredJSON`.
* Every connector repo therefore has constructor located under `test/<PROVIDER_NAME>/connector.go`.
* Connector instantiation code is consistent among connectors.

# Review
Commits were done incrementally as I was applying the same rules over and over. They have nice progression.


Cobalt0s:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1017?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1017** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1017?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1010** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1010?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

test/apollo/connector.go
```diff
@@ -4,41 +4,24 @@ import (
...
 
-	"github.com/amp-labs/connectors/common/scanning"
 	"github.com/amp-labs/connectors/common/scanning/credscanning"
 	"github.com/amp-labs/connectors/providers"
 	"github.com/amp-labs/connectors/providers/apollo"
 	testUtils "github.com/amp-labs/connectors/test/utils"
-	"github.com/amp-labs/connectors/utils"
 )
 
-func GetApolloConnector(ctx context.Context, filePath string) *apollo.Connector {
```
RajatPawar, line 15: How would one override this file path if they needed to?

test/gong/connector.go
```diff
@@ -2,55 +2,50 @@ package gong
...
+	"github.com/amp-labs/connectors/test/utils"
+	"golang.org/x/oauth2"
+	"net/http"
 )
 
-func GetGongConnector(ctx context.Context, filePath string) *gong.Connector {
-	registry := scanning.NewRegistry()
-
-	readers := []scanning.Reader{
-		&scanning.JSONReader{
```
RajatPawar, line 17: Where is this mapping living now? Or is it redundant now?

test/apollo/connector.go
Cobalt0s: Every test file had code as follows:
```
filePath := os.Getenv("MS_CRM_CRED_FILE")
if filePath == "" {
	filePath = "./ms-crm-creds.json"
}
```
These things are preserved. So no changing path via code but via arguments.
Environment variable allows such configuration. With apollo it would be `APOLLO_CRED_FILE`.
See:
https://github.com/amp-labs/connectors/blob/main/common/scanning/credscanning/path.go

test/gong/connector.go
Cobalt0s: For most part it is redundant. We shouldn't specify what fields are to be read from the file. The provider catalog has all the information (at least it should in final catalog version, ex: require workspace flag is yet to be used by all types of connectors, we explicitly say we need this, in future it should be implied as well).

In here we imply fields using `providerInfo`: https://github.com/amp-labs/connectors/blob/3c4deeb3cf1f01e97b9f836b90bdb7d96c7c255e/common/scanning/credscanning/provider.go#L62 (and flags, which are temporary as I said).

Then for every field create reader: https://github.com/amp-labs/connectors/blob/3c4deeb3cf1f01e97b9f836b90bdb7d96c7c255e/common/scanning/credscanning/provider.go#L73

And finally validate required fields. Every field key and value will be cached in the map. https://github.com/amp-labs/connectors/blob/3c4deeb3cf1f01e97b9f836b90bdb7d96c7c255e/common/scanning/credscanning/provider.go#L87

When making `ProviderCredentials.Get` call it will give you that string value. https://github.com/amp-labs/connectors/blob/3c4deeb3cf1f01e97b9f836b90bdb7d96c7c255e/common/scanning/credscanning/provider.go#L135

RajatPawar (approved):
LGTM, please run all of the tests to ensure there's no regression

PR #1019, opened 2024-09-04
update record size response in ListMetadata

Jkarage:
This limits the record response returned during metadata generation to only 1 record. 

- Outreach
- Marketo (forgot to implement this in the previous PR, as you recommended)

Running tests:
Outreach
Screenshot 2024-09-04 at 15 06 00

Marketo
Screenshot 2024-09-04 at 15 42 14

### DIFF
--- a/providers/outreach/metadata.go
+++ b/providers/outreach/metadata.go
@@ -7,6 +7,11 @@ import (
 	"github.com/amp-labs/connectors/common"
 )
 
+var (
+	metadataPageSize      string = "1"          //nolint:gochecknoglobals
+	metadataPageSizeQuery string = "page[size]" //nolint: gochecknoglobals
+)
+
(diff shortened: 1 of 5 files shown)


providers/outreach/metadata.go
```diff
@@ -7,6 +7,11 @@ import (
 	"github.com/amp-labs/connectors/common"
 )
 
+var (
+	metadataPageSize      string = "1"          //nolint:gochecknoglobals
+	metadataPageSizeQuery string = "page[size]" //nolint: gochecknoglobals
```
RajatPawar, line 12: Just making sure, is `size` supposed to be substituted or does this actually look like this?

Jkarage: Looks like that. No substitution.

PR #1021, opened 2024-09-05
fix: Id retrieval Marketo connector

Jkarage:
As I was testing writing, I found an error with some Objects not having a standard `id` as a unique field for the record. Rather they have the `marketoGUID` as the unique field. 

The Objects are: namedAccountLists,namedaccounts and opportunities. All in the Leads Module.
- This updates the write to take this into account.

- I have removed the route for adding the path `id` for updating, Which was mainly for the Assets Module, which we currently do not support.

- Also all rest APIs/Endpoints within Leads Module, can write more than 1 record at a time. But with the current implementation, we are assuming the client writes 1 object at a time. If a client writes multiple we will only look if the first record was created successfully and return it's id, Or return an err  if it was not. 

Running test cases:
- Objects returning `id`
Screenshot 2024-09-11 at 17 10 49

- Objects returning `marketoGUID`
Screenshot 2024-09-11 at 17 11 35

Failing case:
- Writing to opportunities, failing as we have not provided all required fields. This generates a RecordLevel Err. Now we can return this to the client.
Screenshot 2024-09-11 at 17 32 30

### DIFF
--- a/providers/marketo/write.go
+++ b/providers/marketo/write.go
@@ -57,6 +57,7 @@ func constructId(objectName string, resp *writeResponse) (string, error) {
 		recordId = resp.Result[0]["id"]
 		// By default the recordId is returned as a float64
 		recordId, success = recordId.(float64)
+
 		err = checkErr(resp, recordId, success)
 		if err != nil {
 			return "", err
@@ -66,6 +67,7 @@ func constructId(objectName string, resp *writeResponse) (string, error) {
 		recordId = resp.Result[0]["marketoGUID"]
 		// By default the recordId is returned as a string
 		recordId, success = recordId.(string)
+
 		err = checkErr(resp, recordId, success)
 		if err != nil {
 			return "", err


providers/marketo/write.go
```diff
@@ -42,17 +29,57 @@ func (c *Connector) Write(ctx context.Context, config common.WriteParams) (*comm
...
+		recordId any
+		success  bool
+	)
+
+	switch {
+	case usesStandardId(objectName):
+		recordId = resp.Result[0]["id"]
+		// By default the recordId is returned as a float64
+		recordId, success = recordId.(float64)
+		if !success || recordId == 0 {
```
RajatPawar, line 55: Can you please add a case for this failure in the tests?

providers/marketo/write.go
```diff
@@ -42,17 +29,57 @@ func (c *Connector) Write(ctx context.Context, config common.WriteParams) (*comm
...
+			}
+
+			return "", errors.New(message) //nolint: goerr113
+		}
+
+	case usesMarketoGUID(objectName):
+		recordId = resp.Result[0]["marketoGUID"]
+		// By default the recordId is returned as a float64
+		recordId, success = recordId.(string)
+		if !success || recordId == "" {
```
RajatPawar, line 70: Seems like this logic is shared between the cases, so we can extract it out

test/marketo/write/write.go
```diff
@@ -17,15 +17,22 @@ func main() {
 }
 
 func MainFn() int {
-	err := testWrite(context.Background())
+	ctx := context.Background()
```
RajatPawar, line 20: After you update the tests for both kinds of objects, please add screenshots of it in the PR

RajatPawar (approved):
Approved, please resolve all the comments before merging.

PR #1023, opened 2024-09-05
feat: Add interface Supported Objects

Cobalt0s:
# Background
During testing, exploration of Salesforce APIs ended up writing query to get list of supported objects.
Simple API call with some processing. Sample output included for reference as `output.txt`.

# Future
This PR will stay as a draft. Need to discuss value of returning supported objects.


Cobalt0s:
* **#1023** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1023?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Cobalt0s:
@RajatPawar need your opinion on this Interface. When I was working on Salesforce I needed to know what Objects can I use for testing. I ended up writing this script very fast and decided to keep it as a draft PR. Implementing something like this is easy, but requires maintance, therefore, I am not sure how much benefit it provides.

Note: connectors that rely on OpenAPI file can easily implement this interface as static file already has all objects within.

There is a benefit of usage: 
* pick a connector
* call GetSupportedObjects
* pick an object and call ListObjectMetadata
* then use Read/Write, etc
This makes resources discoverable and methods create a small connected ecosystem.

RajatPawar:
> @RajatPawar, I need your opinion on this Interface. When I was working on Salesforce I needed to know what Objects can I use for testing. I ended up writing this script very fast and decided to keep it as a draft PR. Implementing something like this is easy, but requires maintance, therefore, I am not sure how much benefit it provides.
> 
> Note: connectors that rely on OpenAPI file can easily implement this interface as static file already has all objects within.
> 
> There is a benefit of usage:
> 
> * pick a connector
> * call GetSupportedObjects
> * pick an object and call ListObjectMetadata
> * then use Read/Write, etc
>   This makes resources discoverable and methods create a small connected ecosystem.

Interesting, thanks for drafting this up! 

I am a little unclear - does this return the objects that Salesforce supports or the objects that are supported for our specific instance? In the first case, if we make a call, we would basically understand if it is supported. In the latter case, that might be interesting!

Cobalt0s:
> I am a little unclear - does this return all the objects that Salesforce supports or the objects that are supported for our specific instance? In the first case, if we make a call, we would basically understand if it is supported. In the latter case, that might be interesting!

Supported objects include both standard and custom objects defined for each account, which aligns with the first case you described! Each account owner will receive different outputs from this method. I like the discoverablity of APIs and Salesforce fits well into this category as it exposes the right endpoints to achieve this. There you can find:
* the max batch size when querying resource 
* Whether the the Object is custom or standard
* what operations can be applied (search, create, delete)

See both docs:
https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_describeGlobal.htm
https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_describeGlobal.htm


For connectors relying on OpenAPI file, only "standard" objects will be returned. Should the connector have capabilities similar to Salesforce, then we would also add custom objects, resulting with a hybrid approach.
That is the meaning I put under `GetSupportedObjects` - retrieving the objects supported by my account.

RajatPawar:
> > I am a little unclear - does this return all the objects that Salesforce supports or the objects that are supported for our specific instance? In the first case, if we make a call, we would basically understand if it is supported. In the latter case, that might be interesting!
> 
> Supported objects include both standard and custom objects defined for each account, which aligns with the first case you described! Each account owner will receive different outputs from this method. I like the discoverablity of APIs and Salesforce fits well into this category as it exposes the right endpoints to achieve this. There you can find:
> 
> * the max batch size when querying resource
> * Whether the the Object is custom or standard
> * what operations can be applied (search, create, delete)
> 
> See both docs: https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_describeGlobal.htm https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_describeGlobal.htm
> 
> For connectors relying on OpenAPI file, only "standard" objects will be returned. Should the connector have capabilities similar to Salesforce, then we would also add custom objects, resulting in a hybrid approach. That is the meaning I put under `GetSupportedObjects` - retrieving the objects supported by my account.

Understood, thanks. This is quite helpful, but not enough to be a connector type by itself (since we don't know how well this applies across providers, and it is highly likely that most APIs don't support introspection well), so I'd have this as a util/script until then.

PR #1028, opened 2024-09-10
tidy: testroutines improvement with generics

Cobalt0s:
# Description

Discovered that test cases for Read/Write/Delete/ListObjectMetadata can be further improved with generics making the code shorter.

Another advantage is custom `TestCase` can be declared by any connector. This allows to reuse the same testing checks (expected errors, output matching) and turns the problem into type declaration instead of rewriting boiler plate.

Included `README` file for those who are planning on using this package with examples.

### DIFF
--- a/test/utils/testroutines/README.md
+++ b/test/utils/testroutines/README.md
@@ -1,0 +1,81 @@
+# Package testroutines
+
+## Purpose
+
+This package provides base test case for creating `Test Suites`.
+
+## Description
+
+Most common connector methods can be tested using:
+
+* testroutines.Read - Read
+* testroutines.Write - Write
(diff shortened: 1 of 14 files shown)


test/utils/testroutines/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+// Example calling method
+tt.Run(t, func() (connectors.ReadConnector, error) {
+    return constructTestConnector(tt.Server.URL)
+})
+```
+
+### TestCase
+
+A **Test case** consists of:
```
RajatPawar, line 38: Can you please add an example of a particular connector?

RajatPawar (approved):

PR #1030, opened 2024-09-10
test: Salesforce Bulk Write

Cobalt0s:
# Description
Using real requests recorded Provider API output to create mock tests.

Added validation to **BulkWrite** to require **ObjectName**, **ExternalIdField** (enforced by Salesforce, so we do that too), and **CSV reader**.

The test immitiates 3 API calls. 3 different endpoints are called to initiate Bulk Job, upload CSV and finally update bulk job to mark the completion of upload.

### DIFF
--- a/providers/salesforce/bulk-write_test.go
+++ b/providers/salesforce/bulk-write_test.go
@@ -1,0 +1,254 @@
+package salesforce
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/interpreter"
+	"github.com/amp-labs/connectors/test/utils/mockutils"
+	"github.com/amp-labs/connectors/test/utils/mockutils/mockserver"
+	"github.com/amp-labs/connectors/test/utils/testroutines"
+	"github.com/amp-labs/connectors/test/utils/testutils"
+)
+
+func TestBulkWrite(t *testing.T) { //nolint:funlen,gocognit,cyclop,maintidx
+	t.Parallel()
+
+	responseCreateJob := testutils.DataFromFile(t, "bulk/write/launch-job-opportunity.json")
+	responseUpdateJob := testutils.DataFromFile(t, "bulk/write/update-job-opportunity.json")
+
+	bodyRequest := `{
+		"contentType":"CSV",
+		"externalIdFieldName":"external_id__c",
+		"lineEnding":"LF",
+		"object":"Opportunity",
+		"operation":"upsert"
+	}`
+
+	tests := []bulkWriteTestCase{
+		{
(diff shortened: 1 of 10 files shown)


providers/salesforce/bulk-write_test.go
```diff
@@ -0,0 +1,254 @@
...
+			},
+			Server: httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				w.Header().Set("Content-Type", "application/json")
+				w.WriteHeader(http.StatusInternalServerError)
+				_, _ = w.Write([]byte{})
+			})),
+			ExpectedErrs: []error{ErrCreateJob},
+		},
+		{
+			Name: "Create job id must be string",
```
jlimatampersand, line 100: I am trying to understand, what is the significance of testing the responses from Salesforce by returning mock responses?

Is this to detect salesforce API change? We do not govern what kind of responses we get from Salesforce. So whether we get "Open" as state or not is not something we can test on. Just trying to understand reason behind this.

providers/salesforce/bulk-read_test.go
```diff
@@ -0,0 +1,117 @@
...
+			ExpectedErrs: []error{common.ErrMissingObjects},
+		},
+		{
+			Name:         "At least one field is requested",
+			Input:        common.ReadParams{ObjectName: "Orders"},
+			Server:       mockserver.Dummy(),
+			ExpectedErrs: []error{common.ErrMissingFields},
+		},
+		{
+			Name:         "Mime response header expected",
```
jlimatampersand, line 38: I don't think we govern response headers.

jlimatampersand (commented):
Please response to my comments and ping me when it is ready to be re-reviewed.

providers/salesforce/bulk-write_test.go
Cobalt0s: The whole reason I introduced this mock tests which are now present in almost every deep connector is to capture the functionallity inside code. Otherwise, any change would require manually running all scripts. This is not always possible due to license expiration, even connector setup (ex: Salesforce needs custom fields to be declared for bulk tests to run), and it would be time consuming. Therefore, these unit tests imitate what provider would respond and enforce certain output, connector behaviour. You can think of it as regression testing.

providers/salesforce/bulk-read_test.go
Cobalt0s: As I wrote above it is for guarding against regression. We expect our code to handle this execution path. However, it does look repetitive, I am not sure how that can be generalised effectively and be applied to some if not all deep connectors.

Also by looking on these tests you can conclude what the code is doing. Tests preserve this by being a live contract. They are easy to run and done on every CI. Of course when the underlying provider changes due to breaking API the tests would need to follow and adapt. But for that there should be module support and per module tests.

jlimatampersand (approved):

PR #1031, opened 2024-09-10
test: Salesforce Bulk Delete

Cobalt0s:
# Description
Using real requests recorded Provider API output to create mock tests.

`BulkDelete` is very similiar to `BulkWrite` therefore created common `mock server` which is reused by both **TestSuites**.

Added validation to BulkDelete argument params. Also removed a comment that ExternalIdField is required (that is only true for **upserts**).

### DIFF
--- a/providers/salesforce/bulk-delete_test.go
+++ b/providers/salesforce/bulk-delete_test.go
@@ -1,0 +1,88 @@
+package salesforce
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/interpreter"
+	"github.com/amp-labs/connectors/test/utils/mockutils/mockserver"
+	"github.com/amp-labs/connectors/test/utils/testroutines"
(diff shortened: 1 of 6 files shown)


RajatPawar (commented):
Adding Jay as a reviewer to make sure that the test coverage is per what we need

providers/salesforce/bulk-delete_test.go
```diff
@@ -0,0 +1,88 @@
...
+	t.Parallel()
+
+	responseCreateJob := testutils.DataFromFile(t, "bulk/delete/launch-job-opportunity.json")
+	responseUpdateJob := testutils.DataFromFile(t, "bulk/delete/update-job-opportunity.json")
+
+	bodyRequest := `{"object":"Opportunity","operation":"delete"}`
+
+	tests := []bulkDeleteTestCase{
+		{
+			Name: "Mime response header expected",
```
jlimatampersand, line 27: we do not govern response headers.

jlimatampersand (commented):
Same thing, you can ping me when you answer question on the other test PR.

jlimatampersand (approved):

PR #1034, opened 2024-09-11
[ENG-1536] chore: Add list job results endpoint

RajatPawar:
## Context
We can get all job results in a single call instead of calling this endpoint for each job - https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/get_all_jobs.htm

## Testing
When I asked for ~4 jobs that were spread out over ~38 pages, i.e. from page 200 to page 7800
Screenshot 2024-09-11 at 3 20 47 PM
Screenshot 2024-09-11 at 3 20 42 PM

the quota used up was 3 units. I am fairly certain that it's only 1 unit, and the other 2 units were from previous calls. This is because when I tested this function to get ALL jobs, only 1 unit was used up.
Screenshot 2024-09-11 at 3 20 29 PM

## Benchmark
To pick out 7 jobs out of the dataset, it took the function ~15 seconds, going over 40 pages / 8000 records
Screenshot 2024-09-11 at 3 38 08 PM

### DIFF
--- a/providers/salesforce/bulk-info.go
+++ b/providers/salesforce/bulk-info.go
@@ -32,8 +32,9 @@ func (c *Connector) ListIngestJobsInfo(ctx context.Context, jobIds ...string) ([
 	// To collect all the jobs we need to return
 	var jobsInfo []GetJobInfoResult
 
-	// If we have jobIds, we create a set to keep track of the matches we need to find.
-	// Each time we get a match, we remove it from the set. If the set is empty, we can break the loop.
+	// If we have jobIds, we create a set to keep track of the matches we need to find. Each time we get
+	// a match, we remove it from the set. If the set is empty, we can break the loop to save time and unnecessary
+	// pagination.
 	pending := handy.NewSet[string](jobIds)
 
 	// To keep track of pages
@@ -48,7 +49,7 @@ func (c *Connector) ListIngestJobsInfo(ctx context.Context, jobIds ...string) ([
 			)
 		}
 
-		result, err := common.UnmarshalJSON[ListIngestJobsResult](res)
+		response, err := common.UnmarshalJSON[ListIngestJobsResult](res)
 		if err != nil {
 			return nil, fmt.Errorf(
 				"failed to unmarshal list ingest jobs info: %w",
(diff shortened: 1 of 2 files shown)


providers/salesforce/bulk-info.go
```diff
@@ -127,3 +207,14 @@ func (c *Connector) GetSuccessfulJobResults(ctx context.Context, jobId string) (
...
+// contains is a quick helper function to check if a slice contains a specific element.
+func contains[T comparable](s []T, e T) bool {
+	for _, a := range s {
+		if a == e {
+			return true
+		}
+	}
+
+	return false
+}
```
eberle1080, line 220: not necessary, I'd use `slices.Contains` which does the same thing

eberle1080 (approved):

providers/salesforce/bulk-info.go
```diff
@@ -7,8 +7,88 @@ import (
...
+			break
+		}
+
+		// If we aren't done yet, check if we have all the jobs we need
+		if len(jobIds) > 0 && pending.IsEmpty() {
+			break
+		}
+
+		// Else, get the next page
+		domain, err := c.getDomainURL()
```
Cobalt0s, line 80: This can be outside of the `for` loop.

providers/salesforce/bulk-info.go
```diff
@@ -7,8 +7,88 @@ import (
...
+		if err != nil {
+			return nil, err
+		}
+
+		// getDomainURL escapes some of the characters in the nextRecordsURL, so we use the URL directly
+		location = domain.String() + response.NextRecordsURL
```
Cobalt0s, line 86: I am not sure what is the content of `response.NextRecordsURL`.
There is `domain.AddPath(...strings...)` which uses `url.URL.JoinPaith` underneath.
Maybe it can be useful here. If it is then we need would need to keep `getDomainURL` inside loop as we would create a new URL instance and modify it.

RajatPawar: I tried to use that initially, but it kept escaping ? to a %3F

providers/salesforce/bulk-info.go
RajatPawar: Oops, thanks

providers/salesforce/bulk-info.go
RajatPawar: Merging this now, if there's any helper that accomplishes the same, we can change it later

Cobalt0s: Yeah, I also noticed that `url.URL.JoinPaith` doesn't have that many usecases. For resolving this super strict escaping I added `AddEncodingExceptions(...)` but I would rather stick to string concat. I wonder why standard `url.Url` doesnt deal with it.

PR #1037, opened 2024-09-12
feat: Universal Oauth2 error handling

Cobalt0s:
# Background

There is `ErrorPostProcessor` that is used by JSON and XML HTTP Clients. Error handling method is called by those clients before return statement. Some connectors could choose to provide final callback to handle error in some way.

# Changes

Salesforce is the only connector that has custom handling of error coming from Oauth2 library when refresh token is not valid.
This logic is not specific to Salesforce and should be applied universally across all connectors. 
`ErrorPostProcessor` will now own this logic which will trickle down the effects to others.

### DIFF
--- a/common/error.go
+++ b/common/error.go
@@ -5,6 +5,9 @@ import (
 	"fmt"
 	"log/slog"
 	"net/http"
+	"net/url"
+
+	"golang.org/x/oauth2"
 )
 
 // InterpretError interprets the given HTTP response (in a fairly straightforward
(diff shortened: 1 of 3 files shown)


Cobalt0s:
* **#1037** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1037?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

common/error.go
```diff
@@ -73,3 +84,19 @@ func (p ErrorPostProcessor) handleError(err error) error {
 		return err
 	}
 }
+
+// Errors coming from Oauth2 library are mapped here.
```
RajatPawar, line 88: Seems like this is the default, can you please mention it here in that case?

Cobalt0s: I changed and added some more comments. Let me know if it is better.

RajatPawar (approved):
I see that we have a `ErrorPostProcessor` on JSONHTTPClient, and a `ResponseHandler` on HTTPClient. What could be a good idea is to just unify both of those into one post processor that we can use for response/error handling. Approving this for now, feel free to make that change later on or in this PR.

PR #1038, opened 2024-09-12
chore: Linter includes formatting check

Cobalt0s:
# Background

I always run `make lint` but noticed that `make fix` may modify files that were not flagged by `make lint`. I tried clearing the cache using `golangci-lint cache clean`, but the issue persisted.

# Changes

After researching more about `gci`, which is used for formatting, I found the command `gci list .` which performs a "dry run" by returning a list of files that need to be modified.

Now, `make lint` will also print the files that should be changed by running `make fix`. If no files require changes, the exit code will be 0. This preserves the existing behavior while adding an extra check to notify when `make fix` needs running.

# Review

There are 2 commits. One is the result of running `make fix`, the other adds additional check under Makefile and Github action. I used this PR to test github action, such that removing first commit would result into `linter` job failure.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -15,3 +15,14 @@ jobs:
         continue-on-error: false
         with:
           version: v1.54
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: ">=1.21.0"
+          cache: false
+      - name: Install gci
+        run: "go install github.com/daixiang0/gci@latest"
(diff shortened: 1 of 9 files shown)


Cobalt0s:
* **#1038** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1038?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

eberle1080 (approved):

PR #1046, opened 2024-09-23
feat: Add ListMetadata + Read + Write functionality to Attio connector

sanjay-a01:
Response:
image


providers/attio/metadata.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	// Parse the JSON bytes into a map[string]interface{}
+	var responseMap map[string]interface{}
+	if err := json.Unmarshal(bodyBytes, &responseMap); err != nil {
+		return nil, err
+	}
+
+	fieldsMap := map[string]string{}
+	for k := range responseMap {
```
Cobalt0s, line 111: Since we are only collecting "keys", `ajson` library has method that will avoid additional marshall, unmarshall. It would look like this:
```
responseMap, err := body.GetObject()
```
https://github.com/spyzhov/ajson/blob/0d5fdeed023cf269f55ec781a2713153e3f3ca33/node.go#L494


Alternatively, if you want to deal with byte data directly avoiding ajson library, you could use the same UnmarshalJSON you use above but for map type:
```
responseMap, err := common.UnmarshalJSON[map[string]any](resp)
```

providers/attio/metadata.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/spyzhov/ajson"
+)
+
+// This struct is used for when the response data having slice of data.
+type responseObject struct {
+	Result []map[string]any `json:"data"`
+	// Other fields
+}
+
+var errCannotLoadMetadata = errors.New("cannot load metadata")
```
Cobalt0s, line 19: There is a recent `common` error that could fit for use case: `ErrMissingExpectedValues`.
Otherwise, export this error, and maybe even not within connector but from `common` package.

test/attio/metadata/metadata.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+func main() {
+	ctx := context.Background()
+
+	conn := attio.GetAttioConnector(ctx)
+
+	m, err := conn.ListObjectMetadata(ctx, []string{"objects", "lists", "self", "workspace_members", "webhooks", "tasks"})
+	if err != nil {
+		log.Fatal(err)
```
Cobalt0s, line 18: I was told to use `slog` lib and avoid `log`.
There is `utils.Fail()` that does that underneath.

Cobalt0s (commented):
Hi :wave:, some minor comments

providers/attio/metadata.go
Cobalt0s: You are serving Metadata from API and so far we were adding static `schema.json` for those providers that do not have such APIs. Therefore, no action is needed on this one.

.

**PS**: you can check out pipeliner to see how `schema.json` is created and served.
There is OpenAPI yaml or json commited under:
https://github.com/amp-labs/connectors/tree/main/providers/pipeliner/openapi
The `FileManager` loads it (file.go) and is referenced by 2 places:
* Schemas are objects object representation of that file and is used by `ListObjectMetadata` https://github.com/amp-labs/connectors/blob/main/providers/pipeliner/metadata.go.
* And most importantly this is the script which will differ among providers: https://github.com/amp-labs/connectors/blob/main/scripts/openapi/pipeliner/metadata/main.go. It loads OpenAPI file, processes it in custom way and outputs the `schema.json`.

In rare cases `schema.json` is created manually or a website is scrapped, which should be avoided for maintenance cost: https://github.com/amp-labs/connectors/tree/main/scripts/scrapper/salesloft/metadata

Here are some READMEs that I wrote:
* https://github.com/amp-labs/connectors/tree/main/scripts/openapi
* https://github.com/amp-labs/connectors/tree/main/tools/fileconv/api3

sanjay-a01:
updated unmarshal logic, used slog for error handling and common package vars

Cobalt0s (approved):
Everything is great! Some minor follow up comments.

Additionally, could you add mock tests in this or other PR.
For that, please record sample outputs of `c.Client.Get(ctx, url.String())` from Attio including successful response and some kind of failure. Mock tests would provide freedom to make changes to any code knowing that it will have our back and ensure that it functions as you intended.

sanjay-a01:
- Removed code related to the self object ( Not supported as it is a single object and related to Auth info)
- Added mock test for a list of getting metadata.
- Implemented read functionality in the Attio connector

providers/attio/metadata_test.go
```diff
@@ -0,0 +1,164 @@
...
+	tasksresponse := testutils.DataFromFile(t, "tasks.json")
+
+	tests := []testroutines.Metadata{
+		{
+			Name:         "Object must be included",
+			Server:       mockserver.Dummy(),
+			ExpectedErrs: []error{common.ErrMissingObjects},
+		},
+		{
+			Name:  "Successfully describe multiple object with metadata",
```
Cobalt0s, line 34: This is a great test!

providers/attio/objectNames.go
```diff
@@ -0,0 +1,23 @@
...
+	objectNameObjects          = "objects"
+	objectNameLists            = "lists"
+	objectNameWorkspacemembers = "workspace_members"
+	objectNameWebhooks         = "webhooks"
+	objectNameTasks            = "tasks"
+	objectNameNotes            = "notes"
+)
+
+var supportedObjectsByRead = handy.NewSet( //nolint:gochecknoglobals
+	// Object Name	----------	API endpoint path
```
Cobalt0s, line 16: comment is redundant

providers/attio/parse.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+var dummyNextPageFunc = func(*ajson.Node) (string, error) {
+	return "", nil
+}
+
+// recordsWrapperFunc returns the records using the objectName dynamically.
+func recordsWrapperFunc(obj string) common.RecordsFunc {
+	return func(node *ajson.Node) ([]map[string]any, error) {
+		result, err := jsonquery.New(node).Array(obj, true)
```
Cobalt0s, line 17: This method is similar to `common.GetRecordsUnderJSONPath()`. The only difference is that your version is ok with array being null as indicated by the `true` flag.

Maybe there should be 2 methods under common:
* GetRecordsUnderJSONPath -> which would be your version
* MustRecordsUnderJSONPath -> which is current impl GetRecordsUnderJSONPath.

Frankly, I am not sure that distinction is important!
I think before when I was refactoring, if key was not found that indicated that our code is at fault having wrong expectations in payload, thats about it... I have a feeling that you could change `GetRecordsUnderJSONPath` to have a `true` flag and use that method.

providers/attio/read.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	rsp, err := c.Client.Get(ctx, url.String())
+	if err != nil {
+		return nil, err
+	}
+
+	return common.ParseResult(
+		rsp,
+		recordsWrapperFunc("data"),
```
Cobalt0s, line 37: As suggested by above, something along the lines of `common.GetRecordsUnderJSONPath("data")`

Cobalt0s (approved):
Looks awesome! Thanks for adding so many mock responses, this is good.

@RajatPawar please have a look

sanjay-a01:
Used common functionality for unmarshalling the read response data

providers/attio/write.go
```diff
@@ -0,0 +1,58 @@
...
+		return nil, err
+	}
+
+	resp, err := common.UnmarshalJSON[writeResponse](res)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(resp.Data) == 0 {
+		return nil, ErrEmptyResultResponse
```
Cobalt0s, line 48: I think we can return success but with data as nil.

@RajatPawar what do you think about this suggestion? Also, when I was going over deep connectors I started wondering can common.WriteResult just have Success field?

Cobalt0s (approved):
Write functionality looks good to me!

The PR title should be renamed.

Also it would be nice to have multiple dedicate PRs. Write and ListObjectMetadata are fully independent from Read.

providers/attio/connector.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/paramsbuilder"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+	"github.com/amp-labs/connectors/providers"
+)
+
+const (
+	apiVersion = "/v2"
```
RajatPawar, line 11: Can you truncate the `/` out and use a method like filepath.Join (or if we have some other util function) instead wherever we use it? This is easy to make changes and cause bugs with

providers/attio/connector.go
```diff
@@ -0,0 +1,63 @@
...
+		conn = nil
+	})
+
+	params, err := paramsbuilder.Apply(parameters{}, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	// Read provider info
+	providerInfo, err := providers.ReadInfo(providers.Attio)
```
RajatPawar, line 31: Can you use `Provider()` wherever we want Attio? Not a big problem right now, but it'll help in realizing common code paths and generalizing them

providers/attio/parse.go
```diff
@@ -0,0 +1,10 @@
+// nolint
+package attio
+
+import (
+	"github.com/spyzhov/ajson"
+)
+
+var dummyNextPageFunc = func(*ajson.Node) (string, error) {
```
RajatPawar, line 8: Can you move this over to `read`? Would be good to have lesser files whenever possible

providers/attio/types.go
```diff
@@ -0,0 +1,11 @@
+// nolint
+package attio
+
+import "errors"
+
+var ErrEmptyResultResponse = errors.New("writing reponded with an empty result")
+
+type writeResponse struct {
```
RajatPawar, line 8: I'd move all these to `write.go` for the same rationale as my other comment, especially since they also are all about writing

providers/attio/write.go
RajatPawar: Looking at other connectors, like @Cobalt0s said, this should probably be an empty but successful `WriteResult`

providers/attio/write.go
```diff
@@ -0,0 +1,58 @@
+// nolint
+package attio
+
+import (
+	"context"
+
+	"github.com/amp-labs/connectors/common"
+)
+
+// Write creates/updates records in marketo. Write currently supports operations to the leads API only.
```
RajatPawar, line 10:
```suggestion
// Write creates/updates records in attio. Write currently supports operations to the leads API only.
```

providers/attio/metadata.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	metadataResult := common.ListObjectMetadataResult{
+		Result: make(map[string]common.ObjectMetadata),
+		Errors: make(map[string]error),
+	}
+
+	for _, obj := range objectNames {
+		// Constructing the request url.
+		url, err := c.getApiURL(obj)
```
RajatPawar, line 31: If there's a way to limit the response to just 1 object, that'd be great. We can avoid waiting on a larger response size

providers/attio/README.md
```diff
@@ -0,0 +1,16 @@
+# Atio connector
+
+
+## Supported Objects 
+Below is an exhaustive list of the supported Objects in the Attio deep connector with their endpoint resources(ObjectName).
+
+Attio API version : v2
+
+| Object | Resource | Method
```
RajatPawar, line 9: Write seems to be missing as of now?

providers/attio/README.md
```diff
@@ -0,0 +1,16 @@
+# Atio connector
+
+
+## Supported Objects 
+Below is an exhaustive list of the supported Objects in the Attio deep connector with their endpoint resources(ObjectName).
```
RajatPawar, line 5:
```suggestion
Below is an exhaustive list of objects & methods supported on the objects -
```

RajatPawar (approved):
Approving this, please - 
1. Reply/resolve all the comments before merging it.
2. Attach screenshots of all the tests you've added under a 'Testing' section in the PR description
3. Wherever the test does not actually make calls to the API, please attach a screenshot of a live read/write/metadata call.

Other than that, the PR looks great! Congrats on getting your first connector in 🎉

sanjay-a01:
### Getting a List of Metadata for the supporting objects
image

### Test Result for list of Metadata
image

### Read Functionality Responses
image
image
image
image

### Write Functionality Responses
image
image
image
image

### About creating Webhooks
"Target Url" is one of the body parameters used to create the webhook.
To make the target URL, I created a gin server and implemented the POST method to handle the URL.  Then using ngrok to expose my local server and get the public URL.
With the use of the URL, we can create the webhook. While the Event is triggered, the webhook URL automatically triggered by Attio server.
Also, we can monitor the changes in the ngrok dashboard page.

### Adding Query parameter in read functionality
Some supporting objects have pagination in the Attio connector but, we give the pagination value through a query parameter. So I added the url.WithQueryParams() function manually in the read functionality. If the object has a process to handle the query parameter then the query value will be added to the URL or otherwise, it will neglected.

providers/attio/metadata.go
```diff
@@ -0,0 +1,77 @@
+// nolint
```
Jkarage, line 1: Hey @sanjay-a01 ,Is there a reason for disabling linter in these files: metadata, ObjectNames, read, write?

providers/attio/metadata.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"context"
+
+	"github.com/amp-labs/connectors/common"
+)
+
+// This struct is used for when the response data having slice of data.
+type responseObject struct {
+	Result []map[string]any `json:"data"`
```
Jkarage, line 12: I would have named this Data instead of Result.

providers/attio/metadata.go
sanjay-a01: Yeah, I faced the lint issue while raising the PR so only I have disabled the linter.

providers/attio/metadata.go
sanjay-a01: If we specify the Result instead of Data, when we are unmarshalling the response for getting metadata it can't unmarshal because the response data contains "data" as the key.
image

providers/attio/write.go
```diff
@@ -0,0 +1,63 @@
...
+		return nil, err
+	}
+
+	var write common.WriteMethod
+	if len(config.RecordId) == 0 {
+		// writing to the entity without id means
+		// that we are extending 'List' resource and creating a new record
+		write = c.Client.Post
+	} else {
+		// only put is supported for updating 'Single' resource
```
Jkarage, line 39:
```suggestion
		// uses patch in updating 'single' resource.
```

providers/attio/write.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+var ErrEmptyResultResponse = errors.New("writing reponded with an empty result")
+
+type writeResponse struct {
+	Success bool           `json:"success"`
+	Data    map[string]any `json:"data"`
+}
+
+// Write creates/updates records in attio. Write currently supports operations to the leads API only.
```
Jkarage, line 18:
```suggestion
// Write creates/updates records in attio.
```

providers/attio/metadata.go
Cobalt0s: Yeah, I dont think it should be disabled entirely. Some things are to be addressed and some should be ignored line by line. Otherwise, if more inconsistencies are added the linter would not report them, which is not good.

providers/attio/connector.go
Cobalt0s: `urlbuilder` deals with this issues. It is prefered to not have slashes for sure for readability.
```
// Constructor handles path
urlbuilder.New(c.BaseURL, uriPart1, uriPart2, uriPart3, uriPart4)
```

providers/attio/parse.go
Cobalt0s: This function should belong in common. I can see other connectors using it. As a default behaviour -- no pagination.

providers/attio/metadata.go
Cobalt0s: I think I saw a code somewhere that you wrote that does gorotuines, generalizing it and amplying here would be nice.
Requesting in multithreaded fashion.

providers/attio/README.md
Cobalt0s: I think objectNames.go is an exact representation of this README. I would remove README, maybe would add comments to that file instead. There is always an incentive to keep go code up to date unlike `.md`, `.txt` type of files.

providers/attio/metadata.go
Cobalt0s: I think the suggestion is 
```
Data []map[string]any `json:"data"`
```
if understood correctly.

providers/attio/read.go
```diff
@@ -0,0 +1,44 @@
...
+	if !supportedObjectsByRead.Has(config.ObjectName) {
+		return nil, common.ErrOperationNotSupportedForObject
+	}
+
+	url, err := c.getApiURL(config.ObjectName)
+	if err != nil {
+		return nil, err
+	}
+
+	url.WithQueryParam("limit", "1")
```
Cobalt0s, line 29: This doesn't seem right. This is correct during development process. We don't want to query only 1 item at a time!
Also if some objects have offset pagination support that should be implemented instead of `dummyNextPageFunc`.

Cobalt0s (changes requested):
There is an action needed regarding `read.go`. You need to handle pagination or at least remove `per page limit of 1`.

sanjay-a01:
I have changed the pagination logic for the read functionality in the Attio connector

image

In that while using **DumpJSON**  function for printing the response, I faced issue with the next page response like 

image

and also I tried with **Json.MarshalIndent** function again getting the same result.
so I changed the marshalIndent logic like below

image
image

and got the nextpage URL like

image

It is just printing the response so I removed my marshalIndent logic and used **json.MarshalIndent** function for printing the response.

some object in the Attio connector don't require pagination, so I make the restriction using supportLimitAndOffset variable and check in the read functionality like

image
image

If the URL contains query parameters like limit and offset and has a response value, then only next page data will generate in the response otherwise it will return empty value in the next page

image

providers/attio/parse.go
```diff
@@ -0,0 +1,43 @@
...
+		if err != nil {
+			return "", err
+		}
+
+		previousStart := 0
+
+		// Extract the data key value from the response.
+		value := node.MustObject()["data"].MustArray()
+
+		if (url.Query().Has("limit") || url.Query().Has("offset")) && len(value) != 0 {
```
Cobalt0s, line 23: I think it is worth changing urlbuilder.URL to address this issue instead of converting to url.URL.
urlbuilder holds query parameters already:
```
type URL struct {
	delegate           *url.URL
	queryParams        url.Values
	encodingExceptions map[string]string
}
```
Just add `HasQueryParam(name string)` method and delegate the call `u.queryParams.Has(name)`.

providers/attio/parse.go
```diff
@@ -0,0 +1,43 @@
...
+				}
+			}
+
+			nextStart := previousStart + DefaultPageSize
+
+			reqLink.WithQueryParam("limit", strconv.Itoa(DefaultPageSize))
+			reqLink.WithQueryParam("offset", strconv.Itoa(nextStart))
+
+			return reqLink.String(), nil
+		} else {
```
Cobalt0s, line 39:
```suggestion
		}
		
		return "", nil
```
Else keyword is not needed due to `return` under if statement.

Cobalt0s (changes requested):
Everything is great! I see lots of effort and well done!

One important thing is check the comments for `makeNextRecordsURL` method.
You should update `urlbuilder` package. And make use of `jsonquery` to avoid `panic()` calls.

providers/attio/params.go
```diff
@@ -0,0 +1,42 @@
...
+
+type Option = func(params *parameters)
+
+type parameters struct {
+	paramsbuilder.Client
+}
+
+const (
+	// DefaultPageSize is number of elements per page.
+	DefaultPageSize = 10
```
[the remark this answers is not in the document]
sanjay-a01: yes, I have added the default size which is mentioned in the attio documentation

providers/attio/parse.go
sanjay-a01: I removed dummyNextPageFunc function because some object in Attio connector has pagination. Instead of this function, I created new function makeNextRecordsURL function in attio/parse.go file.

Cobalt0s (approved):
Awesome!
LGTM

PR #1047, opened 2024-09-24
handy Unique list

Cobalt0s:

PR #1048, opened 2024-09-24
feat: Define ModuleID

Cobalt0s:
# Background

Relying on the Module URL path as identifiter for that module seems limiting.

# Changes

* Module.Name is renamed to Module.Path
* Module has new field: ID which is of `APIModuleID` type.

# Usage

The developer of a deep connector defines the 'slugs' that the connector supports. These IDs can then be associated with a URL path. However, for some connectors, the endpoints of a module may not share identical URL prefixes, requiring custom mapping and code to resolve the URLs. This is why defining a Module ID adds more abstraction and flexibility for cases like this.

### DIFF
--- a/common/module.go
+++ b/common/module.go
@@ -1,0 +1,33 @@
+package common
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/amp-labs/connectors/common/handy"
+)
+
+// ErrInvalidModuleDeclaration occurs when the identifier used for map indexing in Modules
+// does not match the module's ID.
+var ErrInvalidModuleDeclaration = errors.New("supported modules are not correctly defined")
+
+type ModuleID string
+
+type Modules = handy.Map[ModuleID, Module]
+
+// Module represents a set of endpoints and functionality available by provider.
+// Single provider may support multiple modules, requiring the user to choose a module before making requests.
+// Modules may differ by version or by theme, covering different systems or functionalities.
+type Module struct {
+	ID      ModuleID
+	Label   string // e.g. "crm"
+	Version string // e.g. "v3"
+}
+
+func (a Module) Path() string {
+	if len(a.Label) == 0 {
+		return a.Version
+	}
+
+	return fmt.Sprintf("%s/%s", a.Label, a.Version)
+}
--- a/common/paramsbuilder/module.go
+++ b/common/paramsbuilder/module.go
@@ -1,71 +1,44 @@
 package paramsbuilder
 
 import (
-	"errors"
-	"fmt"
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/handy"
 )
-
-var ErrNoSupportedModule = errors.New("no supported module was chosen")
 
 // Module represents a sub-product of a provider.
 // This is relevant where there are several APIs for different sub-products, and the APIs
 // are versioned differently or have different ways of constructing URLs and requests for reading/writing.
 type Module struct {
-	// Name represents the name of the sub-product. e.g. crm, marketing, etc.
-	Name string
+	Selection common.Module
 	// Connector implementation must provide list of modules that are currently supported.
-	// This defines a list of module a user could switch to. Any out of scope module will be ignored.
-	supported []APIModule
-	// If user supplied unsupported module it will use this fallback.
-	fallback *APIModule
+	// This defines a list of modules a user could switch to.
+	// Validation will check module identifiers are consistent.
+	supported common.Modules
 }
 
 func (p *Module) ValidateParams() error {
-	// making sure the provided module is supported.
-	// If the provided module is not supported, use fallback.
-	if !p.isSupported() {
-		if p.fallback == nil {
-			// not supported and user didn't provide a fallback
-			return ErrNoSupportedModule
-		}
-
-		// replace with fallback module
-		p.Name = p.fallback.String()
-
-		// even fallback is not supported
-		if !p.isSupported() {
-			return ErrNoSupportedModule
+	// Module ID must match the ID which is used to index modules.
+	for id, module := range p.supported {
+		if module.ID != id {
+			return common.ErrInvalidModuleDeclaration
 		}
 	}
 
 	return nil
 }
 
-func (p *Module) WithModule(module APIModule, supported []APIModule, defaultModule *APIModule) {
-	p.Name = module.String()
+// WithModule allows API module selection.
+// Connector implementation must provide list of modules that are currently supported.
+// This defines a list of module a user could switch to. Any out of scope module will be ignored.
+// If user supplied unsupported module, then it will use this fallback.
+func (p *Module) WithModule(moduleID common.ModuleID, supported common.Modules, fallbackID common.ModuleID) {
+	modules := handy.NewDefaultMap(supported,
+		func(unknownID common.ModuleID) common.Module {
+			// Unknown module ids will receive a fallback module.
+			return supported[fallbackID]
+		},
+	)
+
+	p.Selection = modules.Get(moduleID)
 	p.supported = supported
-	p.fallback = defaultModule
-}
-
-func (p *Module) isSupported() bool {
-	for _, mod := range p.supported {
-		if p.Name == mod.String() {
-			return true
-		}
-	}
-
-	return false
-}
-
-type APIModule struct {
-	Label   string // e.g. "crm"
-	Version string // e.g. "v3"
-}
-
-func (a APIModule) String() string {
-	if len(a.Label) == 0 {
-		return a.Version
-	}
-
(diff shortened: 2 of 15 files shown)


Cobalt0s:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1048?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1050** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1050?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1049** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1049?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1048** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1048?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1047** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1047?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

common/paramsbuilder/module.go
```diff
@@ -11,52 +13,57 @@ var ErrNoSupportedModule = errors.New("no supported module was chosen")
...
-		}
-	}
-
-	return false
+func (p *Module) save(id APIModuleID, m APIModule) {
+	p.ID = id
+	p.Path = m.String()
 }
 
+type APIModuleID string
```
RajatPawar, line 63: IMO `API` is redundant in the current context, just `ModuleID` should be great

common/paramsbuilder/module.go
```diff
@@ -11,52 +13,57 @@ var ErrNoSupportedModule = errors.New("no supported module was chosen")
...
+
+type APIModuleMapping handy.Map[APIModuleID, APIModule]
```
RajatPawar, line 65: The mapping is an implementation detail, but for actual users, it would just look like listing out all the possible modules which is why I feel like just `Modules` is a better type name

common/paramsbuilder/module.go
```diff
@@ -11,52 +13,57 @@ var ErrNoSupportedModule = errors.New("no supported module was chosen")
 // This is relevant where there are several APIs for different sub-products, and the APIs
 // are versioned differently or have different ways of constructing URLs and requests for reading/writing.
 type Module struct {
-	// Name represents the name of the sub-product. e.g. crm, marketing, etc.
-	Name string
+	// ID is name of the module.
```
RajatPawar, line 16: 1. There's different schools of thoughts, but `ID` conveys that this is an acronym which is not the case, so we tend to use `Id` instead

2. `Id is a unique identifier for the module. It makes it easier to group multiple endpoints under the same module, even if they do not share a common URL prefix.`

providers/atlassian/modules.go
```diff
@@ -4,21 +4,22 @@ import (
...
+	// ModuleJira is the module used for listing Jira issues.
+	ModuleJira paramsbuilder.APIModuleID = "jira"
+)
 
 // supportedModules represents currently working and supported modules within the Atlassian connector.
 // Any added module should be appended here.
-var supportedModules = []paramsbuilder.APIModule{ // nolint: gochecknoglobals
-	ModuleEmpty,
-	ModuleJira,
+var supportedModules = paramsbuilder.APIModuleMapping{ // nolint: gochecknoglobals
```
RajatPawar, line 16: It'd be quite cool to use a `DefaultMap` such that we specify the fallback module. That way, we don't have to specify the fallback as an argument

providers/marketo/connnector.go
```diff
@@ -10,9 +10,10 @@ import (
 )
 
 type Connector struct {
-	BaseURL string
-	Client  *common.JSONHTTPClient
-	Module  string
+	BaseURL    string
+	Client     *common.JSONHTTPClient
+	ModuleID   paramsbuilder.APIModuleID
```
RajatPawar, line 15: This doesn't feel easy to remember - can you find a way to just have a `Module` concept which gives us access to the `Id` & the `Path`?

RajatPawar (changes requested):
Added a couple of minor change requests

common/paramsbuilder/module.go
Cobalt0s: ID is abbreviation for identifier.
I think creating a difference between ID and Id creates confusion. Lauren pointed out that it should be `Id` for JSON fields. In code I commit this sin. ID visually draws attention and looks like a concept. In the Java world it was always Id, in go world I see ID. I think my mental mapping got messed up since then.

It is a go convention to do it that way. Linters complain about such things (of course not in comments ahaha).

https://google.github.io/styleguide/go/decisions#initialisms, id or ID.

providers/atlassian/modules.go
Cobalt0s: Nice, I will give it a thought

Cobalt0s: So far right now I see that fallback will be used only when `WithModule` is called. But if it is not, then effectively the fallback doesn't kick in. I think I need to think better about it, and `DefaultMap` could come in handy.

common/paramsbuilder/module.go
Cobalt0s: Also in the IDE it highlights field names in comments if the spelling matches. Thats why I wrote ID as caps.

image

common/module.go
```diff
@@ -0,0 +1,33 @@
...
+var ErrInvalidModuleDeclaration = errors.New("supported modules are not correctly defined")
+
+type ModuleID string
+
+type Modules = handy.Map[ModuleID, Module]
+
+// Module represents a set of endpoints and functionality available by provider.
+// Single provider may support multiple modules, requiring the user to choose a module before making requests.
+// Modules may differ by version or by theme, covering different systems or functionalities.
+type Module struct {
```
Cobalt0s, line 21: Module was moved out of `paramsbuilder` to `common` package.
This eliminates name collisions, and this concept doesn't really belong to building parameters. 
The parameter will own it. Module is then propagated and saved under connector struct and is later used for it's `ID` and `Path()`. Less separate data to store as you requested.

@RajatPawar

common/paramsbuilder/module.go
```diff
@@ -1,71 +1,44 @@
...
 }
 
-func (p *Module) WithModule(module APIModule, supported []APIModule, defaultModule *APIModule) {
-	p.Name = module.String()
+// WithModule allows API module selection.
+// Connector implementation must provide list of modules that are currently supported.
+// This defines a list of module a user could switch to. Any out of scope module will be ignored.
+// If user supplied unsupported module, then it will use this fallback.
+func (p *Module) WithModule(moduleID common.ModuleID, supported common.Modules, fallbackID common.ModuleID) {
+	modules := handy.NewDefaultMap(supported,
```
Cobalt0s, line 35: Delegated logic to `DefaultMap`.

common/paramsbuilder/params.go
```diff
@@ -13,7 +13,18 @@ type ParamAssurance interface {
...
 // This is a generalized constructor of parameters used to initialize any connector.
 // To qualify as a parameter one must have data validation laid out.
-func Apply[P ParamAssurance](params P, opts []func(params *P)) (*P, error) {
+//
+// It is possible to give default options which will be applied first.
+// Any other options supplied by the user may override defaults,
+// as they are applied last and therefore take higher precedence.
+func Apply[P ParamAssurance](params P,
+	opts []func(params *P),
+	defaultOpts ...func(params *P),
```
Cobalt0s, line 22: Given connectors can be configured via options the only way to specify defaults a connector should have is to directly pass them. As option function sets values to parameter the last caller wins.

providers/atlassian/modules.go
```diff
@@ -1,24 +1,27 @@
...
-	ModuleEmpty,
-	ModuleJira,
+var supportedModules = common.Modules{ // nolint: gochecknoglobals
+	ModuleEmpty: {
+		ID:      ModuleEmpty,
+		Label:   "",
+		Version: "",
+	},
+	ModuleJira: {
+		ID:      ModuleJira,
```
Cobalt0s, line 23: This ID must be declared on a module and must match the index. `ValidateParams` enforces this, so nobody would forget this.

Cobalt0s (commented):
@RajatPawar I made some changes effecting the whole PR.

RajatPawar:
@laurenzlong I added you on this because of our upcoming module work, and wanted you to take a look at it to ensure it's in line with the general direction

RajatPawar (approved):

PR #1041, opened 2024-09-13
feat: Zendesk modules use URLs from schema.json

Cobalt0s:
# Background

As of now the only supported resources have short URLs. There is no systematic approach for other cases.

# Changes

* The static schema file holds URL paths for every object, allowing nested REST API resources to be requested via Read/Write/Delete methods.
Some connectors have few endpoints with hardcoded mapping between ObjectName and URL. However, in this case `schema.json` was generated from the OpenAPI file, so it naturally holds the mapping. As a result, creating URLs is straightforward.
```go
path, err := metadata.Schemas.LookupURLPath(c.ModuleID, objectName)
url, err := urlbuilder.New(c.BaseURL, path)
```
* I noticed that pagination response format differs between ticketing and help center modules. I updated the parsing so that both pagination formats are handled in a fallback sequence, starting with the more likely format.
* ZendeskSupport BaseURL is modified :exclamation: :exclamation: :exclamation: 

# Test Results
Running Zendesk test scripts for backwards compatability.
image
image
image

# Review
[First commit](https://github.com/amp-labs/connectors/pull/1041/commits/c44521ce0216efbeb7a2e8f4b6dca45609da146c) has main code changes. [Second](https://github.com/amp-labs/connectors/pull/1041/commits/9bc54f7fff8075934648693e3283ec080d11d69c) is mock tests.

PR #1050, opened 2024-09-24
feat: Zendesk add Help Center module

Cobalt0s:
# Background

Zendesk Support consists of 3 modules:
* [Ticketing API](https://developer.zendesk.com/api-reference/ticketing/introduction/)
* [Help Center API](https://developer.zendesk.com/api-reference/help_center/help-center-api/introduction/)
* [Voice API](https://developer.zendesk.com/api-reference/voice/talk-api/introduction/)

# Changes

This PR introduces Help Center module by adding OpenAPI yaml. The script is modified in such way that `schema.json` instead of having default empty name module, it is renamed to ticketing while a newly added module is called help center.

# Server changes

No changes are `required`.
To be more explicit when constructin connector we could add `WithModule` option.

# Notes

While it is possible to merge the two API modules since each has universally unique objects without collission, I decided to keep them as separate modules to better reflect the structure of the official documentation.

image

### DIFF
--- a/common/module.go
+++ b/common/module.go
@@ -36,3 +36,33 @@ func (a Module) Path() string {
 
 	return fmt.Sprintf("%s/%s", a.Label, a.Version)
 }
+
+// ModuleObjectNameToFieldName is a grouping of ObjectName to response field name mappings defined for each Module.
+//
+// Explanation: modules have objects, each object is located under certain field name in the response body.
+// This mapping stores the general relationship between the said ObjectName and FieldName.
+// Those objects that do not follow the pattern described in the fallback method are hard code as exceptions.
+//
+// Ex:
+//
+//	Given:	Connector has 2 modules -- Commerce, Messaging.
+//			Commerce module has objects stored under "data" field name, except "carts".
+//			Messaging module has objects stored under the same name as object, except "chats".
+//	Then:	It will be represented as follows:
+//
+//		ModuleObjectNameToFieldName{
+//			ModuleCommerce: handy.NewDefaultMap(map[string]string{
+//				"carts": "carts",
+//			},
+//				func(objectName string) string {
+//					return "data" // always under "data" field {"data": [{},{},...]}
+//				},
+//			),
+//			ModuleHelpCenter: handy.NewDefaultMap(map[string]string{
+//				"chats":        "active_chats",
+//			}, func(objectName string) string {
+//				fieldName := objectName // Object "messages" is stored under {"messages": [{},{},...]}
+//				return fieldName
+//			}),
+//		}
+type ModuleObjectNameToFieldName map[ModuleID]handy.DefaultMap[string, string]
--- a/providers/zendesksupport/connector.go
+++ b/providers/zendesksupport/connector.go
@@ -22,7 +22,9 @@ func NewConnector(opts ...Option) (conn *Connector, outErr error) {
 		conn = nil
 	})
 
-	params, err := paramsbuilder.Apply(parameters{}, opts)
+	params, err := paramsbuilder.Apply(parameters{}, opts,
+		WithModule(ModuleTicketing), // The module is resolved on behalf of the user if the option is missing.
+	)
 	if err != nil {
 		return nil, err
 	}
@@ -32,6 +34,7 @@ func NewConnector(opts ...Option) (conn *Connector, outErr error) {
 		Client: &common.JSONHTTPClient{
 			HTTPClient: httpClient,
 		},
+		Module: params.Module.Selection,
 	}
 
 	providerInfo, err := providers.ReadInfo(conn.Provider(), &params.Workspace)
--- a/providers/zendesksupport/metadata/schemas.json
+++ b/providers/zendesksupport/metadata/schemas.json
@@ -1,7 +1,140 @@
 {
   "modules": {
-    "root": {
-      "id": "root",
+    "help-center": {
+      "id": "help-center",
+      "path": "/api/v2",
+      "objects": {
+        "article_labels": {
+          "displayName": "Article Labels",
+          "path": "/help_center/articles/labels",
+          "fields": {
+            "created_at": "created_at",
+            "id": "id",
+            "name": "name",
+            "updated_at": "updated_at",
+            "url": "url"
+          }
+        },
+        "articles": {
+          "displayName": "Articles",
+          "path": "/help_center/articles/search",
+          "fields": {
+            "author_id": "author_id",
+            "body": "body",
+            "comments_disabled": "comments_disabled",
+            "content_tag_ids": "content_tag_ids",
+            "created_at": "created_at",
+            "draft": "draft",
+            "edited_at": "edited_at",
+            "html_url": "html_url",
+            "id": "id",
+            "label_names": "label_names",
+            "locale": "locale",
+            "outdated": "outdated",
+            "outdated_locales": "outdated_locales",
+            "permission_group_id": "permission_group_id",
+            "position": "position",
+            "promoted": "promoted",
+            "section_id": "section_id",
+            "source_locale": "source_locale",
+            "title": "title",
+            "updated_at": "updated_at",
+            "url": "url",
+            "user_segment_id": "user_segment_id",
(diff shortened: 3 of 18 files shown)


Cobalt0s:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1050?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1041** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1041?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1050** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1050?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1049** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1049?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1048** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1048?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1047** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1047?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

RajatPawar (commented):
Removed my request for review until we can get #1048 finalized since this depends on it

providers/zendesksupport/connector.go
```diff
@@ -24,7 +24,9 @@ func NewConnector(opts ...Option) (conn *Connector, outErr error) {
 		conn = nil
 	})
 
-	params, err := paramsbuilder.Apply(parameters{}, opts)
+	params, err := paramsbuilder.Apply(parameters{}, opts,
+		WithModule(ModuleTicketing), // The module is resolved on behalf of the user if the option is missing.
```
RajatPawar, line 28: If `NewConnector` is called with another module, would `paramsbuilder.Apply` first apply that and then `WithModule(ModuleTicketing)` over it?

providers/zendesksupport/metadata/schemas.json
```diff
@@ -1,6 +1,137 @@
 {
   "modules": {
-    "": {
+    "help-center": {
+      "objects": {
+        "article_labels": {
+          "displayName": "Article Labels",
+          "urlPath": "/api/v2/help_center/articles/labels",
```
RajatPawar, line 7: Can we pull `/api/v2/` out of modules into the connector itself? If the version ever changes, we'd have to make the change in multiple places for it to take effect. If we have a structure from my comment in #1049 where we can specify multiple properties on the module (path, etc), then the version could be a part of the path. 

If we need a specific object to be a different version, we could add the ability to override the path at an object level (we don't need to add that now, but just to make sure this approach works)

providers/zendesksupport/read.go
```diff
@@ -26,7 +26,7 @@ func (c *Connector) Read(ctx context.Context, config common.ReadParams) (*common
 		return nil, err
 	}
 
-	responseFieldName := ObjectNameToResponseField.Get(config.ObjectName)
+	responseFieldName := ObjectNameToResponseField[c.Module.ID].Get(config.ObjectName)
```
RajatPawar, line 29: This could panic, can you add a check here?

providers/zendesksupport/objectNames.go
```diff
@@ -1,21 +1,34 @@
...
 import (
+	"github.com/amp-labs/connectors/common"
 	"github.com/amp-labs/connectors/common/handy"
 	"github.com/amp-labs/connectors/providers/zendesksupport/metadata"
 )
 
 // Supported object names can be found under schemas.json.
 var supportedObjectsByRead = metadata.Schemas.GetObjectNames() //nolint:gochecknoglobals
 
+// Grouping of ObjectName to response field name mappings by Module.
```
RajatPawar, line 12: Would be nice to comment/refer to some mock data that a viewer could look at to understand this better.

providers/zendesksupport/modules.go
```diff
@@ -0,0 +1,36 @@
...
+	},
+	ModuleTicketing: {
+		ID:      ModuleTicketing,
+		Label:   "",
+		Version: "",
+	},
+	ModuleHelpCenter: {
+		ID:      ModuleHelpCenter,
+		Label:   "",
+		Version: "",
```
RajatPawar, line 34: Could the `/v2/` version be here? I want to simplify version changes when we need to do them, so better to have lesser concepts to deal with.

RajatPawar (changes requested):
Couple of comments / clarifications

RajatPawar:
@Cobalt0s Please don't forget to update the server as soon as you merge this in!

providers/zendesksupport/connector.go
Cobalt0s: Yes. This means there is no need to change `server` repo. There are no breaking changes.
By default the module is `Ticketing`. But if `WithModule` is called by the user it will be applied later and override this value.

Ex, called test, where instantiated connector with `ModuleHelpCenter`: 
image

Note: options design pattern doesn't prohibit multiple calls for the same function, so anybody could do something like this:
```
connector, err := NewConnector(
	WithAuthenticatedClient(http.DefaultClient),
	WithWorkspace("test-workspace"),
	WithModule(ModuleTicketing),
	WithModule(ModuleHelpCenter),
	WithModule(ModuleTicketing),
)
```

`paramsbuilder.Apply` accepts "first order", "default" options which is applied `before` all, providing default working setup.

providers/zendesksupport/read.go
Cobalt0s: During development I thought so too and validated it. In golang it cannot panic.
The reason for this is every object I use is not a reference.

If `ObjectNameToResponseField` map hits a miss then the default empty value is returned, which will be empty `DefaultMap[string,string]` struct. Calling `Get` will look through empty map, then try to use `nil` fallback (there is a check) and would return the empty value of string which is `""`.

Worse case, it cannot break, panic.

common/module.go
```diff
@@ -36,3 +36,33 @@ func (a Module) Path() string {
...
+//				},
+//			),
+//			ModuleHelpCenter: handy.NewDefaultMap(map[string]string{
+//				"chats":        "active_chats",
+//			}, func(objectName string) string {
+//				fieldName := objectName // Object "messages" is stored under {"messages": [{},{},...]}
+//				return fieldName
+//			}),
+//		}
+type ModuleObjectNameToFieldName map[ModuleID]handy.DefaultMap[string, string]
```
Cobalt0s, line 68: @Rajat you have asked about the explanation in different comment thread.

First of all I moved this type into `common`. I see that it will be reused by those connectors with multiple modules and inconsitent `ObjectName` to `FieldName` relationship.

providers/zendesksupport/modules.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/amp-labs/connectors/providers/zendesksupport/metadata"
+)
+
+const (
+	// ModuleTicketing is used for proxying requests through.
+	// https://developer.zendesk.com/api-reference/ticketing/introduction/
+	ModuleTicketing common.ModuleID = "ticketing"
+	// ModuleHelpCenter is Zendesk Help Center.
+	// https://developer.zendesk.com/api-reference/help_center/help-center-api/introduction/
+	ModuleHelpCenter common.ModuleID = "help-center"
```
Cobalt0s, line 14: @RajatPawar 

Empty module is removed. It doesn't make sense to have such.
The old comment which was copied from old code mentions proxying to be the reason for such module, it is misleading.

For example, module selection has no impact on something like this:
```
// From: examples/authconnectors/salesforce/salesforce.go
conn := createAuthConnector(ctx)
response, err := conn.Client.Get(ctx, "/services/data/v61.0/limits/")
```
Read/Write make sense only with resepect to `ticketing` and `help-center`.

RajatPawar (approved):

PR #1053, opened 2024-09-30
feat: Add an option for handling 401 errors

eberle1080:
Some providers (cough, Salesforce) will invalidate OAuth access tokens under certain circumstances (e.g., the refresh token being used by another IP address). If this happens, it will return a 401. Right now, we treat that 401 as fatal, when really it needs to be seen as a chance to forcefully refresh the OAuth token.

This PR provides a new hook which the server code can use to intercept the 401, and call out to the token manager service with `force=true` to obtain a new token.

### DIFF
--- a/common/oauth.go
+++ b/common/oauth.go
@@ -35,6 +35,7 @@ type oauthClientParams struct {
 	config       *oauth2.Config
 	tokenSource  oauth2.TokenSource
 	tokenUpdated func(oldToken, newToken *oauth2.Token) error
+	unauthorized func(token *oauth2.Token, req *http.Request, rsp *http.Response) (*http.Response, error)
 	debug        func(req *http.Request, rsp *http.Response)
 }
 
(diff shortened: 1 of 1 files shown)


common/oauth.go
```diff
@@ -167,6 +181,15 @@ func (t *oauth2Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 		t.Debug(req2, cloneResponse(rsp))
 	}
 
+	// Certain providers return 401 when the token has been invalidated.
+	// This may indicate that the token needs to be forcefully refreshed.
+	// Since this is per-provider, the caller can provide a custom handler.
+	if rsp.StatusCode == http.StatusUnauthorized {
```
RajatPawar, line 187: I could be wrong, but I faintly remember SF sending an error back with an `invalid_grant` message but an empty response

RajatPawar (approved):

PR #1054, opened 2024-09-30
feat: add the TokenSourceContext interface

eberle1080:
Right now the token source has no way to get hints from the caller. Specifically, the kinds of hints I'm talking about are things like "I got a 401 and I need to forcefully refresh". This kind of info can be encoded in to a `context.Context` but as is, the `oauth2.TokenSource` interface doesn't accept a context.

This extends the interface so that the caller may implement this new signature, and get called with the context. This will let me do the 401 "force refresh" logic I need. If the caller doesn't implement this interface, it's fine, it will just fall back to the normal one.

### DIFF
--- a/common/oauth.go
+++ b/common/oauth.go
@@ -8,6 +8,21 @@ import (
 	"golang.org/x/oauth2"
 )
 
+// TokenSourceContext is an interface that extends the oauth2.TokenSource interface
+// with a context. This is useful for token sources that need to be aware of the
+// context in which they are being called. The use of this interface is optional,
+// but if the token source conforms to it, then the context version of the Token
+// method will be called instead of the normal one.
+type TokenSourceContext interface {
+	oauth2.TokenSource
+
+	// TokenContext returns a token or an error.
+	// Token must be safe for concurrent use by multiple goroutines.
+	// The returned Token must not be modified.
+	// Similar semantics to oauth2.TokenSource, but with a context.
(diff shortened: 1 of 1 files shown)


common/oauth.go
```diff
@@ -8,6 +8,21 @@ import (
...
+// but if the token source conforms to it, then the context version of the Token
+// method will be called instead of the normal one.
+type TokenSourceContext interface {
+	oauth2.TokenSource
+
+	// TokenContext returns a token or an error.
+	// Token must be safe for concurrent use by multiple goroutines.
+	// The returned Token must not be modified.
+	// Similar semantics to oauth2.TokenSource, but with a context.
+	TokenContext(ctx context.Context) (*oauth2.Token, error)
```
laurenzlong, line 23: maybe `TokenFromContext`? Since the return value is a token.

laurenzlong (changes requested):

common/oauth.go
eberle1080: I was trying to stay faithful to the typical go naming convention. The other interface just has a function called "Token", and usually in go if you add something taking a context, you just do "NameContext". e.g. `net.Dialer.Dial` vs `net.Dialer.DialContext`.

Also it's not extracting the token from the context, per se. It's using the context to understand its environment better, but the token will still come either from the DB or from the token manager service.

I could maybe do `TokenWithContext` or something like that?

laurenzlong (approved):

PR #1059, opened 2024-10-07
fix: file naming, creates AddSuffixIfNotExists func

Jkarage:
Context: 

- Renamed connector go files from `connnector` to `connector`.
- Moved the AddsJSONSuffix to common.
- Cleaned some code documentations.


common/json.go
```diff
@@ -187,3 +188,12 @@ func addAcceptJSONHeader(headers []Header) []Header {
 
 	return append(headers, Header{Key: "Accept", Value: "application/json"})
 }
+
+// AddJSONSuffix adds the suffix ".json" to an objectName.
+func AddJSONSuffix(objectName string) string {
```
RajatPawar, line 193: If we want to have this in `common`, let's repurpose this to be more generic - `AddSuffixIfNotExists(str string, suffix str)`

providers/marketo/README.md
```diff
@@ -1,11 +1,20 @@
-# Write connnector
+# Marketo Connector
+
+The connector currently supports Leads API only.
 
-Currently the Write connector supports Leads API only. 
+# Write connnector
 
 The objects that we can write to currently are:
```
RajatPawar, line 7: Minor, feel free to do it later, but a table would be better at representing this:

| object | support     |   
|--------|----------|
| lists  | read, write |

providers/marketo/metadata.go
```diff
@@ -11,11 +11,10 @@ import (
 type responseObject struct {
 	Result  []map[string]any `json:"result"`
 	Success bool             `json:"success"`
-	// Other fields
 }
 
-// ListObjectMetadata creates metadata of object via reading objects using Marketo API.
-// If it fails to fretrieve the metadata, It tried using static schema file in metadata dir.
+// ListObjectMetadata creates metadata of object via reading object using Marketo API.
```
RajatPawar, line 16:
```suggestion
// ListObjectMetadata returns metadata for an object by sampling an object from Marketo's API.
```

providers/marketo/metadata.go
```diff
@@ -11,11 +11,10 @@ import (
...
+// If it fails to retrieve the metadata, It retries using marketo's OpenAPI schema files.
```
RajatPawar, line 17:
```suggestion
// If that fails, it generates object metadata by parsing Marketo's OpenAPI files.
```

providers/marketo/write.go
```diff
@@ -29,8 +29,11 @@ func (c *Connector) Write(ctx context.Context, config common.WriteParams) (*comm
 		return nil, err
 	}
 
+	// This should(will) never happen.
+	// But it's checked so as to avoid panics in accesing Result Array(incase the above happens).
+	// The rest WriteResults fields are zero valued.
```
RajatPawar, line 34:
```suggestion
	// In case of an empty result, we return a zero-valued WriteResult.
```

PR #1060, opened 2024-10-07
feat(utils): Conditional mock server

Cobalt0s:
# Background

This is part of PR series dedicated to simplifying expressing the goal of mock server used by test case.
Often times when designing a test scenario you want to express if part of request matches certain criteria (condition) and when it does it should act one way otherwise fallback, following `if...else` paradigm.

# Features
* `mockserver.Reactive{...}.Server()` describes what mock server should do when condition evaluates to true
* `mockond.Condition` is an interface which allows to express nested rules very easily:
  * `Check func(w http.ResponseWriter, r *http.Request) bool` -- basic building block of conditions. Future PRs introduce such things as: if URL matches then, if REST operation is POST then, if has query paremeter then, etc.
  * `Or []Condition` and `And []Condition` allow nesting and expressing more sophisticated rules.

# Example
Let's take bulk write test for Salesforce as an example. The new version can be read as follows: "Create mock server that will react/respond with `200, responseCreateJob` only when URL path matches `**and**` body matches".
image
This is much more readable and understandable than older version:
image

### DIFF
--- a/test/utils/mockutils/mockcond/body.go
+++ b/test/utils/mockutils/mockcond/body.go
@@ -1,0 +1,46 @@
+package mockcond
+
+import (
+	"bytes"
+	"io"
+	"net/http"
+	"strings"
+)
+
+// Body returns a check expecting body to match template text.
+func Body(expectedBody string) Check {
+	return func(w http.ResponseWriter, r *http.Request) bool {
+		reader := r.Body
+
+		body, err := io.ReadAll(reader)
+		if err != nil {
+			return false
+		}
+
+		r.Body.Close()
+		r.Body = io.NopCloser(bytes.NewBuffer(body))
+
+		a := stringCleaner(string(body), []string{"\n", "\t"})
+		b := stringCleaner(expectedBody, []string{"\n", "\t"})
+		match := a == b
+
+		return match
+	}
+}
+
+func stringCleaner(text string, toRemove []string) string {
+	rules := make(map[string]string)
+	for _, remove := range toRemove {
+		rules[remove] = ""
+	}
+
+	return stringReplacer(text, rules)
+}
+
+func stringReplacer(text string, rules map[string]string) string {
+	for from, to := range rules {
+		text = strings.ReplaceAll(text, from, to)
+	}
+
+	return text
+}
--- a/test/utils/mockutils/mockcond/conditions.go
+++ b/test/utils/mockutils/mockcond/conditions.go
@@ -1,0 +1,61 @@
+package mockcond
+
+import "net/http"
+
+// Condition computes whether http.Request meets some rule.
+type Condition interface {
+	EvaluateCondition(w http.ResponseWriter, r *http.Request) bool
+}
+
+// Check is the most basic Condition. It is a function definition that allows custom implementation.
+// There are some out of the box functions in this package that have this signature.
+type Check func(w http.ResponseWriter, r *http.Request) bool
+
+// Or is a composite Condition which evaluates to true if at least one condition is met.
+// Empty list returns false.
+type Or []Condition
+
+// And is a composite Condition which evaluates to true if all conditions are met.
+// Empty list returns false.
+type And []Condition
+
+func (c Check) EvaluateCondition(w http.ResponseWriter, r *http.Request) bool {
+	return c(w, r)
+}
+
+func (o Or) EvaluateCondition(w http.ResponseWriter, r *http.Request) bool {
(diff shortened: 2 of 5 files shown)


Cobalt0s:
* **#1062** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1062?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1061** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1061?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1060** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1060?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

test/utils/mockutils/mockserver/reactive.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"net/http"
+	"net/http/httptest"
+
+	"github.com/amp-labs/connectors/test/utils/mockutils"
+	"github.com/amp-labs/connectors/test/utils/mockutils/mockcond"
+)
+
+// Reactive is a server recipe that describes how mock server should react when conditions are met.
+type Reactive struct {
```
RajatPawar, line 12: Well, every server is reactive, so can we call this `ConditionalServer` to be more explicit that it handles if/else blocks?

test/utils/mockutils/mockcond/conditions.go
```diff
@@ -0,0 +1,61 @@
+package mockcond
```
RajatPawar, line 1: For the sake of simplicity, could we just have mockserver have both the server and conditions?

test/utils/mockutils/mockserver/handlers.go
```diff
@@ -0,0 +1,26 @@
...
+	}
+}
+
+// Response is used to configure server response with HTTP status and body data.
+// Data is optional.
+func Response(status int, data ...[]byte) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(status)
+
+		if len(data) == 1 {
```
RajatPawar, line 18: What happens if len > 1?

test/utils/mockutils/mockcond/conditions.go
```diff
@@ -0,0 +1,61 @@
...
+
+import "net/http"
+
+// Condition computes whether http.Request meets some rule.
+type Condition interface {
+	RunCondition(w http.ResponseWriter, r *http.Request) bool
```
RajatPawar, line 7: `EvaluateCondition`

test/utils/mockutils/mockcond/body.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"bytes"
+	"io"
+	"net/http"
+	"strings"
+)
+
+// Body returns a check expecting body to match template text.
+func Body(expectedBody string) Check {
```
RajatPawar, line 11: Cool stuff! I wonder if we can rename it to be more like primitives that we already are familiar with - 
```
If: And{ 
  IfSuffix(...),
  IfBody(...),
},
// Evaluation success
Then: {
   SendResponse(...),
},
// Evaluation failure
Else: {
}
```
```

RajatPawar, line 11: Might be easier to deal with bytes than a string

RajatPawar (approved):
Approving, please resolve comments before merging

test/utils/mockutils/mockcond/conditions.go
Cobalt0s: It will conflate the meaning. `mockcond` should autocomplete all possible checks one can have on response body. While `mockserver` is about building, setting up the server. There are few options when doing autocomplete, easier to search for methods, no need to remember which goes where.

Relation:
mockserver uses mockcond

test/utils/mockutils/mockserver/handlers.go
Cobalt0s: I will add a panic. This is a "hack" to allow optional field. Not all mocks respond with data. Adding new method would not make it look nice, although proper style.

The reason for panic is to show that whoever created a test did a mistake.

test/utils/mockutils/mockcond/body.go
Cobalt0s: Usually request body is written as a string inside test file.
But I can see that this data may come from file.

PR #1064, opened 2024-10-07
feat: Enable read+write: Gong

Cobalt0s:
# Installation
image

# Read
image
This should count as success. Here are more details for this operation.
image
As you can see it says "no calls found". I can replicate this using PROXY.
image
The message is **identical**. If we modify `Since` parameter to be Jan 2024, since then we have one call:
image
It takes time the whole day for the call to appear after its creation. Querying should happen using "old" date instead of "now".

# Write
Create call.
image
image

# Examples
https://github.com/amp-labs/testdata/pull/16

### DIFF
--- a/providers/gong.go
+++ b/providers/gong.go
@@ -36,9 +36,9 @@ func init() {
 				Delete: false,
 			},
 			Proxy:     true,
-			Read:      false,
+			Read:      true,
 			Subscribe: false,
-			Write:     false,
+			Write:     true,
 		},
 	})
 }


Cobalt0s:
* **#1064** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1064?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Cobalt0s:
@laurenzlong 

"Since" funcitonality is a bit unusual. Gong calls are firstly uploaded, and after waiting for a day for AI processing it becomes available with a yesterday date. Calls object in gong is "Calls History".

Does server make API call with the "last" Since parameter using certain interval? Ex: give objects since 10-14-2024, and after 10 min, same query until success.

laurenzlong (approved):

laurenzlong:
@Cobalt0s that's very interesting, the interval depends on what the builder has specified, so it's really case dependent and can be as frequent as every 10 minutes. If Gong has not yet processed the call yet, does it simply not show up in the results, or is there some indication there's a call that's being processed?

Cobalt0s:
It just doesn't show. I would try to perform media URL upload for the recently created call which doesn't show up and it gives an error message that it is already in progress and media was already provided. So it is not really invisible. You can detect it by it having such "side effects". I wish they would return "calls" with some `status` property.

I was confused for a while that API says create successful. But I cannot verify... I managed to see a call appear on the next day. Not entirely sure how to have it guranteed. It is problem usage of Gong API not really ours. (thats one of the troubles implementing deep connectors I mentioned, that you have to become a true client of that provider).

PR #1068, opened 2024-10-14
feat: Add Pipedrive ListMetadata connector

Jkarage:
This adds the PipeDrive ListMetadata connector:

- Attempts getting the metadata through the API, If it fails (mostly due to 0 records returned), it will use the static schemas.json file.

Test Outputs:
Screenshot 2024-10-14 at 16 23 17


Cobalt0s (commented):
Looks great!

Metadata with schema file fallback is great.

Same thing as with the other PR, would be nice to have mock tests. 
One would immitate API response with response check and the other would have API call fail so the connector would fallback to static schemas.

laurenzlong:
@Cobalt0s  please re-review when you have a chance.

providers/pipedrive/metadata_test.go
```diff
@@ -0,0 +1,160 @@
...
+		},
+		{
+			Name:  "Zero records returned from server",
+			Input: []string{"activities"},
+			Server: httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				w.Header().Set("Content-Type", "application/json")
+				w.WriteHeader(http.StatusOK)
+				_, _ = w.Write([]byte(zeroRecords))
+			})),
+			Expected: &common.ListObjectMetadataResult{
```
Cobalt0s, line 63: You could shrink the output by comparing only the subset of fields: 
```
Comparator: func(baseURL string, actual, expected *common.ListObjectMetadataResult) bool {
    return mockutils.MetadataResultComparator.SubsetFields(actual, expected)
},
```

providers/pipedrive/objectNames.go
```diff
@@ -0,0 +1,12 @@
+package pipedrive
+
+import (
+	"github.com/amp-labs/connectors/common/handy"
+)
+
+// ObjectNameToResponseField maps ObjectName to the response field name which contains that object.
+var ObjectNameToResponseField = handy.NewDefaultMap(map[string]string{}, //nolint:gochecknoglobals
```
Cobalt0s, line 8: This declaration is not needed, for OpenAPI tool please use: `api3.IdenticalObjectCheck` which does exactly as the method.
Ex:
```
objects, err := explorer.GetBasicReadObjects(
	ignoreEndpoints, nil, nil, api3.IdenticalObjectCheck,
)
```

Cobalt0s (approved):
LGTM

providers/pipedrive/metadata_test.go
Jkarage: Cool, Looking into updating/cleaning these tests, As am working on adding Mock Tests for the other connectors.

scripts/oauth/token.go
```diff
@@ -272,7 +272,7 @@ func setup() *OAuthApp {
 
 	substitutions, err := registry.GetMap(SubstitutionsFieldName)
 	if err != nil {
-		slog.Warn("no substitutions, ensure that the provider info doesn't have any {{variables}}", "error", err)
+		slog.Warn("no substitutions, ensure that the provider info doesn't have any variables", "error", err)
```
Cobalt0s, line 275: I understand why this line was changed as I had troubles myself. But the brackets have to stay as it is referes to catalog variables, ex: `{{.workspace}}`.

PR #1076, opened 2024-10-15
feat: Add Pipedrive Read Connector

Jkarage:
This Adds the Pipedrive Read connector

Output Tests:
Screenshot 2024-10-15 at 14 55 14


providers/pipedrive/read.go
```diff
@@ -0,0 +1,53 @@
...
+		return nil, err
+	}
+
+	resp, err := c.Client.Get(ctx, url.String())
+	if err != nil {
+		return nil, err
+	}
+
+	return common.ParseResult(resp,
+		getRecords,
```
Cobalt0s, line 27: This method can be replaced with general method for get records.
```go
common.GetRecordsUnderJSONPath("data")
```
It expects an array of objects under key field "data".

Cobalt0s (commented):
It would be nice to have mock tests. 
1. One for successful fields reading with next page 
2. and the other for some error parsing. Pipedrive should have some error response formats.

For reference Zendesk has 3 major response formats: https://github.com/amp-labs/connectors/blob/main/providers/zendesksupport/errors.go.
If error response has `description` key in payload, then it will use `DescriptiveResponseError` struct, if `status` then `StatusResponseError` format and otherwise `MessageResponseError`, each struct converts provider response into golang error.

providers/pipedrive/read.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func (c *Connector) buildReadURL(config common.ReadParams) (*urlbuilder.URL, error) {
+	// If NextPage is set, then we're reading the next page of results.
+	// The NextPage URL has all the necessary parameters.
+	if len(config.NextPage) > 0 {
+		return urlbuilder.New(config.NextPage.String())
+	}
+
+	url, err := c.getApiURL(config.ObjectName)
```
RajatPawar, line 41: Please choose either of `getApiUrl` or `getAPIURL`

providers/pipedrive/read.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// begin fetching objects at provided start date
+	if !config.Since.IsZero() {
+		startDate := config.Since.Format(time.DateOnly)
```
RajatPawar, line 48: Is there an API limitation about supporting time? If so, could you please document it here

RajatPawar (approved):
LGTM!

providers/pipedrive/read_test.go
```diff
@@ -0,0 +1,210 @@
...
+							"due_date":                      "2024-10-30",
+							"due_time":                      "",
+							"duration":                      "",
+							"busy_flag":                     false,
+							"add_time":                      "2024-10-16 12:16:02",
+							"marked_as_done_time":           "",
+							"last_notification_time":        nil,
+							"last_notification_user_id":     nil,
+							"notification_language_id":      nil,
+							"subject":                       "I usually can't come up with words",
```
Cobalt0s, line 79: Same :smile: :wink: 

Speaking of `nil` values. You can remove all of them and modify the equality operation to check that all the fields you mention must be present in response. Just like you do in the test below.
```
			Comparator: func(baseURL string, actual, expected *common.ReadResult) bool {
				return mockutils.ReadResultComparator.SubsetFields(actual, expected) &&
					mockutils.ReadResultComparator.SubsetRaw(actual, expected) &&
					actual.NextPage.String() == expected.NextPage.String() &&
					actual.Rows == expected.Rows &&
					actual.Done == expected.Done
			}
```
I will be updating some utilities later on to have this as single utility function replacing several such usages.

Cobalt0s (approved):
Cool! :tada:

PR #1082, opened 2024-10-16
feat: Intercom Read via search

Cobalt0s:
# Changes

* First of all static schema for conversation originally had the right fields then they were replaced with `newsfeed` object. It seems this issue was not noticed as it affects "selectable fields" for Read. The right conversation schema is selected from OpenAPI (coming from POST search conversations)
* Some objects were not properly handled. A better distinction of `Ampersand Object Name` and matching URL path is introduced for exceptions.
image
* Tickets was added as a supported object. It can be queried via POST only. For `builders` this means they must pass Since parameter. This is added into [documentation](https://github.com/amp-labs/docs/pull/78).
* Conversations, Contacts, Tickets fallback to search endpoint when Since parameter is provided.
* Activity Logs is now an only GET endpoint that uses `Since` param.

This should complete full support of searchable objects by `updated_at` property.

### DIFF
--- a/providers/intercom/metadata/schemas.json
+++ b/providers/intercom/metadata/schemas.json
@@ -144,21 +144,32 @@
     "conversations": {
       "displayName": "Conversations",
       "fields": {
-        "body": "body",
-        "cover_image_url": "cover_image_url",
+        "admin_assignee_id": "admin_assignee_id",
+        "ai_agent": "ai_agent",
+        "ai_agent_participated": "ai_agent_participated",
+        "contacts": "contacts",
+        "conversation_parts": "conversation_parts",
+        "conversation_rating": "conversation_rating",
         "created_at": "created_at",
-        "deliver_silently": "deliver_silently",
+        "custom_attributes": "custom_attributes",
+        "first_contact_reply": "first_contact_reply",
         "id": "id",
-        "labels": "labels",
-        "name": "name",
-        "newsfeed_assignments": "newsfeed_assignments",
-        "reactions": "reactions",
-        "sender_id": "sender_id",
+        "linked_objects": "linked_objects",
+        "open": "open",
+        "priority": "priority",
+        "read": "read",
+        "sla_applied": "sla_applied",
+        "snoozed_until": "snoozed_until",
+        "source": "source",
         "state": "state",
+        "statistics": "statistics",
+        "tags": "tags",
+        "team_assignee_id": "team_assignee_id",
+        "teammates": "teammates",
         "title": "title",
         "type": "type",
         "updated_at": "updated_at",
-        "workspace_id": "workspace_id"
+        "waiting_since": "waiting_since"
       }
     },
     "data_attributes": {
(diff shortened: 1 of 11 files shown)


Cobalt0s:
* **#1082** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1082?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

providers/intercom/metadata/schemas.json
```diff
@@ -328,6 +339,30 @@
         "updated_at": "updated_at",
         "workspace_id": "workspace_id"
       }
+    },
+    "tickets": {
+      "displayName": "Ticket List",
```
RajatPawar, line 344: This can be `Tickets` since this seems to be the only way to get all tickets from the API anyway

providers/intercom/objectNames.go
```diff
@@ -6,6 +6,7 @@ import (
 )
 
 // Supported object names can be found under schemas.json.
+// NOTE: tickets can be queried only by using non-empty `Since` parameter.
```
RajatPawar, line 9: What is the behavior of the tickets search API when we pass in no filters? Would that list all the tickets? If that is the case, I think that's better behavior and would be useful for builders to receive a backfill of all tickets, and then receive only updated tickets. 

i.e. when Since is not present, call `POST /tickets/search` with no request body, else call `POST /tickets/search WHERE updated_at > Since`

providers/intercom/objectNames.go
```diff
@@ -18,9 +19,32 @@ var ObjectNameToResponseField = handy.NewDefaultMap(map[string]string{ //nolint:
...
+	"activity_logs": "/admins/activity_logs",
+	"collections":   "/help_center/collections",
+	"help_centers":  "/help_center/help_centers",
+	"news_items":    "/news/news_items",
+	"newsfeeds":     "/news/newsfeeds",
+}, func(obj string) string {
+	return obj
+})
+
+var incrementalSearchObjectPagination = handy.NewDefaultMap(map[string]int{ //nolint:gochecknoglobals
```
RajatPawar, line 41: You can add `nolint:gomnd` on the line above the var and it should not lint the var (so that you don't have to repeat it)

RajatPawar (changes requested):
Added a comment about how we can list all objects

providers/intercom/objectNames.go
Cobalt0s: This is a great hack!
image
Other objects can be retrieved using GET endpoints. Tickets is the only exception that requires `Search` to see any.

This deep connector has many exceptions now.

providers/intercom/metadata/schemas.json
Cobalt0s: Good catch. I will add a mapping to openapi script

providers/intercom/objectNames.go
Cobalt0s: That worked!

providers/intercom/objectNames.go
Cobalt0s: I have updated the docs removing the NOTE on tickets: https://github.com/amp-labs/docs/pull/78.

Also the full historic read works (temporary replaced "conversations" with "tickets", without Since param):
image

Cobalt0s:
@RajatPawar addressed your comments in the second commit: https://github.com/amp-labs/connectors/pull/1082/commits/1e66d010e2c17f916646493ace3ee7c4ac824bda

providers/intercom/search.go
```diff
@@ -0,0 +1,116 @@
...
+
+	return rsp, url, nil, true
+}
+
+func (c *Connector) createSearchPayload(params common.ReadParams) (*searchReqPayload, error) {
+	if params.Since.IsZero() && params.ObjectName == ticketsObjectName {
+		// Perform full read for tickets using POST query.
+		// This is a hack, query is designed to return all objects.
+		return &searchReqPayload{
+			Query: searchQuery{
```
RajatPawar, line 51: @Cobalt0s this can also be done for a couple more objects, right? Off the top of my head, articles would be one of them

RajatPawar (approved):

providers/intercom/search.go
Cobalt0s: Articles search is done via GET and has no time parameter to benefit `Since` filter.

Docs: https://developers.intercom.com/docs/references/rest-api/api.intercom.io/articles/searcharticles

PR #1085, opened 2024-10-16
fix: Error Handler soft media requirement

Cobalt0s:
fix: Error Handler soft media requirement

remove tests


Cobalt0s:
* **#1085** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1085?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

common/interpreter/error.go
```diff
@@ -18,16 +17,22 @@ type FaultyResponseHandler interface {
...
+	UnknownMedia FaultyResponseHandler
+	Fallback     FaultyResponseHandler
 }
 
-func (h ErrorHandler) Handle(res *http.Response, body []byte) error {
+func (h ErrorHandler) Handle(res *http.Response, body []byte) error { // nolint:cyclop
 	mediaType, _, err := mime.ParseMediaType(res.Header.Get("Content-Type"))
-	if err != nil {
-		return fmt.Errorf("%w: %w", ErrMissingContentType, err)
+	if err != nil && h.UnknownMedia != nil {
```
jlimatampersand, line 32: This essentially is just adding another layer of complexity in my opinion. Why are we

jlimatampersand: This is essentially requires the handler to accept another method for UnknowMedia, and Fallback. This does not prevent the error messages to be wrongly parsed when there is no specific handlers.

Cobalt0s: It provides customization to the deep connector. Salesforce is so far the only one that needs to handle absent media type in error response. How is this handled is up to the Salesforce package.

Let me explain, the path how we ended up here. Consider the base case, where all error response for all connectors are parsed using `common.InterpretError(res, body)`. What it does is use the whole body as error text and enhance with some trivial errors based on status code.
In most cases we will get 400 error like such:
``
error caller: Arbitrary text from response body
``

To improve this we can stop treating error response as `opaque bytes`. To do this we need to know the language.
Most are JSON, sometimes HTML may popup in the inconsitent APIs. Thats why media type is checked to judge if we can reason about `opaque string` in a valuable manner.

For Salesforce I would imagine a separate PR just to specify how unknown media should be handled. 
Otherwise, it will be treated as opaque string via `common.InterpretError(res, body)`, thats how all errors were processed long time ago and a present fallback.

I think `Fallback` method shouldn't really be used, so that can be removed. Ideally, new media should be added as a property to error handler.

jlimatampersand: I think in this case, we can have one function that returns error type `UnparseableError` which will include raw responses. If there is no mimetype or fails to get mimetype. But let's not make it method. We can do something like 

```
if err != nil || mediaType == "" {
  return common.HandleUnparseableHTTPResponseErrror(res, body) // return type is common.UnparseableHTTPResponseError 
}
```

so that we don't have to attach define `UnknownMedia` every time

jlimatampersand: cc @eberle1080

Cobalt0s: I cannot replicate the error because the script needs some configuring which is hard to know what parts are common and what unique per salesfroce account.


Special error wrapper makes sense. I will add a commit attempting to do this.

Cobalt0s: With or without the changes in the current PR it seems that I can run the `event-flow` without any issues 
image

Since Salesforce event flow script doesn't return any errors it doesn't make the change about missing media type irrelevant. Therefore, I added mock tests to demonstrate how error handler would behave.

Related changes are grouped into 3 commits.

common/interpreter/error.go
```diff
@@ -2,32 +2,41 @@ package interpreter
...
-	XML  FaultyResponseHandler
-	HTML FaultyResponseHandler
+	JSON         FaultyResponseHandler
+	XML          FaultyResponseHandler
+	HTML         FaultyResponseHandler
+	UnknownMedia FaultyResponseHandler
 }
 
-func (h ErrorHandler) Handle(res *http.Response, body []byte) error {
+func (h ErrorHandler) Handle(res *http.Response, body []byte) error { // nolint:cyclop
```
jlimatampersand, line 28: I think we can rewrite this whole function like below


```
func (h ErrorHandler) Handle(res *http.Response, body []byte) error {
	mediaType, _, err := mime.ParseMediaType(res.Header.Get("Content-Type"))
	if err == nil {
		if h.JSON != nil && mediaType == "application/json" {
			return h.JSON.HandleErrorResponse(res, body)
		}

		if h.XML != nil && (mediaType == "text/xml" || mediaType == "application/xml") {
			return h.XML.HandleErrorResponse(res, body)
		}

		if h.HTML != nil && (mediaType == "text/html" || mediaType == "application/html") {
			return h.HTML.HandleErrorResponse(res, body)
		}
	}

	return common.InterpretError(res, body)
}

```

jlimatampersand (changes requested):

common/interpreter/error.go
Cobalt0s: I will do that, but I thought you wanted to treat unknown media type error handling in a special way, as per thread above: 
```
return common.HandleUnparseableHTTPResponseErrror(res, body) // return type is common.UnparseableHTTPResponseError 
```

jlimatampersand: Initially, I thought so. But I looked at the entire function and common.InterpretError is independent of the mime type while XML, JSON, and HTML are just some special (or common) ones that we care. The goal of the initial task was to correct the error message when we are incorrectly parsing responses. `common.InterpretError` just returns a status code and returns string message from providers, that should be sufficient I think.

jlimatampersand (approved):

common/interpreter/error.go
Cobalt0s: That makes sense. I merging PR.

PR #1087, opened 2024-10-21
fix: handle recordId field in the Attio write

sanjay-a01:
…nality


laurenzlong:
@sanjay-a01 please fix linter then merge

PR #1090, opened 2024-10-22
feat: Add ZohoCRM ListMetadata connnector

Jkarage:
This adds the zohocrm ListMetadata connector.
Uses specific object endpoint for retrieving metadata fields.

Test Output:
Screenshot 2024-10-22 at 18 42 09

### DIFF
--- a/providers/zoho.go
+++ b/providers/zoho.go
@@ -9,8 +9,9 @@ func init() {
 		AuthType:    Oauth2,
 		BaseURL:     "https://www.zohoapis.com",
 		Oauth2Opts: &Oauth2Opts{
-			GrantType:                 AuthorizationCode,
-			AuthURL:                   "https://accounts.zoho.com/oauth/v2/auth",
+			GrantType: AuthorizationCode,
+			AuthURL:   "https://accounts.zoho.com/oauth/v2/auth",
+			// ref: https://www.zoho.com/analytics/api/v2/authentication/generating-code.html
 			AuthURLParams:             map[string]string{"access_type": "offline"},
 			TokenURL:                  "https://accounts.zoho.com/oauth/v2/token",
 			ExplicitScopesRequired:    true,
--- a/providers/zoho.go
+++ b/providers/zoho.go
@@ -11,6 +11,7 @@ func init() {
 		Oauth2Opts: &Oauth2Opts{
 			GrantType:                 AuthorizationCode,
 			AuthURL:                   "https://accounts.zoho.com/oauth/v2/auth",
+			AuthURLParams:             map[string]string{"access_type": "offline"},
 			TokenURL:                  "https://accounts.zoho.com/oauth/v2/token",
 			ExplicitScopesRequired:    true,
 			ExplicitWorkspaceRequired: false,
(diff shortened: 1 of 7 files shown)


providers/zohocrm/metadata.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil, common.ErrMissingObjects
+	}
+
+	objMetadata := common.ListObjectMetadataResult{
+		Result: make(map[string]common.ObjectMetadata),
+		Errors: make(map[string]error),
+	}
+
+	for _, obj := range objectNames {
+		url, err := c.getAPIURL(restMetadataEndpoint)
```
jlimatampersand, line 33: Since these calls in loop doesn't need to be in order, we can do this is go func and use wait group to optimize time.

providers/zohocrm/metadata.go
```diff
@@ -0,0 +1,82 @@
...
+		}
+
+		metadata.DisplayName = capObj
+		objMetadata.Result[obj] = *metadata
+	}
+
+	return &objMetadata, nil
+}
+
+func metadataMapper(resp *common.JSONHTTPResponse) (*common.ObjectMetadata, error) {
```
jlimatampersand, line 63: I think this can be named `parseMetadata` or `parseMetadataFromResponse`

providers/zohocrm/metadata.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"context"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/naming"
+)
+
+// restMetadataEndpoint is the resource for retrieving metadata details.
+const restMetadataEndpoint = "settings/fields"
```
jlimatampersand, line 11: nit: add link to doc in comment please.

providers/zoho.go
```diff
@@ -11,6 +11,7 @@ func init() {
 		Oauth2Opts: &Oauth2Opts{
 			GrantType:                 AuthorizationCode,
 			AuthURL:                   "https://accounts.zoho.com/oauth/v2/auth",
+			AuthURLParams:             map[string]string{"access_type": "offline"},
```
jlimatampersand, line 14: Can you add link to doc in comment.

jlimatampersand (changes requested):
Mostly looking good. Please add doc links, and some minor changes. Also, use go func and wait group for getting object metadata to optimize. Feel free to ping me when done.

providers/zohocrm/metadata.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (c *Connector) ListObjectMetadata(ctx context.Context,
+	objectNames []string,
+) (*common.ListObjectMetadataResult, error) {
+	var (
+		wg sync.WaitGroup //nolint: varnamelen
+		mu sync.Mutex     //nolint: varnamelen
+	)
+
+	wg.Add(len(objectNames))
```
jlimatampersand, line 30: This can be move to after `if len(objectNames) == 0` check since `wg` is not referred before the loop.

jlimatampersand (approved):
Pre-approving, just move the wg to few lines below, then feel free to merge.

PR #1096, opened 2024-10-23
feat:Add ZohoCRM Read connector

Jkarage:
This Adds ZohoCRM Read connector.

Manual Test Outputs:
Testing the `If-Modified-Since` header functionality. Works fine considering the timestamp provided.
Screenshot 2024-10-23 at 14 15 35

Read(After Adjusting it):
Screenshot 2024-10-23 at 14 16 10

### DIFF
--- a/providers/zohocrm/errors.go
+++ b/providers/zohocrm/errors.go
@@ -1,0 +1,45 @@
+package zohocrm
+
+import (
+	"bytes"
+	"encoding/json"
+	"io"
+	"net/http"
+)
+
+/*
+The normal standard response from ZohoCRM look like this:
+{
+	"data": [{...}, {...}],
+	"info": {...}
+}
+*/
+
+// responseHandler wraps the http.StatusNotModified response into http.StatusOK
+// and sends back an empty data field response.
+func responseHandler(resp *http.Response) (*http.Response, error) { //nolint:cyclop
+	// responseData represents data we will send back when the response status code 304.
+	responseData := map[string]any{
+		"data": []any{},
+		"info": map[string]any{
+			"page":         1,
+			"more_records": false,
+		},
+	}
+
+	// When the ZohoCRM API responds with 304 status code,
+	// this indicates there is no data modified since the modification time provided.
+	// We modify the response to return 200, And an empty data field response.
+	if resp.StatusCode == http.StatusNotModified {
+		// Build an empty Response Result (Mimicking ZohoResponse with empty data)
+		data, err := json.Marshal(responseData)
+		if err != nil {
+			return nil, err
+		}
+
+		resp.StatusCode = http.StatusOK
+		resp.Body = io.NopCloser(bytes.NewBuffer(data))
+	}
+
+	return resp, nil
+}
--- a/providers/zohocrm/parse.go
+++ b/providers/zohocrm/parse.go
@@ -1,0 +1,57 @@
+package zohocrm
+
+import (
+	"strconv"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/jsonquery"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+	"github.com/spyzhov/ajson"
+)
+
+/*
+   "data": {...},
+   "info": {
+        "call": false,
+        "per_page": 5,
+        "next_page_token": "c8582xx9e7c7",
+        "count": 5,
+        "sort_by": "id",
+        "page": 1,
+        "previous_page_token": null,
+        "page_token_expiry": "2022-11-11T15:08:14+05:30",
+        "sort_order": "desc",
+        "email": false,
+        "more_records": true
+    }
+*/
+
+func getNextRecordsURL(url *urlbuilder.URL) common.NextPageFunc {
+	return func(node *ajson.Node) (string, error) {
+		more, err := jsonquery.New(node, "info").Bool("more_records", false)
+		if err != nil {
+			return "", err
+		}
+
+		if *more {
+			nextPageToken, err := jsonquery.New(node, "info").Str("next_page_token", true)
+			if err != nil {
+				return "", err
+			}
+
+			currPage, err := jsonquery.New(node, "info").Integer("page", false)
+			if err != nil {
+				return "", err
+			}
+
+			nextPage := *currPage + 1
+
+			url.WithQueryParam("page_token", *nextPageToken)
+			url.WithQueryParam("page", strconv.FormatInt(nextPage, 10))
+
+			return url.String(), nil
+		}
+
+		return "", nil
+	}
+}
--- a/providers/zohocrm/read.go
+++ b/providers/zohocrm/read.go
@@ -28,14 +28,9 @@ func (c *Connector) Read(ctx context.Context, config common.ReadParams) (*common
 	fields := strings.Join(config.Fields.List(), ",")
 	url.WithQueryParam("fields", fields)
 
-	// Add the `If-Modified-Since` header if provided.
-	// All Objects(Modules in Zoho terms) supports this.
-	modHeader := common.Header{
-		Key:   "If-Modified-Since",
-		Value: config.Since.Format(time.RFC3339),
-	}
+	modHeader := modificationHeaders(config)
 
-	res, err := c.Client.Get(ctx, url.String(), modHeader)
+	res, err := c.Client.Get(ctx, url.String(), modHeader...)
 	if err != nil {
 		return nil, err
 	}
@@ -46,4 +41,19 @@ func (c *Connector) Read(ctx context.Context, config common.ReadParams) (*common
 		common.GetMarshaledData,
 		config.Fields,
 	)
+}
+
+func modificationHeaders(config common.ReadParams) []common.Header {
+	// Add the `If-Modified-Since` header if provided.
+	// All Objects(or Modules in ZohoCRM terms) supports this.
+	if !config.Since.IsZero() {
+		modHeader := common.Header{
+			Key:   "If-Modified-Since",
+			Value: config.Since.Format(time.RFC3339),
+		}
+
+		return []common.Header{modHeader}
+	}
+
+	return []common.Header{}
 }
--- a/providers/zohocrm/read.go
+++ b/providers/zohocrm/read.go
@@ -1,0 +1,49 @@
+package zohocrm
+
+import (
+	"context"
+	"strings"
+	"time"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/naming"
+)
+
+// Read retrieves data based on the provided common.ReadParams configuration parameters.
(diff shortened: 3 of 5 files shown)


providers/zohocrm/errors.go
```diff
@@ -0,0 +1,45 @@
...
+	// responseData represents data we will send back when the response status code 304.
+	responseData := map[string]any{
+		"data": []any{},
+		"info": map[string]any{
+			"page":         1,
+			"more_records": false,
+		},
+	}
+
+	// When the ZohoCRM API responds with 304 status code,
```
jlimatampersand, line 30: Can you add a link to API documentation for the endpoint we are using as a reference?

providers/zohocrm/errors.go
```diff
@@ -0,0 +1,45 @@
...
+
+// responseHandler wraps the http.StatusNotModified response into http.StatusOK
+// and sends back an empty data field response.
+func responseHandler(resp *http.Response) (*http.Response, error) { //nolint:cyclop
+	// responseData represents data we will send back when the response status code 304.
+	responseData := map[string]any{
+		"data": []any{},
+		"info": map[string]any{
+			"page":         1,
+			"more_records": false,
```
jlimatampersand, line 26: for clarity, we can keep this inside of the if statement below.

providers/zohocrm/parse.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/jsonquery"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+	"github.com/spyzhov/ajson"
+)
+
+/*
+   "data": {...},
```
jlimatampersand, line 11: is this object or a list? This is contradicting from what I saw when you handle 304 status code. Lets have API link here as reference.

providers/zohocrm/errors.go
```diff
@@ -0,0 +1,45 @@
...
+		},
+	}
+
+	// When the ZohoCRM API responds with 304 status code,
+	// this indicates there is no data modified since the modification time provided.
+	// We modify the response to return 200, And an empty data field response.
```
jlimatampersand, line 32: let's change the comment as below

```
// When there is no new record after the specified time `since`, ZohoCRM returns `304 Status Not Modified`.  Then we wrap this response to 200 Status Okay with empty array data. 
```

providers/zohocrm/errors.go
```diff
@@ -0,0 +1,45 @@
...
+	"info": {...}
+}
+*/
+
+// responseHandler wraps the http.StatusNotModified response into http.StatusOK
+// and sends back an empty data field response.
+func responseHandler(resp *http.Response) (*http.Response, error) { //nolint:cyclop
+	// responseData represents data we will send back when the response status code 304.
+	responseData := map[string]any{
+		"data": []any{},
```
jlimatampersand, line 23: so this is not data. This is the object name in lowercase.

providers/zohocrm/errors.go
```diff
@@ -0,0 +1,45 @@
...
+	"bytes"
+	"encoding/json"
+	"io"
+	"net/http"
+)
+
+/*
+The normal standard response from ZohoCRM look like this:
+{
+	"data": [{...}, {...}],
```
jlimatampersand, line 13: the key should be "user" or "profile" or something not "data"

jlimatampersand (changes requested):

providers/zohocrm/errors.go
```diff
@@ -0,0 +1,45 @@
...
+		"info": map[string]any{
+			"page":         1,
+			"more_records": false,
+		},
+	}
+
+	// When the ZohoCRM API responds with 304 status code,
+	// this indicates there is no data modified since the modification time provided.
+	// We modify the response to return 200, And an empty data field response.
+	if resp.StatusCode == http.StatusNotModified {
```
jlimatampersand, line 33: @RajatPawar @laurenzlong I don't think we need to handle this, do we? We can just return 304 with data `nil`. Then it will be taken care of by the caller, right?

jlimatampersand (changes requested):
There are many parts I am not sure of. Please respond to my comments and resolve if necessary.

providers/zohocrm/errors.go
Jkarage: This is specific to Reading Records API. Updated comments to include this and a link.

providers/zohocrm/errors.go
Jkarage: This should be resolved with the above comment, If Accepted.

providers/zohocrm/parse.go
Jkarage: A list, Thanks.

providers/zohocrm/errors.go
RajatPawar: I agree, Read()'s return logic should take a nil response and convert it into an empty map

providers/zohocrm/read.go
```diff
@@ -0,0 +1,55 @@
...
+		common.GetRecordsUnderJSONPath("data"),
+		getNextRecordsURL(url),
+		common.GetMarshaledData,
+		config.Fields,
+	)
+}
+
+func constructHeader(config common.ReadParams, url *urlbuilder.URL) []common.Header {
+	// Retrieve the since from the url.
+	since, ok := url.GetFirstQueryParam("since")
```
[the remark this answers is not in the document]
Jkarage: In order for us to support since header, we need to preserve the value of the header(If-Modified-Since) used during the first call. This is bacause the next_page_token generated depends on that value.  

Sample Examples: 

First call Example:

Read(ctx, common.ReadParams{
    ObjectName: "Leads",
    Since: time.Now(-10*time.Hour)  // Lets say value is `2009-11-10 13:00:00 +0000`
    // Other Fields
})

The next_page_token generatedd after this call depends on the value passed in the Since field. So in the next calls we have to pass the same value. Changing this invalidates the token. 

Next Call Examples:

Read(ctx, common.ReadParams{
    ObjectName: "Leads",
    Since: "2009-11-10 13:00:00 +0000" // In all next-page requests this value should never change & is Required.
    NextPageToken: xxxxxxxxxxxxxxxxxxxx
})

So to preseve this value sent during the first request, I created a custom parameter (since). This param has nothing to do with Zoho. Am using this parameter to preserve the Since value used in the first request, And am returning it in the Next Page URL. So that whenever we make Next Page Requests. I can get it and Add the value in the header as well.

providers/zohocrm/url.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	// Adds the fields requirement parameter.
+	fields := strings.Join(config.Fields.List(), ",")
+	url.WithQueryParam("fields", fields)
+
+	// This will be added during the first call.
+	// This is a custom parameter, we add this in the request
+	// So as we do not loose the state of the since parameter value.
+	// `since` is just a preferred name parameter.
```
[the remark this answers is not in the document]
Jkarage: Since is not Recognized parameter in Zoho, It's a custom parameter.I Use it to pass the Since value passed during the first call. I couldn't figure out another way to save that Since field Value passed during the first call. So am passing it as a query parameter in the URLs(First call URL and NextPage URLs). So I can use it during the calls.

Jkarage:
@jlimatampersand  Resolved some of the comments. Responded to some comments as well.

providers/zohocrm/errors.go
jlimatampersand: got it, thanks for the explanation and thanks for the doc link!

providers/zohocrm/read.go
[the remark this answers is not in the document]
jlimatampersand: Okay, so what you are saying that next_page_token alone does not work? and it only works if the `If-Modified-Since` is matching from previous call?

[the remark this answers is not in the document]
jlimatampersand: That sounds really odd to me that Zoho does that. For sanity check, can you make a first call with `If-Modified-Since` header, then make 2 second calls as below and share screenshots of results? 
1. second page with `If-Modified-Since` 
2. second page without `If-Modified-Since`

I like the creative approach you are taking, but I am afraid that at some point, they might use or try to parse that param and return unexpected results. So just want to double check.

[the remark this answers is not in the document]
jlimatampersand: Feel free to ping me directly on slack when you are done with this. We don't need to wait for another day to be re-reviewed.

providers/zohocrm/read.go
```diff
@@ -0,0 +1,55 @@
...
+	if !ok {
+		return []common.Header{}
+	}
+
+	// Add the `If-Modified-Since` header if provided.
+	// All Objects(or Modules in ZohoCRM terms) supports this.
+	if !config.Since.IsZero() {
+		modHeader := common.Header{
```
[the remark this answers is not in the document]
jlimatampersand: also just `header` should be good enough.

jlimatampersand (changes requested):

providers/zohocrm/errors.go
Jkarage: Okay, Just for clarification. If we don't handle this then we will be returning an error to the caller when there is no new records to return(when we receive 304). Since we are assuming any non 200>=x<299 status are errors. won't this be an issue to the caller?

providers/zohocrm/read.go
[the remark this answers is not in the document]
Jkarage: Header Tests:
First Call with the If-Modified-Since Header: 
Result:
Screenshot 2024-10-24 at 22 20 46

Second-Call with the `If-Modified-since` Header:
Screenshot 2024-10-24 at 22 22 02

Second-Call  with no `If-Modified-since` Header, Responds with an error:
Screenshot 2024-10-24 at 22 22 54

providers/zohocrm/errors.go
jlimatampersand: Actually, your approach makes sense and good point on how we are handling the error. Yeah, wrapping with 200 ok and empty array is good.

jlimatampersand: Okay, so after discussing with others in my team. This is not the correct way to handle it. We should handle 3XX error correctly for all connectors. But for the scope of this task, we can merge it as is. I will create a new task to handle 3XX errors in general, but it will not be your task to do. 

TLDR, this PR is good for now.

jlimatampersand (changes requested):
Please take a look at my response on `Since`. We can remove that logic.

providers/zohocrm/read.go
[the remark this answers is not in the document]
Jkarage: Okay, Noted.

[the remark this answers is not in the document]
Jkarage: Also removed the variable, so am returning directly. I had misunderstood the named variable comments.

jlimatampersand (approved):

PR #1100, opened 2024-10-25
feat: Add ZohoCRM Write Connector

Jkarage:
This Adds the ZohoCRM write conector. 

Writing to zoho supports upto 100 records per API Call. So the the RecordData should be a slice of maps.
Supports Updating 1 record per API Call. 

They have an API for updating several records as well.(We currently do not support this)

We do not return the recordId as one may create several records per call.

Tests:
Screenshot 2024-10-25 at 08 11 24

On the last call, The request was a success but there are some records that failed to insert.

### DIFF
--- a/providers/zohocrm/write.go
+++ b/providers/zohocrm/write.go
@@ -31,6 +31,7 @@ Sample Response Data:
             "message": "record added",
             "status": "success"
         },
+		{...}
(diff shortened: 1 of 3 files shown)


jlimatampersand (approved):
LGTM

PR #1106, opened 2024-10-30
feat: Add CloseCRM Read connector

Jkarage:
Adds the CloseCRM Read Connector. The Close API offers the Listing and Searching APIs.
The searching API supports incremental Read by Adding Filtering Queries in the request body. I have added the switching route to search if the Since Field is provided. 

The search API has a limitation of reading up to 10K with pagination.
The Cursor for pagination expires in 30 seconds.

Test Outputs:
Screenshot 2024-10-30 at 20 21 31

### DIFF
--- a/providers/closecrm/parse.go
+++ b/providers/closecrm/parse.go
@@ -19,7 +19,7 @@ Read Response Schema:
 
 */
 
-var (
+const (
 	defaultPageSize = "100"      // nolint:gochecknoglobals
 	limitQuery      = "_limit"   // nolint:gochecknoglobals
 	skipQuery       = "_skip"    // nolint:gochecknoglobals
@@ -30,6 +30,7 @@ var (
 var ErrSkipFailure = errors.New("error: failed to create next page url")
 
 // nextRecordsURL builds the next-page url func.
+// https://developer.close.com/topics/pagination/
 func nextRecordsURL(url *urlbuilder.URL) common.NextPageFunc {
 	return func(node *ajson.Node) (string, error) {
 		// check if there is more items in the collection.
@@ -54,17 +55,23 @@ func nextRecordsURL(url *urlbuilder.URL) common.NextPageFunc {
 	}
 }
 
-func searchNextRecords() common.NextPageFunc {
-	return func(node *ajson.Node) (string, error) {
-		crs, err := jsonquery.New(node).Str("cursor", true)
-		if err != nil {
-			return "", err
-		}
+/*
+Search Response schema:
 
-		if crs == nil {
-			return "", nil
-		}
-
-		return *crs, nil
+	{
+		 "data": [{...},{...}],
+		 "cursor": "..."
 	}
+*/
+func getNextRecordCursor(node *ajson.Node) (string, error) {
+	crs, err := jsonquery.New(node).Str("cursor", true)
+	if err != nil {
+		return "", err
+	}
+
+	if crs == nil {
+		return "", nil
+	}
+
+	return *crs, nil
 }
--- a/providers/closecrm/parse.go
+++ b/providers/closecrm/parse.go
@@ -1,0 +1,42 @@
+package closecrm
+
+import (
+	"strconv"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/jsonquery"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+	"github.com/spyzhov/ajson"
+)
+
+/*
+Response Schema:
+{
+    "has_more": false,
+    "total_results": 1,
+    "data": [
+        {...},
+		{...}
+    ]
+}
+
+*/
+
+// nextRecordsURL builds the next-page url func.
+func nextRecordsURL(url *urlbuilder.URL) common.NextPageFunc {
+	return func(node *ajson.Node) (string, error) {
+		// check if there is more items in the collection.
+		hasMore, err := jsonquery.New(node).Bool("has_more", false)
+		if err != nil {
+			return "", err
+		}
+
+		if *hasMore {
+			url.WithQueryParam("start", strconv.FormatInt(*startValue, 10))
+
+			return url.String(), nil
+		}
+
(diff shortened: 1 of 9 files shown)


providers/closecrm/search.go
```diff
@@ -0,0 +1,99 @@
...
+// searchEndpoint represents the search api endpoint.
+// ref: https://developer.close.com/resources/advanced-filtering/
+const searchEndpoint = "data/search"
+
+// Search reads data through searching API. Supports advanced filtering using the filters field.
+// The NextPage Token generated takes 30 seconds to expire.
+//
+// doc: https://developer.close.com/resources/advanced-filtering
+func (c *Connector) Search(ctx context.Context, config SearchParams) (*common.ReadResult, error) {
+	input, err := buildUpdatedDateFilter(config)
```
jlimatampersand, line 31: let's rename this to searchFilter to make it more clear.

providers/closecrm/types.go
```diff
@@ -0,0 +1,40 @@
...
+
+type SearchParams struct {
+	ObjectName string
+	Fields     handy.StringSet
+	Since      time.Time
+	NextPage   common.NextPageToken
+	Filters    Filters
+}
+
+type Filters struct {
```
jlimatampersand, line 18: This should be singular `Filter`

providers/closecrm/types.go
```diff
@@ -0,0 +1,40 @@
...
+	FilterQueries FilterQueries       `json:"query"`
```
jlimatampersand, line 19: We can keep the field name as `Query` to match json schema required by provider.

providers/closecrm/types.go
```diff
@@ -0,0 +1,40 @@
...
+	Cursor        any                 `json:"cursor"`
+	Limit         int                 `json:"_limit"`  //nolint:tagliatelle
+	Fields        map[string][]string `json:"_fields"` //nolint:tagliatelle
+}
+
+type FilterQueries struct {
```
jlimatampersand, line 25: This can be just `Query`

test/closecrm/read/read.go
```diff
@@ -0,0 +1,107 @@
...
+	ctx, done := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
+	defer done()
+
+	// Set up slog logging.
+	utils.SetupLogging()
+
+	conn := closecrm.GetCloseConnector(ctx)
+	defer utils.Close(conn)
+
+	// if err := readActivities(ctx, conn); err != nil {
```
jlimatampersand, line 28: is there reason this is commented out?

providers/closecrm/search.go
```diff
@@ -0,0 +1,99 @@
...
+	return common.ParseResult(
+		resp,
+		common.GetRecordsUnderJSONPath("data"),
+		searchNextRecords(),
+		common.GetMarshaledData,
+		config.Fields,
+	)
+}
+
+func buildUpdatedDateFilter(config SearchParams) (Filters, error) {
```
jlimatampersand, line 55: rename config to `params`

providers/closecrm/types.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"time"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/handy"
+)
+
+type SearchParams struct {
+	ObjectName string
+	Fields     handy.StringSet
```
jlimatampersand, line 12: This can be just []string. Unnecessary use of this set. There is no logic in this PR that uses the benefits of this package. I actually don't like the abuse of this handy package happening in this repo. Please refrain from using it unless it is really necessary.

providers/closecrm/parse.go
```diff
@@ -0,0 +1,70 @@
...
+			url.WithQueryParam(limitQuery, defaultPageSize)
+
+			return url.String(), nil
+		}
+
+		return "", nil
+	}
+}
+
+func searchNextRecords() common.NextPageFunc {
```
jlimatampersand, line 57: So this function does not take an argument to return a function. So this is an unnecessary layer. We can just have and directly pass to `common.ParseResult(`

example below. 
```
func getNextRecordCursor(node *ajson.Node) (string, error) {...}
``

File: providers/closecrm/search.go [hunk 9]
```diff
@@ -0,0 +1,99 @@
...
+
+	resp, err := c.Client.Post(ctx, url.String(), input)
+	if err != nil {
+		return nil, err
+	}
+
+	return common.ParseResult(
+		resp,
+		common.GetRecordsUnderJSONPath("data"),
+		searchNextRecords(),
```
jlimatampersand, line 49: unnecessary function invocation to get an anonymous function. See the comment near `searchNextRecords` definition and fix this.

providers/closecrm/parse.go
```diff
@@ -0,0 +1,70 @@
...
+Read Response Schema:
+{
+    "has_more": false,
+    "total_results": 1,
+    "data": [{...},{...}]
+}
+
+*/
+
+var (
```
jlimatampersand, line 22: change to const

providers/closecrm/types.go
```diff
@@ -0,0 +1,40 @@
...
+	Fields        map[string][]string `json:"_fields"` //nolint:tagliatelle
+}
+
+type FilterQueries struct {
+	Type    string           `json:"type"`
+	Queries []map[string]any `json:"queries"`
+}
+
+// nolint:gochecknoglobals
+var (
```
jlimatampersand, line 31: const

jlimatampersand (changes requested):

test/closecrm/read/read.go
Jkarage: A mistake, Thanks.

providers/closecrm/parse.go
Jkarage: Yeah, Thanks.

jlimatampersand (approved):

PR #1114, opened 2024-11-01
feat: Package Goutils

Cobalt0s:
# Background

Handy package primarily was created to hold data structures, but some general utility methods slipped in.

# Changes
I think it makes more sense to split this package in 2 distinct pieces.
This PR is branching first package: `goutils`. The other would then become a simple rename.
Both packages live under `internal` folder.

Package includes:
* Generics struct cloning
* Error checks (used by scripts, not appropriate for connectors)
* Pointer returners to primitive types.

These function are used for golang convenience, syntactic sugar.


Cobalt0s:
* **#1115** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1115?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1114** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1114?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

internal/goutils/errors.go
```diff
@@ -1,4 +1,4 @@
-package handy
+package goutils
 
 func Must(err error) {
```
laurenzlong, line 3: Can you rename to `MustBeNil`? `Must` is not intuitive.

internal/goutils/pointers.go
```diff
@@ -1,4 +1,4 @@
-package handy
+package goutils
```
laurenzlong, line 1: Can you remove the `Str`, `Bool` and `Int64` functions? They don't really save any lines of code and makes the code that uses these functions more confusing the read since someone has to figure out what these functions do. Simply using the `&` operator is more intuitive.

internal/goutils/cloning.go
```diff
@@ -1,4 +1,4 @@
-package handy
+package goutils
```
laurenzlong, line 1: Seems like `Clone` is not used at all in the codebase, can you remove?

laurenzlong (changes requested):
@Cobalt0s I think you have to update main to the latest and then run `git rebase main`, now that the datautils PR has been merged. 

Also I'm not sure I like the name `goutils` (not sure why it's prefixed with `go`, since the whole codebase is in Go, so prefixing with `go` isn't adding any additional info). What about simply `utils` for now? Feel free to suggest other prefixes @jlimatampersand 

I also add some comments on the functions that can be removed, our general philosophy is that utility functions should only be introduced if they save lines of code and are a common use case. Otherwise it adds to the learning curve for contributors who are developing new connectors (and makes existing connectors harder to read and reason about). After you remove the functions I suggest removing, the only ones left will be: MustBeNil and IsTrue, so it might be easier to collapse them into one file.

jlimatampersand:
> @Cobalt0s I think you have to update main to the latest and then run `git rebase main`, now that the datautils PR has been merged.
> 
> Also I'm not sure I like the name `goutils` (not sure why it's prefixed with `go`, since the whole codebase is in Go, so prefixing with `go` isn't adding any additional info). What about simply `utils` for now? Feel free to suggest other prefixes @jlimatampersand
> 
> I also add some comments on the functions that can be removed, our general philosophy is that utility functions should only be introduced if they save lines of code and are a common use case. Otherwise it adds to the learning curve for contributors who are developing new connectors (and makes existing connectors harder to read and reason about). After you remove the functions I suggest removing, the only ones left will be: MustBeNil and IsTrue, so it might be easier to collapse them into one file.

I think we can just have these be part of utils library which we already have.

jlimatampersand (changes requested):
you can rename the package to just `utils`.

internal/goutils/errors.go
Cobalt0s: Must is a pattern in golang: https://stackoverflow.com/questions/12434565/built-in-helper-for-must-pattern-in-go
@eberle1080 I think there was some PR a while ago about some other method, and prefixing a method with `Must` that you suggested. What do you think about this case?

Must be nil is a bit confusing to me as it is using negation. Must be empty == without an error == successful. The closest equivalent I can think of is `assert ...` in JAVA.

Ex: Must(loadFile("data.json"))

Cobalt0s: Maybe we could take inspiration from this stackoverflow or blog: https://www.willem.dev/code-snippets/generic-must-function/
To imrpove the `Must` method.

internal/goutils/pointers.go
Cobalt0s: In golang you cannot do this. Address operator cannot be applied to literals: (reference: https://go.dev/ref/spec#Address_operators)
Ex: 
```var textPtr *string =  &"some text" // illegal```

Therefore, there will be 2 lines to express this. First create a variable and then take address. This is the reason I named it `goutils`. It is purely related to golang "issues".

Alternativly, there is a library: https://pkg.go.dev/github.com/icza/gog#Ptr but adding dependency for this case is an overkill in my opinion.

Just by reading more about this it seems that we can replace all functions with one generic. These methods were created just a long time ago before I started using generics in connectors lib.

File: internal/goutils/cloning.go (hunk 3 above)
Comment (Cobalt0s) [reply internal/goutils/cloning.go:1]: This method lived under `providers` package and I moved it into handy. Reason: providers is an "external" package that shouldn't be clumped by other utilitise.

Only 2 usages exist:
```
// Location: connectors/providers/utils.go

catalogCopy, err := goutils.Clone[*CatalogWrapper](catalogInstance)
providerInfo, err := goutils.Clone[ProviderInfo](pInfo)
```

Copying structs could be `goutils` level. Almost a standard/general package. In Java that would be `Cloneable` but for structs. There could be a dedicated coning package for various purposes. But I don't see the requirement for that yet. That would be the best naming but an overkill ahaha.

**Comment by Cobalt0s** [comment]
> you can rename the package to just `utils`.

Most methods from `utils` packages were catogarized into specialized packages to this date. I was trying to limit the usage of `utils` to avoid making it a God pacakge.

So far there are only 2 `utils` package:
image
One for testing scripts and the other for example scripts. For this context it makes sense.

On the other hand `connectors`, `providers` don't have utils pacakge. I suppose `goutils` rename to `utils` would fill the category.

NOTE: When I was working with `test` repo, the scripts there had imports of 2 utils which meant `alias importing` due to naming collision, therefore I would rather prefix utils with something by default.

**Comment by Cobalt0s** [comment]
Rebased and addressed some comments, used ideas after reading more about it.
First commit(important): https://github.com/amp-labs/connectors/pull/1114/commits/533b3bda6c80d971065c4a2e1897390f21d8716f
Second commit(repetative follow up): https://github.com/amp-labs/connectors/pull/1114/commits/e888c52cbbf4d8ebe9ef79afb97ee5976e9b2850

File: internal/goutils/utils.go [hunk 4]
```diff
@@ -0,0 +1,27 @@
...
+
+// MustBeNil - panics on non-empty error.
+func MustBeNil(err error) {
+	if err != nil {
+		panic(err)
+	}
+}
+
+// Must succeed - panics on non-empty error, returns one value.
+func Must[T any](obj T, err error) T { // nolint:ireturn
```
laurenzlong, line 16: Please remove Must and Must2, it is just as easy to write the following and it's way more readable.

```
a, b, err := someFunction(...)
MustBeNil(err)
```

internal/goutils/utils.go
```diff
@@ -0,0 +1,27 @@
+package goutils
+
+// Ptr returns a pointer to the given value.
+func Ptr[T any](input T) *T {
```
laurenzlong, line 4: Can you actually spell this out and call this function `Pointer` for clarity?

For context: https://google.github.io/styleguide/go/decisions#variable-names
- refer to the bullet that says "Do not simply drop letters to save typing. For example Sandbox is preferred over Sbx, particularly for exported names."

laurenzlong (approved):
Thanks for the context @Cobalt0s ! I think prefixing makes sense to avoid package collision, and I can't really think of anything better than `goutils` so let's go with it for now. Added more comments. Feel free to merge after you address them.

internal/goutils/utils.go
laurenzlong: Ah I see your other comment that links to https://www.willem.dev/code-snippets/generic-must-function/, that's probably why you introduced these 2 functions.

The thing about the Connectors library is that we cannot be as pedantic about Go conventions as we are in our closed source repos. Future contributors will not always be well-versed in Go. So in some cases, we have to deviate about what the "Go way" of doing things is in order to make this codebase more beginner-friendly. 

If you think `MustBeNil` is confusing, then feel free to rename to `MustNotErr`, I think that will read nicely, e.g.

```
MustNotErr(loadFile("data.json"))
```

But please remove `Must` and `Must2`, they're not intuitive to someone who hasn't done a lot of Go development.

internal/goutils/errors.go
laurenzlong: See https://github.com/amp-labs/connectors/pull/1114#discussion_r1830437721

internal/goutils/utils.go
Cobalt0s: Yes, I thought those posts had some good generalized ideas that ringed with our utils.
Your motivation is clear! I kind of didn't picture non-golang developers in a mix, which should be a mistake.

internal/goutils/utils.go
Cobalt0s: In my coding practice I try not to do this, the reason it was shortned is that I relied on one of the posts and for some reason the shortened version didn't trigger me.

In general I prefer longer function names and variables so that it becomes sort of an abstract English text, "cooking recipe instruction" ahaha.

jlimatampersand (approved):

PR #1117, opened 2024-11-04
[ENG-1652] Introduce a method to transform the webhook payload into a ReadResultRow

jlimatampersand:
## About this PR

This PR implements a Connector method for HubSpot that transforms webhook messages to `WebhookResult` which will contain original record data. In order to achieve that, this PR introduces `GetRecord(...)` method that takes `recordId` and `objectName` as arguments and returns `ReadResultRow` which will have raw data. 

## Test Result
```
jaehyunlim@Jaehyuns-MacBook-Pro connectors % go run test/hubspot/record/main.go  
time=2024-11-04T16:29:05.992-08:00 level=DEBUG msg="loading credentials file" path=./hubspot-creds.json
{
  "webhookMessage": {
    "appId": 4210286,
    "eventId": 100,
    "subscriptionId": 2902227,
    "portalId": 44237313,
    "occurredAt": 1730750483646,
    "subscriptionType": "contact.propertyChange",
    "attemptNumber": 0,
    "objectId": 75032904814,
    "changeSource": "CRM",
    "propertyName": "message",
    "propertyValue": "sample-value"
  },
  "record": {
    "fields": null,
    "raw": {
      "archived": false,
      "createdAt": "2024-11-05T00:29:06.427Z",
      "id": "75032904814",
      "properties": {
        "createdate": "2024-11-05T00:29:06.427Z",
        "email": "evertaufderhar@robel.org",
        "firstname": "Geovanni",
        "hs_object_id": "75032904814",
        "lastmodifieddate": "2024-11-05T00:29:06.427Z",
        "lastname": "Veum"
      },
      "updatedAt": "2024-11-05T00:29:06.427Z"
    }
  }
}
```

### DIFF
--- a/providers/hubspot/errors.go
+++ b/providers/hubspot/errors.go
@@ -52,11 +52,7 @@ type ErrLinks struct {
 
 func (c *Connector) interpretError(res *http.Response, body []byte) error {
 	mediaType, _, err := mime.ParseMediaType(res.Header.Get("Content-Type"))
-	if err != nil {
-		return fmt.Errorf("mime.ParseMediaType failed: %w", err)
-	}
-
-	if mediaType == "application/json" {
+	if err == nil && mediaType == "application/json" {
 		return c.interpretJSONError(res, body)
 	}
 
--- a/providers/hubspot/record.go
+++ b/providers/hubspot/record.go
@@ -11,6 +11,15 @@ import (
 	"github.com/amp-labs/connectors/internal/datautils"
 )
 
+//nolint:gochecknoglobals
+var (
+	getRecordSupportedObjectsSet = datautils.NewStringSet(
+		"company", "contact", "deal", "ticket", "line_item", "product",
+	)
+
+	errGerRecordNotSupportedForObject = errors.New("getRecord is not supproted for the object")
+)
+
 /*
    docs:
    https://developers.hubspot.com/beta-docs/reference/api/crm/objects/companies
@@ -21,15 +30,7 @@ import (
    https://developers.hubspot.com/beta-docs/reference/api/crm/objects/products
 */
 
-//nolint:gochecknoglobals
-var (
-	getRecordSupportedObjectsSet = datautils.NewStringSet(
-		"company", "contact", "deal", "ticket", "line_item", "product",
-	)
-
-	errGerRecordNotSupportedForObject = errors.New("getRecord is not supproted for the object")
-)
-
+// GetRecord returns a record from the object with the given ID and object name.
 func (c *Connector) GetRecord(ctx context.Context, objectName string, recordId string) (*common.ReadResultRow, error) {
 	if !getRecordSupportedObjectsSet.Has(objectName) {
 		return nil, fmt.Errorf("%w %s", errGerRecordNotSupportedForObject, objectName)
--- a/providers/hubspot/record.go
+++ b/providers/hubspot/record.go
@@ -31,7 +31,7 @@ var getRecordSupportedObjectsToPathMap = map[string]string{
 
 var ErrUnsupportedObject = errors.New("unsupported object")
 
-func (c *Connector) GetRecord(ctx context.Context, objectName string, recordId string) (*common.ReadResult, error) {
+func (c *Connector) GetRecord(ctx context.Context, objectName string, recordId string) (*common.ReadResultRow, error) {
 	objectNameInPath, ok := getRecordSupportedObjectsToPathMap[objectName]
 	if !ok {
 		return nil, fmt.Errorf("%w: %s", ErrUnsupportedObject, objectName)
@@ -49,10 +49,7 @@ func (c *Connector) GetRecord(ctx context.Context, objectName string, recordId s
 		return nil, fmt.Errorf("error parsing record: %w", err)
 	}
 
-	return &common.ReadResult{
-		Rows: 1,
-		Data: []common.ReadResultRow{
-			{Raw: *record},
-		},
+	return &common.ReadResultRow{
+		Raw: *record,
 	}, nil
 }
--- a/providers/hubspot/record.go
+++ b/providers/hubspot/record.go
@@ -7,6 +7,7 @@ import (
 	"path"
 
 	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/naming"
 )
 
 /*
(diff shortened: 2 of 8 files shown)


providers/hubspot/errors.go
```diff
@@ -52,11 +52,7 @@ type ErrLinks struct {
 
 func (c *Connector) interpretError(res *http.Response, body []byte) error {
 	mediaType, _, err := mime.ParseMediaType(res.Header.Get("Content-Type"))
-	if err != nil {
```
jlimatampersand, line 55: Similar to issue handled in this PR https://github.com/amp-labs/connectors/pull/1085

Made change for hubspot as well.

providers/hubspot/webhook.go
```diff
@@ -0,0 +1,47 @@
...
+	OccurredAt       int    `json:"occurredAt"`
+	SubscriptionType string `json:"subscriptionType"`
+	AttemptNumber    int    `json:"attemptNumber"`
+	ObjectId         int    `json:"objectId"`
+	ChangeSource     string `json:"changeSource"`
+	PropertyName     string `json:"propertyName"`
+	PropertyValue    string `json:"propertyValue"`
+}
+
+type WebhookResult struct {
```
jlimatampersand, line 25: @laurenzlong I wasn't too sure if this return type makes sense. I think we do need to get the original record. The way we standardize in `Read` is to use read result. but `GetRecord` is just reading a single record and not sure using the same structure makes sense. 

Maybe it might be better to introduce another type 
```
type Record struct {
   RecordId string
   Data any
}
```

or an interface

type Record interface {
  RecordId() string
  Raw any
}

test/hubspot/record/main.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// Set up slog logging.
+	utils.SetupLogging()
+
+	// Get the Hubspot connector.
+	conn := connTest.GetHubspotConnector(ctx)
+	defer utils.Close(conn)
+
+	// Write an artificial contact to Hubspot.
+	writeresult, err := conn.Write(ctx, common.WriteParams{
```
Cobalt0s, line 45:
```suggestion
	writeResult, err := conn.Write(ctx, common.WriteParams{
```

providers/hubspot/url.go
```diff
@@ -1,10 +1,11 @@
...
 import (
-	"strings"
+	"path"
 )
 
 // getURL is a helper to return the full URL considering the base URL & module.
 func (c *Connector) getURL(arg string) string {
-	return strings.Join([]string{c.BaseURL, c.Module.Path(), arg}, "/")
+	// TODO: use url package to join paths and avoid issues with slashes in another PR
+	return c.BaseURL + "/" + path.Join(c.Module.Path(), arg)
```
Cobalt0s, line 10:
```go
       // args - list of URI parts that will be joined using slashes

	url, err := urlbuilder.New(c.BaseURL, c.Module.Path(), args...)
	if err != nil {
		return nil, err
	}
```

All other places in this PR would revolve around this way of building URLs.

providers/hubspot/errors.go
```diff
@@ -52,11 +52,7 @@ type ErrLinks struct {
...
-		return fmt.Errorf("mime.ParseMediaType failed: %w", err)
-	}
-
-	if mediaType == "application/json" {
+	if err == nil && mediaType == "application/json" {
```
Cobalt0s, line 55: I can create a follow up PR but hubspot lacks mock tests to make such changes and I don't want to break prod.

I reviewed the code and the error handling can be replaced like so:
```
conn.Client.HTTPClient.ErrorHandler = interpreter.ErrorHandler{
	JSON: interpreter.NewFaultyResponder(errorFormats, statusCodeMapping),
}.Handle


var errorFormats = interpreter.NewFormatSwitch(
	[]interpreter.FormatTemplate{
 		{
			MustKeys: nil,
			Template: func() interpreter.ErrorDescriptor { return &HubspotError{} },
		},
	}...,
)

var statusCodeMapping = map[int]error{
	http.StatusBadGateway:      common.ErrLimitExceeded,
        ....
}
```

providers/hubspot/record.go
```diff
@@ -0,0 +1,55 @@
...
+	"line_item": "line_items",
+	"product":   "products",
+}
+
+var ErrUnsupportedObject = errors.New("unsupported object")
+
+func (c *Connector) GetRecord(ctx context.Context, objectName string, recordId string) (*common.ReadResultRow, error) {
+	objectNameInPath, ok := getRecordSupportedObjectsToPathMap[objectName]
+	if !ok {
+		return nil, fmt.Errorf("%w: %s", ErrUnsupportedObject, objectName)
```
Cobalt0s, line 37: I am not sure if the common error fits in here:
```
common.ErrOperationNotSupportedForObject
```

providers/hubspot/record.go
```diff
@@ -0,0 +1,55 @@
...
+   https://developers.hubspot.com/beta-docs/reference/api/crm/objects/contacts
+   https://developers.hubspot.com/beta-docs/reference/api/crm/objects/deals
+   https://developers.hubspot.com/beta-docs/reference/api/crm/objects/tickets
+   https://developers.hubspot.com/beta-docs/reference/api/crm/objects/line_items
+   https://developers.hubspot.com/beta-docs/reference/api/crm/objects/products
+*/
+
+//nolint:gochecknoglobals
+var getRecordSupportedObjectsToPathMap = map[string]string{
+	"company":   "companies",
```
Cobalt0s, line 24: Instead of a map this can be a set and URL part could be created using `naming.NewPluralString(objectName).String()`.

What about the variable name: `supportedObjectsByWebhook`.

providers/hubspot/record.go
```diff
@@ -0,0 +1,55 @@
...
+	"contact":   "contacts",
+	"deal":      "deals",
+	"ticket":    "tickets",
+	"line_item": "line_items",
+	"product":   "products",
+}
+
+var ErrUnsupportedObject = errors.New("unsupported object")
+
+func (c *Connector) GetRecord(ctx context.Context, objectName string, recordId string) (*common.ReadResultRow, error) {
```
Cobalt0s, line 34: What is the meaning of `GetRecord` method?

When should we use `Search` and when `GetRecord`. Is `GetRecord` more general or relevant in the context of Webhook?

test/hubspot/record/main.go
```diff
@@ -0,0 +1,80 @@
...
+		utils.Fail("error writing to hubspot", "error", err)
+	}
+
+	propMsg := hubspot.WebhookMessage{}
+
+	if err := json.Unmarshal([]byte(samplePropertyChange), &propMsg); err != nil {
+		utils.Fail("error unmarshalling property change message", "error", err)
+	}
+
+	recordId, err := strconv.Atoi(writeresult.RecordId)
```
Cobalt0s, line 67: I see that Identifier is converted from string to int to later inside `GetWebhookResultFromWebhookMessage` to be converted back to streing. Should we drop the int format?

Cobalt0s (commented):
Left some comments.

providers/hubspot/record.go
jlimatampersand: actually I looked into the error, and the error test wasn't clearly fitting. GetRecord is not much tied to `operation`. But I changed the text.

providers/hubspot/record.go
Cobalt0s: Ok, I will think of it as "Singular Read" on demand

providers/hubspot/record.go
Cobalt0s: The `handy` package (soon `datautils`) has: 
```
supported := handy.NewSet("calls", "orders", "customers")
if !supported.Has("orders") {
 return Err..
}
```
But that is syntactic sugar underneath it is `map[string]struct{}`.

providers/hubspot/record.go
Cobalt0s: Yeah the error message is a bit distant. But if we are planning to have `GetRecord` it would be nice to have an error in `common` package to use the same language among connectors

providers/hubspot/record.go
```diff
@@ -0,0 +1,57 @@
...
+	"contact":   true,
+	"deal":      true,
+	"ticket":    true,
+	"line_item": true,
+	"product":   true,
+}
+
+var errGerRecordNotSupportedForObject = errors.New("getRecord is not supproted for the object")
+
+func (c *Connector) GetRecord(ctx context.Context, objectName string, recordId string) (*common.ReadResultRow, error) {
```
laurenzlong, line 35: Can you please add a comment to explain what this function does?

providers/hubspot/webhook.go
laurenzlong, line 25: What's the rationale for introducing a WebhookResult type? It seems a bit unusual that the function below simply puts the `msg` parameter inside of `WebhookMessage`, since the caller of this method will already have `msg`. It would be simpler if the function just returned `*common.ReadResultRow`. Also `WebhookMessage` will look different for each provider, whereas `common.ReadResultRow` will look the same across providers, so by returning only `*common.ReadResultRow`, we can more easily make this method into an interface later when we support subscribe actions for other providers.

Also the function name is too verbose, maybe something like `GetRecordFromWebhook`?

So then the signature becomes:

```
func (c *Connector) GetRecordFromWebhook(
  ctx context.Context, msg *WebhookMessage,
 )(*common.ReadResultRow, error)
```

laurenzlong (changes requested):

PR #1118, opened 2024-11-05
[CON-210] feat: Add Kit ListMetadata connector

sanjay-a01:
### Response

image


providers/kit/metadata.go
```diff
@@ -0,0 +1,90 @@
...
+	for _, obj := range objectNames {
+		// Constructing the request url.
+		url, err := c.getApiURL(obj)
+		if err != nil {
+			return nil, err
+		}
+
+		url.WithQueryParam(perPage, metadataPageSize)
+
+		resp, err := c.Client.Get(ctx, url.String())
```
Cobalt0s, line 38: This approach assumes that every resource returns an array with at least one object.
I don't think that this is a reliable source of Metadata fields.

I tried to search for OpenAPI spec and it is not there, even postman collection linked from their documentation is gone.
The easiest approach would be going over [kit API documentation](https://developers.kit.com/v4#kit-api-accounts) and copying example response to produce the static `schema.json` file manually. There is a possiblity for scrapper if it is easy but I think manual scrapping would be enough.

I think it is important to hear opinion of other reviewers on this matter: @RajatPawar @eberle1080

eberle1080 (approved):

eberle1080 (approved):

providers/kit/metadata.go
sanjay-a01: @eberle1080 Do you have any suggestion for this or can I mark it as resolved conversation

Cobalt0s: 1. I was working on one connector a while ago, where I noticed an interesting behaviour with some objects. The `test` script started failing for some. And I created this method with a comment: https://github.com/amp-labs/connectors/blob/811c1332919f0782360cdad2dd0414ee3377bb67/test/utils/mockutils/metadataResult.go#L46-L50

**Rule:** `Fields(from read) ⊆ Fields(from metadata)`

Therefore, some fields may not exist at a time of making ListObjectMetadata call. But after API writes as a side effect it will return different, new fields. It seems that OpenAPI (given it is up to date) returns largest possible set of fields.

2. Another point is `dynamicscrm` had an issue when the read response had fields which couldn't be queried and Read would error out.

**Rule:** `Fields(for read query-params) ⊆ Fields(from read resp)`

Therefore, in rare cases not all response fields can be used to construct API call, meaning ListObjectMetadata must return only "safe" fields.

Cobalt0s (approved):

providers/kit/metadata_test.go
```diff
@@ -0,0 +1,224 @@
...
+	"github.com/amp-labs/connectors/test/utils/mockutils/mockserver"
+	"github.com/amp-labs/connectors/test/utils/testroutines"
+	"github.com/amp-labs/connectors/test/utils/testutils"
+)
+
+func TestListObjectMetadata(t *testing.T) { // nolint:funlen,gocognit,cyclop
+	t.Parallel()
+
+	broadcastsresponse := testutils.DataFromFile(t, "broadcasts.json")
+	Customfieldsresponse := testutils.DataFromFile(t, "custom_fields.json")
```
Cobalt0s, line 21: lower camel case prefered.
For consistency it should be at least `customfieldsresponse`

providers/kit/metadata_test.go
```diff
@@ -0,0 +1,224 @@
...
+							"published_at":      "published_at",
+							"send_at":           "send_at",
+							"subject":           "subject",
+							"subscriber_filter": "subscriber_filter",
+							"thumbnail_alt":     "thumbnail_alt",
+							"thumbnail_url":     "thumbnail_url",
+						},
+					},
+					"custom_fields": {
+						DisplayName: "custom_fields",
```
Cobalt0s, line 99: The display name should be:
`Custom Fields`
The above broadcasts would be:
`Broadcasts`.

I don't see the usage of OpenAPI tool, where `api3.WithDisplayNamePostProcessors` option would help with formatting display names, hence I am thinking you created static script yourself. There are some connectors that are like that, could you then edit `schema.json` to have display names as described? 

These display names are what clients will be seeing in UI.

Cobalt0s (approved):
Please, change the display names for objects.
(The field display names are good as is, they are usually very "programmatic" for +90% of connectors)

Cobalt0s (approved):

eberle1080 (approved):

PR #1124, opened 2024-11-07
feat: Keap (Read)

Cobalt0s:
# Notes

Some error response from provider are either in JSON or HTML format. 
`schema.json` produced from OpenAPI is added illustrating the support for read objects.

Depending on the Module the Read is affected such that `limit` == `page_size` query parameter. Next page is located under either "next" or "next_page_token".
Since parameter is available for V1 only.

# Test Results
Listing emails:
image

### DIFF
--- a/providers/keap/connector.go
+++ b/providers/keap/connector.go
@@ -1,0 +1,75 @@
+package keap
+
+import (
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/interpreter"
+	"github.com/amp-labs/connectors/common/paramsbuilder"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+	"github.com/amp-labs/connectors/providers"
+	"github.com/amp-labs/connectors/providers/keap/metadata"
+)
+
+const ApiPathPrefix = "crm/rest"
+
+type Connector struct {
+	BaseURL string
+	Client  *common.JSONHTTPClient
+	Module  common.Module
+}
+
+func NewConnector(opts ...Option) (conn *Connector, outErr error) {
+	defer common.PanicRecovery(func(cause error) {
+		outErr = cause
+		conn = nil
+	})
+
+	params, err := paramsbuilder.Apply(parameters{}, opts, WithModule(ModuleV1))
+	if err != nil {
+		return nil, err
+	}
+
+	httpClient := params.Client.Caller
+	conn = &Connector{
+		Client: &common.JSONHTTPClient{
+			HTTPClient: httpClient,
+		},
+		Module: params.Selection,
+	}
+
+	// Read provider info
+	providerInfo, err := providers.ReadInfo(conn.Provider())
+	if err != nil {
+		return nil, err
+	}
+
+	// connector and its client must mirror base url and provide its own error parser
+	conn.setBaseURL(providerInfo.BaseURL)
+	conn.Client.HTTPClient.ErrorHandler = interpreter.ErrorHandler{
+		JSON: interpreter.NewFaultyResponder(errorFormats, nil),
+		HTML: &interpreter.DirectFaultyResponder{Callback: conn.interpretHTMLError},
+	}.Handle
+
+	return conn, nil
+}
+
+func (c *Connector) getURL(objectName string) (*urlbuilder.URL, error) {
+	path, err := metadata.Schemas.LookupURLPath(c.Module.ID, objectName)
+	if err != nil {
+		return nil, err
+	}
+
+	return urlbuilder.New(c.BaseURL, ApiPathPrefix, path)
+}
+
+func (c *Connector) setBaseURL(newURL string) {
+	c.BaseURL = newURL
+	c.Client.HTTPClient.Base = newURL
+}
+
+func (c *Connector) Provider() providers.Provider {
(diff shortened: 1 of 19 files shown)


Cobalt0s:
* **#1124** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1124?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1125](https://github.com/amp-labs/connectors/pull/1125) <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1125?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1126](https://github.com/amp-labs/connectors/pull/1126) <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1126?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈
* `main`

<h2></h2>

providers/keap/connector.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/interpreter"
+	"github.com/amp-labs/connectors/common/paramsbuilder"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+	"github.com/amp-labs/connectors/providers"
+	"github.com/amp-labs/connectors/providers/keap/metadata"
+)
+
+const ApiPathPrefix = "crm/rest"
```
RajatPawar, line 12: Is this the same for all modules and the entire API? Are there other products that have differing API prefixes?

providers/keap/errors.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func (r ResponseError) CombineErr(base error) error {
+	return fmt.Errorf("%w: %v", base, r.Detail)
+}
+
+func (c *Connector) interpretHTMLError(res *http.Response, body []byte) error {
+	base := interpreter.DefaultStatusCodeMappingToErr(res, body)
+
+	document, err := goquery.NewDocumentFromReader(bytes.NewReader(body))
```
RajatPawar, line 33: Can you add some references here to the kind of errors keap sends?

providers/keap/objectNames.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/amp-labs/connectors/internal/datautils"
+	"github.com/amp-labs/connectors/providers/keap/metadata"
+)
+
+// Supported object names can be found under schemas.json.
+var supportedObjectsByRead = metadata.Schemas.ObjectNames() //nolint:gochecknoglobals
+
+// ObjectNameToResponseField maps ObjectName to the response field name which contains that object.
+var ObjectNameToResponseField = datautils.NewDefaultMap(map[string]string{ //nolint:gochecknoglobals
+	"merchants":       "merchant_accounts",
```
RajatPawar, line 13: Would this be handled by the `locatedAt` PR?

test/keap/read/main.go
```diff
@@ -0,0 +1,42 @@
...
+		Fields:     connectors.Fields("id", "subject", "sent_from_address"),
+	})
+	if err != nil {
+		utils.Fail("error reading from microsoft CRM", "error", err)
+	}
+
+	fmt.Println("Reading emails..")
+	utils.DumpJSON(res, os.Stdout)
+
+	if res.Rows > keap.DefaultPageSize {
```
RajatPawar, line 39: Is this something you expect to happen?

RajatPawar (approved):

providers/keap/connector.go
Cobalt0s: So while working on metadata PR, I dicsovered that I covered one of the two of OpenAPI files. First is V1 and the other much smaller is V2. I revisited the read from this persective and added some changes, in particular 2 modules.

To answer above question, yes, both share `crm/rest`.
Ex:
`crm/rest/v1` - first module
`crm/rest/v2` - second module

providers/keap/errors.go
Cobalt0s: I added the comment referncing test directory for samples and describing the nature of HTML format for Keap.

providers/keap/objectNames.go
Cobalt0s: I moved it to Metadata PR. The answer is yes and no ahaha. Yes, since now on it will be served from the file which conveniently summarizes everything about that object. But this map will live inside the OpenAPI script which produces the file.

Schema.json in this way is a documentation of what is happenning by deep connector implementation. We still need to tell the tool the "exceptions", but it automatically figures out responses with array on top JSON level. If it is an object we need to tell where do we expect the array to be located at. (hence the old name `locatedAt`).

test/keap/read/main.go
Cobalt0s: I keep up with the established purpose of the `test` directory which holds test. There is no good assert for reading and thats what we have for 7 deep connectors.
image

Honestly, test scripts are rather Read/Write/ListObjectMetadata examples or just plain scripts using combination of those methods to demonstrate functionality.

Cobalt0s:
@RajatPawar there have been some important changes, could you take a look at the code again. You reviewed Read focused on module V1. I have added small support for V2.

Tests capture read differences for pagination.

PR #1147, opened 2024-11-15
[ENG-1637] Parse webhook types from message for hubspot

jlimatampersand:
Added webhook types including create,delete,update, pathThrough

added hubspot webhook type parsing and unit test.

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -247,3 +247,12 @@ type PostAuthInfo struct {
 	CatalogVars *map[string]string
 	RawResponse *JSONHTTPResponse
 }
+
+type WebhookEventType string
+
+const (
+	WebhookEventTypeCreate      WebhookEventType = "create"
+	WebhookEventTypeUpdate      WebhookEventType = "update"
+	WebhookEventTypeDelete      WebhookEventType = "delete"
+	WebhookEventtypePassThrough WebhookEventType = "passThrough"
+)
--- a/common/types.go
+++ b/common/types.go
@@ -251,8 +251,8 @@ type PostAuthInfo struct {
 type WebhookEventType string
 
 const (
-	WebhookEventTypeCreate      WebhookEventType = "create"
-	WebhookEventTypeUpdate      WebhookEventType = "update"
-	WebhookEventTypeDelete      WebhookEventType = "delete"
-	WebhookEventtypePassThrough WebhookEventType = "passThrough"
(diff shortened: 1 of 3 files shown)


providers/hubspot/webhook.go
```diff
@@ -40,6 +40,24 @@ func (c *Connector) GetRecordFromWebhookMessage(
 	return c.GetRecord(ctx, objectName, recordId)
 }
 
+func (c *Connector) ExtractEventTypeFromWebhookMessage(msg *WebhookMessage) (common.WebhookEventType, error) {
+	parts := strings.Split(msg.SubscriptionType, ".")
+	if !getRecordSupportedObjectsSet.Has(parts[0]) {
```
laurenzlong, line 45: Please remove this check from all the Extract functions in this file, just because we cannot get a record for that particular object, we should not error out. We can just leave the `result` blank when we send the webhook to the builder. We want people to be able to use Subscribe Actions for ALL Hubspot event types and not unnecessarily limit them.

laurenzlong (approved):

providers/hubspot/webhook_test.go
```diff
@@ -66,3 +67,107 @@ func TestWebhook(t *testing.T) {
...
+	}
+
+	conn := &Connector{}
+
+	evtTypeCreate, err := conn.ExtractEventTypeFromWebhookMessage(createMessage)
+	if err != nil {
+		t.Fatalf("error extracting object name from webhook message: %s", err)
+	}
+
+	assert.Equal(t, evtTypeCreate, common.WebhookEventTypeCreate, "object type should be parsedCorrectly")
```
laurenzlong, line 96: Please update the message to "object type should be create", same for the other ones in this file.

common/types.go
```diff
@@ -247,3 +247,12 @@ type PostAuthInfo struct {
...
 	RawResponse *JSONHTTPResponse
 }
+
+type WebhookEventType string
+
+const (
+	WebhookEventTypeCreate      WebhookEventType = "create"
+	WebhookEventTypeUpdate      WebhookEventType = "update"
+	WebhookEventTypeDelete      WebhookEventType = "delete"
+	WebhookEventtypePassThrough WebhookEventType = "passThrough"
```
laurenzlong, line 257: Can you name this `other`? Passthrough is not as user friendly.

PR #1135, opened 2024-11-14
feat(scripts): PKCE for AuthCode grant type

Cobalt0s:
# Description

Modified `token` script to generate authentication token using AuthCode grant type with PKCE. 
I am able to generate auth token for Klaviyo. Additionally checked ConstantContact which for backwards compatability, and the script works as expected.

For `proxy` the script was modified to allow requests through without the proper refresh strategy. For local testing I just recreate access token manually. For deep Klaviyo connector implementation  it is sufficient.

PR #1134, opened 2024-11-12
feat(jsonquery): Soft string requirement

Cobalt0s:
# Description

Identifiers in provider responses may be of either string or integer types. This adds a utility that surfaces that value as string.

The strict attitude of what JSON data type we expect is not relevant. Instead with a new method we are asking give me data as "Text" regardless of what it is.

PR #1097, opened 2024-10-23
feat: CustomerJourneysApp (Write)

Cobalt0s:
# Description

There is a difference in support among objects between Create/Update/Delete. 
2 objects are made up mapping to URL path:
`customer_exports`  refers to "/exports/customers";
`deliveries_exports` refers to "/exports/deliveries".

Response key holding record of interest differes among objects and such exceptions are also recorded, for more refer to `objectNames.go`.

# Test results

image

PR #1150, opened 2024-11-20
[CON-186] Keap - read with custom fields

Cobalt0s:
# Description

Include custom fields as part of `Read` response alongside the native fields.

[How to create/update/delete custom fields?](https://help.keap.com/help/create-custom-fields) This can be done via dashboard. I noticed that update and delete are absent in REST APIs, therefore managing custom fields via dashboard is preferred.

# Review

Check the `README.md` for how the read output is effected, as well as tests.

### DIFF
--- a/providers/keap/README.md
+++ b/providers/keap/README.md
@@ -1,0 +1,62 @@
+# Custom Fields
+
+The `Read` operation will return custom fields alongside other native properties.
+Keap API indexes custom fields using numbered identifiers without including human-readable names.
+This issue is addressed in the implementation.
+
+
+## Example
+```json
+{
+  "id": 22,
+  "given_name": "Erica",
+  "jobtitle": "Product Owner" // custom field
+} 
+```
+
+
+## Explanation
+[List Contacts](https://developer.keap.com/docs/rest/#tag/Contact/operation/listContactsUsingGET)
+When requesting the contacts resource, the response includes custom fields as follows:
+```json
+{
+  "contacts": [
+    {
+      "id": 22,
+      "given_name": "Erica",
+      "custom_fields": [
+        {
+          "id": 6, // Obscure field name.
+          "content": "Product Owner"
+        },
+      ]
+    }
+  ]
+}
+```
+[Contacts Model](https://developer.keap.com/docs/rest/#tag/Contact/operation/retrieveContactModelUsingGET)
+To determine the meaning of `"id": 6`, you can query the contacts model, which provides additional details:
+```json
+{
+  "custom_fields": [
+    {
+      "id": 6,
+      "label": "title",
+      "options": [],
+      "record_type": "CONTACT",
+      "field_type": "Text",
+      "field_name": "jobtitle", // Prefered field name.
+      "default_value": null
+    },
+  ],
+  "optional_properties": []
+}
+```
+Combining both results, the connector's `Read` will return the following:
+```json
+{
+  "id": 22,
+  "given_name": "Erica",
+  "jobtitle": "Product Owner"
+} 
+```
--- a/providers/keap/README.md
+++ b/providers/keap/README.md
@@ -5,7 +5,7 @@ Keap API indexes custom fields using numbered identifiers without including huma
 This issue is addressed in the implementation.
 
 
-## Example
+## Original record
 ```json
 {
   "id": 22,
@@ -15,7 +15,7 @@ This issue is addressed in the implementation.
 ```
 
 
-## Explanation
+## Normal response
 [List Contacts](https://developer.keap.com/docs/rest/#tag/Contact/operation/listContactsUsingGET)
 When requesting the contacts resource, the response includes custom fields as follows:
 ```json
--- a/providers/keap/parse.go
+++ b/providers/keap/parse.go
@@ -18,7 +18,8 @@ func makeNextRecordsURL(moduleID common.ModuleID) common.NextPageFunc {
 	}
 }
 
-// Before parsing the records if applicable API will be called one time to get model describing custom fields.
+// Before parsing the records, if any custom fields are present (without a human-readable name),
+// this will call the correct API to extend & replace the custom field with human-readable information.
 // Object will then be enhanced using model.
 func (c *Connector) parseReadRecords(
 	ctx context.Context, config common.ReadParams, jsonPath string,
@@ -34,18 +35,11 @@ func (c *Connector) parseReadRecords(
 			return nil, err
 		}
 
-		result := make([]map[string]any, len(arr))
-
-		for index, object := range arr {
-			item, err := c.enhanceObjectWithCustomFieldNames(object, customFields)
-			if err != nil {
-				return nil, err
-			}
-
-			result[index] = item
+		if len(customFields) == 0 {
+			return jsonquery.Convertor.ArrayToMap(arr)
 		}
 
-		return result, nil
+		return enhanceObjectsWithCustomFieldNames(arr, customFields)
 	}
 }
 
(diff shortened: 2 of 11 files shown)


Cobalt0s:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1150?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1151** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1151?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1150** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1150?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1150?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1125** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1125?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1124** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1124?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1126](https://github.com/amp-labs/connectors/pull/1126) <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1126?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1134** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1134?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2 other dependent PRs ([#1097](https://github.com/amp-labs/connectors/pull/1097) <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1097?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1135](https://github.com/amp-labs/connectors/pull/1135) <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1135?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`

<h2></h2>

providers/keap/README.md
```diff
@@ -0,0 +1,62 @@
+# Custom Fields
+
+The `Read` operation will return custom fields alongside other native properties.
+Keap API indexes custom fields using numbered identifiers without including human-readable names.
+This issue is addressed in the implementation.
+
+
+## Example
```
RajatPawar, line 8:
```suggestion
## Original record
```

providers/keap/README.md
```diff
@@ -0,0 +1,62 @@
...
+```json
+{
+  "id": 22,
+  "given_name": "Erica",
+  "jobtitle": "Product Owner" // custom field
+} 
+```
+
+
+## Explanation
```
RajatPawar, line 18:
```suggestion
## Normal response
```

providers/keap/parse.go
```diff
@@ -15,3 +17,70 @@ func makeNextRecordsURL(moduleID common.ModuleID) common.NextPageFunc {
 		return jsonquery.New(node).StrWithDefault("next_page_token", "")
 	}
 }
+
+// Before parsing the records if applicable API will be called one time to get model describing custom fields.
```
RajatPawar, line 21:
```suggestion
// Before parsing the records, if any custom fields are present (without a human readable name), this will call the correct API to extend & replace the custom field with human readable information.
```

providers/keap/parse.go
```diff
@@ -15,3 +17,70 @@ func makeNextRecordsURL(moduleID common.ModuleID) common.NextPageFunc {
...
+func (c *Connector) parseReadRecords(
+	ctx context.Context, config common.ReadParams, jsonPath string,
+) common.RecordsFunc {
+	return func(node *ajson.Node) ([]map[string]any, error) {
+		arr, err := jsonquery.New(node).Array(jsonPath, true)
+		if err != nil {
+			return nil, err
+		}
+
+		customFields, err := c.requestCustomFields(ctx, config.ObjectName)
```
RajatPawar, line 32: What happens when there are no custom fields on an object that we expect to have?

providers/keap/parse.go
```diff
@@ -15,3 +17,70 @@ func makeNextRecordsURL(moduleID common.ModuleID) common.NextPageFunc {
...
+	}
+
+	// Replace identifiers with human-readable field names which were found by making a call to "/model".
+	for _, field := range customFieldsResponse.CustomFields {
+		if model, ok := fields[field.ID]; ok {
+			object[model.FieldName] = field.Content
+		}
+	}
+
+	delete(object, "custom_fields")
```
RajatPawar, line 83: Can you keep `custom_fields`? If a builder is going by the API docs, I wouldn't want them to be confused about why there's missing information in the raw object. As a rule of thumb, please add information, not remove / replace

providers/keap/parse.go
RajatPawar, line 32: If this process (of enhancing the custom fields) fails, then the whole read operation fails. Can you make this optional? i.e. If we are able to add in human-readable names to the data, great, else, we should just log it and move on instead of completely failing

RajatPawar (approved):
Couple of comments, please resolve those in code before merging

providers/keap/parse.go
Cobalt0s: Sure. Other connectors have flattening of nested objects, and in those cases I had removed the nested object. Recent example would be Klaviyo:
```
	    {
	        "type": "",
	        "id": "",
	        "attributes": {
	            "test_account": false,
	            "contact_information": {},
	            "locale": ""
	        },
	        "links": {}
	    }
	],
```
Attributes object removed and it's properties moved to parent.
Should those cases be addressed as conceptually it is similar?

providers/keap/parse.go
Cobalt0s: Since, this is not so apparent I will change the code to be more explicit.

https://github.com/amp-labs/connectors/blob/4bdd68dcf7450bdb6c4716d4ba9888a10c621e53/providers/keap/parse.go#L67

Cobalt0s: This one is interconnected with the ListObjectMetadata PR.

Few impressions/ideas I have at the moment:
* ListObjectMetadata returns fields that client uses to select :ballot_box_with_check: to configure reading objects. If unlucky the UI will display some fields. Assumption: This doesn't happen often.
* In case of incomplete success it would be misleading to only return custom_fields without inlined property. This will blur why particular custom field is missing, because of actual null or because of hidden error that we only know of. Assumption: clients rely on `ReadResultRow.Fields`.

Let me know, if my judgment is relevant. I had a feeling that it counts as an error during development. But now I am not sure.

providers/keap/parse.go
RajatPawar: Since this object is what we also send to the user as the 'Raw' object later on, I'd want to have at least a copy of the original untouched response. For Klaviyo, IMO we should still have the attributes object to maintain the original payload signature, but I'll let @laurenzlong take the call since this is builder facing

providers/keap/parse.go
RajatPawar: cc @laurenzlong

RajatPawar (approved):
@Cobalt0s for now let's go ahead with - 
1. Flattened custom fields in the `Fields`
2. Raw can be as it is, or with flattened custom fields depending on what you already have, but please make sure that the original response is not deleted

PR #1159, opened 2024-11-25
[ENG-1652] Implement method to extract id from raw record

jlimatampersand:
This pr creates connector method to extract id from a raw hubspot record. This will be used for subscribe action.

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -170,7 +170,7 @@ type ReadResultRow struct {
 	Fields map[string]any `json:"fields"`
 	// Raw is the raw JSON response from the provider.
 	Raw map[string]any `json:"raw"`
-	// RecordId is the ID of the record. Currently only used for hubspot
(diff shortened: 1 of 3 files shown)


common/types.go
```diff
@@ -170,6 +170,8 @@ type ReadResultRow struct {
 	Fields map[string]any `json:"fields"`
 	// Raw is the raw JSON response from the provider.
 	Raw map[string]any `json:"raw"`
+	// RecordId is the ID of the record. Currently only used for hubspot
```
laurenzlong, line 173:
```suggestion
	// RecordId is the ID of the record. Currently only populated for hubspot GetRecord and GetRecordsWithId function
```

laurenzlong (approved):

common/types.go
laurenzlong: also please add functionality to GetRecord function too.

PR #1169, opened 2024-11-27
feat: Klaviyo flatten WriteParams.RecordData

Cobalt0s:
# Changes
Predefined fields are kept on the top JSON node level. Unknown fields are moved to nested "attributes" test.
Unit test for this method was added. Input represents what should be supplied to the connector while the output is what will be send to provider APIs. Examples are taken from API docs.

# Test Results
After updating "Tag" payload for the script the test still passes:
image


Cobalt0s:
* **#1169** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1169?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1169?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

test/klaviyo/write-delete/main.go
```diff
@@ -69,13 +56,9 @@ func main() {
...
 	updateTag(ctx, conn, tagID, &TagPayload{
-		Tag: Tag{
-			ID:   tagID,
-			Type: "tag",
-			Attributes: TagAttributes{
-				Name: newName,
-			},
-		},
+		ID:   tagID,
+		Type: "tag",
```
laurenzlong, line 60: We should not require the user to provide `Type` either. Since the object name is already captured in WriteParams

Also "ID" should be "Id", see https://docs.withampersand.com/define-integrations/write-actions#update-an-existing-record for the Write API request that customers make.

laurenzlong: I'm conditionally approving this, but please revise the code to be able to accept this as the TagPayload:

```
&TagPayload{
   Id: tagID,
   Name: newName,
 }
```

laurenzlong (approved):

test/klaviyo/write-delete/main.go
Cobalt0s: Ok, it took some time to verify all exceptions. I added a new mapping between ObjectName and the Type property for Write payload.

The payload is even better, it is just 
```
&TagPayload{
   Name: newName,
 }
```
RecordId will be moved to the payload for all update cases automatically. This way payload holds essentials and whatever the logic in the server to populate WriteParams.RecordId will end up as "id" property in the payload an in the URL path.

PR #1177, opened 2024-12-02
[CON-603] feat: enable Lemlist proxy

Jkarage:
### GET

Screenshot 2024-12-02 at 16 33 28

### POST
Screenshot 2024-12-02 at 15 34 23

### PATCH
Screenshot 2024-12-02 at 15 35 41

### DELETE
Screenshot 2024-12-02 at 15 38 22

### Invalid requests
As observed while testing the API, Invalid request responds with 200 OK
Screenshot 2024-12-02 at 15 49 26

## Pagination
Pagination is only supported in the campaigns resource.
Screenshot 2024-12-02 at 15 37 17

Screenshot 2024-12-02 at 15 37 48

## Successful console operation (operation & events)
Screenshot 2024-12-02 at 15 51 20

## Failed console operation (operation & events)
Screenshot 2024-12-02 at 15 52 33

### DIFF
--- a/providers/lemlist.go
+++ b/providers/lemlist.go
@@ -31,7 +31,7 @@ func init() {
 				Upsert: false,
 				Delete: false,
 			},
-			Proxy:     false,
+			Proxy:     true,
 			Read:      false,
 			Subscribe: false,
 			Write:     false,


RajatPawar (approved):
Can you please mention the 200 error code in the ticket as well? We might have to mention this in our docs. Also - if we use the API directly, does it still return a 200 for errors?

Jkarage:
Added it in ticket. Both the API and the Proxy responds the same. 200 for a not available/unknown resource. 404 for an available resource but no such record. I have added the behaviours in the slab as well as in the ticket.

PR #1180, opened 2024-12-02
[ENG-1637] Reimplement Hubspot connector's webhook parsing logic into webhook methods. 

jlimatampersand:


jlimatampersand:
@laurenzlong After changing the webhook url to include integrationId instead of installationId, this change was needed because we need to parse webhook first in order to get connection/connector. Also makes sense for those messages to to have it's own method when there is no connector needed to communicate with provider while parsing messeages.

common/types.go
```diff
@@ -266,3 +266,10 @@ const (
 	WebhookEventTypeDelete WebhookEventType = "delete"
 	WebhookEventTypeOther  WebhookEventType = "other"
 )
+
+type WebhookMessage interface {
```
laurenzlong, line 270: Can you add a comment to explain that this is the webhook that Hubspot sends for a subscription?

laurenzlong (approved):

PR #1186, opened 2024-12-04
fix: Update hubspot.GetResultId to look for ID in the raw object

RajatPawar:
* We added the `Raw` field to `ReadResult` after this function was originally introduced. At the time, it looked for ID inside the fields object
*  This function needs to look for the ID of a record in the `Raw` object, and not the `Fields` object, since `Fields` may or may not have the ID field
* The `Raw` response will always have the ID inside `Raw.properties`

This test should help clarify the required output - https://github.com/amp-labs/connectors/pull/1186/files#diff-24f929d222aa9a04a0837d0f97a1b5c230629c1420ae254486acefc9b10f2a00R61

### DIFF
--- a/providers/hubspot/parse.go
+++ b/providers/hubspot/parse.go
@@ -141,6 +141,7 @@ func getMarshalledData(records []map[string]interface{}, fields []string) ([]com
 }
 
 // GetLastResultId returns the last row's id from a result.
+// nolint:cyclop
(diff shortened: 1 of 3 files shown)


jatinsandilya (approved):
Understand the new logic, approving as a result. Although admittedly I'm a little unfamiliar with this change so trusting your judgement here :)

PR #1189, opened 2024-12-04
feat: Allow a module to override base URL

RajatPawar:
Atlassian connect has a different base URL than the usual atlassian API (read more @ https://developer.atlassian.com/cloud/jira/platform/rest/v3/intro/#authentication), but we need this to function in the same connector. This PR enables the atlassian connector to use a module which overrides the base URL and lets it makes calls to the Connect API. If we like this pattern, we can add it into modules

### DIFF
--- a/providers/atlassian/connector.go
+++ b/providers/atlassian/connector.go
@@ -7,6 +7,7 @@ import (
 	"github.com/amp-labs/connectors/common"
 	"github.com/amp-labs/connectors/common/interpreter"
 	"github.com/amp-labs/connectors/common/paramsbuilder"
+	"github.com/amp-labs/connectors/common/substitutions/catalogreplacer"
 	"github.com/amp-labs/connectors/common/urlbuilder"
 	"github.com/amp-labs/connectors/providers"
 )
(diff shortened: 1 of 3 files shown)


RajatPawar:
* **#1190** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1190?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1189** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1189?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1189?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

laurenzlong (approved):
This works for now, but I think a better long term solution is to have different base URLs for different modules defined in the catalog file itself, rather than doing this override logic, which makes the code harder to debug & understand.

Can you name this module `jira-atlassian-connect` or something? So it's clear that it's scoped to Jira (and the read/write methods will only interact with Jira and not Confluence)

RajatPawar:
> This works for now, but I think a better long term solution is to have different base URLs for different modules defined in the catalog file itself, rather than doing this override logic, which makes the code harder to debug & understand.
> 
> Can you name this module `jira-atlassian-connect` or something? So it's clear that it's scoped to Jira (and the read/write methods will only interact with Jira and not Confluence)

Agreed, I just didn't want to get into thinking about a module's config & its relationship with a provider, so this should be okay for now.

PR #1190, opened 2024-12-04
[ENG-1729] feat: Atlassian JWT client

RajatPawar:
- Adds a JWT client that can be used with Atlassian's Connect module 
- This client, for every request, generates a signed token based on the current time, the actual request and some fixed claim data (`iss` to be exact)
- **Note:** The current connector ONLY supports `issues` though, so this will also only support issues for now. 

I need to do more testing around writes, but reads work just like usual.

## Test
Screenshot 2024-12-05 at 4 27 54 AM

### DIFF
--- a/common/header.go
+++ b/common/header.go
@@ -59,18 +59,21 @@ func WithHeaderUnauthorizedHandler(
 
 // WithDynamicHeaders sets a function that will be called on every request to
 // get additional headers to use. Use this for things like time-based tokens
-// or loading headers from some external authority.
-func WithDynamicHeaders(f func() ([]Header, error)) HeaderAuthClientOption {
+// or loading headers from some external authority. The function can access a
+// copy of the request object to use its metadata for generating headers.
+func WithDynamicHeaders(f DynamicHeadersGenerator) HeaderAuthClientOption {
 	return func(params *headerClientParams) {
 		params.dynamicHeaders = f
 	}
 }
 
+type DynamicHeadersGenerator func(http.Request) ([]Header, error)
+
 // oauthClientParams is the internal configuration for the oauth http client.
 type headerClientParams struct {
 	client         *http.Client
 	headers        []Header
-	dynamicHeaders func() ([]Header, error)
+	dynamicHeaders DynamicHeadersGenerator
 	debug          func(req *http.Request, rsp *http.Response)
 	unauthorized   func(hdrs []Header, req *http.Request, rsp *http.Response) (*http.Response, error)
 }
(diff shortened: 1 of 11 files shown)


RajatPawar:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1190?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1190** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1190?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1190?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1189** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1189?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

common/header.go
```diff
@@ -59,18 +59,21 @@ func WithHeaderUnauthorizedHandler(
...
-func WithDynamicHeaders(f func() ([]Header, error)) HeaderAuthClientOption {
+// or loading headers from some external authority. The function can access a
+// copy of the request object to use its metadata for generating headers.
+func WithDynamicHeaders(f DynamicHeadersGenerator) HeaderAuthClientOption {
 	return func(params *headerClientParams) {
 		params.dynamicHeaders = f
 	}
 }
 
+type DynamicHeadersGenerator func(http.Request) ([]Header, error)
```
eberle1080, line 70: I'd use `*http.Request` rather than `http.Request`, because I'm not sure what copy-by-value is going to do with certain fields (e.g. body).

providers/utils.go
```diff
@@ -253,6 +253,8 @@ func (i *ProviderInfo) NewClient(ctx context.Context, params *NewClientParams) (
 		}
 
 		return createApiKeyHTTPClient(ctx, params.Client, params.Debug, i, params.ApiKey)
+	case Jwt:
+		fallthrough
```
eberle1080, line 257: Seems like you're gonna want to implement this. All of the clients we use in the server side call this function.

eberle1080 (changes requested):
Overall looks good, with one major missing thing: the providers.NewClient implementation. If you don't implement this, our normal client creation flow will break. If there's some reason this won't be the case, let me know. But this seems incomplete otherwise.

RajatPawar:
I haven't added it because Lauren thinks that this is just a one-off, and not actually supporting it as an auth scheme by itself. Do we want to support providers with JWT as their primary authScheme? cc @laurenzlong

common/header.go
RajatPawar: Idea was for anyone to not be able to mutate the request (accidentally or not) - it would create hard-to-debug behavior

providers/utils.go
laurenzlong: This is fine for now, since we'll special case Atlassian in the server. Can you please add a comment to explain this?

eberle1080:
OK she explained it to me. I'm changing my review because I understand the exception.

eberle1080 (approved):

PR #1201, opened 2024-12-06
feat(tests): Common unit test comparators

Cobalt0s:
# Description
Refactored repetitive, duplicated comparison methods for unit tests.

# Review
First commit has all the important changes with an updated README.
Second commit is using these methods for all providers.

# Future considerations
To simplify debugging of test I think it would be nice to add loging within comparators.
This should tell the exact reason of what mistmatched speeding up unit test development.

### DIFF
--- a/common/parse.go
+++ b/common/parse.go
@@ -57,6 +57,12 @@ func ParseResult(
 		nextPage = ""
 	}
 
+	if len(marshaledData) == 0 {
+		// Either a JSON array is empty or it was nil.
+		// For consistency return nil for missing records.
+		marshaledData = nil
+	}
+
 	return &ReadResult{
 		Rows:     int64(len(marshaledData)),
 		Data:     marshaledData,
--- a/providers/atlassian/metadata_test.go
+++ b/providers/atlassian/metadata_test.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/amp-labs/connectors"
 	"github.com/amp-labs/connectors/common"
-	"github.com/amp-labs/connectors/test/utils/mockutils"
 	"github.com/amp-labs/connectors/test/utils/mockutils/mockserver"
 	"github.com/amp-labs/connectors/test/utils/testroutines"
 	"github.com/amp-labs/connectors/test/utils/testutils"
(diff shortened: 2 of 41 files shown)


Cobalt0s:
* **#1201** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1201?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1201?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

common/parse.go
```diff
@@ -57,6 +57,12 @@ func ParseResult(
 		nextPage = ""
 	}
 
+	if len(marshaledData) == 0 {
+		// Either a JSON array is empty or it was nil.
+		// For consistency return nil for missing records.
+		marshaledData = nil
```
Jkarage, line 63: Why not []ReadResultRow{}, I believe most providers will return [].
Also are GetRecordsUnderJSONPath and GetOptionalRecordsUnderJSONPath not enough, as they take care of this.

Cobalt0s: `GetRecordsUnderJSONPath` and `GetOptionalRecordsUnderJSONPath` will return either `[]` or `nil` and if excpectation is not met it would return an error. I would say these utilities are a bit low level. Maybe we should modify them to return empty initialized slice. Do you think this abstraction should rather live under `GetRecordsUnderJSONPath`?

We could return `[]ReadResultRow{}` in either case. I chose nil because it is a "default" golang value for a slice. Creating tests would require less fields.

From end user perspective: 
* looping over a nil works technically the same as empty slice
* there is one case, if user will try to modify and append a nil slice it would panic, but it should be a read only.

Jkarage: I think there is no need to move it there, Here should be fine in my opinion, if someone uses another function for parsing records for any reason, he shouldn't re-write this. My thinking was these function will only return a nil response when there is an error, else it will be a list whether empty or with data. That's the reason I prefered [] rather nil.

Cobalt0s: I decided to read some things about it, I would like to share:
https://stackoverflow.com/a/44305910/9628238 - check `Under the hood` section, very interesting.
https://gosamples.dev/empty-vs-nil-slice/ - article also describes behaviour. They don't mention the spread operator, but spreading nil slice works as "expected" and doesn't panic.

It seems that [] and nil are almost the same. Declaring nil is easier, thats what I found in unit tests (I even don't need to include the struct property, it is already nil).

PR #1206, opened 2024-12-10
feat: (apollo) update sequences object support

Jkarage:
This adds
-  support for reading sequence (emailer_campaigns) object in apollo. It uses reading by search but does not have the 50K display limitations. So this can currently be supported.

- reading the tasks object, can now be supported by the Search method.

Testing reading the `emailer_campaigns`
Screenshot 2024-12-13 at 21 27 13

### DIFF
--- a/providers/apollo/types.go
+++ b/providers/apollo/types.go
@@ -11,16 +11,14 @@ var (
 
 type ObjectType string
 
-// postSearchObjects represents the objects that uses the searching endpoint,
-// POST method for requesting records.
+// postSearchObjects represents the objects that uses the searching endpoint,POST method.
 var postSearchObjects = []ObjectType{ //nolint:gochecknoglobals
 	"mixed_people", "mixed_companies", "contacts",
-	"accounts", "emails_campaigns",
+	"accounts", "tasks",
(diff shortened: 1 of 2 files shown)


providers/apollo/types.go
```diff
@@ -11,16 +11,14 @@ var (
 
 type ObjectType string
 
-// postSearchObjects represents the objects that uses the searching endpoint,
-// POST method for requesting records.
+// postSearchObjects represents the objects that uses the searching endpoint,POST method.
 var postSearchObjects = []ObjectType{ //nolint:gochecknoglobals
 	"mixed_people", "mixed_companies", "contacts",
-	"accounts", "emails_campaigns",
```
RajatPawar, line 18: Are we unable to support `emails_campaigns` anymore? Or was this accidentally removed?

Jkarage: We can, I have moved this to the `getSearchObjects` Because it does not have the limitations as the others in here. Using get with it works fine. The name is now `emailer_campaigns` or I may had it as a typo before. But it's now `emailer_campaigns` from their docs.

PR #1210, opened 2024-12-12
[ENG-1707] Hubspot connector method to verify webhook. 

jlimatampersand:
Webhooks from providers need to be verified. This PR creates types for the verification params, and defines method for Hubspot to verify webhook.

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
 	"time"
 
 	"github.com/amp-labs/connectors/internal/datautils"
@@ -274,4 +275,13 @@ type SubscriptionEvent interface {
 	RawEventName() (string, error)
 	ObjectName() (string, error)
 	Workspace() (string, error)
+}
+
+// WebhookVerificationParameters is a struct that contains the parameters required to verify a webhook.
+type WebhookVerificationParameters struct {
+	Headers      http.Header
+	Body         []byte
+	URL          string
+	ClientSecret string
+	Method       string
 }
--- a/connectors.go
+++ b/connectors.go
@@ -86,7 +86,7 @@ type AuthMetadataConnector interface {
 	GetPostAuthInfo(ctx context.Context) (*common.PostAuthInfo, error)
 }
 
-type WebhookVerifiableConnector interface {
+type WebhookVerifierConnector interface {
 	Connector
 
 	// VerifyWebhookMessage verifies the signature of a webhook message.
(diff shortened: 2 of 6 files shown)


providers/hubspot/subscriptionEvent.go
```diff
@@ -40,6 +43,28 @@ func (c *Connector) GetRecordFromSubscriptionEvent(
 	return c.GetRecord(ctx, objectName, recordId)
 }
 
+// GetRecordFromSubscriptionEvent fetches a record from the Hubspot API using the data from a subscription event.
```
RajatPawar, line 46: The comment seems to be not for this method

providers/hubspot/subscriptionEvent.go
```diff
@@ -40,6 +43,28 @@ func (c *Connector) GetRecordFromSubscriptionEvent(
...
+func (c *Connector) VerifyWebhookMessage(
+	_ context.Context, params *common.WebhookVerificationParameters,
+) (bool, error) {
+	ts := params.Headers.Get("X-Hubspot-Request-Timestamp")
```
RajatPawar, line 50: Could you please add these header strings to `hubspot/types.go`?

common/types.go
```diff
@@ -275,3 +276,12 @@ type SubscriptionEvent interface {
 	ObjectName() (string, error)
 	Workspace() (string, error)
 }
+
+// WebhookVerificationParameters is a struct that contains the parameters required to verify a webhook.
+type WebhookVerificationParameters struct {
```
RajatPawar, line 281: Can we redefine this to have `http.Request` along with the client secret instead? That way, if another provider needs a different kind of verification (with some additional fields) for the webhook, we don't have to change this type, and we also get all the handy methods + libraries on top of http.Request

RajatPawar (changes requested):

connectors.go
```diff
@@ -86,6 +86,13 @@ type AuthMetadataConnector interface {
 	GetPostAuthInfo(ctx context.Context) (*common.PostAuthInfo, error)
 }
 
+type WebhookVerifiableConnector interface {
```
RajatPawar, line 89: I think `WebhookVerifierConnector` makes more sense since the connector is the one verifying the message

RajatPawar (approved):
Approving this to unblock

common/types.go
jlimatampersand: http.Request is not meant to be marshaled due to some fields e.g) `Body` is not marshallable because it is a function interface. 

Really, what we need is all here, headers, body, URL, method. Also the most we will be using are getting headers, then json marshal body. So having the entire HTTP library will be overkill.

PR #1217, opened 2024-12-14
[CON-743] feat: Add ZohoDesk Connector

immdipu:
## Checklist
- [x] Ran Linter

### GET 
URL: localhost:4444/api/v1/organizations
image

### POST
URL: localhost:4444/api/v1/contacts
image

### PATCH 
URL: localhost:4444/api/v1/contacts/1060099000000334001
image

### DELETE
URL:  localhost:4444/api/v1/organizations/872965074/logo
image

### Pagination
Pagination  is done by using **from** (starting point) and **limit** (number of records). By default, it starts from record 1 and fetches 10 at a time, with a maximum of 50 per request.

image

# Error 

Attempting to delete a profile picture when there is none.
image

Invalid ticket id used
image

Invalid OrgId provided in the header
image

Invalid payload provided
image

### DIFF
--- a/providers/zohoDesk.go
+++ b/providers/zohoDesk.go
@@ -1,0 +1,46 @@
+package providers
+
+const ZohoDesk Provider = "zohoDesk"
+
+func init() {
+	SetInfo(ZohoDesk, ProviderInfo{
+		DisplayName: "ZohoDesk",
+		AuthType:    Oauth2,
+		BaseURL:     "https://desk.zoho.com",
+		Oauth2Opts: &Oauth2Opts{
+			GrantType: AuthorizationCode,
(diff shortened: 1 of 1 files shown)


providers/zohoDesk.go
```diff
@@ -0,0 +1,46 @@
+package providers
+
+const ZohoDesk Provider = "zohoDesk"
+
+func init() {
+	SetInfo(ZohoDesk, ProviderInfo{
+		DisplayName: "ZohoDesk",
```
RajatPawar, line 7:
```suggestion
		DisplayName: "Zoho Desk",
```

RajatPawar (approved):

PR #1222, opened 2024-12-16
feat: Add associations options to HubSpot connector

eberle1080:
This adds associations to Hubspot's read action, as well as the GetRecordsWithIds method

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -124,6 +124,10 @@ type ReadParams struct {
 	//		Note: timing is already handled by Since argument.
 	//		Reference: https://developers.klaviyo.com/en/docs/filtering_
 	Filter string // optional
+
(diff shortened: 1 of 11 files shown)


common/types.go
```diff
@@ -124,6 +124,10 @@ type ReadParams struct {
 	//		Note: timing is already handled by Since argument.
 	//		Reference: https://developers.klaviyo.com/en/docs/filtering_
 	Filter string // optional
+
+	// AssociatedObjects is a list of associated objects to fetch along with the main object.
+	// Only supported by HubSpot connector.
```
laurenzlong, line 129:
```suggestion
	// Only supported by HubSpot connector Read (not Search)
```

laurenzlong (approved):

PR #1234, opened 2024-12-18
[CON-793] feat: Add Discourse connector

immdipu:
## Checklist
- [x] Ran Linter
- [x] Added tests

## Catalog variables
* {{workspace}}

### GET
URL: localhost:4444/posts.json
image

### POST
URL: localhost:4444/posts.json
image

### PUT
URL: localhost:4444/posts/10.json
image

### DELETE
URL: localhost:4444/posts/10.json
image

### Pagination
Pagination is implemented by passing the page number as a query parameter. Each page returns 30 items.
image

Next page:
image

### Errors

Input duplicate data
image

provide invalid data
image

provide an incorrect dynamic ID in the URL.
image

### DIFF
--- a/providers/discourse.go
+++ b/providers/discourse.go
@@ -6,7 +6,8 @@ func init() {
 	SetInfo(Discourse, ProviderInfo{
 		DisplayName: "Discourse",
 		AuthType:    ApiKey,
-		//Discourse is self-hosted, and users can choose their own custom domains for their hosted instances. The hosted domain serves as the base URL.
+		// Discourse is self-hosted, and users can choose their own custom domains for their hosted instances.
+		// The hosted domain serves as the base URL.
 		BaseURL: "https://{{.workspace}}",
 		ApiKeyOpts: &ApiKeyOpts{
 			AttachmentType: Header,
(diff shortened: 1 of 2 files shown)


providers/discourse.go
```diff
@@ -0,0 +1,40 @@
+package providers
+
+const Discourse Provider = "discourse"
+
+func init() {
+	SetInfo(Discourse, ProviderInfo{
+		DisplayName: "Discourse",
+		AuthType:    ApiKey,
+		BaseURL:     "https://{{.workspace}}",
```
RajatPawar, line 9: The base url doesn't seem right

RajatPawar (changes requested):

providers/discourse.go
immdipu: Discourse is self-hosted, and users can choose their own custom domains for their hosted instances. Since Discourse doesn’t have a fixed base URL, it’s super flexible and can work with any domain.

providers/discourse.go
```diff
@@ -0,0 +1,41 @@
+package providers
+
+const Discourse Provider = "discourse"
+
+func init() {
+	SetInfo(Discourse, ProviderInfo{
+		DisplayName: "Discourse",
+		AuthType:    ApiKey,
+		// Discourse is self-hosted The hosted domain serves as the base URL.
```
RajatPawar, line 9:
```suggestion
		// Discourse is self-hosted, and the domain on which it is hosted serves as the base URL.
```

RajatPawar (approved):

PR #1235, opened 2024-12-19
[CON-423] feat: (zoho)update write input dynamically

Jkarage:
### Issue
-  Zoho expects a list of objects that's are to be created/updated.  
- Ampersand API sends a map[string]any. A single object

This PR wraps the object into a slice and sends it. Checks if the received data is not a list.

Tests:
Screenshot 2024-12-19 at 12 03 16

For more context: https://linear.app/ampersand/issue/CON-420/zoho-crm-deep-connector

### DIFF
--- a/providers/zohocrm/write.go
+++ b/providers/zohocrm/write.go
@@ -44,6 +44,8 @@ type writeResponse struct {
 // A maximum of 100 records can be inserted per API call.
 // https://www.zoho.com/crm/developer/docs/api/v6/insert-records.html
 func (c *Connector) Write(ctx context.Context, config common.WriteParams) (*common.WriteResult, error) {
+	var body any
+
 	if err := config.ValidateParams(); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


providers/zohocrm/write.go
```diff
@@ -67,10 +69,13 @@ func (c *Connector) Write(ctx context.Context, config common.WriteParams) (*comm
...
 
-	// ZohoCRM requires everything to be wrapped in a "data" object.
-	// RecordData should be a list of map[string]any
-	body := map[string]any{
-		"data": config.RecordData,
+	// Checking if the config.RecordData asserts to []map[string]any
+	// If not, we wrap the data in a slice, else we send it as is.
+	recordDataCopy, ok := config.RecordData.([]map[string]any)
+	if !ok {
+		body = map[string]any{"data": []any{config.RecordData}}
```
jlimatampersand, line 76: I would also assert `config.RecordData.(map[string]any)` here and error out if not ok.

jlimatampersand (approved):
Preapproving to unblock, but make sure implement type assertion and return error if needed.

Jkarage:
@jlimatampersand Addressed your review, and returned `common.ErrBadRequest` if the assertion of the map fails. Not so comfortable with the error type returned.  Merging so I can continue with Integration testing. Will update this if you feel the need to.

providers/zohocrm/write.go
```diff
@@ -83,18 +84,30 @@ func (c *Connector) Write(ctx context.Context, config common.WriteParams) (*comm
...
+		Errors:  errs,
 	}, nil
 }
+
+func constructWritePayload(payload any) (any, error) {
+	v, ok := payload.([]map[string]any)
+	if !ok {
+		objectData, ok := payload.(map[string]any)
+		if !ok {
+			return nil, common.ErrBadRequest
```
jlimatampersand, line 106: @Jkarage Can you make an update PR that wraps this error?

```
return nil, fmt.Errorf("%w: unexpected input type. Expecting either a json object or a list of objects, but received type '%T'"), payload) 
```

PR #1237, opened 2024-12-19
feat: Add associations options to HubSpot connector

eberle1080:
This reverts commit 812ae16d4e22fe81a0bf4db8d3470b6b6748b850.

Prior PR is here: https://github.com/amp-labs/connectors/pull/1222

This fixes a bug which caused object names to be invalid.

read test:

Screenshot 2024-12-19 at 3 11 03 PM

read-write test:

Screenshot 2024-12-19 at 3 17 19 PM

Screenshot 2024-12-19 at 3 17 35 PM

Screenshot 2024-12-19 at 3 17 47 PM


providers/hubspot/read.go
```diff
@@ -60,8 +59,14 @@ func (c *Connector) Read(ctx context.Context, config common.ReadParams) (*common
 	} else {
 		// If NextPage is not set, then we're reading the first page of results.
 		// We need to construct the query and then make the request.
-		relativeURL := strings.Join([]string{"objects", config.ObjectName, "?" + makeQueryValues(config)}, "/")
```
eberle1080, line 63: If there's a line of code to blame for yesterday's outage, this is the line. Notice how it's encoding the query param directly in to the string? And then that string gets passed to getURL? Well getURL already has its own handling for query params, and so the encoded path+query param being re-encoded likely caused issues. Either there were 2 question marks, which isn't technically valid, or it escaped the question mark - which would have made the object name itself invalid. Which actually seems more like what happened.

providers/hubspot/read.go
```diff
@@ -60,8 +59,14 @@ func (c *Connector) Read(ctx context.Context, config common.ReadParams) (*common
...
-		rsp, err = c.Client.Get(ctx, c.getURL(relativeURL))
+		relativeURL := strings.Join([]string{"objects", config.ObjectName + "/"}, "/")
```
laurenzlong, line 62: This line can now be simplified to the following, which is more readable:
```
relativeURL := "objects/" + config.ObjectName
```

(I omitted the last `/` since I'm assuming c.getURL handles that. )

laurenzlong (approved):
Can you please add screenshots of the result of running `test/hubspot/record/main.go` and `test/hubspot/write/main.go`? (and please rename the folder to read-write since it actually tests both Write and Read methods).

providers/hubspot/read.go
eberle1080: I kept that there because it's emulating what the old code did. I tried to keep it exact.

eberle1080: but I did simplify it, per your suggestion

PR #1238, opened 2024-12-19
[ENG-1759] Hubspot webhook timestamp and recordId

jlimatampersand:
Added few methods and fields to support event collapsing

Blocking https://github.com/amp-labs/server/pull/2275

## Test Result
Screenshot 2024-12-19 at 2 19 45 PM

### DIFF
--- a/providers/hubspot/subscriptionEvent.go
+++ b/providers/hubspot/subscriptionEvent.go
@@ -22,9 +22,9 @@ type SubscriptionEvent struct {
 	OccurredAt       int    `json:"occurredAt"` // in milliseconds
 	SubscriptionType string `json:"subscriptionType"`
 	AttemptNumber    int    `json:"attemptNumber"`
-	ObjectId         int    `json:"objectId"`
 	ChangeSource     string `json:"changeSource"`
 	// Optional fields
+	ObjectId   *int    `json:"objectId,omitempty"`
 	ChangeFlag *string `json:"changeFlag,omitempty"`
 	// Property Change Fields
 	PropertyName  *string `json:"propertyName,omitempty"`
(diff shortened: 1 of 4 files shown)


providers/hubspot/subscriptionEvent.go
```diff
@@ -18,13 +19,22 @@ type SubscriptionEvent struct {
...
+	OccurredAt       int    `json:"occurredAt"` // in milliseconds
 	SubscriptionType string `json:"subscriptionType"`
 	AttemptNumber    int    `json:"attemptNumber"`
-	ObjectId         int    `json:"objectId"`
 	ChangeSource     string `json:"changeSource"`
-	PropertyName     string `json:"propertyName"`
-	PropertyValue    string `json:"propertyValue"`
+	// Optional fields
+	ObjectId   *int    `json:"objectId,omitempty"`
+	ChangeFlag *string `json:"changeFlag,omitempty"`
```
laurenzlong, line 28: can you turn these `*string` types back into `string`? References to strings are annoying to deal with, since you have to check that it's not nil before dereferencing and it doesn't convey any additional information than a simple string type. With a string type, you can easily check `if ChangeFlag === ""` to see if something is an empty string. (As opposed to something like to a bool value, where `*bool` is actually useful since it allows you to differentiate between an empty value and false).

I know we currently have a lot of `*string` in our codebase, because the openapi generator turns optional strings into `*string` by default, unless you add the `x-go-type-skip-optional-pointer: true` annotation. But we should actively try to avoid introducing more of it into the codebase.

laurenzlong (approved):

providers/hubspot/subscriptionEvent.go
laurenzlong: Never mind, this is a good approach since it allows us to more easily strip out empty fields when sending the webhook message to builders, see https://github.com/amp-labs/connectors/pull/1251#issuecomment-2566067227

laurenzlong: Omitempty will still omit empty strings when you JSON marshal. See https://go.dev/play/p/zU4Q3eohfi4, so `*string` is not necessary, `string` type will work.

PR #1246, opened 2024-12-20
fix: (zohoCRM) enhance error handling for invalid inputs

Jkarage:
@jlimatampersand some changes from the suggested review changes. There is no error to wrap, so just excluded the first sentence.

### DIFF
--- a/providers/zohocrm/write.go
+++ b/providers/zohocrm/write.go
@@ -2,6 +2,7 @@ package zohocrm
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/amp-labs/connectors/common"
 	"github.com/amp-labs/connectors/common/naming"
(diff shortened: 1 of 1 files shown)


providers/zohocrm/write.go
```diff
@@ -103,7 +104,7 @@ func constructWritePayload(payload any) (any, error) {
 	if !ok {
 		objectData, ok := payload.(map[string]any)
 		if !ok {
-			return nil, common.ErrBadRequest
+			return nil, fmt.Errorf("expecting either a json object or a list of objects, but received type '%T'", payload) //nolint:err113,lll
```
jlimatampersand, line 107: This will give us linter error. 

You can either wrap `common.ErrBadRequest` or define new error such as

var errUnexpectedInputTyp = errors.New("unexpected input type") 

you can use one or the other. Don't use `fmt.Errorf` alone.

PR #1251, opened 2024-12-30
Make webhook message to map type, recordId and timestamp from webhook

jlimatampersand:
This PR does 2 things
1. Getter for RecordID, and Timstamps in nano seconds
2. Change the webhook type to map[string]any 

About point 2: 

Strictly typed webhook messages give empty values to each field. Sometimes this is unexpected and can be misinterpreted. Look at the below screenshot.

Screenshot 2024-12-30 at 3 31 31 PM

In another [PR](https://github.com/amp-labs/connectors/pull/1238), I attempted to use pointers so that this can be indicated by null values, but @laurenzlong raised concern about that it can be complex to nill checks and conversion. We can prevent this by using map[string]any type. Test result shows that no empty fields are sent out. 

Test Result

Screenshot 2024-12-30 at 3 26 13 PM

```
jaehyunlim@Jaehyuns-MacBook-Pro connectors % go run ./test/hubspot/record/main.go
time=2024-12-30T11:13:37.128-08:00 level=DEBUG msg="loading credentials file" path=./hubspot-creds.json
time=2024-12-30T11:13:37.130-08:00 level=DEBUG msg="HTTP request" connector=hubspot method=POST url=https://api.hubapi.com/crm/v3/objects/contacts headers="[{Key:Accept Value:application/json}]" bodySize=233
time=2024-12-30T11:13:37.907-08:00 level=DEBUG msg="HTTP request" connector=hubspot method=POST url=https://api.hubapi.com/crm/v3/objects/contacts/batch/read headers="[{Key:Accept Value:application/json}]" bodySize=51
[
  {
    "fields": null,
    "raw": {
      "archived": false,
      "createdAt": "2024-12-30T19:13:37.628Z",
      "id": "88049383345",
      "properties": {
        "createdate": "2024-12-30T19:13:37.628Z",
        "email": "wendygleason@hammes.info",
        "firstname": "Jonathan",
        "hs_object_id": "88049383345",
        "lastmodifieddate": "2024-12-30T19:13:37.628Z",
        "lastname": "Feil"
      },
      "updatedAt": "2024-12-30T19:13:37.628Z"
    },
    "id": "88049383345"
  }
]
```


laurenzlong:
@jlimatampersand sorry I didn't realize that's why you were using pointer types, I actually prefer https://github.com/amp-labs/connectors/pull/1238 over this approach. Strictly typing is preferred over using `map[string]any` since the code is more readable. I'm going to close this PR, feel free to merge https://github.com/amp-labs/connectors/pull/1238

laurenzlong (changes requested):
@jlimatampersand I actually prefer https://github.com/amp-labs/connectors/pull/1238 over this approach. Strictly typing is preferred over using map[string]any since the code is more readable. I didn't mean that you shouldn't use pointer types, I meant that you shouldn't use `*string`. Even if you use `string` type, omitempty will still omit empty strings when you JSON marshal. See https://go.dev/play/p/eUyr5ixemPj

providers/hubspot/subscriptionEvent.go
```diff
@@ -9,24 +9,40 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/amp-labs/connectors/common"
 )
 
+/*
 type SubscriptionEvent struct {
```
laurenzlong, line 18: can you remove this commented out struct? otherwise it'll be tempting for others to revise the code to use this struct in the future, if you wanted to document what the message looks like, then maybe you can provide a sample message in a comment block.

providers/hubspot/subscriptionEvent.go
```diff
@@ -9,24 +9,40 @@ import (
...
+	FromObjectId         *int    `json:"fromObjectId,omitempty"`
+	ToObjectId           *int    `json:"toObjectId,omitempty"`
+	AssociationRemoved   *bool   `json:"associationRemoved,omitempty"`
+	IsPrimaryAssociation *bool   `json:"isPrimaryAssociation,omitempty"`
 }
 
+
+*/
+
+type SubscriptionEvent map[string]any
```
laurenzlong, line 44: As discussed, going with a map is a good approach in order to preserve the full raw event body to send to builders (whereas JSON marshaling to/from a struct will lose the fields that are not captured in the struct definition). Can you please add a comment to explain this?

laurenzlong (approved):

providers/hubspot/subscriptionEvent.go
```diff
@@ -75,23 +76,63 @@ func (evt *SubscriptionEvent) EventType() (common.SubscriptionEventType, error)
...
 
 	return parts[0], nil
 }
 
-func (evt *SubscriptionEvent) Workspace() (string, error) {
-	return strconv.Itoa(evt.PortalId), nil
+func (evt SubscriptionEvent) Workspace() (string, error) {
+	m := evt.asMap()
+
+	portalId, err := m.AsInt("portalId")
```
jlimatampersand, line 104: Here as int returns int64, so below at line 109, the value should be casted to int as strconv.Itoa accepts int.

common/map.go
```diff
@@ -0,0 +1,168 @@
+package common
+
+import (
+	"errors"
+	"fmt"
+	"reflect"
+)
+
+type GenericMap map[string]any
```
laurenzlong, line 9: Can you name this `StringMap`? Just because "generic" has a specific meaning in Go.

common/map.go
```diff
@@ -0,0 +1,168 @@
...
+
+func (m GenericMap) Keys(key string) any {
```
laurenzlong, line 11: This should return `[]string`, not any

common/map.go
```diff
@@ -0,0 +1,168 @@
...
+func (m GenericMap) Get(key string) (any, error) {
+	val, ok := m[key]
+	if !ok {
+		return nil, fmt.Errorf("%w: %s", errKeyNotFound, key)
+	}
+
+	return val, nil
+}
+
+var (
```
laurenzlong, line 30: can you put these at the top of the file?

common/map.go
```diff
@@ -0,0 +1,168 @@
...
+func (m GenericMap) GetString(key string) (string, error) {
+	val, err := m.Get(key)
+	if err != nil {
+		return "", err
+	}
+
+	return assertType[string](val)
+}
+
+var errFieldTypeMismatch = errors.New("field type mismatch")
```
laurenzlong, line 93: can you put this at the top of the file?

common/map.go
```diff
@@ -0,0 +1,168 @@
...
+
+//nolint:ireturn
+func assertType[T any](val any) (T, error) {
```
laurenzlong, line 96: can you put this function below the other functions?

common/map.go
```diff
@@ -0,0 +1,168 @@
...
+	return assertType[bool](val)
+}
+
+func (m GenericMap) Has(key string) bool {
+	_, ok := m[key]
+
+	return ok
+}
+
+func (m GenericMap) Values() []any {
```
laurenzlong, line 120: can you put the Values and Len function at the top, close to Keys? It makes more sense than interspersing among the the `Get` functions.

common/map.go
```diff
@@ -0,0 +1,168 @@
...
+	values := make([]any, 0, len(m))
+
+	for _, v := range m {
+		values = append(values, v)
```
laurenzlong, line 124: Similar to my comments on the Keys function, it's more efficient to not use append.

common/map.go
```diff
@@ -0,0 +1,168 @@
...
+import (
+	"errors"
+	"fmt"
+	"reflect"
+)
+
+type GenericMap map[string]any
+
+func (m GenericMap) Keys(key string) any {
+	keys := make([]any, 0, len(m))
```
laurenzlong, line 12: keys are guaranteed to be strings, so this should be `[]string` not `[]any`

common/map.go
```diff
@@ -0,0 +1,168 @@
...
+
+	for k := range m {
+		keys = append(keys, k)
```
laurenzlong, line 15: nit: append is less efficient. You can re-write as:

```
keys := make ([]string, len(m))
i := 0
for k := range m {
  keys[i] = k
  i++
}
return keys
```

common/map_test.go
```diff
@@ -0,0 +1,188 @@
...
+import (
+	"slices"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+//nolint:funlen,gocognit, cyclop
+func TestGenericMap(t *testing.T) {
```
laurenzlong, line 12: once you rename GenericMap to StringMap, don't forget to update this test function also.

common/map_test.go
```diff
@@ -0,0 +1,188 @@
...
+		got3, err3 := m.Get("map")
+
+		require.NoError(t, err3)
+		assert.Equal(t, want3, got3)
+
+		passes := []string{"string", "int", "float", "map", "bool"}
+
+		for k := range m {
+			got, err := m.Get(k)
+			if !slices.Contains(passes, k) {
```
laurenzlong, line 50: I'm confused by what this is testing, this if statement will never be entered into, since you are iterating through `m` in the outer for loop, so `k` will always be one of the elements of `passes`

common/map_test.go
```diff
@@ -0,0 +1,188 @@
...
+	t.Run("GetInt", func(t *testing.T) {
+		t.Parallel()
+
+		want := int64(1)
+		got, err := m.GetInt("int")
+
+		require.NoError(t, err)
+		assert.Equal(t, want, got)
+
+		passes := []string{"int"}
```
laurenzlong, line 66: can you rename `passes` something like `keysWithIntValue` for clarity? Similarly for the other test cases in this file.

common/map_test.go
laurenzlong: After reading through the other test cases, this makes more sense, but it's better to just remove it since it's not testing anything for the regular `Get` function.

providers/hubspot/subscriptionEvent.go
```diff
@@ -9,23 +9,19 @@ import (
...
-	SubscriptionType string `json:"subscriptionType"`
-	AttemptNumber    int    `json:"attemptNumber"`
-	ObjectId         int    `json:"objectId"`
-	ChangeSource     string `json:"changeSource"`
-	PropertyName     string `json:"propertyName"`
-	PropertyValue    string `json:"propertyValue"`
-}
+/*
+Note:
+SubscriptionEvent is a map[string]any on purpose because the structure of the event is not known.
```
laurenzlong, line 19: can you add "as opposed to a typed struct"

common/types.go
```diff
@@ -295,3 +297,5 @@ type WebhookVerificationParameters struct {
 	ClientSecret string
 	Method       string
 }
+
+type SubscriptionEventMap GenericMap
```
laurenzlong, line 301: SubscriptionEventMap isn't used anywhere, and you already have SubscriptionEvent in the `subscriptionEvent.go` file, this should be removed.

laurenzlong (approved):

common/stringMap.go
```diff
@@ -0,0 +1,157 @@
...
+	"golang.org/x/exp/maps"
+)
+
+var (
+	errKeyNotFound       = errors.New("key not found")
+	errNotANumber        = errors.New("not a number")
+	errFieldTypeMismatch = errors.New("field type mismatch")
+)
+
+type StringMap map[string]any
```
Cobalt0s, line 17: I think this type should live under `datautils` package as it is a new/similiar data type.

PR #1253, opened 2024-12-31
[CON-933] feat: Add HeyReach Connector

Jkarage:
## Catalog variables
None

### GET
Screenshot 2024-12-31 at 13 17 26

### POST
Screenshot 2024-12-31 at 13 16 29

No Other Methods apart from `GET` and `POST`

# Logo & Icons
Screenshot 2024-12-31 at 11 11 44

### DIFF
--- a/providers/heyreach.go
+++ b/providers/heyreach.go
@@ -1,0 +1,42 @@
+package providers
+
+const HeyReach Provider = "heyreach"
+
+func init() {
+	// Hive Connector Configuration
+	SetInfo(Hive, ProviderInfo{
+		DisplayName: "heyreach",
+		AuthType:    ApiKey,
+		BaseURL:     "https://api.heyreach.io/api",
+		ApiKeyOpts: &ApiKeyOpts{
+			AttachmentType: Header,
+			Header: &ApiKeyOptsHeader{
+				Name: "X-API-KEY",
+			},
+			DocsURL: "https://documenter.getpostman.com/view/23808049/2sA2xb5F75",
+		},
+		//nolint:lll
+		Media: &Media{
+			DarkMode: &MediaTypeDarkMode{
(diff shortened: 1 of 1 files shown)


laurenzlong (approved):

providers/heyreach.go
```diff
@@ -0,0 +1,41 @@
...
+	SetInfo(HeyReach, ProviderInfo{
+		DisplayName: "heyreach",
+		AuthType:    ApiKey,
+		BaseURL:     "https://api.heyreach.io/api",
+		ApiKeyOpts: &ApiKeyOpts{
+			AttachmentType: Header,
+			Header: &ApiKeyOptsHeader{
+				Name: "X-API-KEY",
+			},
+			DocsURL: "https://documenter.getpostman.com/view/23808049/2sA2xb5F75",
```
laurenzlong, line 15: This link isn't very informative, since it doesn't tell the user how to actually get the API key, is there a better link? if not, please remove this line. It's better to not have a docsURL than to have one that isn't helpful.

providers/heyreach.go
```diff
@@ -0,0 +1,41 @@
+package providers
+
+const HeyReach Provider = "heyreach"
+
+func init() {
+	SetInfo(HeyReach, ProviderInfo{
+		DisplayName: "heyreach",
```
laurenzlong, line 7: Even though their logo is all lower case, I think the display name should probably be "HeyReach", since that's how it appears in text throughout their website, examples:
Screenshot 2025-01-02 at 4 25 04 PM
b.com/user-attachments/assets/3629f9ee-02a2-437f-969e-5b23dcc555f3)

providers/heyreach.go
Jkarage: No better link, Removing this.

PR #1255, opened 2025-01-02
Add emailbison connector

sanjay-a01:


providers/emailBison.go
```diff
@@ -6,14 +6,13 @@ func init() {
 	SetInfo(EmailBison, ProviderInfo{
 		DisplayName: "EmailBison",
 		AuthType:    ApiKey,
-		BaseURL:     "https://dedi.emailbison.com",
+		BaseURL:     "https://{{.workspace}}.emailbison.com",
```
jlimatampersand, line 9: Where did you get this workspace reference in the docs? I couldn't find one. it seems like the previous baseURL might be the correct one. Can you provide what the problem was before and why this change is made? It is confusing

jlimatampersand (changes requested):

jlimatampersand:
Can you provide description on why this change is needed?

providers/emailBison.go
sanjay-a01: @jlimatampersand For this, I have contacted the support team and they said the value will change for each customer

emailBison workspace mail chat

jlimatampersand: Okay, so from the last response, it seems like each customer will have a full URL, which seems like those are self-hosted services and even the `emailbison.com` portion will not be there. This is not common case. Can you double-check with them about the full URL and subdomain situation? 

@laurenzlong @RajatPawar This seems like a unique situation, what would be the best approach?

RajatPawar: I think in this case, let's have the entire base URL be `workspace`

laurenzlong: @sanjay-a01 can you please add a comment above this line to explain this? Also that means this connector is blocked from integration testing since we don't support collection of workspace for API key connectors yet. Can you please mark the Linear ticket as blocked and comment with this as the reason?

laurenzlong (approved):

providers/emailBison.go
sanjay-a01: @laurenzlong FYI, I have added a comment and also changed the status in linear

PR #1256, opened 2025-01-03
[CON-892] feat: (apollo) support API product names

Jkarage:
Apollo sometimes uses different names in the URL vs in the description of the API documentation:

[Emailer Campaigns are called Sequences](https://docs.apollo.io/reference/search-for-sequences)
[Opportunities are called Deals](https://docs.apollo.io/reference/list-all-deals)
[Opportunity Stages are called Deal Stages](https://docs.apollo.io/reference/list-deal-stages)
[Lists and Tags and called Labels](https://docs.apollo.io/reference/get-a-list-of-all-liststags)

This Adds support for these product names.

Testing(Reading):
Screenshot 2025-01-03 at 09 15 45

Writing:
Screenshot 2025-01-04 at 10 17 57

Reading Metadata:
Screenshot 2025-01-04 at 10 26 56

### DIFF
--- a/providers/apollo/types.go
+++ b/providers/apollo/types.go
@@ -39,6 +39,9 @@ var productNameToObjectName = map[string]string{
 	"lists_and_tags": "labels",
 }
 
+// Apollo uses mismatched API object names and display names in the documentation.
+// We want to support both naming conventions. This function checks whether the provided objectName
+// is a display name, and if so, maps it to the corresponding API object name.
 func constructSupportedObjectName(obj string) string {
 	// we want to update the objectName if the provided objectName
 	// is the product name from the API docs to the supported objectName.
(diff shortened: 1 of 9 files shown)


providers/apollo/types.go
```diff
@@ -13,14 +17,38 @@ var (
...
+//
+//nolint:gochecknoglobals,lll
+var productNameToObjectName = map[string]string{
+	"sequences":      "emailer_campaigns",
+	"deals":          "opportunities",
+	"deal_stages":    "opportunity_stages",
+	"lists_and_tags": "labels",
+}
+
+func constructSupportedObjectName(obj string) string {
```
laurenzlong, line 42: Can you add a comment to explain what this function does and why it's needed? Explain that Apollo has mismatched API object names and display names used in docs, and that we want to support both.

PR #1261, opened 2025-01-07
feat: (zoho) Add support for lowercase fieldnames

Jkarage:
Currently Reading Field Names are case sensitive. With an exception in `id`. This updates the connector so as it can receiver lower cased field names. 
This also updates the Metadata Connector to return lower case field names.

Testing:
- Read
Screenshot 2025-01-07 at 15 15 57

- Metadata:
Screenshot 2025-01-07 at 15 07 53

- Write
Screenshot 2025-01-08 at 12 47 50

For more ctx: https://github.com/amp-labs/samples/pull/7

### DIFF
--- a/providers/zohocrm/write.go
+++ b/providers/zohocrm/write.go
@@ -99,15 +99,38 @@ func (c *Connector) Write(ctx context.Context, config common.WriteParams) (*comm
 }
 
 func constructWritePayload(payload any) (any, error) {
-	v, ok := payload.([]map[string]any)
+	data, ok := payload.([]map[string]any)
 	if !ok {
 		objectData, ok := payload.(map[string]any)
 		if !ok {
 			return nil, common.ErrBadRequest
 		}
 
+		// Capitalize words in the data fields for Creation/Updating
+		for k, data := range objectData {
+			fld := constructFieldNames([]string{k})
+			objectData[fld] = data
+			// Remove the previous field key in the map, as it's no longer required.
+			if fld != k {
+				delete(objectData, k)
+			}
+		}
+
 		return map[string]any{"data": []map[string]any{objectData}}, nil
 	}
 
-	return map[string]any{"data": v}, nil
+	// Range Over the Slice for every map, Capitalize them.
+	for _, v := range data {
+		// Capitalize words in the data fields for Creation/Updating
+		for k, d := range v {
+			fld := constructFieldNames([]string{k})
+			v[fld] = d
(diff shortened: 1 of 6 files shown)


laurenzlong (changes requested):
@Jkarage Just to confirm, have you tried just using lower case field names and see if the Zoho API accepts it? We've found that many APIs actually treat field names in a case-insensitive way, so it might not be necessary to do all the conversion. 

We also need to make sure that Write supports lower case field names also. Can you modify `test/zohocrm/write/write.go` so that some of the requests use upper case field names and some of the requests use lower case field names?

Jkarage:
@laurenzlong yes I tried  using lower case field names and the Zoho API failed.

jlimatampersand:
Are we trying to support only lowercase inputs? or do we want to support regardless of case? If we want to provide case-insensitive support, then we want to keep both lowercase testing and uppercase testing for regression.

providers/zohocrm/write.go
```diff
@@ -99,15 +99,38 @@ func (c *Connector) Write(ctx context.Context, config common.WriteParams) (*comm
...
 
+		// Capitalize words in the data fields for Creation/Updating
+		for k, data := range objectData {
+			fld := constructFieldNames([]string{k})
+			objectData[fld] = data
+			// Remove the previous field key in the map, as it's no longer required.
+			if fld != k {
+				delete(objectData, k)
+			}
+		}
```
jlimatampersand, line 117: we can take this one out and put it in separate function and reuse here and below.

test/zohocrm/read/read.go
```diff
@@ -41,8 +41,8 @@ func main() {
 func readContacts(ctx context.Context, conn *zohocrm.Connector) error {
 	config := connectors.ReadParams{
 		ObjectName: "contacts",
-		Since:      time.Now().Add(-30 * time.Hour),
-		Fields:     connectors.Fields("Assistant", "Created_By", "Full_Name", "id", "Created_Time"),
+		Since:      time.Now().Add(-3000 * time.Hour),
+		Fields:     connectors.Fields("assistant", "created_by", "full_name", "id", "created_time", "enrich_status__s", "skype_id"),
```
jlimatampersand, line 45: Let's keep some capital cased fields in tests.

Jkarage:
@jlimatampersand  Added both test cases.

jlimatampersand (approved):

laurenzlong (approved):

PR #1266, opened 2025-01-08
[ENG-1662] refactor: add components for quicker initialization

RajatPawar:
- Adds base components that let you build a base connector component with - 
  - Provider information (can be overriden)
  - Clients (simple, JSON/XML for now)
  - Support for various endpoints
  - Options to add specific response or error handlers

- Adds common parameters that are self-validating whenever they are included (thanks Constantin!)

### DIFF
--- a/common/params.go
+++ b/common/params.go
@@ -19,13 +19,8 @@ type Parameters struct {
 	// Metadata is a map of key-value pairs that can be used to pass in additional information to the connector.
 	// Generally this is used to substitute placeholders in the providerInfo, like workspace, server, etc, which is
 	// information that is specific to the connection.
-	Metadata Metadata
+	Metadata map[string]string
 }
-
-type (
-	Metadata    map[string]string
-	MetadataKey string
-)
 
 var (
 	ErrValidationFailed  = errors.New("validation failed")
@@ -37,48 +32,34 @@ var (
 // ValidateParameters sees which interfaces conn implements, calls the relevant validation methods.
 // nolint:cyclop
 func ValidateParameters(conn any, params Parameters) error {
-	var errs []error
+	var errs error
 
-	if r, ok := conn.(requireWorkspace); ok {
-		if err := r.validateWorkspace(params); err != nil {
-			errs = append(errs, err)
-		}
+	if r, ok := conn.(workspaceValidator); ok {
+		errs = errors.Join(errs, r.validateWorkspace(params))
 	}
 
-	if r, ok := conn.(requireAuthenticatedClient); ok {
-		if err := r.validateAuthenticatedClient(params); err != nil {
-			errs = append(errs, err)
-		}
+	if r, ok := conn.(authenticatedClientValidator); ok {
+		errs = errors.Join(errs, r.validateAuthenticatedClient(params))
 	}
 
-	if r, ok := conn.(requireMetadata); ok {
-		if err := r.validateMetadata(params); err != nil {
-			errs = append(errs, err)
-		}
+	if r, ok := conn.(metadataValidator); ok {
+		errs = errors.Join(errs, r.validateMetadata(params))
 	}
 
-	if r, ok := conn.(requireModule); ok {
-		if err := r.validateModule(params); err != nil {
-			errs = append(errs, err)
-		}
+	if r, ok := conn.(moduleValidator); ok {
+		errs = errors.Join(errs, r.validateModule(params))
 	}
 
-	if len(errs) > 0 {
-		err := ErrValidationFailed
-
-		for _, e := range errs {
-			err = fmt.Errorf("%w: %w", err, e)
-		}
-
-		return err
+	if errs != nil {
+		return fmt.Errorf("%w: %w", ErrValidationFailed, errs)
 	}
 
 	return nil
 }
 
-// RequireAuthenticatedClient is an interface that requires an authenticated client to be set in the parameters.
+// authenticatedClientValidator is an interface that requires an authenticated client to be set in the parameters.
 
-type requireAuthenticatedClient interface {
+type authenticatedClientValidator interface {
 	validateAuthenticatedClient(parameters Parameters) error
 }
 
@@ -92,9 +73,9 @@ func (RequireAuthenticatedClient) validateAuthenticatedClient(parameters Paramet
 	return nil
 }
 
-// RequireWorkspace is an interface that requires a workspace to be set in the parameters.
+// workspaceValidator is an interface that requires a workspace to be set in the parameters.
 
-type requireWorkspace interface {
+type workspaceValidator interface {
 	validateWorkspace(parameters Parameters) error
 }
 
@@ -108,9 +89,9 @@ func (RequireWorkspace) validateWorkspace(parameters Parameters) error {
 	return nil
 }
 
-// RequireMetadata is an interface that requires metadata to be set in the parameters.
+// metadataValidator is an interface that requires metadata to be set in the parameters.
 
-type requireMetadata interface {
+type metadataValidator interface {
 	validateMetadata(parameters Parameters) error
 }
 
@@ -132,9 +113,9 @@ func (RequireMetadata) validateMetadata(parameters Parameters) error {
 	return nil
 }
 
-// RequireModule is an interface that requires a module to be set in the parameters.
+// moduleValidator is an interface that requires a module to be set in the parameters.
 
-type requireModule interface {
+type moduleValidator interface {
 	validateModule(parameters Parameters) error
 }
 
--- a/common/params.go
+++ b/common/params.go
@@ -34,7 +34,8 @@ var (
 	ErrMissingWorkspace  = errors.New("workspace not given")
 )
 
-// ValidateParameters sees which interfaces conn implements, calls the relevant validation methods
+// ValidateParameters sees which interfaces conn implements, calls the relevant validation methods.
+// nolint:cyclop
 func ValidateParameters(conn any, params Parameters) error {
 	var errs []error
 
@@ -87,6 +88,7 @@ func (RequireAuthenticatedClient) validateAuthenticatedClient(parameters Paramet
 	if parameters.AuthenticatedClient == nil {
 		return ErrMissingAuthClient
 	}
+
 	return nil
 }
 
@@ -102,6 +104,7 @@ func (RequireWorkspace) validateWorkspace(parameters Parameters) error {
 	if parameters.Workspace == "" {
 		return ErrMissingWorkspace
 	}
+
 	return nil
 }
 
@@ -141,5 +144,6 @@ func (RequireModule) validateModule(parameters Parameters) error {
 	if parameters.Module == "" {
 		return ErrMissingModule
 	}
+
 	return nil
 }
--- a/common/params.go
+++ b/common/params.go
@@ -1,0 +1,145 @@
+package common
+
+import (
+	"errors"
+	"fmt"
+)
+
+// Parameters can be used to pass input parameters to the connector.
+type Parameters struct {
+	Module ModuleID
+
+	// AuthenticatedClient is a client for the connector that knows how to handle authentication for the provider.
+	AuthenticatedClient AuthenticatedHTTPClient
+
+	// Workspace is the provider workspace that the connector is connecting to. It could be a part of the metadata
+	// map, but it is a common enough field that it is included here.
+	Workspace string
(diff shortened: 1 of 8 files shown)


RajatPawar:
* **#1268** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1268?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1267** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1267?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1266** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1266?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1266?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

internal/components/components.go
```diff
@@ -0,0 +1,51 @@
+package components
+
+import (
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/internal/goutils"
+	"github.com/amp-labs/connectors/providers"
+)
+
+type Setup[T any] func(component *ConnectorComponent) (*T, error)
```
jlimatampersand, line 9: Can we rename this to something like ConnectorConstructor? `Setup` is too vague. Then in
or `ConnectorInitializer` to clearly note that this is to initialize connector. 


`Initialize` function, function variable name can be a constructor.

jlimatampersand: `T any` is too broad. We should have some bare minimum connector interface to start with.

internal/components/client.go
```diff
@@ -0,0 +1,64 @@
...
+		return nil, err
+	}
+
+	clientComponent := &ClientComponent{
+		JSON: &common.JSONHTTPClient{
+			HTTPClient: &common.HTTPClient{
+				Client: params.AuthenticatedClient,
+
+				// ErrorHandler is set to a default, but can be overridden using options.
+				ErrorHandler: interpreter.ErrorHandler{
```
Cobalt0s, line 30: Ok, I now get the idea you are trying to do here.

To simplify error handler there are 2 "Null Objects" options you can do which are **equivalent**, just pick a prefered method:
* ErrorHandler: common.InterpretError
* ErrorHandler: interpreter.ErrorHandler{}.Handle

This alleviates from `errorFormats(), statusCodeMapping()` usage.

FYI: The handler is just a callback `func(rsp *http.Response, body []byte) error`. interpreter.ErrorHandler is used to handle provider error response based on the Mime Type.

internal/components/components.go
jlimatampersand: `T baseConnector` or something to start with. Then each other interfaces can inherit the base connector. 
```
type requireMetadata interface {
  baseConnector
  requireMetadata func(....)....
} 
```

internal/components/components.go
```diff
@@ -0,0 +1,51 @@
...
+
+	clientComponent, err := NewClientComponent(provider, params)
+	if err != nil {
+		return nil, err
+	}
+
+	connectorComponent := &ConnectorComponent{ClientComponent: *clientComponent}
+
+	// Apply options *before* the setup, so that setups can override if needed
+	for _, opt := range opts {
```
Cobalt0s, line 37: Could you create a constructor for ConnectorComponent()?

This will recieve `goutils.PanicRecovery` and `opts loop` from this `Initialize` and hide details as that is only important for ConnectorComponent.

common/params.go
```diff
@@ -0,0 +1,149 @@
...
+
+		return err
+	}
+
+	return nil
+}
+
+// RequireAuthenticatedClient is an interface that requires an authenticated client to be set in the parameters.
+
+type requireAuthenticatedClient interface {
```
jlimatampersand, line 81: I might be viewing this in narrower perspective, so please correct me if I am lacking some information. I am sure I will lack some. But below is my concern. 

If all of these `requireOOO` interfaces will have only 1 struct each that adopts the interface, then I think making these interfaces will be overkill. None of these struct methods refer to field values in the structs either, so there's no need to require these as structs unless you foresee actual use cases of field values of the structs. 

I would suggest that we use boolean values in providerInfo for each module, then just run functions if those boolean values are run the functions without creating an unnecessary struct that only uses the method.

jlimatampersand: Even if we end up using different validateAuthenticatedClient functions, I still think it might be better to attach them in providerApp as a function and have function interface to be defined rather than structs because I don't foresee much use of field values of these `RequireAuthenticatedClient` structs

jlimatampersand (changes requested):

Cobalt0s (commented):
I left some comments and suggestions.

* ProviderEndpointSupport looks like a new feature worthy of dedicated PR
* Everything else looks great, it would be good to merge it! :rocket:

internal/components/option.go
```diff
@@ -0,0 +1,35 @@
+package components
+
+import (
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/interpreter"
+)
+
+type Option func(*ConnectorComponent)
+
+func WithErrorHandler(handler interpreter.ErrorHandler) Option {
```
jlimatampersand, line 10: maybe we should separate parameters for xml and json as they might be very different from each other. We can just pass same handler to both if one handler can be used for both. Or we can define handler params
```
{
   JSONErrHandler: handler1,
   XMLErrHandler: handler2, 
}
```

Cobalt0s: I think we should rename interpreter.ErrorHandler into `interpreter.MimeErrorHandler`. Internally it has Mime subdivisions for JSON, XML, custom, etc.

internal/components/client.go
RajatPawar: Thanks!

internal/components/provider.go
```diff
@@ -0,0 +1,49 @@
+package components
+
+import (
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/paramsbuilder"
+	"github.com/amp-labs/connectors/providers"
+)
+
+// ProviderComponent is a component that adds provider information to a connector.
+type ProviderComponent struct {
```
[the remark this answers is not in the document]
RajatPawar: Agree. I resorted to components because it seemed closest to what these want to be. Since these are pretty much internal at this point, we can always come back and rename these when we have better terms.

internal/components/components.go
RajatPawar: Renamed this to `ConnectorConstructor` & `InitializeConnector`, added comments. And since this itself is responsible for building `connectors.Connector`, which is the most basic kind of connector we have, I can't think of anything that is narrower. Since this is internal, and everywhere it's used will need assertion, we should be okay / can change it in the future

common/params.go
```diff
@@ -0,0 +1,149 @@
...
+
+// Parameters can be used to pass input parameters to the connector.
+type Parameters struct {
+	Module ModuleID
+
+	// AuthenticatedClient is a client for the connector that knows how to handle authentication for the provider.
+	AuthenticatedClient AuthenticatedHTTPClient
+
+	// Workspace is the provider workspace that the connector is connecting to. It could be a part of the metadata
+	// map, but it is a common enough field that it is included here.
```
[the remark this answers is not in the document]
RajatPawar: Well, what I meant is that you could avoid having `Workspace` as a parameter and expect it to be a part of the `metadata` map, but it's a common parameter that most connectors need. That's the reason it's a first class parameter. Technically, you could pass in whatever you want in `metadata`, and if the connector digs into the `metadata` map for it, you're good

common/params.go
```diff
@@ -0,0 +1,149 @@
...
+	Workspace string
+
+	// Metadata is a map of key-value pairs that can be used to pass in additional information to the connector.
+	// Generally this is used to substitute placeholders in the providerInfo, like workspace, server, etc, which is
+	// information that is specific to the connection.
+	Metadata Metadata
+}
+
+type (
+	Metadata    map[string]string
```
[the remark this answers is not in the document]
RajatPawar: My bad, this is a left over, but I actually decided to just have Metadata as `map[string]string` for now - additional types introduce complexity that isn't justified right now. Being aggressive with deleting until we need to add it back!

common/params.go
```diff
@@ -0,0 +1,149 @@
...
+			err = fmt.Errorf("%w: %w", err, e)
+		}
+
+		return err
+	}
+
+	return nil
+}
+
+// RequireAuthenticatedClient is an interface that requires an authenticated client to be set in the parameters.
```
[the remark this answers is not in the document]
RajatPawar: * Updated the interface names to be in line with convention
* That's a good observation, we always do need auth client, maybe we can eventually move it to `ConnectorComponent`

common/params.go
```diff
@@ -0,0 +1,149 @@
...
+	ErrMissingWorkspace  = errors.New("workspace not given")
+)
+
+// ValidateParameters sees which interfaces conn implements, calls the relevant validation methods.
+// nolint:cyclop
+func ValidateParameters(conn any, params Parameters) error {
+	var errs []error
+
+	if r, ok := conn.(requireWorkspace); ok {
+		if err := r.validateWorkspace(params); err != nil {
```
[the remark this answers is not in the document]
RajatPawar: Neat, thanks!

internal/components/components.go
RajatPawar: If you mean moving those lines out into a function of its own, sure. Let me know otherwise.

internal/components/support.go
```diff
@@ -0,0 +1,111 @@
...
+import (
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/providers"
+	"github.com/gobwas/glob"
+)
+
+// ProviderEndpointSupport defines the support for endpoints of a provider. All matches are merged to provide the final
+// support level for an endpoint/object. Explicit denies aren't considered *for now*, because this answers
+// 'What do we want to allow?'.
+type ProviderEndpointSupport struct {
```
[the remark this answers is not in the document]
RajatPawar: Sounds good!

RajatPawar:
@jlimatampersand @Cobalt0s addressed the reviews

internal/components/client.go
```diff
@@ -0,0 +1,56 @@
...
+	providerComponent, err := newProviderComponent(provider, params.Module, params.Workspace, params.Metadata)
+	if err != nil {
+		return nil, err
+	}
+
+	clientComponent := &ClientComponent{
+		JSON: &common.JSONHTTPClient{
+			HTTPClient: &common.HTTPClient{
+				Client: params.AuthenticatedClient,
+
```
Cobalt0s, line 27:
```suggestion
Base: providerComponent.ProviderInfo().BaseURL
```
This would remove field modifier after this struct initialization:
`clientComponent.JSON.HTTPClient.Base = providerComponent.ProviderInfo().BaseURL`

internal/components/components.go
```diff
@@ -0,0 +1,64 @@
...
+	params common.Parameters,
+	constructor ConnectorConstructor[T],
+	opts ...Option,
+) (conn *T, err error) {
+	defer goutils.PanicRecovery(func(cause error) {
+		err = cause
+		conn = nil
+	})
+
+	connectorComponent, err := initializeConnectorComponent(nil, provider, params)
```
Cobalt0s, line 31: It is confusing that we are passing nil. Maybe remove the argument.

I was thinking that `goutils.PanicRecovery` is conceptually connected to `for _, opt := range opts` as thats the only source of panics. So pulling these 2 together and isolating in a different method would be nice. Otherwise is will be confusing to whoever looks at the method and sees PanicRecovery. Side effects should be grouped, otherwise reader would need to dig to find reasons of panicRecovery.

Cobalt0s (approved):
I re-read everything. It is good to go!

There is one concern about `PanicRecovery`. It used to be under each constructor init but then I noticed it shouldn't be and refactored into params.Apply. I think it is good to seal pandora box like this.
https://github.com/amp-labs/connectors/blob/39b56039f8bdabf1f76da2f86f36c59851e0fc03/common/paramsbuilder/params.go#L22-L40

internal/components/components.go
RajatPawar: This is a mistake, thanks for catching it

PR #1267, opened 2025-01-08
[ENG-1662] refactor: add CRUD operations and simple strategies

RajatPawar:
- Adds read, write, delete & metadata functions on to the base connector so that it satisfies Read/Write/Metadata/DeleteConnector interfaces
- Adds simple strategies that accept a client, a request builder and a response parser, for lack of a better term (feel free to suggest alternatives)


RajatPawar:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1267?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1268** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1268?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1267** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1267?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1267?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1266** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1266?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

internal/components/del/delete.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+type (
+	RequestBuilder func(context.Context, common.DeleteParams) (*http.Request, error)
+	ResponseParser func(context.Context, common.DeleteParams, *http.Response) (*common.DeleteResult, error)
+)
+
+var ErrFailedToDeleteObject = errors.New("failed to delete object")
+
+type SimpleDeleteStrategy struct {
```
jlimatampersand, line 20: Can you give some some examples of a DeleteStrategy that is not simple?

internal/components/metadata.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+
+	"github.com/amp-labs/connectors/common"
+)
+
+// MetadataStrategy describes an object's schema / metadata retrieval strategy. It can be done using a REST API,
+// OpenAPI file, etc.
+type MetadataStrategy interface {
+	GetObjectMetadata(ctx context.Context, objects ...string) (*common.ListObjectMetadataResult, error)
+	fmt.Stringer
```
jlimatampersand, line 14: what is the significance of this `fmt.Stringer`?

jlimatampersand (commented):
Can you add some read kind of documentation on what Strategy is and what they are for? I still not fully understanding what it means to say `Strategy` I think this will be question for every contributors.

internal/components/metadata/multiple.go
```diff
@@ -0,0 +1,56 @@
...
+	for _, strategy := range m.Strategies {
+		metadata, err := strategy.GetObjectMetadata(ctx, objects...)
+		if err != nil {
+			slog.Error("Strategy failed with error", "strategy", strategy, "error", err)
+
+			continue
+		}
+
+		for object, objectMetadata := range metadata.Result {
+			result.Result[object] = objectMetadata
```
Cobalt0s, line 45: Maybe we could shorten:
```
result.Result = metadata.Result
result.Errors = metadata.Result
```
This of course assumes that strategy returns a copy which is not shared.

I think the strategy fallback shouldn't just rely on err from the whole procedure. What if we called API and got an error, but we could reach out to the static file strategy and still return data that way? There should be documentation when to return an error and how result struct values could be overriden due to srategy fallback.

internal/components/metadata/openapi.go
```diff
@@ -0,0 +1,30 @@
...
+	module  common.ModuleID
+	schemas *staticschema.Metadata
+}
+
+func NewOpenAPIStrategy(schemas *staticschema.Metadata, module common.ModuleID) *OpenAPIStrategy {
+	return &OpenAPIStrategy{module: module, schemas: schemas}
+}
+
+func (i *OpenAPIStrategy) String() string {
+	return "metadata.OpenAPIStrategy"
```
Cobalt0s, line 22: This is great for debugging.

internal/components/components.go
```diff
@@ -13,6 +13,12 @@ type ConnectorComponent struct {
 	// are allowed on the connector.
 	ClientComponent
 	ProviderEndpointSupport
+
+	// These are the strategies that define how the connector interacts with the API for particular operations.
+	MetadataStrategy
+	ReadStrategy
+	WriteStrategy
```
Cobalt0s, line 20: Ok, I have returned from the smartlead PR. Should these be specified on case to case basis?

We could have BulkWriteStrategy, etc. Adding to the "parent class" would add it to all even if it will be a no-op it would be misleading to call it capable of bulk operations. I am not sure about the realstic implications, it could be premature Java lang induced OCD ahaha :wink: .

internal/components/del/delete.go
RajatPawar: It could be anything - batch deleting objects, you might have to send a post request with a request body instead of a DELETE, etc

RajatPawar: `Simple` is not the best word for it, maybe just `Base`?

internal/components/metadata.go
RajatPawar: Helps us identify what strategy is being used in case we use multiple strategies

internal/components/components.go
RajatPawar: It's a simple wrapper over a get request. Maybe we'll find a good composable form in between, or maybe we won't, but that's the idea!

internal/components/metadata/singleObject.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/amp-labs/connectors/common"
+)
+
+type (
+	RequestBuilder func(ctx context.Context, object string) (*http.Request, error)
+	ResponseParser func(ctx context.Context, response *http.Response) (*common.ObjectMetadata, error)
+)
+
+type SingleObjectEndpointStrategy struct {
```
Jkarage, line 15: This suggests we are fetching metadata for a single object in my head. What about `EndpointMetadataStrategy` or `EndpointStrategy`. The change could also be reflected in the filename. I get that it uses a single strategy.

internal/components/write/write.go
```diff
@@ -0,0 +1,57 @@
...
+) *SimpleWriteStrategy {
+	return &SimpleWriteStrategy{
+		Client:               client,
+		createRequestBuilder: createRequestBuilder,
+		updateRequestBuilder: updateRequestBuilder,
+		responseParser:       responseParser,
+	}
+}
+
+func (s *SimpleWriteStrategy) WriteObject(ctx context.Context, config common.WriteParams) (*common.WriteResult, error) {
```
Jkarage, line 36: Special reasons for `WriteObject`? Why not `Write`. Do we intend for this to be used for a single object writing?

Jkarage: Ok, I think I get this, If it's to avoid the `ConnectorComponent` to implement the WriteStratergy interface.

internal/components/metadata/singleObject.go
Cobalt0s: Good point. We could say that "single object" phrase is reserved by a different concept you are refering to.

RajatPawar: I've removed this for now, the only supported one is OpenAPI, will push this in a separate PR

internal/components/del/delete.go
```diff
@@ -0,0 +1,61 @@
...
+func (s *SimpleDeleteStrategy) DeleteObject(
+	ctx context.Context,
+	params common.DeleteParams,
+) (*common.DeleteResult, error) {
+	req, err := s.requestBuilder(ctx, params)
+	if err != nil {
+		return nil, err
+	}
+
+	rsp, err := s.Client.Do(req)
```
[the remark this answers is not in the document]
RajatPawar: Because we only have AuthenticatedClient for running requests/responses right now, this is what it is, but I think we could clean up and define a single interface for all the JSON/XML/CSV/HTTP clients and directly the interface methods like get/post/patch in here

PR #1268, opened 2025-01-08
[ENG-1662] refactor: use components for smartleadv2

RajatPawar:
Actually uses the previously defined components for implementing a smartleadv2 connector. I didn't want to make changes to the current connector to avoid breaking anything, so v2 it is.

## Test results
Existing tests pass -

### DIFF
--- a/common/json.go
+++ b/common/json.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"mime"
 	"net/http"
 	"strings"
@@ -113,6 +114,15 @@ func (j *JSONHTTPClient) Delete(ctx context.Context, url string, headers ...Head
 	}
 
 	return parseJSONResponse(res, body)
+}
+
+func ParseJSONResponse(res *http.Response) (*JSONHTTPResponse, error) {
+	b, err := io.ReadAll(res.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read response body: %w", err)
+	}
+
+	return parseJSONResponse(res, b)
 }
 
 // parseJSONResponse parses the given HTTP response and returns a JSONHTTPResponse.
--- a/providers/smartleadv2/errors.go
+++ b/providers/smartleadv2/errors.go
@@ -1,0 +1,61 @@
+package smartleadv2
+
+import (
+	"bytes"
+	"fmt"
+	"net/http"
+
+	"github.com/PuerkitoBio/goquery"
+	"github.com/amp-labs/connectors/common/interpreter"
+)
+
+var errorFormats = interpreter.NewFormatSwitch( // nolint:gochecknoglobals
+	[]interpreter.FormatTemplate{
+		{
+			MustKeys: []string{"message"},
+			Template: func() interpreter.ErrorDescriptor { return &ResponseMessageError{} },
+		}, {
+			MustKeys: []string{"error"},
+			Template: func() interpreter.ErrorDescriptor { return &ResponseBasicError{} },
+		},
+	}...,
+)
+
+type ResponseMessageError struct {
+	Message string `json:"message"`
+}
+
+func (r ResponseMessageError) CombineErr(base error) error {
+	if len(r.Message) != 0 {
+		return fmt.Errorf("%w: %s", base, r.Message)
+	}
+
+	return base
+}
+
+type ResponseBasicError struct {
+	Error string `json:"error"`
+}
+
+func (r ResponseBasicError) CombineErr(base error) error {
+	if len(r.Error) != 0 {
+		return fmt.Errorf("%w: %s", base, r.Error)
+	}
+
+	return base
+}
+
+func interpretHTMLError(res *http.Response, body []byte) error {
+	base := interpreter.DefaultStatusCodeMappingToErr(res, body)
+
+	document, err := goquery.NewDocumentFromReader(bytes.NewReader(body))
+	if err != nil {
+		// ignore HTML that cannot be understood
+		return base
+	}
+
+	// Message is located under the <pre></pre> tag
+	message := document.Find("pre").Text()
+
+	return fmt.Errorf("%w: %v", base, message)
+}
--- a/providers/smartleadv2/new.go
+++ b/providers/smartleadv2/new.go
@@ -1,0 +1,314 @@
+// nolint
+package smartleadv2
+
+import (
+	"bytes"
+	"context"
+	_ "embed"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strconv"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/interpreter"
+	"github.com/amp-labs/connectors/common/jsonquery"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+	"github.com/amp-labs/connectors/internal/components"
+	"github.com/amp-labs/connectors/internal/components/del"
+	"github.com/amp-labs/connectors/internal/components/metadata"
+	"github.com/amp-labs/connectors/internal/components/read"
+	"github.com/amp-labs/connectors/internal/components/write"
+	"github.com/amp-labs/connectors/providers"
+	"github.com/amp-labs/connectors/tools/fileconv"
+	"github.com/amp-labs/connectors/tools/scrapper"
+	"github.com/spyzhov/ajson"
+)
+
+// This reads the OpenAPI schema from the embedded file.
+var (
+	//go:embed schema.json
+	smartleadSchemas []byte
+
+	FileManager = scrapper.NewMetadataFileManager(smartleadSchemas, fileconv.NewSiblingFileLocator()) // nolint:gochecknoglobals
+
+	// Schemas is cached Object schemas.
+	Schemas = FileManager.MustLoadSchemas() // nolint:gochecknoglobals
+)
+
+// TODO: Could this be a part of the providerInfo object?
+const (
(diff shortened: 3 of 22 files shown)


RajatPawar:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1268?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1268** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1268?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1268?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1267** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1267?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1266** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1266?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

providers/smartleadv2/errors.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"bytes"
+	"fmt"
+	"net/http"
+
+	"github.com/PuerkitoBio/goquery"
+	"github.com/amp-labs/connectors/common/interpreter"
+)
+
```
RajatPawar, line 11: We can do some refactoring with errors to make it a little more intuitive, but I am reusing existing code for now.

common/json.go
```diff
@@ -115,6 +116,15 @@ func (j *JSONHTTPClient) Delete(ctx context.Context, url string, headers ...Head
 	return parseJSONResponse(res, body)
 }
 
+func ParseJSONResponse(res *http.Response) (*JSONHTTPResponse, error) {
```
Cobalt0s, line 119: I think it could be confusing when we are using `conn.JSON.HTTPClient.Client` alongside this method.
When do we make JSON requests via client and when via this method. 

This method must stay private and instead the caller should delegate to the Client. There should be one "enterance", not many.

Reason: any customization of JSON Client would be shared across.

providers/smartleadv2/errors.go
Cobalt0s: We should discuss 2 layers.
* Ampersand errors.
* Provider errors.

First one needs normalized list of errros across connectors. Thats to be defined.
Current file makes a first step to bridge `Provider errors` to `Ampersand errors`.

providers/smartleadv2/new.go
```diff
@@ -0,0 +1,314 @@
...
+	)
+
+	// Behavior overrides can go in here.
+	return conn, nil
+}
+
+// support returns the currently supported endpoints for the Smartlead connector.
+func support() (*components.ProviderEndpointSupport, error) {
+	a := map[common.ModuleID][]components.EndpointSupport{
+		"": {
```
Cobalt0s, line 110: `staticschema.RootModuleID`

providers/smartleadv2/new.go
```diff
@@ -0,0 +1,314 @@
...
+			{
+				Endpoint: "{campaigns,client,email-accounts}",
+				Support:  components.WriteSupport,
+			},
+
+			// Everything else is rejected.
+		},
+	}
+
+	return components.NewProviderEndpointSupport(a)
```
Cobalt0s, line 128: `a` variable name is short.

providers/smartleadv2/new.go
```diff
@@ -0,0 +1,314 @@
...
+	s, err := support()
+	if err != nil {
+		return nil, err
+	}
+
+	return components.Initialize(
+		providers.Smartlead,
+		params,
+		setup,
+		components.WithProviderEndpointSupport(*s),
```
Cobalt0s, line 66: Maybe WithProviderEndpointSupport should internally call `components.NewProviderEndpointSupport(a)`.
It seems like you want that method to be called eventually.

My goal is to remove `if err` check after support(). It would be nice to just give the `map[common.ModuleID][]components.EndpointSupport`

providers/smartleadv2/new.go
```diff
@@ -0,0 +1,314 @@
...
+	)
+}
+
+// setup is a constructor for the Smartlead connector.
+func setup(connectorComponent *components.ConnectorComponent) (*Connector, error) {
+	conn := &Connector{
+		ConnectorComponent: *connectorComponent,
+	}
+
+	conn.MetadataStrategy = metadata.NewMultipleStrategy(
```
Cobalt0s, line 80: Just a preference.

```
return &Connector{
	ConnectorComponent: *connectorComponent,
	ReadStrategy: read.NewSimpleReadStrategy(
		...
	),
	WriteStrategy: write.NewSimpleWriteStrategy(
		...
	),
	...
}
```

Cobalt0s (commented):
Smartlead has some tests, it would be nice if you could copy them over and run them. If they pass then the functionality should match with existing smartlead.

providers/smartleadv2/new.go
```diff
@@ -0,0 +1,314 @@
...
+		conn.JSON.HTTPClient.Client,
+		deleteRequestBuilder(conn),
+	)
+
+	// Behavior overrides can go in here.
+	return conn, nil
+}
+
+// support returns the currently supported endpoints for the Smartlead connector.
+func support() (*components.ProviderEndpointSupport, error) {
```
jlimatampersand, line 108: This support function seems to be an unnecessary layer. The map below is just a static map which I should be just part of providerInfo or some predefined variable and either passed down or imported and used here. But if that is the case, I would rather just directly call `components.NewProviderEndpointSupport(mapParamter)` from `NewConnector`

Do you foresee this map becoming dynamically manipulated later in this support function?

providers/smartleadv2/new.go
```diff
@@ -0,0 +1,314 @@
...
+			jsonResponse,
+			getRecords,
+			getNextRecordsURL,
+			common.GetMarshaledData,
+			params.Fields,
+		)
+	}
+}
+
+func createRequestBuilder(conn *Connector) write.RequestBuilder {
```
jlimatampersand, line 190: After looking into this, can we make this part of map of operationToRequestBuilder and map of operationToResponseBuilder? Then caller can just import 1 map and just use builders from the map?

```
const operationCreate = "create"
const operationRead = "read"
RequestBuilders ;= map[string]RequestBuilder{
  operationRead: ...
  operationCreate: ...
  ...
}
```

providers/smartleadv2/new.go
```diff
@@ -0,0 +1,314 @@
...
+		components.WithProviderEndpointSupport(*s),
+		components.WithErrorHandler(interpreter.ErrorHandler{
+			JSON: interpreter.NewFaultyResponder(errorFormats, nil),
+			HTML: &interpreter.DirectFaultyResponder{Callback: interpretHTMLError},
+		}),
+	)
+}
+
+// setup is a constructor for the Smartlead connector.
+func setup(connectorComponent *components.ConnectorComponent) (*Connector, error) {
```
jlimatampersand, line 75: This should be GetConnector or ConstructConnector for clarity.

providers/smartleadv2/new.go
```diff
@@ -0,0 +1,314 @@
...
+
+			// Everything else is rejected.
+		},
+	}
+
+	return components.NewProviderEndpointSupport(a)
+}
+
+// metadataSampleRequestBuilder makes a request to sample an object.
+// TODO: What happens when response is empty?
```
jlimatampersand, line 132: you mean when request is empty?

providers/smartleadv2/new.go
jlimatampersand: This make it easier for the caller to get right builder and less error prone.

providers/smartleadv2/new.go
jlimatampersand, line 110: what is the significance of this empty key?

jlimatampersand (changes requested):

providers/smartleadv2/new.go
Cobalt0s: Some connectors have multiple modules. For consistency, we consider that each connector has at least one module being the default. Empty module id indicates a default, one and only module for that connector. Instead of empty string we used `staticschema.RootModuleID` which is just "root" module. In code we treat "" and "root" as alias.

common/json.go
RajatPawar: For the time being, I've assumed a JSON client and returned it internally ([related comment](https://github.com/amp-labs/connectors/pull/1267#discussion_r1944238231)) but I agree, plenty of opportunity to follow up with a refactor here

providers/smartleadv2/errors.go
RajatPawar: Added a TODO to visit this in a follow up PR

providers/smartleadv2/new.go
RajatPawar: There's cases where we need to dynamically build this from OpenAPI files. If there's none, we can use it directly too

providers/smartleadv2/new.go
RajatPawar: this is a part of reorganization, let's tackle this in follow up PRs

providers/smartleadv2/new.go
RajatPawar: When we a sample a row from a read response to collect metadata, its possible that there is no data and we fail to collect metadata

PR #1272, opened 2025-01-09
[ENG-1792] fix: add workaround for hubspot when too many fields are requested

RajatPawar:
# Problem
When too many fields are requested in a normal GET call, the URL length can exceed web server limits, causing an error. 

# Fix
Hubspot has a `includeAllProperties` param, but it hasn't been implemented for all endpoints. The easier fix is to route potentially very long queries to the search endpoint where we list properties in the request body.

# Testing
## Normal read call when under the limit 
Screenshot 2025-01-09 at 6 20 26 PM

# Read call when over the limit 
Screenshot 2025-01-09 at 6 20 45 PM

### DIFF
--- a/providers/hubspot/read.go
+++ b/providers/hubspot/read.go
@@ -10,7 +10,7 @@ import (
 
 var (
 	// maxHubspotReadQueryLength is the maximum length of a query that should be sent to Hubspot. This is around 2-3K,
-	// but we're leaving room for error.
+	// but we add some buffer to be safe.
 	maxHubspotReadQueryLength = 1800
 )
 
--- a/providers/hubspot/read.go
+++ b/providers/hubspot/read.go
@@ -10,8 +10,8 @@ import (
 
 var (
 	// maxHubspotReadQueryLength is the maximum length of a query that should be sent to Hubspot. This is around 2-3K,
-	// but we're leaving some room for error.
-	maxHubspotReadQueryLength = 1
+	// but we're leaving room for error.
+	maxHubspotReadQueryLength = 1800
 )
 
 // Read reads data from Hubspot. If Since is set, it will use the
--- a/providers/hubspot/read.go
+++ b/providers/hubspot/read.go
@@ -8,6 +8,12 @@ import (
 	"github.com/amp-labs/connectors/common/logging"
 )
 
+var (
+	// maxHubspotReadQueryLength is the maximum length of a query that should be sent to Hubspot. This is around 2-3K,
+	// but we're leaving some room for error.
+	maxHubspotReadQueryLength = 1
+)
+
 // Read reads data from Hubspot. If Since is set, it will use the
 // Search endpoint instead to filter records, but it will be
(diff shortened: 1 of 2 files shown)


providers/hubspot/read.go
```diff
@@ -8,6 +8,12 @@ import (
 	"github.com/amp-labs/connectors/common/logging"
 )
 
+var (
+	// maxHubspotReadQueryLength is the maximum length of a query that should be sent to Hubspot. This is around 2-3K,
+	// but we're leaving room for error.
```
laurenzlong, line 13:
```suggestion
	// but we add some buffer to be safe.
```

providers/hubspot/read.go
```diff
@@ -105,3 +94,32 @@ func makeQueryValues(config common.ReadParams) []string {
...
 }
+
+func buildSearchParams(config common.ReadParams) SearchParams {
+	searchParams := SearchParams{
+		ObjectName: config.ObjectName,
+		SortBy: []SortBy{
+			BuildSort(ObjectFieldHsObjectId, SortDirectionAsc),
+		},
+		NextPage: config.NextPage,
+		Fields:   config.Fields,
```
laurenzlong, line 105: maybe i'm missing something but I don't actually see the code where you're replacing the fields list with `includeAllProperties`

providers/hubspot/read.go
```diff
@@ -26,28 +32,11 @@ func (c *Connector) Read(ctx context.Context, config common.ReadParams) (*common
...
-			},
-			SortBy: []SortBy{
-				BuildSort(ObjectFieldHsObjectId, SortDirectionAsc),
-			},
-			NextPage: config.NextPage,
-			Fields:   config.Fields,
-		}
+	// If filtering is required, then we have to use the search endpoint. This is also the case when the query is
+	// too long. The Search endpoint has a 10K record limit. In case this limit is reached, the sorting allows the
+	// caller to continue in another call by offsetting until the ID of the last record that was successfully fetched.
```
laurenzlong, line 37: So it sounds like only the search endpoint supports includeAllProperties, not the List endpoint? Can you add that to the comments if that's the case?

Also I'm forgetting why we even call List for backfills. Your statement "In case this limit is reached, the sorting allows the caller to continue in another call by offsetting until the ID of the last record that was successfully fetched." seems to indicate that we would be fine with calling the Search endpoint always, even for backfills?

laurenzlong (changes requested):

providers/hubspot/read.go
RajatPawar: Added this in the description - ```Hubspot has a includeAllProperties param, but it hasn't been implemented for all endpoints. The easier fix is to route potentially very long queries to the search endpoint where we list properties in the request body.
```

For search, we don't need to change anything since the request body can handle as many fields as required

RajatPawar:
Closing this because hubspot supports `includeAllFields` for only one or two objects, and the search endpoint also has a similar limitation around request body length (3K chars)

PR #1273, opened 2025-01-10
feat: Add Mock Tests Zoho CRM connector

Jkarage:
Thanks @Cobalt0s for the mockserver package. Really helpful.

### DIFF
--- a/providers/zohocrm/metadata.go
+++ b/providers/zohocrm/metadata.go
@@ -42,6 +42,7 @@ func (c *Connector) ListObjectMetadata(ctx context.Context,
 	for _, object := range objectNames {
 		go func(object string) {
 			object = naming.CapitalizeFirstLetterEveryWord(object)
+
 			metadata, err := c.getMetadata(ctx, object)
 			if err != nil {
 				mu.Lock()
--- a/providers/zohocrm/metadata.go
+++ b/providers/zohocrm/metadata.go
@@ -41,8 +41,6 @@ func (c *Connector) ListObjectMetadata(ctx context.Context,
 
 	for _, object := range objectNames {
 		go func(object string) {
-			object = naming.CapitalizeFirstLetterEveryWord(object)
-
 			metadata, err := c.getMetadata(ctx, object)
 			if err != nil {
 				mu.Lock()
--- a/providers/zohocrm/metadata.go
+++ b/providers/zohocrm/metadata.go
@@ -41,6 +41,7 @@ func (c *Connector) ListObjectMetadata(ctx context.Context,
 
 	for _, object := range objectNames {
 		go func(object string) {
+			object = naming.CapitalizeFirstLetterEveryWord(object)
 			metadata, err := c.getMetadata(ctx, object)
 			if err != nil {
 				mu.Lock()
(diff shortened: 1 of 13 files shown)


providers/zohocrm/metadata.go
```diff
@@ -40,6 +40,8 @@ func (c *Connector) ListObjectMetadata(ctx context.Context,
 
 	for _, object := range objectNames {
 		go func(object string) {
+			object = naming.CapitalizeFirstLetterEveryWord(object)
```
Cobalt0s, line 43: Could you add a comment why object name should be capital? Did it not work without this line?

Also object doesn't have spaces in between so it would be better to create a function that capitalizes the first letter only, because it "gives" a wrong hint about the object variable.

providers/zohocrm/read_test.go
```diff
@@ -0,0 +1,106 @@
...
+			Input:        common.ReadParams{ObjectName: "deals"},
+			Server:       mockserver.Dummy(),
+			ExpectedErrs: []error{common.ErrMissingFields},
+		},
+		{
+			Name:  "Unsupported object",
+			Input: common.ReadParams{ObjectName: "arsenal", Fields: connectors.Fields("assistant")},
+			Server: mockserver.Fixed{
+				Setup:  mockserver.ContentJSON(),
+				Always: mockserver.ResponseString(http.StatusBadRequest, string(unsupportedResponse)),
```
Cobalt0s, line 39:
```suggestion
				Always: mockserver.Response(http.StatusBadRequest, unsupportedResponse),
```
There is an option for `[]byte`. Please, use the same for other occurences in this PR

providers/zohocrm/metadata_test.go
```diff
@@ -0,0 +1,112 @@
...
+							"Record_Image":           "Record_Image",
+							"Record_Status__s":       "Record_Status__s",
+							"Sales_Cycle_Duration":   "Sales_Cycle_Duration",
+							"Stage":                  "Stage",
+							"Type":                   "Type",
+							"id":                     "id",
+						},
+					},
+				},
+				Errors: map[string]error{
```
Cobalt0s, line 79: Maybe for the `Errors` it would be good to have a dedicated test if you want. This test title is `successfully describe ...` which assumes no errors.
As I understand you would need to mock some request case in `mockserver.Switch` to remove this error for this test.

The other test could omit `Result` and demonstrate that `Errors` property is correctly constructed.

PS: I was thinking to create `testroutines.NEW_METHOD_FOR_METADATA_ERRORS` to handle comparison, but all deep connectors I was working with didn't have it. In tests, it is not good to depend on deeply nested errors, what I mean is `common.ErrCaller` is not important while `string(arsenalResponse)` is what we are after. Thats the goal of utility method.

Cobalt0s (approved):
I left some comments, but overall the tests look great, well done!

providers/zohocrm/metadata.go
Cobalt0s: This could be a breaking functionality change, or a fix, please let me know what is it.

Jkarage: Thanks, this was for some tests.

PR #1276, opened 2025-01-13
feat: Add MockTests Outreach connector

Jkarage:


providers/outreach/metadata_test.go
```diff
@@ -0,0 +1,88 @@
...
+	tests := []testroutines.Metadata{
+		{
+			Name:         "At least one object name must be provided",
+			Input:        nil,
+			Server:       mockserver.Dummy(),
+			ExpectedErrs: []error{common.ErrMissingObjects},
+		},
+		{
+			Name:  "Successfully describe supported objects",
+			Input: []string{"mailings", "opportunities"},
```
Cobalt0s, line 29: You should either remove mailings or include mailings in the expected output. Mock server would need to handle `/v2/mailings` as well.

PS: This means that `testroutines.ComparatorSubsetMetadata` has a "bug". While it should compare only subset of fields it shouldn't skip over the objects.

providers/outreach/write_test.go
```diff
@@ -0,0 +1,254 @@
...
+			},
+		},
+		{
+			Name:         "RecordID must be convertable to integers",
+			Input:        common.WriteParams{ObjectName: "prospects", RecordData: map[string]any{}, RecordId: "xseedesrt"},
+			Server:       mockserver.Dummy(),
+			ExpectedErrs: []error{ErrIdMustInt},
+		},
+		{
+			Name: "Successfully creation of a sequence",
```
Cobalt0s, line 57:
```suggestion
			Name: "Successful creation of a sequence",
```

providers/outreach/write_test.go
```diff
@@ -0,0 +1,254 @@
...
+						"finishOnReply":               true,
+						"interestedCount":             nil,
+						"lastUsedAt":                  "2025-01-13T12:47:31.000Z",
+						"locked":                      false,
+						"lockedAt":                    nil,
+						"maxActivations":              float64(150),
+						"name":                        "string",
+						"negativeReplyCount":          float64(0),
+						"neutralReplyCount":           float64(0),
+						"numContactedProspects":       float64(0),
```
Cobalt0s, line 97:
```suggestion
```
Since you have already specified `Comparator: testroutines.ComparatorSubsetWrite,` this means you could remove this line and it will still work. Of course if you change float64(0) to float64(37) then the test will break as expected.
You could remove most such fields as they don't hold much value.

providers/outreach/write_test.go
```diff
@@ -0,0 +1,254 @@
...
+						"throttlePausedAt":      nil,
+						"transactional":         false,
+						"updatedAt":             "2025-01-13T12:47:31.000Z",
+					},
+					"id": float64(18),
+					"links": map[string]any{
+						"self": "https://api.outreach.io/api/v2/sequences/18",
+					},
+
+					"relationships": map[string]any{
```
Cobalt0s, line 129: This is a very complicated field. This should be removed. Other fields already demonstrate the write capability.

Cobalt0s (approved):
LGTM

providers/outreach/write_test.go
Jkarage: Well these are mandatory unless, I remove the whole `attributes` field. Since if field is available DeepEqual is used to check for it's equality.

PR #1291, opened 2025-01-20
feat: Static schema supports V1 and V2 Fields Metadata

Cobalt0s:
# Background

There were recent changes to the format of the ObjectMetadata that is returned by ListObjectMetadata.

# Changes
To accommodate these updates, newly created static schema files may need to follow a revised structure. Each reference to the static file structure must now specify a generic type of either `ObjectFieldsV1` or `ObjectFieldsV2`.

* Created new type ObjectFields which is either ObjectFieldsV1(map[string]string) or ObjectFieldsV2 (map[string]common.FieldMetadata)
* Each connector script that parses OpenAPI was affected and now uses the existing `ObjectFieldsV1`.
* Files located under `.../<provider_name>/metadata` were previously named `scrapper.go`. While we are at it, renamed them to be `schemas.go`, because they do not perform scraping but instead reference a constant metadata structure. For context, originally first files were created using scrapping.

# Review
Main code changes are under `internal/staticschema/core.go` file.

# Tests
Ran each OpenAPI script. It produces the same output as before.

### DIFF
--- a/internal/staticschema/convertors.go
+++ b/internal/staticschema/convertors.go
@@ -1,18 +1,14 @@
 package staticschema
 
 import (
-	"errors"
 	"fmt"
 
 	"github.com/amp-labs/connectors/common"
-	"github.com/amp-labs/connectors/internal/datautils"
 )
-
-var ErrObjectNotFound = errors.New("object not found")
 
 // Select will look for object names under the module and will return metadata result for those objects.
 // NOTE: empty module id is treated as root module.
-func (r *Metadata) Select(
+func (m *Metadata[F]) Select(
 	moduleID common.ModuleID, objectNames []string,
 ) (*common.ListObjectMetadataResult, error) {
 	if len(objectNames) == 0 {
@@ -22,7 +18,7 @@ func (r *Metadata) Select(
 	moduleID = moduleIdentifier(moduleID)
 
 	// Convert and return only listed objects
-	module, ok := r.Modules[moduleID]
+	module, ok := m.Modules[moduleID]
 	if !ok {
 		return nil, fmt.Errorf("%w: connector is using unknown module [%v]", common.ErrMissingModule, moduleID)
 	}
@@ -36,12 +32,9 @@ func (r *Metadata) Select(
 	for _, objectName := range objectNames {
 		if v, ok := module.Objects[objectName]; ok {
 			// move metadata from scrapper object to common object
-			list.Result[objectName] = common.ObjectMetadata{
-				DisplayName: v.DisplayName,
-				FieldsMap:   datautils.FromMap(v.FieldsMap).ShallowCopy(),
-			}
+			list.Result[objectName] = *v.getObjectMetadata()
 		} else {
-			return nil, fmt.Errorf("%w: unknown object [%v]", ErrObjectNotFound, objectName)
+			return nil, fmt.Errorf("%w: unknown object [%v]", common.ErrObjectNotSupported, objectName)
 		}
 	}
 
@@ -50,29 +43,20 @@ func (r *Metadata) Select(
 
 // SelectOne reads one object metadata from the static file.
 // NOTE: empty module id is treated as root module.
-func (r *Metadata) SelectOne(
+func (m *Metadata[F]) SelectOne(
 	moduleID common.ModuleID, objectName string,
 ) (*common.ObjectMetadata, error) {
 	moduleID = moduleIdentifier(moduleID)
 
 	// Convert and return only listed objects
-	module, ok := r.Modules[moduleID]
+	module, ok := m.Modules[moduleID]
 	if !ok {
 		return nil, fmt.Errorf("%w: connector is using unknown module [%v]", common.ErrMissingModule, moduleID)
 	}
 
-	mtd := common.ObjectMetadata{
-		FieldsMap: make(map[string]string),
-	}
-
 	if v, ok := module.Objects[objectName]; ok {
-		mtd = common.ObjectMetadata{
-			DisplayName: v.DisplayName,
-			FieldsMap:   v.FieldsMap,
-		}
-	} else {
-		return nil, fmt.Errorf("%w: unknown object [%v]", ErrObjectNotFound, objectName)
+		return v.getObjectMetadata(), nil
 	}
 
-	return &mtd, nil
+	return nil, fmt.Errorf("%w: unknown object [%v]", common.ErrObjectNotSupported, objectName)
 }
--- a/internal/staticschema/core.go
+++ b/internal/staticschema/core.go
@@ -60,6 +60,9 @@ type (
 	}
 )
 
+// All maps must be copies of the original map, which remains immutable.
+// The immutable map serves as a unified registry of metadata for the connector.
+// ListObjectMetadata returns a copy that consumers are allowed to modify as needed.
 func (o Object[F]) getObjectMetadata() *common.ObjectMetadata {
 	if fieldsMap, isV1 := (any(o.Fields)).(FieldMetadataMapV1); isV1 {
 		return &common.ObjectMetadata{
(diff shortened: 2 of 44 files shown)


Cobalt0s:
@RajatPawar I decided to add static schema file to Hubspot. It needs to follow a new ObjectMetedata schema. Could you have a look?

internal/staticschema/core.go
```diff
@@ -45,39 +45,102 @@ type Object struct {
 	ResponseKey string `json:"responseKey"`
 
 	// FieldsMap is a map of field names to field display names
-	FieldsMap map[string]string `json:"fields"`
+	FieldsMap F `json:"fields"`
```
jlimatampersand, line 48: let's rename `FieldMap` to `FieldMetadataMap`

internal/staticschema/core.go
```diff
@@ -45,39 +45,102 @@ type Object struct {
...
+}
+
 // Add will appropriately store the data, abiding to data structure rules.
 // NOTE: empty module id is treated as root module.
-func (r *Metadata) Add(
+func (r *Metadata[F]) Add(
 	moduleID common.ModuleID,
-	objectName, objectDisplayName, fieldName, urlPath, responseKey string,
+	objectName, objectDisplayName, urlPath, responseKey string,
+	records F,
```
jlimatampersand, line 88: records is misleading

scripts/openapi/constantcontact/metadata/main.go
```diff
@@ -70,7 +70,7 @@ func main() {
 	)
 	goutils.MustBeNil(err)
 
-	schemas := staticschema.NewMetadata()
+	schemas := staticschema.NewMetadata[staticschema.ObjectFieldsV1]()
```
jlimatampersand, line 73: rename this to metadata

scripts/openapi/customerio/journeysapp/metadata/main.go
```diff
@@ -47,7 +47,7 @@ func main() {
 	)
 	goutils.MustBeNil(err)
 
-	schemas := staticschema.NewMetadata()
+	schemas := staticschema.NewMetadata[staticschema.ObjectFieldsV1]()
```
jlimatampersand, line 50: same here, metadata

scripts/openapi/gong/metadata/main.go
```diff
@@ -69,7 +69,7 @@ func main() {
 	)
 	goutils.MustBeNil(err)
 
-	schemas := staticschema.NewMetadata()
+	schemas := staticschema.NewMetadata[staticschema.ObjectFieldsV1]()
```
jlimatampersand, line 72: same here and elsewhere

internal/staticschema/core.go
jlimatampersand: fieldMetadata or fieldMatadataMap

internal/staticschema/core.go
```diff
@@ -45,39 +45,102 @@ type Object struct {
...
 
 	// DocsURL points to docs endpoint. Optional.
 	DocsURL *string `json:"docs,omitempty"`
 }
 
+type (
+	ObjectFieldsV1 map[string]string
+	ObjectFieldsV2 map[string]common.FieldMetadata
+
+	ObjectFields interface {
```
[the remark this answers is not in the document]
jlimatampersand: Also, I think it would be so much easier if we have this interface to have method 

Version() string

which will return "v1" or "v2" accordingly. 

Then all of the version checking logic in below should be so much simpler. 

then for v1, we can just assert map[string]string, then we don't need `datautils.FromMap(fieldsMap).ShallowCopy()` this copy call since they are the same type

internal/staticschema/core.go
```diff
@@ -45,39 +45,102 @@ type Object struct {
...
+		ObjectFieldsV1 | ObjectFieldsV2
+	}
+)
+
+func (v Object[F]) getObjectMetadata() *common.ObjectMetadata {
```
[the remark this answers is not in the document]
Cobalt0s: Of course! I believe the struct was renamed and recievers were not. It is weird to even read it.

scripts/openapi/constantcontact/metadata/main.go
Cobalt0s: This will collide with the package name metadata.
At the end of the script that extracted metadata from OpenAPI we save into the file: 
```
goutils.MustBeNil(metadata.FileManager.SaveSchemas(schemas))
```

Cobalt0s: I think if we are to add this renaming change then we should rename `schema.json` file into `metadata.json`.
There are so many things called metadata that it is hard to tell the difference between concepts

internal/staticschema/core.go
Cobalt0s: I will rename it to match `common.ObjectMetadata`.
Version 1 was `FieldsMap map[string]string`
Version 2 is `Fields map[string]FieldMetadata`

internal/staticschema/convertors.go
```diff
@@ -61,18 +57,9 @@ func (r *Metadata) SelectOne(
...
-			DisplayName: v.DisplayName,
-			FieldsMap:   v.FieldsMap,
-		}
-	} else {
-		return nil, fmt.Errorf("%w: unknown object [%v]", ErrObjectNotFound, objectName)
+		return v.getObjectMetadata(), nil
 	}
 
-	return &mtd, nil
+	return nil, fmt.Errorf("%w: unknown object [%v]", ErrObjectNotFound, objectName)
```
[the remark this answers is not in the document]
Cobalt0s: common package already has an error which would replace the outdated and also not appropriate not found.
Change:
```
return nil, fmt.Errorf("%w: unknown object [%v]", common.ErrObjectNotSupported, objectName)

For reference:
ErrObjectNotSupported = errors.New("operation is not supported for this object")
```

internal/staticschema/core.go
```diff
@@ -45,39 +45,102 @@ type Object struct {
...
+			fieldsMap[k] = v
+		}
+
+		for k, v := range (any(records)).(ObjectFieldsV1) { // nolint:forcetypeassert
+			fieldsMap[k] = v
+		}
+
+		object.FieldsMap = any(fieldsMap).(F) // nolint:forcetypeassert
+
+		return
```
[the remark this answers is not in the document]
Cobalt0s: Yes, they do look identical but I cannot simplify it for multiple reasons.
Generics is used to change struct version very fast without the need to create a duplicate manually but that comes with a price.

First of all while `F` is either map of string to string or map of string to FieldMetadata struct. In either case it is a map but golang doesn't support range 
image
Reason: Range expressions must have a core type.

Secondly, type assertions and conversion from concrete type back to the most general any allows conversion to compiste `F`. This requires `isV1` type check. One check is more than enough for method to be valid. But the rest are suggested by linter. Some I ignored with `// nolint:forcetypeassert`.

internal/staticschema/core.go
```diff
@@ -45,39 +45,102 @@ type Object struct {
...
+	ObjectFieldsV1 map[string]string
+	ObjectFieldsV2 map[string]common.FieldMetadata
+
+	ObjectFields interface {
+		ObjectFieldsV1 | ObjectFieldsV2
+	}
+)
+
+func (v Object[F]) getObjectMetadata() *common.ObjectMetadata {
+	if fieldsMap, isV1 := (any(v.FieldsMap)).(ObjectFieldsV1); isV1 {
```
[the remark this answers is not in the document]
Cobalt0s: Here we must have type assertion to bridge from generic type to concrete type.
If generic F is v1 then set `common.ObjectMetadata.FieldsMap` map[string]string otherwise set `common.ObjectMetadata.Fields` map[string]FieldMetadata.

Shallow copy is needed because while we store static metadata for the whole connector this information should be `**immutable**` and ListObjectMetadata should always grab a copy. There was once a bug because of this.

internal/staticschema/core.go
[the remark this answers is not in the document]
Cobalt0s: Shallow copy is required as connectors registry of metadata should stay immutable to avoid bugs caused by reference modifications.

jlimatampersand (approved):

PR #1279, opened 2025-01-14
[CON-972] feat(Hubspot): Read Lists object

Cobalt0s:
# Background

Lists object while being part of CRM is not part of Objects API. I put it under `crmObjectsOutsideTheObjectAPI` category.

Lists can be quired using 
* `crm/v3/lists` - relies on query params and knowing ids beforehand, otherwise always returns empty array
* `crm/v3/lists/search` - returns all lists even on empty payload. There is no incremental read. Payload gives control over pagination.

# Changes

Read for `Lists` object defaults to SearchCRM method calling `crm/v3/search`. Next Page token is the offset integer that is used by `Lists` pagination.

# Tests
image

Modifying page size to 2 and giving NextPage as `2`, meaning offset 2 returns the last page with last third list.
image

## Old Tests
Read: Contacts
image
Read+Write Contacts
image
As you can see the reading in read and write doesn't always return contacts, I noticed the same on main. After running the test again I get 
image

# Docs

Updated docs: https://github.com/amp-labs/docs/pull/147

PR #1278, opened 2025-01-14
tidy(Hubspot): Bootstrap Read tests

Cobalt0s:
# Changes

* Given more API endpoints will be called I have seperated building existing API endpoints into dedicated functions and shortening the conditional branching.
* Bootstrap mock read tests.
* Added dedicated hubpost contacts read. See screenshots below.

# Testing
image

PR #1292, opened 2025-01-20
[CON-972] feat(Hubspot): Support Lists object by ListObjectMetadata

Cobalt0s:
# Problem

Lists is an object that lives under CRM module but is outside Objects APIs. Current implementation of `ListObjectMetadata` focusses on CRM Objects API that uses field discovery endpoint. This is not applicable for `Lists`.

# Solution
As with other connectors the approach is to make read request for 1 item and use response fields as metadata fields. As a fallback rely on static schema file for the object metadata. This same approach takes place in this PR. Lists object is added under schema.json for the fallback.

# Tests - ListObjectMetadata
Lists
image
image
Contacts
image
image

### DIFF
--- a/providers/hubspot/metadata.go
+++ b/providers/hubspot/metadata.go
@@ -5,11 +5,12 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
-	"time"
 
+	"github.com/amp-labs/connectors"
 	"github.com/amp-labs/connectors/common"
 	"github.com/amp-labs/connectors/common/logging"
 	"github.com/amp-labs/connectors/internal/datautils"
+	"github.com/amp-labs/connectors/providers/hubspot/metadata"
 )
 
 type objectMetadataResult struct {
@@ -42,7 +43,7 @@ func (c *Connector) ListObjectMetadata( // nolint:cyclop,funlen
 
 	for _, objectName := range objectNames {
 		go func(object string) {
-			objectMetadata, err := c.describeObject(ctx, object)
+			objectMetadata, err := c.getObjectMetadata(ctx, object)
 			if err != nil {
 				errChannel <- &objectMetadataError{
 					ObjectName: object,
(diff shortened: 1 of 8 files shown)


Cobalt0s:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1292?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1292** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1292?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1292?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1291** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1291?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1283** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1283?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1279** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1279?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1278** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1278?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

jlimatampersand (approved):
Mostly looking good. 

Pre-approving, but please take a look at my comment and change the method names so that we can make the library more consistent. 

Thanks!

providers/hubspot/metadata.go
```diff
@@ -80,6 +81,19 @@ func (c *Connector) ListObjectMetadata( // nolint:cyclop,funlen
 
 // describeObject returns object metadata for the given object name.
 func (c *Connector) describeObject(ctx context.Context, objectName string) (*common.ObjectMetadata, error) {
```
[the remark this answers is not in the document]
Cobalt0s: I agree. I went alone with the existing names but I honestly cannot tell what they do without checking the call stack.
Also, `lists` object is interesting as it is part of CRM but is not part of ObjectsAPI. There are too many sets to name in the "Venn diagram". I even thought about flipping the names, where all existing would be prefixed with ObjectsAPI while the rest would become the new "normal/default".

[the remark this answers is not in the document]
Cobalt0s: Ok, what do you think of:
getObjectMetadata ->
* getObjectMetadataCRM -- this should be objects outside ObjectAPI but part of CRM. For now it is only `Lists`.
* getObjectMetadataCRMCoreObjects -- these are all supported objects by Hubspot deep connector prior to introducing `Lists`. Calling them core could be misleading but calling them just `objects` as in docs would be hard to differentiate :sob: . 
 
image

providers/hubspot/metadata.go
```diff
@@ -78,8 +79,21 @@ func (c *Connector) ListObjectMetadata( // nolint:cyclop,funlen
...
 
-// describeObject returns object metadata for the given object name.
-func (c *Connector) describeObject(ctx context.Context, objectName string) (*common.ObjectMetadata, error) {
+// getObjectMetadata returns object metadata for the given object name.
+func (c *Connector) getObjectMetadata(ctx context.Context, objectName string) (*common.ObjectMetadata, error) {
+	if crmObjectsOutsideTheObjectAPI.Has(objectName) {
+		return c.getObjectMetadataCRM(ctx, objectName)
+	}
+
+	return c.getObjectMetadataCRMCoreObjects(ctx, objectName)
```
jlimatampersand, line 88: I think we can do something like

`getObjectMetadataFromPropetyAPI`
`getObjectMetadataFromObjectAPI`

jlimatampersand: I don't actually like it either because it is verbose, but at least it is clear

Cobalt0s: Sometimes I am ok with Java styled method names, the choice of words can get hard as some ideas cannot be described with one word. Combining such gives a very crazy sentence ahaha.

Cobalt0s: The names you proposed I think make the most sense. I believe the core idea is represented.

PR #1301, opened 2025-01-23
feat: Add Marketo Mock Tests

Jkarage:


providers/marketo/test/channels.json
```diff
@@ -0,0 +1,563 @@
+{
```
Cobalt0s, line 1: The file is quite big. Likely most of the items are not needed. You should shrink it. This will be sampling the real response.

providers/marketo/metadata_test.go
```diff
@@ -0,0 +1,83 @@
...
+			Name:  "Successfully describe supported object",
+			Input: []string{"channels"},
+			Server: mockserver.Switch{
+				Setup: mockserver.ContentJSON(),
+				Cases: []mockserver.Case{{
+					If:   mockcond.PathSuffix("v1/channels.json"),
+					Then: mockserver.Response(http.StatusOK, channelsResponse),
+				}},
+			}.Server(),
+			// Comparator: testroutines.ComparatorSubsetMetadata,
```
Cobalt0s, line 37: Remove comment.

providers/marketo/read_test.go
```diff
@@ -0,0 +1,101 @@
...
+			Input:        common.ReadParams{ObjectName: "deals"},
+			Server:       mockserver.Dummy(),
+			ExpectedErrs: []error{common.ErrMissingFields},
+		},
+		{
+			Name:  "Unsupported object",
+			Input: common.ReadParams{ObjectName: "arsenal", Fields: datautils.NewStringSet("testField")},
+			Server: mockserver.Fixed{
+				Setup:  mockserver.ContentJSON(),
+				Always: mockserver.ResponseString(http.StatusBadRequest, string(unsupportedResponse)),
```
Cobalt0s, line 40:
```suggestion
				Always: mockserver.Response(http.StatusBadRequest, unsupportedResponse),
```
It is already of `[]byte` type. True for other places in the PR.

providers/marketo/read_test.go
```diff
@@ -0,0 +1,101 @@
...
+						"active":        false,
+						"createdAt":     "2024-08-23T12:09:55Z",
+						"id":            float64(1023),
+						"name":          "Meme",
+						"type":          "batch",
+						"updatedAt":     "2024-08-23T12:09:55Z",
+						"workspaceName": "Default",
+					},
+				}},
+				NextPage: "",
```
Cobalt0s, line 85: There should be a test, or just modify this one to check nextPage is correctly implied or extracted from the provider response.
The implementation is looking for `nextPageToken` response field.

providers/marketo/write_test.go
```diff
@@ -0,0 +1,88 @@
...
+			}.Server(),
+			ExpectedErrs: []error{
+				common.ErrRetryable,
+				errors.New(string(unsupportedResponse)), // nolint:goerr113
+			},
+		},
+		{
+			Name: "Successfully creation of a lead",
+			Input: common.WriteParams{ObjectName: "leads", RecordData: map[string]any{
+				"input": []map[string]any{
```
Cobalt0s, line 51: Often when you specify concrete, real input payload it means that you want to assert data will be passed to the mock server in a certain form.
Ex: the information coming from payload will end up in the URL. Thats how catalogs are implemented for iterable. So the check later asserts this expectation.
https://github.com/amp-labs/connectors/blob/f8c7b390e5d8f24040c78f69e9882981ae6766c5/providers/iterable/write_test.go#L87-L98

Usually the concrete content of the payload we pass has no impact on a test, therefore we ignore it. This would be helpful for the sample script to demonstrate how to make a request but sadly not for the mock test.

Cobalt0s (approved):

providers/marketo/write_test.go
Jkarage: Thanks, Added the condition for the Url created.

PR #1303, opened 2025-01-25
[CON-972] fix(Hubspot): Search makes a call to Lists

Cobalt0s:
# Changes

* Read works for both groups of objects. For the group that for now only has lists it delegates to internal search.
* Search works for both groups of objects. For lists branches out and calls internal search (similar to read) 
* Added dedicated tests for lists ("new" object) and contacts(existing object).
image

# Tests

### Contacts
image
image
### Deals
image
### Lists
It depends on Read and therefore on `searchCRMOutsideThePropertiesAPI`. Metadata is returned as expected.
image

### Contacts
image
image
### Lists
image
image

### Contacts
image
### Lists
image

### Contacts
image
### Lists
image


Cobalt0s:
* **#1303** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1303?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1303?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

providers/hubspot/metadata.go
```diff
@@ -127,7 +127,7 @@ func (c *Connector) getObjectMetadataFromPropertyAPI(
 func (c *Connector) getObjectMetadataFromObjectAPI(
 	ctx context.Context, objectName string,
 ) (*common.ObjectMetadata, error) {
-	readResult, err := c.SearchCRM(ctx, SearchCRMParams{
+	readResult, err := c.searchCRMOutsideThePropertiesAPI(ctx, searchCRMParams{
```
Cobalt0s, line 130: SearchCRM renamed to local private method. It deals with `Lists`.

providers/hubspot/read.go
```diff
@@ -24,10 +24,10 @@ func (c *Connector) Read(ctx context.Context, config common.ReadParams) (*common
 	if crmObjectsOutsideThePropertiesAPI.Has(config.ObjectName) {
 		// Objects outside ObjectAPI have different endpoint while both are part of CRM module.
 		// For instance Lists are fully returned only via Search endpoint.
-		return c.SearchCRM(ctx, SearchCRMParams{
+		return c.searchCRMOutsideThePropertiesAPI(ctx, searchCRMParams{
```
Cobalt0s, line 27: Method rename as above. The input param is also renamed and made private.

providers/hubspot/search.go
```diff
@@ -21,6 +21,16 @@ func (c *Connector) Search(ctx context.Context, config SearchParams) (*common.Re
 		return nil, err
 	}
 
+	if crmObjectsOutsideThePropertiesAPI.Has(config.ObjectName) {
```
Cobalt0s, line 24: If somebody wants to call Search with `Lists` it is possible and will call the private method that deals exclusively with lists search.

providers/hubspot/search.go
```diff
@@ -50,19 +60,15 @@ func (c *Connector) Search(ctx context.Context, config SearchParams) (*common.Re
...
+func (c *Connector) searchCRMOutsideThePropertiesAPI(
+	ctx context.Context, config searchCRMParams,
+) (*common.ReadResult, error) {
 	url, err := c.getCRMSearchURL(config)
 	if err != nil {
 		return nil, err
 	}
 
-	payload, err := config.Payload()
+	payload, err := config.payload()
```
Cobalt0s, line 71: Struct is private and its methods are now private for consistency.

providers/hubspot/search.go
```diff
@@ -50,19 +60,15 @@ func (c *Connector) Search(ctx context.Context, config SearchParams) (*common.Re
 //   - Search always returns an array of items, unlike the usual "read" operation.
 //     Therefore, the "retrieve" API endpoint is not used
 //     https://developers.hubspot.com/docs/guides/api/crm/lists/overview#retrieve-lists
-func (c *Connector) SearchCRM(ctx context.Context, config SearchCRMParams) (*common.ReadResult, error) {
-	ctx = logging.With(ctx, "connector", "hubspot")
-
-	if err := config.ValidateParams(); err != nil {
```
jlimatampersand, line 56: is there reason we got rid of the validation?

providers/hubspot/search.go
```diff
@@ -21,6 +21,16 @@ func (c *Connector) Search(ctx context.Context, config SearchParams) (*common.Re
 		return nil, err
 	}
 
+	if crmObjectsOutsideThePropertiesAPI.Has(config.ObjectName) {
+		// Objects outside ObjectAPI have different endpoint while both are part of CRM module.
+		// For instance such object is Lists.
+		return c.searchCRMOutsideThePropertiesAPI(ctx, searchCRMParams{
```
jlimatampersand, line 27: it seems like unnecessary to define new `searchCRMParams` just to copy over from existing struct. 

we can just config and use it as is.

providers/hubspot/search.go
jlimatampersand: if `func (p searchCRMParams) payload()` is the only method you are using with this `searchCRMParams` is this `payload`, then I would just refactor and use `SearchParams` as config type and just create 1 locally scoped function create payload from `SearchParams` object.

jlimatampersand (changes requested):

providers/hubspot/search.go
Cobalt0s: This is a private method and caller performs validation it is similar to `Search()`. I think, since we would keep `searchCRMParams` maybe for consistency I would return the validation back

providers/hubspot/search.go
Cobalt0s: This is not the primary reason for `searchCRMParams`. I also wanted to get rid of it and collapse it into usage of one `SearchParams`.

The main reason is `ListObjectMetadata` calls `searchCRMOutsideThePropertiesAPI` and adds page size set to 1, a special case for reasing.

I think to better communicate this intent I will use embedded structs.

providers/hubspot/search.go
jlimatampersand, line 24: okay, so I thought about some of these objects and function names and I think we can make it little more clear.
outside the properties API sounds other objects can be searched from properties API, but we are just getting metadata. So below can be more clear. 

`crmObjectsOutsideThePropertiesAPI` -> `properiesAPIUnsupportedCRMObjects`

providers/hubspot/search.go
```diff
@@ -50,19 +60,15 @@ func (c *Connector) Search(ctx context.Context, config SearchParams) (*common.Re
...
 //     Therefore, the "retrieve" API endpoint is not used
 //     https://developers.hubspot.com/docs/guides/api/crm/lists/overview#retrieve-lists
-func (c *Connector) SearchCRM(ctx context.Context, config SearchCRMParams) (*common.ReadResult, error) {
-	ctx = logging.With(ctx, "connector", "hubspot")
-
-	if err := config.ValidateParams(); err != nil {
-		return nil, err
-	}
-
+func (c *Connector) searchCRMOutsideThePropertiesAPI(
```
jlimatampersand, line 63: This can stay as `searchCRM` since this method has nothing to do with Properties API, but

providers/hubspot/search.go
jlimatampersand: nvm, I see why this is needed. good call.

providers/hubspot/metadata.go
jlimatampersand: so we can `searchCRMOutsideThePropertiesAPI` change this back to `searchCRM` since this method is just searching CRM. 

and the caller `getObjectMetadataFromObjectAPI` can be renamed to `getObjectMetadataFromCRMSearch`

providers/hubspot/search.go
jlimatampersand: `CRMObjectsWithoutPropertiesAPISupport`

jlimatampersand (approved):
Looks good, overall and thanks for explaining things. 

I came up with better object/method namings for clarify

crmObjectsOutsideThePropertiesAPI -> properiesAPIUnsupportedCRMObjects or CRMObjectsWithoutPropertiesAPISupport

searchCRMOutsideThePropertiesAPI -> back to `searchCRM`

`getObjectMetadataFromObjectAPI` -> `getObjectMetadataFromCRMSearch`

Feel free to merge after renaming these

providers/hubspot/search.go
Cobalt0s: This is good! I will keep access private for now.

Cobalt0s:
@jlimatampersand , sorry I got a bit interrupted between writing the comments and pushing the code.

Thank you for review. To integrate your ideas I decided to embed one search params into the other. Check finalized commit with all the changes: https://github.com/amp-labs/connectors/pull/1303/commits/e57a93ddb181d63cbacaaf2721e288282ce5f2c1

This should communicate a direct intent of search params being the same, where there is only pageSize difference. Also the choice of names is better across vars/methods.

PR #1304, opened 2025-01-27
feat: Add ListMetadata + Read + Write to Asana connector 

immdipu:
# Testing

-  List Metadata
image

- Read 
image

- Write
image


providers/asana/metadata_test.go
```diff
@@ -0,0 +1,97 @@
...
+			Name:         "Object must be included",
+			Server:       mockserver.Dummy(),
+			ExpectedErrs: []error{common.ErrMissingObjects},
+		},
+
+		{
+			Name:         "Unknown object requested",
+			Input:        []string{"groot"},
+			Server:       mockserver.Dummy(),
+			ExpectedErrs: []error{staticschema.ErrObjectNotFound},
```
Cobalt0s, line 28: This error can be replaced with `common.ErrObjectNotSupported`.

providers/asana/read_test.go
```diff
@@ -0,0 +1,123 @@
...
+					},
+				},
+				Done: true,
+			},
+			ExpectedErrs: nil,
+		},
+		{
+			Name:  "Read list of all tags",
+			Input: common.ReadParams{ObjectName: "tags", Fields: connectors.Fields("")},
+			Server: mockserver.Fixed{
```
Cobalt0s, line 92: Could you add a conditional mock server to check that deep connector produces the URL path as expected. In case somebody changes that part of code this test would break/notify.

Ex: 
https://github.com/amp-labs/connectors/blob/c548efa81b6a714bc4339c95bdd8742e90ddd30b/providers/iterable/read_test.go#L65-L69

There were some production issues, where I concluded that checking URLs would be nice. Note: existing tests in the codebase may not always do it, they need updating.

providers/asana/write_test.go
```diff
@@ -0,0 +1,112 @@
...
+				Errors:   nil,
+				Data: map[string]any{
+					"data": map[string]any{
+						"gid":           "15",
+						"resource_type": "project",
+						"name":          "Stuff to buy",
+						"archived":      false,
+						"color":         "light-green",
+						"created_at":    "2012-02-22T02:06:58.147Z",
+						"current_status_update": map[string]any{
```
Cobalt0s, line 61: You can omit some fields to shorten the test. Additionally, specifying nested objects can become quite challenging.

The main piece of code I want to point out is:
https://github.com/amp-labs/connectors/blob/c548efa81b6a714bc4339c95bdd8742e90ddd30b/test/utils/testroutines/outline.go#L44-L52

This means you can replace `full deep equals` comparator with `custom` comparator. The one that is commonly used is 
https://github.com/amp-labs/connectors/blob/c548efa81b6a714bc4339c95bdd8742e90ddd30b/providers/pipeliner/write_test.go#L104

This comparator ensures that any fields explicitly mentioned in the test must be present in the output, which is why it is named subset.

PS: During developement if you have any issues with the test put some debug points in this file `test/utils/testroutines/outline.go`.

test/asana/write/main.go
```diff
@@ -0,0 +1,63 @@
...
+		slog.Error(err.Error())
+	}
+
+	// Print the results
+	jsonStr, err := json.MarshalIndent(res, "", "  ")
+	if err != nil {
+		return fmt.Errorf("error marshalling JSON: %w", err)
+	}
+
+	_, _ = os.Stdout.Write(jsonStr)
```
Cobalt0s, line 59: There is a utility that "dumps" json struct into the stdout.

https://github.com/amp-labs/connectors/blob/c548efa81b6a714bc4339c95bdd8742e90ddd30b/test/salesforce/metadata/read/accounts/main.go#L34

providers/asana/metadata/scraping.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/amp-labs/connectors/tools/scrapper"
+)
+
+var (
+	// Static file containing a list of object metadata is embedded and can be served.
+	//
+	//go:embed schemas.json
+	schemas []byte
+
+	FileManager = scrapper.NewMetadataFileManager(schemas, fileconv.NewSiblingFileLocator()) // nolint:gochecknoglobals
```
Cobalt0s, line 16: The signiture of `NewMetadataFileManager` recently changed. For reference: https://github.com/amp-labs/connectors/pull/1291

The fix would be to use `FieldMetadataMapV1`.
```
FileManager = scrapper.NewMetadataFileManager[staticschema.FieldMetadataMapV1](schemas, fileconv.NewSiblingFileLocator()) // nolint:gochecknoglobals
```

providers/asana/metadata/schemas.json
```diff
@@ -0,0 +1,63 @@
+{
+    "modules": {
```
Cobalt0s, line 2: This file could be generated using the openapi file you have uploaded. If you need help with it let me know!

There are many such files located under `scripts/openapi/<providerName>/metadata/main.go`:
https://github.com/amp-labs/connectors/blob/main/scripts/openapi/pipeliner/metadata/main.go

There are few readmes:
* https://github.com/amp-labs/connectors/blob/main/scripts/openapi/README.md
* https://github.com/amp-labs/connectors/blob/main/tools/fileconv/api3/README.md

providers/asana.go
```diff
@@ -7,7 +7,7 @@ func init() {
 	SetInfo(Asana, ProviderInfo{
 		DisplayName: "Asana",
 		AuthType:    Oauth2,
-		BaseURL:     "https://app.asana.com/api",
+		BaseURL:     "https://app.asana.com",
```
Cobalt0s, line 10: This could be a breaking change. This could break existing proxy requests. They would need to add `/api` suffix.

@RajatPawar / @eberle1080 are there any clients relying on Asana tha could be affected?

Cobalt0s (commented):
Good work! Left some suggestions. Feel free to reach out to me.

providers/asana.go
immdipu: added the “/api” back to the BaseURL.

providers/asana/objectNames.go
```diff
@@ -0,0 +1,25 @@
...
+	objectNameWorkspaces,
+	objectNameTags,
+)
+
+var supportedObjectsByWrite = datautils.NewSet( //nolint:gochecknoglobals
+	objectNameProjects,
+)
+
+var supportLimitAndOffset = datautils.NewSet( //nolint:gochecknoglobals
+	objectNameWorkspaces,
```
Cobalt0s, line 24: I downloaded your branch revision and ran OpenAPI script. There is a file it produces but we is part of gitignore. It is located under `~/asana/metadata/queryParamStats.json`.
image

First of all, `tags` and `projects` must belong to `supportLimitAndOffset`.

Secondly, the `queryParamStats.json` doesn't include workspace, and I will add this as a todo action for myself. From the first look it seems to be an issue with the underlying library not loading all the parameters from the file. Not sure though.

Manually checked official documentation and indeed workspace/tags/projects/users all of them support limit and offset parameters.

providers/asana/objectNames.go
```diff
@@ -0,0 +1,25 @@
...
+import "github.com/amp-labs/connectors/internal/datautils"
+
+const (
+	objectNameProjects   = "projects"
+	objectNameTags       = "tags"
+	objectNameUsers      = "users"
+	objectNameWorkspaces = "workspaces"
+)
+
+var supportedObjectsByRead = datautils.NewSet( //nolint:gochecknoglobals
```
Cobalt0s, line 12: Since `schema.json` describes the whole domain space for `ListObjectMetadata` it therefore describes all the `Read`.
Once the `schema.json` file changes and gets/looses an object this would require to come back and edit this Set.

Instead create a supported `set` from object names located in the file like so:
```
var supportedObjectsByRead = metadata.Schemas.ObjectNames()
```

providers/asana/write_test.go
```diff
@@ -0,0 +1,84 @@
...
+			ExpectedErrs: []error{
+				common.ErrOperationNotSupportedForObject,
+			},
+		},
+		{
+			Name:  "Create projects as POST",
+			Input: common.WriteParams{ObjectName: "projects", RecordData: "dummy"},
+			Server: mockserver.Conditional{
+				Setup: mockserver.ContentJSON(),
+				If:    mockcond.MethodPOST(),
```
Cobalt0s, line 46: Could you add another test that given `RecordId` is not empty it will call `MethodPUT`? I think you could use the same `responsePortfolios` provider response and omit most expected fields.

scripts/openapi/asana/metadata/main.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/amp-labs/connectors/tools/fileconv/api3"
+	"github.com/amp-labs/connectors/tools/scrapper"
+)
+
+var ignoreEndpoints = []string{ // nolint:gochecknoglobals
+	"/allocations",
+	"/attachments",
+	"/custom_fields",
+	"/custom_field_settings",
+	"/goals",
```
Cobalt0s, line 20: I see that goals could be supported: [List Goals docs](https://developers.asana.com/reference/getgoals).
Is there a reason deep connector doesn't support `goals` object? I haven't checked each object but the same idea applies to others.

If it can be supported but for some reason we choose not to, could you group these endpoints and leave a comment?

Those that trully cannot be supported would be in different group for example [List Users](https://developers.asana.com/reference/getteamsforuser), where it requires path ID. Also for such case we don't need to ignore the endpoint explicitly the OpenAPI tool by default omits endpoints with nested `{}` path identifiers.

Cobalt0s (approved):
The PR looks good!

I have a comment regarding supported objects. It looks like few more could be supported: `allocations`, `goals`, maybe others.

providers/asana/objectNames.go
immdipu: image

You are correct. Workspace, tags, projects, and users do support limit and offset parameters, but not by default. To enable these parameters, you must pass the workspace ID.

scripts/openapi/asana/metadata/main.go
immdipu: image

Those endpoints you mentioned require us to pass either the team ID or the workspace. Although the API documentation doesn’t explicitly state that these fields are mandatory for fetching data, they are.

Cobalt0s: This makes sense. Sadly, there are a few deep connectors that do the same. I am not sure why providers do it, it confuses me ahah :laughing:

Cobalt0s: Could you add a comment above the script the same as in this PR? In future when we will come back to this connector it would give us context.

providers/asana/objectNames.go
Cobalt0s: That was my main concern, anyway, this means the PR is ready for merging.

scripts/openapi/asana/metadata/main.go
immdipu: Sure

PR #1308, opened 2025-01-28
[CON-988] feat: Add ChiliPiper Metadata Connector

Jkarage:
Tests:
Screenshot 2025-01-28 at 16 58 40

### DIFF
--- a/providers/chilipiper/metadata.go
+++ b/providers/chilipiper/metadata.go
@@ -2,7 +2,6 @@ package chilipiper
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/amp-labs/connectors/common"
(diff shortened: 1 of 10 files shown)


providers/chilipiper/metadata.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+
+		res, err := common.UnmarshalJSON[Response](resp)
+		if err != nil {
+			metadataResults.Errors[object] = err
+
+			continue
+		}
+
+		for fld := range res.Results[0] {
```
RajatPawar, line 65: Need to check if results has any length

RajatPawar (approved):

PR #1309, opened 2025-01-28
[CON-989] feat: Add ChiliPiper Read Connector

Jkarage:
Tests:
Screenshot 2025-01-30 at 15 55 27

This Adds Read Connector:
- To support more objects, I have created a map of objects to their paths.

### DIFF
--- a/providers/chilipiper/url.go
+++ b/providers/chilipiper/url.go
@@ -1,44 +1,23 @@
 package chilipiper
 
 import (
-	"github.com/amp-labs/connectors/common"
 	"github.com/amp-labs/connectors/common/urlbuilder"
 )
 
 const restAPIVersionPrefix = "api/fire-edge/v1/org"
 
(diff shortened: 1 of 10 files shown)


providers/chilipiper/url.go
```diff
@@ -1,9 +1,44 @@
...
+import (
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+)
 
 const restAPIVersionPrefix = "api/fire-edge/v1/org"
 
-func (conn *Connector) buildReadURL(object string) (*urlbuilder.URL, error) {
-	return urlbuilder.New(conn.BaseURL, restAPIVersionPrefix, object)
+func (conn *Connector) buildReadURL(config common.ReadParams) (*urlbuilder.URL, error) {
```
RajatPawar, line 10: minor: buildURL can have a page size as a parameter, and we can reduce these two into one function that can be used after checking for a next page.

RajatPawar (approved):

PR #1311, opened 2025-01-29
[CON-990] feat: Add ChiliPiper Write Connector

Jkarage:
Tests:
Screenshot 2025-01-30 at 11 55 42

- Similar to Read Connector, I have added a map of objects to their paths, to support more objects.

### DIFF
--- a/providers/chilipiper/types.go
+++ b/providers/chilipiper/types.go
@@ -22,10 +22,32 @@ var objectReadPath = map[string]string{ //nolint:gochecknoglobals
 	// "export_meetings": "meeting/meetings/export",
 }
 
+var objectWritePath = map[string]string{
+	"remove_users_workspace":     "workspace/users/remove",
+	"remoe_users_all_workspaces": "workspace/users/remove-from-all",
+	"add_users_workspace":        "workspace/users/add",
+	"remove_users_team":          "team/users/remove",
+	"add_user_team":              "team/users/add",
+	"distribution":               "distribution", // Allows updates only.
+	"update_license_users":       "user/licenses",
+	"invite_users":               "user/invite",
+}
+
 // supportsRead returns a unique path for reading the object.
-// or an error if the provided object is not supported.
+// errors out if the provided object is not supported.
 func supportsRead(object string) (string, error) {
 	path, ok := objectReadPath[object]
+	if !ok {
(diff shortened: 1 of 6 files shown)


providers/chilipiper/types.go
```diff
@@ -22,8 +22,19 @@ var objectReadPath = map[string]string{ //nolint:gochecknoglobals
 	// "export_meetings": "meeting/meetings/export",
 }
 
+var objectWritePath = map[string]string{ //nolint:gochecknoglobals
+	"remove_users_workspace":     "workspace/users/remove",
+	"remoe_users_all_workspaces": "workspace/users/remove-from-all",
```
RajatPawar, line 27: are these names from the official docs, or are these something you've shortened?

Jkarage: No modules, They are not from their official docs, So I was thinking since we share the supported objects in the docs we can add these as objects and route them to theirs specific paths/resources.

RajatPawar: Let's support these as they are named natively in the API - `workspace/users/remove`

RajatPawar (commented):
Change for the object required, otherwise looks good

RajatPawar (changes requested):
Change for the object required, otherwise looks good

providers/chilipiper/types.go
```diff
@@ -17,13 +17,24 @@ var objectReadPath = map[string]string{ //nolint:gochecknoglobals
 	"workspace":       "workspace",
 	"team":            "team",
 	"distribution":    "distribution",
-	"workspace_users": "workspace/users",
+	"workspace/users": "workspace/users",
 	// "meetings":        "meetings/meetings",
 	// "export_meetings": "meeting/meetings/export",
 }
 
+var objectWritePath = map[string]string{ //nolint:gochecknoglobals
```
RajatPawar, line 25: Feel free to ignore if I am missing some context, but since the key & value are now the same, we can probably get rid of it and directly use the objectName wherever required (or use an array)

RajatPawar (approved):

providers/chilipiper/types.go
Jkarage: Yeah, makes sense. My bad. I will update it to use set.

PR #1317, opened 2025-01-31
Add solar wind service desk connector

sanjay-a01:
## Checklist

## Catelog variables
	SolarWindsServiceDeskUS Provider = "solarWindsServiceDeskUS"
	SolarWindsServiceDeskEU Provider = "solarWindsServiceDeskEU"
	SolarWindsServiceDeskAU Provider = "solarWindsServiceDeskAU"

## Notes
According to the documentation, the API supports [pagination](https://apidoc.samanage.com/#section/General-Concepts/Pagination), but none of the API modules include the pagination parameters. Additionally, to access the APIs, you need to append .json or .xml to the endpoint, depending on your preferred format. For example: https://api.samanage.com/hardwares.xml or https://api.samanage.com/hardwares.json?per_page=100."

### GET ALL
URL: http://localhost:4444/categories.json
image

### GET by ID
URL: http://localhost:4444/categories/4259049.json
image

### POST
URL: http://localhost:4444/categories.json
image

### PUT
URL: http://localhost:4444/categories/4259049.json
image

### DELETE
URL: http://localhost:4444/categories/4259799.json
image

### DIFF
--- a/providers/solarWindsServiceDesk.go
+++ b/providers/solarWindsServiceDesk.go
@@ -1,8 +1,10 @@
 package providers
 
-const SolarWindsServiceDeskUS Provider = "solarWindsServiceDeskUS"
-const SolarWindsServiceDeskEU Provider = "solarWindsServiceDeskEU"
-const SolarWindsServiceDeskAU Provider = "solarWindsServiceDeskAU"
+const (
+	SolarWindsServiceDeskUS Provider = "solarWindsServiceDeskUS"
+	SolarWindsServiceDeskEU Provider = "solarWindsServiceDeskEU"
+	SolarWindsServiceDeskAU Provider = "solarWindsServiceDeskAU"
+)
 
 //nolint:funlen
 func init() {
--- a/providers/solarWindsServiceDesk.go
+++ b/providers/solarWindsServiceDesk.go
@@ -1,0 +1,119 @@
+package providers
+
+const SolarWindsServiceDeskUS Provider = "solarWindsServiceDeskUS"
+const SolarWindsServiceDeskEU Provider = "solarWindsServiceDeskEU"
+const SolarWindsServiceDeskAU Provider = "solarWindsServiceDeskAU"
+
+//nolint:funlen
+func init() {
+	// SolarWindsServiceDesk US(region) configuration
+	SetInfo(SolarWindsServiceDeskUS, ProviderInfo{
+		DisplayName: "solarWindsServiceDeskUS",
+		AuthType:    ApiKey,
+		BaseURL:     "https://api.samanage.com",
+		ApiKeyOpts: &ApiKeyOpts{
+			AttachmentType: Header,
+			Header: &ApiKeyOptsHeader{
+				Name:        "X-Samanage-Authorization",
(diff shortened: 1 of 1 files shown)


providers/solarWindsServiceDesk.go
```diff
@@ -0,0 +1,121 @@
...
+	SolarWindsServiceDeskUS Provider = "solarWindsServiceDeskUS"
+	SolarWindsServiceDeskEU Provider = "solarWindsServiceDeskEU"
+	SolarWindsServiceDeskAU Provider = "solarWindsServiceDeskAU"
+)
+
+//nolint:funlen
+func init() {
+	// SolarWindsServiceDesk US(region) configuration
+	SetInfo(SolarWindsServiceDeskUS, ProviderInfo{
+		DisplayName: "solarWindsServiceDeskUS",
```
Cobalt0s, line 13: Display name could have "spaces". This should be the human-readable label which is displayed somewhere in the UI.

providers/solarWindsServiceDesk.go
```diff
@@ -0,0 +1,121 @@
...
+				LogoURL: "https://res.cloudinary.com/dycvts6vp/image/upload/v1738322158/media/solarwinds.com_1738322159.svg",
+			},
+		},
+	})
+
+	// SolarWindsServiceDesk EU(region) configuration
+	SetInfo(SolarWindsServiceDeskEU, ProviderInfo{
+		DisplayName: "solarWindsServiceDeskEU",
+		AuthType:    ApiKey,
+		BaseURL:     "https://apieu.samanage.com",
```
Cobalt0s, line 52: I think that it can be one connector but it would have a variable, like a "workspace" which will take on values "api", "apieu", "apiau" representing region. Domain prefix is the only difference I see among connectors.

Idk the approach to take and if it is settled at all. Likely for now only USA related connector is to be added.
@jlimatampersand you have more information on it, could you comment on this?

providers/solarWindsServiceDesk.go
jlimatampersand: yeah, I would also capitalize the first character.

providers/solarWindsServiceDesk.go
```diff
@@ -0,0 +1,121 @@
...
+			},
+			Regular: &MediaTypeRegular{
+				IconURL: "https://res.cloudinary.com/dycvts6vp/image/upload/v1738322100/media/solarwinds.com_1738322099.svg",
+				LogoURL: "https://res.cloudinary.com/dycvts6vp/image/upload/v1738322158/media/solarwinds.com_1738322159.svg",
+			},
+		},
+	})
+
+	// SolarWindsServiceDesk EU(region) configuration
+	SetInfo(SolarWindsServiceDeskEU, ProviderInfo{
```
jlimatampersand, line 49: Chatted with @laurenzlong let's keep just US for now. We don't currently support any non-US customers yet. But thanks for working on this ahead of time.

jlimatampersand (changes requested):
Let me know when done resolving comments :)

jlimatampersand (approved):

PR #1321, opened 2025-01-31
feat(common): Make ParseResult use generic Record type

Cobalt0s:
# Description 

Currently, marshaling records that require extra processing—such as flattening—is not straightforward. Working with records of type `map[string]any` can be cumbersome in such cases, whereas using `*ajson.Node` simplifies the process.

# Solution
`common.ParseResult` relies on a generic record type:
* `map[string]any` – Convenient when raw and field values are similar.
* `*ajson.Node` – Requires less code for JSON manipulation and should be used when extracting specific fields.

# Utilities

Additionaly added a `common.RecordFlattener` method, which moves fields from nested object to the root.
Outreach and Klaviyo connectors produce response with similiar structure and would benefit from this method.
The algorithm is the same.


Cobalt0s:
* **#1320** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1320?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1321** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1321?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1321?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

RajatPawar:
Let's hold off on this until we get around to refactoring some of our HTTP clients and have a good understanding of where parsing lives there - maybe it can be a part of that.

Cobalt0s:
> Let's hold off on this until we get around to refactoring some of our HTTP clients and have a good understanding of where parsing lives there - maybe it can be a part of that. Also, type constraints come with certain limitations. Until then feel free to add functions to common to flatten records, etc

@RajatPawar `common.ParseResult` had its limitation that `marshalFunc` only operates on `map[string]any`. For those connectors where raw differed from fields it made `marshalFunc` unnecessary complex, where it either had to operate on `map[string]any` by undoing work done by `recordsFunc` or create higher order function.

Instead I am suggesting to match output of `recordsFunc` with input of `marshalFunc`. By having this the connector implementation becomes a one liner: `common.MakeMarshaledDataFunc(common.RecordFlattener("attributes"))`. 

Could we have `ParseResult` live in parallel :pray: to the refactored version, in whenever place it will be or level?

The generic could be changed to type `any` if you dont like type constraints. The implementation goal is to preserve the matching chain of types from `recordsFunc` to `marshalFunc`. Not having this kind of seperation produced a bug where raw and fields were processed in `recordsFunc` affecting both.

common/parse.go
```diff
@@ -9,21 +9,40 @@ import (
...
 
 type (
+	// NextPageFunc extracts the next page token/URL from the response body.
 	NextPageFunc func(*ajson.Node) (string, error)
-	RecordsFunc  func(*ajson.Node) ([]map[string]any, error)
+
+	// RecordsFunc extracts a list of records as map[string]any from the response body.
+	RecordsFunc func(*ajson.Node) ([]map[string]any, error)
+	// NodeRecordsFunc extracts a list of records as ajson.Node from the response body.
+	NodeRecordsFunc func(*ajson.Node) ([]*ajson.Node, error)
```
RajatPawar, line 18:
```suggestion
	RecordsAsNodeFunc func(*ajson.Node) ([]*ajson.Node, error)
```

common/parse.go
```diff
@@ -9,21 +9,40 @@ import (
...
+
+	// RawReadRecordFunc processes a raw record, applying additional transformations if needed.
+	// There is one common example of flattening nested fields provided by RecordFlattener.
+	RawReadRecordFunc func(node *ajson.Node) (map[string]any, error)
+	// MarshalFunc converts a list of map[string]any records into ReadResultRow format.
+	MarshalFunc func(records []map[string]any, fields []string) ([]ReadResultRow, error)
+	// NodeMarshalFunc converts a list of ajson.Node records into ReadResultRow format.
+	NodeMarshalFunc func(records []*ajson.Node, fields []string) ([]ReadResultRow, error)
```
RajatPawar, line 26:
```suggestion
	MarshalFromNodeFunc func(records []*ajson.Node, fields []string) ([]ReadResultRow, error)
```

common/parse.go
```diff
@@ -9,21 +9,40 @@ import (
...
 )
 
+// RawReadRecord defines the types of records that ParseResult can process.
+// It determines which callback function is used for parsing.
+type RawReadRecord interface {
```
RajatPawar, line 31:
```suggestion
type ProviderReadResponseType interface {
```

common/parse.go
```diff
@@ -9,21 +9,40 @@ import (
...
-	RecordsFunc  func(*ajson.Node) ([]map[string]any, error)
+
+	// RecordsFunc extracts a list of records as map[string]any from the response body.
+	RecordsFunc func(*ajson.Node) ([]map[string]any, error)
+	// NodeRecordsFunc extracts a list of records as ajson.Node from the response body.
+	NodeRecordsFunc func(*ajson.Node) ([]*ajson.Node, error)
+
+	// RawReadRecordFunc processes a raw record, applying additional transformations if needed.
+	// There is one common example of flattening nested fields provided by RecordFlattener.
+	RawReadRecordFunc func(node *ajson.Node) (map[string]any, error)
```
RajatPawar, line 22: can you please take the help of chatGpt to come up with some alternative names + comments for all of these functions? these are not very intuitive to understand, and it'd be daunting for a new user

common/parse.go
```diff
@@ -107,7 +126,35 @@ func GetMarshaledData(records []map[string]any, fields []string) ([]ReadResultRo
 	return data, nil
 }
 
-func GetRecordsUnderJSONPath(jsonPath string, nestedPath ...string) RecordsFunc {
+// MakeMarshaledDataFunc produces ReadResultRow where raw record differs from the fields.
+// This usually includes a set of actions to preprocess, usually to flatten the raw record and then extract
+// fields requested by the user.
+func MakeMarshaledDataFunc(nodeRecordFunc RawReadRecordFunc) NodeMarshalFunc {
```
RajatPawar, line 132: This is a complex function that takes in a func as input and outputs a func - is there a simpler way to express this so that it's reusable without a lot of cognitive overhead?

common/parse.go
```diff
@@ -9,21 +9,40 @@ import (
...
 // The sizeFunc returns the total number of records in the response.
 // The recordsFunc returns the records in the response.
 // The nextPageFunc returns the URL for the next page of results.
 // The marshalFunc is used to structure the data into an array of ReadResultRows.
 // The fields are used to populate ReadResultRow.Fields.
-func ParseResult(
+func ParseResult[R RawReadRecord](
 	resp *JSONHTTPResponse,
-	recordsFunc func(*ajson.Node) ([]map[string]any, error),
+	recordsFunc func(*ajson.Node) ([]R, error),
```
RajatPawar, line 43:
```suggestion
	extractRecords func(*ajson.Node) ([]R, error),
```

common/parse.go
```diff
@@ -9,21 +9,40 @@ import (
...
 	nextPageFunc func(*ajson.Node) (string, error),
```
RajatPawar, line 44:
```suggestion
	extractNextPage func(*ajson.Node) (string, error),
```

common/parse.go
```diff
@@ -107,7 +126,35 @@ func GetMarshaledData(records []map[string]any, fields []string) ([]ReadResultRo
...
+				Fields: ExtractLowercaseFieldsFromRaw(fields, record),
+				Raw:    raw,
+			}
+		}
+
+		return data, nil
+	}
+}
+
+func GetRecordsUnderJSONPath(jsonPath string, nestedPath ...string) func(*ajson.Node) ([]map[string]any, error) {
```
RajatPawar, line 157:
```suggestion
func ExtractRecordsFromPath(jsonPath string, nestedPath ...string) func(*ajson.Node) ([]map[string]any, error) {
```

common/parse.go
```diff
@@ -135,3 +182,54 @@ func getRecords(optional bool, jsonPath string, nestedPath ...string) RecordsFun
...
+//	            "test_account": false,
+//	            "contact_information": {},
+//	            "locale": ""
+//	        },
+//	        "links": {}
+//	    }
+//	],
+//
+// The resulting fields for the above will be: [ type, id, test_account, contact_information, locale, links ].
+func RecordFlattener(nestedKey string) RawReadRecordFunc {
```
RajatPawar, line 207:
```suggestion
func FlattenNestedFields(nestedKey string) RawReadRecordFunc {
```

common/parse.go
RajatPawar: Here's what I came up with - 
```
/ RecordTransformer is a function that processes a JSON node and transforms it
// into a map representation, potentially applying structural modifications.
//
// Common use cases include:
// - Flattening nested objects (see FlattenNestedFields)
// - Filtering out unwanted fields
// - Converting field formats
// - Renaming fields
// - Adding computed fields
//
// Example usage:
//
//     transformer := FlattenNestedFields("attributes")
//     resultMap, err := transformer(jsonNode)
type RecordTransformer func(node *ajson.Node) (map[string]any, error)```

RajatPawar (approved):
LGTM with the logic, but please resolve all the comments about naming to make it very clear / explicit

common/parse.go
RajatPawar: Screenshot 2025-04-04 at 11 51 45 PM

common/parse.go
Cobalt0s: Indeed, this is the biggest problem in this design.

It is either:
* `common.GetMarshaledData` when Fields == Raw, or
* `common.MakeMarshaledDataFunc` which requires a method handling the difference.

We could factor out the common marshaling so we will end up with either nil or `RawReadRecordFunc`.
But then it assumes that each connector process raw identically. Maybe we could split that in 2. But that will be a new method. My goal was to adapt existing `common.ParseResult` with least breaking changes to existing connectors.

PR #1322, opened 2025-02-03
feat: update Marketo Read Connector to support Leads

Jkarage:
- Marketo Connector requires FilteringValues (Passed as Query Params) to Read `Leads`, `Companies` and  `SalesPersons` Objects. 
- The `Ids` in these objects are numericals starting from 1.
-  I have used the `id` as a filteringType for these Objects and thus we can now provide FilteringValues.

With this Marketo can now support Reading from `Leads` `Companies` and `SalesPersons`. 

Tests:
Screenshot 2025-02-03 at 13 35 45

### DIFF
--- a/providers/marketo/parse.go
+++ b/providers/marketo/parse.go
@@ -1,6 +1,9 @@
 package marketo
 
 import (
+	"strconv"
+
+	"github.com/amp-labs/connectors/common"
 	"github.com/amp-labs/connectors/common/jsonquery"
 	"github.com/spyzhov/ajson"
 )
@@ -8,6 +11,36 @@ import (
 // getNextRecordsURL returns the URL for the next page of results.
 func getNextRecordsURL(node *ajson.Node) (string, error) {
 	return jsonquery.New(node).StrWithDefault("nextPageToken", "")
+}
+
+func constructNextRecordsURL(object string) common.NextPageFunc {
+	if filtersByIDs(object) {
+		// We use a new func for generating the next page url.
+		return func(node *ajson.Node) (string, error) {
+			jsonParser := jsonquery.New(node)
+
+			data, err := jsonParser.Array("result", false)
+			if err != nil {
+				return "", err
+			}
+
+			// If the records returned matches the maximum batchsize, there is a high probability of having more records.
+			// We'd have to check for the next page records, also due deletes the is also a probability of having more records
+			// even if the size do not reach 300.
+			if len(data) > 0 {
+				id, err := jsonquery.New(data[len(data)-1]).Integer("id", false)
+				if err != nil {
+					return "", err
+				}
+
+				return strconv.Itoa(int(*id) + 1), nil
+			}
+
+			return "", nil
+		}
+	}
+
+	return getNextRecordsURL
 }
 
 // getRecords returns the records from the response.
--- a/providers/marketo/types.go
+++ b/providers/marketo/types.go
@@ -3,7 +3,10 @@ package marketo
 import (
 	"errors"
 	"fmt"
+	"slices"
 )
+
+const batchSize = 300 // nolint:gochecknoglobals
 
 type writeResponse struct {
 	Result  []map[string]any `json:"result"`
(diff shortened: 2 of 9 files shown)


providers/marketo/parse.go
```diff
@@ -10,6 +13,36 @@ func getNextRecordsURL(node *ajson.Node) (string, error) {
...
+			data, err := jsonParser.Array("result", false)
+			if err != nil {
+				return "", err
+			}
+
+			// If the records returned matches the maximum batchsize, there is a high probability of having more records.
+			// We'd have to check for the next page records, also due deletes the is also a probability of having more records
+			// even if the size do not reach 300.
+			if len(data) > 0 {
+				id, err := jsonquery.New(data[len(data)-1]).Integer("id", false)
```
Cobalt0s, line 31: `id` => `lastRecordID`

providers/marketo/types.go
```diff
@@ -27,3 +34,9 @@ func constructErrMessage(a any) (string, error) {
 
 	return fmt.Sprint(s[0]["reasons"]), nil
 }
+
+func filtersByIDs(object string) bool {
```
Cobalt0s, line 38: For some reason the term filtering confuses me a bit. I understand that provider uses it, but looking over the PR it looks like we use it in the context of paginating by ID.

providers/marketo/types.go
```diff
@@ -3,21 +3,28 @@ package marketo
...
 	Success bool             `json:"success"`
 	Errors  []map[string]any `json:"errors"`
 }
 
-// IdResponseObjects represents a list of objects that uses `id` as a unique field in the response.
-var IdResponseObjects = []string{"leads", "companies", "salespersons"} //nolint:gochecknoglobals
+// nolint:gochecknoglobals
+var (
+	// IdResponseObjects represents a list of objects that uses `id` as a unique field in the response.
+	IdResponseObjects = []string{"leads", "companies", "salespersons"}
```
Cobalt0s, line 20: It looks like this var is not used and is identical/similiar to `idFilteringObjects`.

providers/marketo/url.go
```diff
@@ -34,3 +35,43 @@ func (c *Connector) getURL(params common.ReadParams) (*urlbuilder.URL, error) {
 
 	return url, nil
 }
+
+func addFilteringIDQueries(urlbuilder *urlbuilder.URL, startIdx string) error {
```
Cobalt0s, line 39: Could you add a documentation link to any object that uses `filterValues` or `filterType` for a quick reference, to whoever is reading this method.

providers/marketo/url.go
```diff
@@ -34,3 +35,43 @@ func (c *Connector) getURL(params common.ReadParams) (*urlbuilder.URL, error) {
...
+	queryIDs := strings.Join(ids, ",")
+	urlbuilder.WithQueryParam("filterValues", queryIDs)
+	urlbuilder.WithQueryParam("filterType", "id")
+
+	return nil
+}
+
+func constructURLQueries(url *urlbuilder.URL, params common.ReadParams) error {
+	if filtersByIDs(params.ObjectName) && len(params.NextPage) == 0 {
+		if err := addFilteringIDQueries(url, "1"); err != nil {
```
Cobalt0s, line 60: Question: does it mean that for the first page we `must` include `filterValues` and `filterType` query params, if so then makes sense.

providers/marketo/url.go
```diff
@@ -34,3 +35,43 @@ func (c *Connector) getURL(params common.ReadParams) (*urlbuilder.URL, error) {
 
 	return url, nil
 }
+
+func addFilteringIDQueries(urlbuilder *urlbuilder.URL, startIdx string) error {
+	ids := make([]string, batchSize)
+
+	idx, err := strconv.Atoi(startIdx)
+	if err != nil {
+		return err
```
Cobalt0s, line 44: Could you wrap this error with `common.ErrNextPageInvalid`?

providers/marketo/url.go
```diff
@@ -34,3 +35,43 @@ func (c *Connector) getURL(params common.ReadParams) (*urlbuilder.URL, error) {
...
+	if filtersByIDs(params.ObjectName) && len(params.NextPage) == 0 {
+		if err := addFilteringIDQueries(url, "1"); err != nil {
+			return err
+		}
+	}
+
+	// If NextPage is set, then we're reading the next page of results.
+	if len(params.NextPage) > 0 {
+		if filtersByIDs(params.ObjectName) {
+			if err := addFilteringIDQueries(url, params.NextPage.String()); err != nil {
```
Cobalt0s, line 68: You dont need to check if the err is nil to return it.

This can be simplified into:
```
return addFilteringIDQueries(url, params.NextPage.String())
```
As a side effect, `else` statement could be removed. Minimizing `else` statements is encouraged.

Cobalt0s (changes requested):
Left some comments. Also it would be nice if you could add one mock test for this new branch in logic.
For example:
* object `leads` + NextPage token
* then connector impl must send query param `filterValues` to provider/mock-server
* and connector impl builds next page token using last record ID from response

providers/marketo/types.go
Jkarage: Well It's for Objects that allows filtering by id. Are you suggesting `paginatesByID`?

providers/marketo/url.go
Jkarage: Yes this is for objects requiring passing filterIds in all calls.

providers/marketo/types.go
Cobalt0s: Makes sense, thank you.

Cobalt0s (approved):
Perfect!

PR #1330, opened 2025-02-04
[ENG-1892] Define subscribe action connector types and interface

jlimatampersand:
- Defining types for subscription and registration for the action.

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -428,7 +428,7 @@ type SubscriptionResult struct { // this corresponds to each API call.
 
 // SubscribeConnector has 2 main responsibilities:
 // 1. Register a subscription with the provider.
-// Registering a subscription is a one-time operation that is requried
+// Registering a subscription is a one-time operation that is required
 // by providers that hold some master registration of all subscriptions.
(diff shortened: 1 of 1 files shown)


common/types.go
```diff
@@ -383,3 +383,75 @@ func inferDeprecatedFieldsMap(fields map[string]FieldMetadata) map[string]string
...
+	Events          []SubscriptionEventType
+	// ["create", "update", "delete"]
+	// our regular CRUD operation events we translate to provider-specific names contact.creation
+	UpdateFields []string
+	// ["email", "fax"]
+	PassThroughEvents []string
+	// provider specific events ["contact.merged"] for hubspot or ["jira_issue:restored", "jira_issue:archived"] for jira.
+}
+
+type SubscribeConnector interface {
```
eberle1080, line 429: Can you add a line or two above each signature explaining the function? Looks good but I want to make sure interfaces especially are well-documented.

eberle1080 (approved):
One request but no blockers. LGTM.

PR #1331, opened 2025-02-05
feat(openapi): Return field enums

Cobalt0s:
# Description
OpenAPI files may contain [enumeration properties](https://swagger.io/docs/specification/v3_0/data-models/enums/).

This PR extends the Field struct to include new properties derived from OpenAPI files:
* Type
* List of enumeration options

# Purpose
Many connectors can benefit from enumeration data present in OpenAPI files. All of them can migrate to `field metadata version 2`.

PR #1314, opened 2025-01-30
feat(tool): Google-Discovery to metadata extractor

Cobalt0s:
# Description

This is a new tool similar to the OpenAPI metadata extractor.
It operates on Google Discovery files. Initially, I considered converting Google Discovery files to OpenAPI format and using the existing script, but I couldn't find such tools online. Since writing a custom extractor is relatively simple, I opted for this approach.

# Purpose
This tool will be useful for adding more Google API modules. Implementing ListObjectMetadata will simply involve downloading the Google Discovery file, making slight script adjustments, and specifying the module ID. This process will generate a schema.json file, mapping each module to its metadata.

# Review

[Next PR ](https://github.com/amp-labs/connectors/pull/1315) in the Graphite stack uses this tool to produce metadata for Calendar Google module.

PR #1313, opened 2025-01-30
tidy(metadata): Separate Schema definition from tools

Cobalt0s:
# Description

Schema struct is a model that is used by `OpenAPI` scripts and one `Scrapper`.
With the upcoming introduction of the GoogleDiscovery extractor, the following considerations are clear:
* Metadata can originate from multiple sources (OpenAPI/GoogleDiscovery/Scrapper).
* A single model should be sufficient to represent all sources. (`metadatadef.Schema`)
* The model should exist within its own package, separate from the tools. (`metadatadef`)

This PR restructures the code to separate data models from tools.

Additonally, fields extractor does not just return the field names as array of strings but array of `metadatadef.Field{}`.

PR #1332, opened 2025-02-05
[CON-971] feat(Pipedrive): Metadata Fields

Cobalt0s:
# Description
The static `schema.json` file has been updated to V2 for Pipedrive.
* `ListObjectMetadata` no longer falls back to the static file; instead, it now uses it by default.
* A single record is still requested, and if the read succeeds, the deprecated `Fields` property is overridden.
Note: Enums cannot be inferred from read responses, making OpenAPI the preferred strategy.

Additionally, some OpenAPI tool warnings have been removed by:
* Ignoring certain endpoints.
* Adding `organization_collection` and `person_collection` objects to the supported list (these are in beta and require admin access).

# Test
Modified metadata script to request `leadLabels` which has enum values.
image


Cobalt0s:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1332?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1332** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1332?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1332?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1331** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1331?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1313** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1313?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1314](https://github.com/amp-labs/connectors/pull/1314) <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1314?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

scripts/openapi/pipedrive/metadata/main.go
```diff
@@ -22,6 +23,18 @@ var ignoreEndpoints = []string{ // nolint:gochecknoglobals
...
+	"/goals/find",
+	// Not an array.
+	"/filters/helpers",
+	"/userConnections",
+	"/userSettings",
+}
+
+var objectEndpoints = map[string]string{ // nolint:gochecknoglobals
+	"/organizations/collection": "organization_collection",
+	"/persons/collection":       "person_collection",
```
Cobalt0s, line 37: These are `beta` endpoints which can be requested by admins only.

providers/pipedrive/metadata.go
```diff
@@ -67,18 +67,17 @@ func metadataMapper(
 		return nil, err
 	}
 
-	mdt := &common.ObjectMetadata{
-		FieldsMap: make(map[string]string),
+	mdt, err := metadata.Schemas.SelectOne(moduleID, obj)
```
Cobalt0s, line 70: Static file lookup is done by default. We are after the enums.

providers/pipedrive/read.go
```diff
@@ -40,16 +40,22 @@ func (c *Connector) buildReadURL(config common.ReadParams) (*urlbuilder.URL, err
...
 	if err != nil {
 		return nil, err
 	}
 
 	// begin fetching objects at provided start date
 	// Supporting objects are: Activities & Notes only.
 	if !config.Since.IsZero() {
-		startDate := config.Since.Format(time.DateOnly)
-		url.WithQueryParam("start_date", startDate)
+		since := config.Since.UTC().Format(time.DateTime)
```
RajatPawar, line 51: Can you remove `UTC` here? `Since` should be accepted & passed as is without modifying it

scripts/openapi/pipedrive/metadata/main.go
RajatPawar: Are these being renamed? If not, can you explain what this map does? And if they are being renamed, can you please revert this? We don't want to create a discrepancy for objects, and should avoid owning any terminologies ourselves.

Quick question: Have we renamed objects in any other providers?

RajatPawar (dismissed):
Added comments that require clarification

scripts/openapi/pipedrive/metadata/main.go
Cobalt0s: As discussed in Slack, yes other providers do that. I tried to limit such cases. But that is the best way to resolve name collisions and abstract URL building logic. ObjectName is not part of URL it is a concept.

providers/pipedrive/read.go
Cobalt0s: Time format should respect the time zone. Creation time of an object at 3pm in one time zone is different from other time zone. I think I found this issue in one of the providers and since then I try to be explicit and convert to UTC first so that timestamp includes it.
Example: https://github.com/amp-labs/connectors/blob/4af8b2f92e4c7d5508b0d8e7b4752affda7b35e1/internal/datautils/timing.go#L17-L19

scripts/openapi/pipedrive/metadata/main.go
laurenzlong: Based on https://developers.pipedrive.com/docs/api/v1/Organizations#getOrganizationsCollection, it seems like `/api/v2/organizations/collection` is just a new API for getting organizations (it is an alternative to "/api/v2/organizations"). So the object name is actually "organization", it is not a distinct object from what the "organizations" endpoint returns. Please remove these endpoints, since they are beta and require admin access, and collide with the "organizations" endpoint.

providers/pipedrive/read.go
laurenzlong: When the server passes in `Since` it is already UTC so there's no need to convert it.

laurenzlong:
Hey @Cobalt0s can you please address the review comments when you get a chance? Thanks!

Cobalt0s:
> Hey @Cobalt0s can you please address the review comments when you get a chance? Thanks!

This PR is next on my list. I will address the comments.
Also could you review a [prerequisite PR](https://github.com/amp-labs/connectors/pull/1352).

scripts/openapi/pipedrive/metadata/main.go
Cobalt0s: You are right. I will remove these endpoints. I saw some logs by the OpenAPI script and these were some of the objects that were neither included nor explicitly ignored. I think I shouldve been even more granular and created a dedicated PR to address these logs, but thats even more open PRs.

providers/pipedrive/read.go
Cobalt0s: There was some connector that was sensitive to UTC. Since I have discovered this bug Chris told me it is not really critical because server sends correct time which is UTC formated. 

I am trying to covert to UTC time for consistency, which shouldn't be expensive operation. When `connectors` library would be used by open source community the deep connector should just work regardles of any `time` object.

laurenzlong (approved):

PR #1350, opened 2025-02-11
[CON-1045] feta: Add ServiceNow Metadata Connector

Jkarage:
Supports Table API Only.(ModuleTable)
Tests:
Screenshot 2025-02-11 at 13 10 45


RajatPawar (commented):
This looks good. Can you quickly evaluate (maybe spend an hour or so) to see if you can implement this using the `internal/components` package we recently introduced? You can use `smartleadv2` as a reference.  Please let me know on slack if you run into any issues - thanks

RajatPawar:
> Okay, Let me look into it. My Initial thought was the metadata Component for introspective endpoints or querying-data approach was not ready.

The implementation is still something we'd have to do - we can over time find ways to abstract it out, but just curious if this even fits well

providers/servicenow/module.go
```diff
@@ -0,0 +1,19 @@
...
+
+const (
+	// ModuleTable is the module/API used for accessing and manging table records.
+	ModuleTable common.ModuleID = "table"
+)
+
+// supportedModules represents currently working and supported modules within the Marketo connector.
+// Any added module should be appended here.
+var supportedModules = common.Modules{ // nolint: gochecknoglobals
+	ModuleTable: {
```
RajatPawar, line 13: can you add a comment above this with the module's expected URL suffix?

providers/servicenow/connector.go
```diff
@@ -0,0 +1,51 @@
...
+	common.RequireAuthenticatedClient
+	// Require workspace
+	common.RequireWorkspace
+	// Require module
+	common.RequireModule
+
+	// Supported operations
+	components.SchemaProvider
+	components.Reader
+	components.Writer
```
RajatPawar, line 29: Please add reader, writer and deleters only when you implement them, because IIRC, they will pass this - 
```
dc, ok := conn.(DeleteConnector)
if ok {
  // servicenow can get here
}
```

providers/servicenow/connector.go
```diff
@@ -0,0 +1,51 @@
...
+	return components.Initialize(providers.ServiceNow, params, constructor)
+}
+
+func constructor(base *components.Connector) (*Connector, error) {
+	connector := &Connector{Connector: base}
+
+	// Set the metadata provider for the connector
+	connector.SchemaProvider = schema.NewObjectSchemaProvider(
+		connector.HTTPClient().Client,
+		schema.FetchModeParallel,
```
RajatPawar, line 43: Can you make sure that this is not prone to 429s? If it is, can you please add a TODO here for that?

providers/servicenow/handlers.go
```diff
@@ -0,0 +1,61 @@
...
+// 1. Support the whole or parts of the `now` namespace.
+// 2. Look into adding more namespace (modules) supports.
+
+type responseData struct {
+	Result []map[string]any `json:"result"`
+	// Other fields
+}
+
+func (c *Connector) buildSingleObjectMetadataRequest(ctx context.Context, objectName string) (*http.Request, error) {
+	module := supportedModules[c.Module()]
```
RajatPawar, line 22: We should definitely add a supported modules concept in components

providers/servicenow/handlers.go
```diff
@@ -0,0 +1,61 @@
...
+	res, err := common.UnmarshalJSON[responseData](response)
+	if err != nil {
+		return nil, common.ErrFailedToUnmarshalBody
+	}
+
+	if len(res.Result) == 0 {
+		return nil, common.ErrMissingExpectedValues
+	}
+
+	if len(res.Result) == 0 {
```
RajatPawar, line 51: Check is duplicated

providers/servicenow/handlers.go
```diff
@@ -0,0 +1,61 @@
...
+		FieldsMap:   make(map[string]string),
+		DisplayName: naming.CapitalizeFirstLetterEveryWord(objectName),
+	}
+
+	res, err := common.UnmarshalJSON[responseData](response)
+	if err != nil {
+		return nil, common.ErrFailedToUnmarshalBody
+	}
+
+	if len(res.Result) == 0 {
```
RajatPawar, line 47: Is there no openapi spec we can fallback on from servicenow?

RajatPawar (changes requested):

providers/servicenow/handlers.go
Jkarage: Couldn't find their OpenAPI specs files. They have some files in their API Explorer, but they dont have the schema.

RajatPawar (approved):

PR #1337, opened 2025-02-07
[CON-617] feat(google): Calendar - Write

Cobalt0s:
- [x] Connector uses `internal/components`
- [x] Raw response is returned as is, no formatting or flattening is performed.
- [x] Write payloads should accept what `ReadResults.Fields` is returning. Any unnecessary nesting around the input is removed.
- [ ] Provider errors are mapped if non-standard (errors with 200 response code are converted to 4XX)
- [ ] Custom fields, if not human readable names, are resolved to readable names.
- [x] Unit tests cover write logic (placed in /tests/<provider>)
- [x] Appropriate object names are used. Objects need to be resources, not actions (`jobs` and not `jobs.list`).

# Live Tests
## calendarList
image

## events
image

PR #1354, opened 2025-02-11
[CON-615] feat: Google Calendar ListObjectMetadata

Cobalt0s:


Cobalt0s:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1354?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1354** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1354?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1354?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1353** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1353?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1337](https://github.com/amp-labs/connectors/pull/1337) <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1337?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

providers/google/internal/calendar/schemas.json
```diff
@@ -2,11 +2,11 @@
   "modules": {
     "calendar": {
       "id": "calendar",
-      "path": "calendar/v3/users/me",
+      "path": "",
       "objects": {
-        "calendarList": {
+        "users/me/calendarList": {
```
laurenzlong, line 7: Can you actually revise this file to have the following objects supported?
- object name "calendarList": path is `users/me/calendarList`
- object name "events": path is `calendars/primary/events`
- object name "settings": path is `users/me/settings`

So basically for settings and calendarList, revert to the previous state. 

I know you're trying to abide by the philosophy of preserving the path, but in this case `users/me` isn't really a part of the resource name, it is a "magic string" to indicate the current user, so it doesn't need to be preserved.

providers/google/internal/calendar/adapter.go
```diff
@@ -0,0 +1,27 @@
+package calendar
+
+import (
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/internal/components"
+	"github.com/amp-labs/connectors/internal/components/schema"
+	"github.com/amp-labs/connectors/providers"
+)
+
+type Adapter struct {
```
laurenzlong, line 10: I find this pattern a little confusing - if you and @RajatPawar have already discussed this and have agreed then it's fine, but here is my reaction:
- `Adapter` doesn't seem to have anything that's Calendar specific, so why is it being defined here in this file as opposed to in a shared file like `internal/components.go`? 
- None of the code here seems very Google specific or calendar specific, so this should be abstracted away in a common file, so we can reduce the boilerplate code for each new module.
- "ModuleComponent" or something like that would be a clearer struct name. Adapter makes me think that this is coercing one interface into another interface.

None of this is blocking this PR from being merged, but we should get aligned on the best way forward before introducing this pattern to more connectors/modules.

laurenzlong (approved):

providers/google/internal/calendar/schemas.json
Cobalt0s: I was thinking that we fully migrated to extensive object names. Yes `users/me` as URI part carries special meaning.
I will change it as you say, but that raises questions on how to do object names consistently in other PRs/connectors, I have a small ticket to revisit older connectors just for that.

providers/google/internal/calendar/adapter.go
Cobalt0s: I wouldn't say that this pattern is fully confirmed. But as experimental it was done for AWS and was especially helpful for existing connectors like Salesforce/Atlassian.

`Adapter` refers to the "Outgoing Adapter" in Hexagonal DDD architecture. The name can change. I was thinking that `google.Connector` will have a list of API adapters: `calendar.Adapter` & `contacts.Adapter` & `drive.Adapter`. Only one is active at a time. The logic across modules is not mixed and doesn't bloat. But I think the common methods can be moved into local `providers/google/internal/core` package but I didn't want to create that prematuraly. As more modules will be added the shared methods will become more aparent.

Essentially I view it as `sub-connector`. I agree that the first idea that comes into mind about adapter is that it converts one plug to fit into some other socket.
This is an idea to organize code in multi-modulus conenctors. For example if we add a module to Salesforce using this approach we can be confident that existing functionality is fully isolated and thats the main goal, motivation I had. Please, let me know if you have alternative ideas.
@RajatPawar, let me know your ideas too!

laurenzlong: Thanks for explaining! Makes sense for now, I'll let you know if I think if anything better.

PR #1355, opened 2025-02-12
[CON-1155] feat: Add Zoho Projects + Zoho BugTracker Connector

immdipu:
## Checklist
- [x] Ran Linter

## Testing 

# Zoho Projects Testing

### GET 
URL: localhost:4444/restapi/portals/
image

### POST
URL: localhost:4444/api/v3/portal/878609240/portal_export/
image

### DELETE
URL: localhost:4444/restapi/portal/878609240/projects/2438969000000057049/tasklists/2438969000000057167/
image

### Pagination
Zoho Projects uses index-based pagination. you must pass `index` (starting position) and `range` (number of records to fetch) 
image

# Error 

Invalid Project id provided in the url
image

incomplete payload 
image

Invalid url 
image

# Zoho BugTracker Testing

### GET
URL: localhost:4444/restapi/portal/878609240/projects/2438969000000057049/users/
image

### POST
URL: localhost:4444/restapi/portal/878609240/projects/2438969000000057049/statuses/?content=Build Awaiting Alpha Test
image

### DELETE
URL: localhost:4444/api/v3/portal/878609240/tags/2438969000000057311
image

### Pagination
Zoho BugTracker uses index-based pagination. you must pass `index` (starting position) and `range` (number of records to fetch) 
image

# Error 

Invalid url
image

provide incomplete payload
image

Invalid method
image


RajatPawar (commented):
@immdipu can we use the same oauth app that is used for `zoho` to also access bugtracker & projects? If we can use the same oauth app & the creds, then these should be modules on zoho rather than their own connectors, and we'll add support for modules soon - can you please add the context in the ticket, close this PR and move it to blocked until then? Thanks!

If we require separate oauth apps for all 3, then the current pr makes sense and they should be different connectors and you can merge this PR in. Approving to unblock

RajatPawar (approved):

providers/zoho.go
```diff
@@ -21,6 +32,37 @@ func init() {
...
+					Read:      true,
+					Subscribe: false,
+					Write:     true,
+				},
+			},
+			ModuleZohoProjects: {
+				BaseURL:     "https://projectsapi.zoho.com",
+				DisplayName: "Zoho Projects",
+				Support: Support{
+					Read:      true,
```
Cobalt0s, line 50: Shouldn't this be false? I don't think that it is used as of now though.

Cobalt0s (approved):
Looks good!

providers/zoho.go
immdipu: oh yeah. it should be false. Thanks for catching that

PR #1357, opened 2025-02-12
tidy(proxy): Split file, utils package

Cobalt0s:
# Background

The proxy for the OAuth2 password grant type no longer works. However, reverting to an earlier commit when it was introduced confirms that it previously functioned.
The solution is straightforward (one liner): the password grant type should delegate and construct a proxy identical to the authorization code grant type (not client credentials).

While I am at it, refactored the script.

# Motivation
The file is quite large, and while it didn’t take long to locate the issue, it was slightly longer than expected. Splitting the file into multiple smaller files helps to isolate similar-looking functions. There is no point to refactor more than that for little return.

# Changes
1. All proxies are now created using the same set of arguments. To avoid excessively long function signatures, these values are encapsulated in a struct `proxyserv.Factory{}`.
2. The `Proxy` struct and its methods were already well-organized, so moving them to a separate file was straightforward. The `Start` method is now a member function of the `Proxy` struct.
3. Other files are now grouped by **authentication type**, with each extracting values from registry and creating the corresponding proxy using `factory` struct.

### DIFF
--- a/scripts/proxy/proxy.go
+++ b/scripts/proxy/proxy.go
@@ -3,28 +3,17 @@ package main
 
 import (
 	"context"
-	"errors"
 	"flag"
-	"fmt"
 	"log"
 	"log/slog"
-	"net"
-	"net/http"
-	"net/http/httputil"
-	"net/url"
 	"os"
 	"os/signal"
-	"strings"
 
-	"github.com/amp-labs/connectors/common"
 	"github.com/amp-labs/connectors/common/paramsbuilder"
 	"github.com/amp-labs/connectors/common/scanning"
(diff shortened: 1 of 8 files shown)


Cobalt0s:
* **#1357** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1357?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1357?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

laurenzlong (approved):

PR #1359, opened 2025-02-13
[CON-1092] feat(ConstantContact): Human-readable custom fields

Cobalt0s:
# Background
The Contacts resource includes custom fields identified by UUIDs.

By making an API call to `contact_custom_fields`, we can retrieve an "explanation" for these fields. Using this information, both `ListObjectMetadata` and `Read` now return human-readable custom fields alongside the base fields.

# Tests
ListObjectMetadata(Contacts):
image
Read(Contacts):
image


laurenzlong (approved):
Nice!

PR #1360, opened 2025-02-13
refactor: Hubspot getMarshalledData -> getDataMarshaller

laurenzlong:
No functional change, some refactors to make code easier to understand & reduce repetitive code:
- renamed getMarshalledData to getDataMarshaller to make it more obvious it's a higher order function
- revised getDataMarshaller to handle case when len(fields) is 0, so that GetRecordsFromIds doesn't need to implement it that logic
- also modified the test the check that full raw record is returned, I'm surprised it wasn't failing before

### DIFF
--- a/providers/hubspot/parse.go
+++ b/providers/hubspot/parse.go
@@ -118,8 +118,9 @@ func getRecords(node *ajson.Node) ([]map[string]interface{}, error) {
 	return out, nil
 }
 
-// getMarshalledData accepts a list of records and returns a list of structured data ([]ReadResultRow).
-func (c *Connector) getMarshalledData(
+// getDataMarshaller returns a function that accepts a list of records and fields
+// and returns a list of structured data ([]ReadResultRow).
+func (c *Connector) getDataMarshaller(
 	ctx context.Context,
 	objName string,
 	associatedObjects []string,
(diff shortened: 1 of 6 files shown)


Cobalt0s (approved):
LGTM

providers/hubspot/parse.go
```diff
@@ -118,8 +118,9 @@ func getRecords(node *ajson.Node) ([]map[string]interface{}, error) {
 	return out, nil
 }
 
-// getMarshalledData accepts a list of records and returns a list of structured data ([]ReadResultRow).
-func (c *Connector) getMarshalledData(
+// getDataMarshaller returns a function that accepts a list of records and fields
+// and returns a list of structured data ([]ReadResultRow).
+func (c *Connector) getDataMarshaller(
```
Cobalt0s, line 123: @RajatPawar [this PR](https://github.com/amp-labs/connectors/pull/1321) is related to this function. 

Hubspot is on the same list as Klaviyo/Atlassian/Keap, etc. The implementation would become more consize if this function would work with `*ajson.Node` which is easier for parsing into the struct or queried using custom `jsonquery` package.

Currently, `common.ParseResult` enforces the use of `map[string]any`, adding unnecessary complexity to the code for some connectors.

Once the decision is made regarding `common.ParseResult` I will revisit to Hubspot implementation to simplify it even further. Essentially we need to convert `ajson.Node` into struct `{"id":"", "properties": {}}`.

PR #1362, opened 2025-02-14
feat(token): Write access/refresh tokens to creds.json

Cobalt0s:
# Background

Generating a token and repeatedly filling in `accessToken/refreshToken` is a tedious and repetitive task that should be automated.

# Requirements
* The `creds.json` file should be updated when a new access token is acquired.
* The `<provider>-creds.json` file should be updated as well. Ex: `keap-creds.json`.
* `creds.json` and `keap-creds.json` should be updated independently. Ex: if `clientId` is different in both files it should stay that way. We focus only on `accessToken/refreshToken/expiry/expiryFormat`.
* The `expiry` field should be populated to enable proper use of the refresh token (manually it is never updated making it overlooked).
* Other fields should remain unchanged.
* The order of lines should be preserved to allow users to group related fields as desired. 
image

# Implementation
To handle file loading, modification, and order preservation efficiently, this implementation uses [gitlab.com/c0b/go-ordered-json](https://gitlab.com/c0b/go-ordered-json). This library provides an ordered map[string]any that works well with the JSON marshaller.

Note: As a side effect, any extra new lines in the file will be removed. However, all other requirements are met.

# Usage
This functionality is NOT applied by default. Credentials files will be updated only if the script will be started using `--writeCreds` flag.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,6 +18,7 @@ require (
 	github.com/spf13/viper v1.19.0
 	github.com/spyzhov/ajson v0.9.6
 	github.com/stretchr/testify v1.10.0
+	gitlab.com/c0b/go-ordered-json v0.0.0-20201030195603-febf46534d5a
 	golang.org/x/exp v0.0.0-20250103183323-7d7fa50e5329
 	golang.org/x/oauth2 v0.26.0
 	golang.org/x/text v0.22.0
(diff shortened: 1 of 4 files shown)


Cobalt0s:
* **#1362** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1362?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1362?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

laurenzlong (approved):

PR #1370, opened 2025-02-17
[CON-1061] feat: Add Freshdesk Metadata Connector

Jkarage:
There are 3 objects having metadata resources(*object_fields*). But they return fewer fields, than querying data.
- [ticket_fields](https://developers.freshdesk.com/api/#list_all_company_fields) 
- [contact_fields](https://developers.freshdesk.com/api/#list_all_contact_fields)
- [company_fields](https://developers.freshdesk.com/api/#list_all_company_fields)

so currently we don't use these resources to generate the metadata.

Tests:
Screenshot 2025-02-17 at 11 07 38


providers/freshdesk/metadata.go
```diff
@@ -0,0 +1,76 @@
+package freshdesk
+
+import (
+	"context"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/naming"
+)
+
+const restAPIPrefix = "api/v2"
```
jlimatampersand, line 10: I think this should be in `connector.go`?

jlimatampersand (approved):
one minor comment, please resolve it, then feel free to merge.

Thanks!

PR #1371, opened 2025-02-17
[CON-1062] feat: Add Freshdesk Read Connector

Jkarage:
Tests
Screenshot 2025-02-17 at 13 26 10

### DIFF
--- a/common/parse.go
+++ b/common/parse.go
@@ -135,24 +135,3 @@ func getRecords(optional bool, jsonPath string, nestedPath ...string) RecordsFun
 		return jsonquery.Convertor.ArrayToMap(arr)
 	}
 }
-
-// ParseLinkHeader extracts the next page URL from the Link Header response.
-func ParseNexPageLinkHeader(linkHeader string) string {
-	var url string
-
-	if linkHeader == "" {
-		return "" // this indicates we're done.
-	}
-
-	links := strings.Split(linkHeader, ",")
-	// [<https://dev269415.service-now.com/api/now/v2/table/incident?sysparm_limit=1&sysparm_offset=0>;rel="next" ...]
-	for _, link := range links {
-		if strings.Contains(link, `rel="next"`) {
-			urls := strings.Split(link, ";")
-			url = strings.TrimPrefix(urls[0], "<")
-			url = strings.TrimRight(url, ">")
-		}
-	}
-
-	return url
-}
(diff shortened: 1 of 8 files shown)


providers/freshdesk/metadata.go
```diff
@@ -34,8 +34,7 @@ func (conn *Connector) ListObjectMetadata(ctx context.Context,
 			return nil, err
 		}
 
-		url.WithQueryParam("page", "1")
```
jlimatampersand, line 37: what is the reason this was removed? was this "page" never used?

common/parse.go
```diff
@@ -125,3 +125,24 @@ func getRecords(optional bool, jsonPath string, nestedPath ...string) RecordsFun
 		return jsonquery.Convertor.ArrayToMap(arr)
 	}
 }
+
+// ParseLinkHeader extracts the next page URL from the Link Header response.
```
jlimatampersand, line 129: this seems to be in wrong package. This is specific to Freshdesk, so it should not be in the common package.

jlimatampersand: Probably should be in `parse.go`

providers/freshdesk/parse.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"github.com/amp-labs/connectors/common"
+	"github.com/spyzhov/ajson"
+)
+
+func nextRecordsURL(link string) common.NextPageFunc {
+	return func(n *ajson.Node) (string, error) {
+		// if link is available, this indicates there's a next page url else there isn't.
+		// link format `<https://ampersand.freshdesk.com/api/v2/contacts?per_page=1&page=2>; rel="next"`
+		return common.ParseNexPageLinkHeader(link), nil
```
jlimatampersand, line 12: wrong package to import from since this is specific to freshdesk

common/parse.go
Jkarage: Okay, Noted. Updating this.

providers/freshdesk/parse.go
```diff
@@ -0,0 +1,37 @@
...
+// ParseLinkHeader extracts the next page URL from the Link Header response.
+func ParseNexPageLinkHeader(linkHeader string) string {
+	var url string
+
+	if linkHeader == "" {
+		return "" // this indicates we're done.
+	}
+
+	links := strings.Split(linkHeader, ",")
+	// [<https://dev269415.service-now.com/api/now/v2/table/incident?sysparm_limit=1&sysparm_offset=0>;rel="next" ...]
```
jlimatampersand, line 27: Since now this is just for this provider only, this example in comment should be updated.

jlimatampersand (approved):
I left comment on the parser function with regards to the example in comment. Feel free to merge after resolving the comment. Thanks!

PR #1372, opened 2025-02-17
[CON-1162] feat (zoom) : Add ListMetadata + Read + Write

immdipu:
# Testing

-  ListMetadata
image

- Read 
image

- Write
image


providers/zoom/objectNames.go
```diff
@@ -0,0 +1,50 @@
...
+var supportedObjectsByRead = metadata.Schemas.ObjectNames() // nolint: gochecknoglobals
+var (
+	ObjectNameContactGroup  = "contacts_groups" // nolint: gochecknoglobals
+	ObjectNameUser          = "users"           // nolint: gochecknoglobals
+	ObjectNameGroup         = "groups"          // nolint: gochecknoglobals
+	objectNameTrackingField = "tracking_fields" // nolint: gochecknoglobals
+)
+
+// ObjectNameToResponseField maps ObjectName to the response field name which contains that object.
+var ObjectNameToResponseField = common.ModuleObjectNameToFieldName{ // nolint: gochecknoglobals
```
Cobalt0s, line 18: Ok, this mapping is no longer needed. `schema.json` evolved over time and you can see that it now has `responseKey` field. It is essentially the mapping you have but part of `schema.json`. I will create a follow PR removing usage of `common.ModuleObjectNameToFieldName` in the READ operations.

Instead you should request the static file to resolve a field name that holds the data for you. Example:
https://github.com/amp-labs/connectors/blob/ff272b8534545b87b6528b92f8cadbda804c4f2d/providers/constantcontact/read.go#L32

providers/zoom/read.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	rsp, err := c.Client.Get(ctx, url.String())
+	if err != nil {
+		return nil, err
+	}
+
+	responseFieldName := ObjectNameToResponseField[c.Module.ID].Get(config.ObjectName)
```
Cobalt0s, line 30: As mentioned above, use: 
https://github.com/amp-labs/connectors/blob/ff272b8534545b87b6528b92f8cadbda804c4f2d/providers/constantcontact/read.go#L32

providers/zoom/write.go
```diff
@@ -0,0 +1,81 @@
...
+	"context"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/jsonquery"
+	"github.com/spyzhov/ajson"
+)
+
+// Each object has different fields that represent the record id.
+// This map is used to get the record id field for each object.
+var recordIdPaths = map[string]string{ //nolint:gochecknoglobals
```
Cobalt0s, line 13: Use `common.ModuleObjectNameToFieldName`. This data type is still relevant for WRITE operations, but for READ would rather use `schema.json`.
https://github.com/amp-labs/connectors/blob/ff272b8534545b87b6528b92f8cadbda804c4f2d/providers/keap/objectNames.go#L188-L198

test/zoom/write/main.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// Set up slog logging.
+	utils.SetupLogging()
+
+	conn := connTest.GetZoomConnector(ctx, zoom.ModuleUser)
+
+	slog.Info("> TEST Create User")
+	slog.Info("Creating a user...")
+
+	createUser(ctx, conn, &CreateUserPaylod{
```
Cobalt0s, line 45: Are there any objects that support both create and update? It would be nice to have an update functionality included as well.
PS: I also include the delete, but it is really not a required operation. It just makes it easier for the script to create/update and cleanup, so I could run it many times without worrying about data accumilation.

Cobalt0s (approved):
Good job!

providers/zoom/objectNames.go
Cobalt0s: A follow up PR: https://github.com/amp-labs/connectors/pull/1375
Zendesk is the only one that used `ModuleObjectNameToFieldName` for reading. Fallback to `schema.json`.

test/zoom/write/main.go
immdipu: I'll implement update and delete functionality iteratively.

PR #1373, opened 2025-02-17
[CON-1063] feat: Add Freshdesk Write Connector

Jkarage:
Tests:
Screenshot 2025-02-17 at 15 14 49


providers/freshdesk/types.go
```diff
@@ -4,34 +4,60 @@ import "slices"
...
+	"tickets",
+	"time_entries",
 }
 
-// objectReadPath represents a mapping of an object to it's read path.
-var objectReadPath = map[string]string{ //nolint:gochecknoglobals
-	"mailboxes": "email/mailboxes",
-	"settings":  "settings/helpdesk",
-	"skills":    "admin/skills",
+var writeSupportedObjects = []string{ //nolint:gochecknoglobals
```
jlimatampersand, line 30: We can make this a set or map for instant lookup and optimize.

providers/freshdesk/types.go
```diff
@@ -4,34 +4,60 @@ import "slices"
 
 // readSupportedObjects represents a list of objects supported by Read Connector.
 var readSupportedObjects = []string{ //nolint:gochecknoglobals
```
jlimatampersand, line 6: We can make this a set or map and optimize

providers/freshdesk/write.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	return constructWriteResponse(resp), nil
+}
+
+func constructWriteResponse(resp *common.JSONHTTPResponse) *common.WriteResult {
+	res, err := common.UnmarshalJSON[map[string]any](resp)
+	if err != nil {
+		return &common.WriteResult{
+			Success: true,
```
jlimatampersand, line 44: How does `err != nil` resolve to a success? Can you add an explanation here in the comment?

jlimatampersand: Are we expecting no body? Can what do you get from log when you hit this?

jlimatampersand (changes requested):
that JSON unmarshal error handling seems odd to me. Please respond to the comment or resolve if there is any to change. Feel free to ping me when you resolve or answer questions. Thanks!

providers/freshdesk/write.go
Jkarage: So it's a precaution, just incase the response is not a map, and we can't unmarshal the data to a map. But since we reached here, This means we got 2XX status code. So we wont get the RecordId, but we have created/updated the data. It shouldn't happen, but it's just a precaution.

jlimatampersand (approved):

PR #1378, opened 2025-02-18
feat: Construct proxy connector with metadata+catalogVars

Cobalt0s:
# Background
We are not limited to workspace catalog variables. The Google connector requires support for multiple modules, each differing by subdomain. This can only be communicated via metadata.
AWS connector also relies uppon region and some service specific metadata.

# Changes
* `connectors.NewConnector` now accepts `WithMetadata` option.
* `paramsbuilder.Metadata` returns all values as potential catalog variables.
* Proxy instances, regardless of auth type, capture substitutions and use the `WithMetadata` option.

### DIFF
--- a/connector/params.go
+++ b/connector/params.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/amp-labs/connectors/common"
 	"github.com/amp-labs/connectors/common/paramsbuilder"
+	"github.com/amp-labs/connectors/common/substitutions/catalogreplacer"
 	"github.com/amp-labs/connectors/providers"
 	"golang.org/x/oauth2"
 )
@@ -25,6 +26,7 @@ type parameters struct {
 	provider providers.Provider
 	paramsbuilder.Client
 	paramsbuilder.Workspace
+	paramsbuilder.Metadata
 }
 
 func (p parameters) ValidateParams() error {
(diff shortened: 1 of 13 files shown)


Cobalt0s:
* **#1378** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1378?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1378?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

eberle1080 (approved):

laurenzlong:
@Cobalt0s please don't merge this until @RajatPawar has had a chance to review - since he's doing some related work on provider metadata.

connector/params.go
```diff
@@ -61,3 +63,22 @@ func WithWorkspace(workspaceRef string) Option {
 		params.WithWorkspace(workspaceRef)
 	}
 }
+
+// WithMetadata sets authentication metadata expected by connector.
```
RajatPawar, line 67: We should also convert this to use common.Parameters soon - but looks good for now!

RajatPawar (approved):
Looks good! Please make sure that you also make the equivalent changes in the server to supply this metadata.

connector/params.go
Cobalt0s: `common.Parameters` already has metadata. So it is already resolved :smile:. By setting common.Parameters.Metadata everything is initialized properly. Also validation is applied correctly. Thats what I used for AWS connector.

This PR is from February and was initialy created for Google. That was posponed a bit and I didn't add reviewers until now. I remembered that I worked on it and its done so I based Graphite stack this way.

Cobalt0s:
> Looks good! Please make sure that you also make the equivalent changes in the server to supply this metadata.

There is no connector that supplies metadata this way. This would be relevant for proxy connectors that require metadata. I will save this as todo to see if it can be generalized in server.

PR #1379, opened 2025-02-19
[CON-1201] feat: add Google Meet connector

immdipu:
## Checklist
- [x] Ran Linter

### GET
URL: localhost:4444/v2/conferenceRecords
image

### GEt
URL: localhost:4444/v2/conferenceRecords/c2416e0b-11ec-40b5-8ec2-ec85fc647d5f
image

### POST
URL: localhost:4444/api/v2/spaces
image

### PATCH
URL: localhost:4444/v2/spaces/uUHtIF4l-3cB
image

### Pagination
The list method supports pagination, returning up to 25 files per page by default (max 100 via `pageSize`). Uses `nextPageToken` from the response as `pageToken` in the query to fetch the next page.

image

Next page:
image

### Errors

provide invalid data
image

provide an incorrect dynamic ID in the URL.
image

### DIFF
--- a/providers/googleMeet.go
+++ b/providers/googleMeet.go
@@ -1,0 +1,47 @@
+package providers
+
+const (
+	GoogleMeet Provider = "googleMeet"
+)
+
+//nolint:funlen
+func init() {
+	SetInfo(GoogleMeet, ProviderInfo{
+		DisplayName: "Google Meet",
+		AuthType:    Oauth2,
+		BaseURL:     "https://meet.googleapis.com",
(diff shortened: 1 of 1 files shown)


laurenzlong (commented):
This should be a module of the Google Connector instead of its own connector. So this cannot be merged or released. I'm marking the Linear ticket as blocked on module support. I'm closing this PR for now but it's linked in Linear so we can transition it into a module later. Thanks for the work here @immdipu , sorry for the miscommunication that led to this.

PR #1387, opened 2025-02-21
[CON-1247] feat(gong): add read support for transcripts object

immdipu:
### Testing 
image


providers/gong/read_test.go
```diff
@@ -196,6 +197,30 @@ func TestRead(t *testing.T) { //nolint:funlen,gocognit,cyclop
...
+			Server: mockserver.Fixed{
+				Setup:  mockserver.ContentJSON(),
+				Always: mockserver.Response(http.StatusOK, responseTranscripts),
+			}.Server(),
+			Comparator: testroutines.ComparatorSubsetRead,
+			Expected: &common.ReadResult{
+				Rows: 1,
+				Data: []common.ReadResultRow{{
+					Fields: map[string]any{
+						"callId": "7782342274025937895",
```
Cobalt0s, line 213: Ok, so changing the `callId` to `callid` fixes the build.
The Fields is lowered because of this line: https://github.com/amp-labs/connectors/blob/93cb96712a6e97975ba77363198fdfd86ecde3b2/common/parse.go#L102

providers/gong/read.go
```diff
@@ -17,23 +17,21 @@ func (c *Connector) Read(ctx context.Context, config common.ReadParams) (*common
...
-	}
+	var res *common.JSONHTTPResponse
 
-	if !config.Since.IsZero() {
-		// This time format is RFC3339 but in UTC only.
-		// See calls or users object for query parameter requirements.
-		// https://gong.app.gong.io/settings/api/documentation#get-/v2/calls
-		url.WithQueryParam("fromDateTime", datautils.Time.FormatRFC3339inUTC(config.Since))
+	if postReadObjects.Has(config.ObjectName) {
+		body := buildReadbody(config)
```
Cobalt0s, line 28: `buildReadbody` -> `buildReadBody`

scripts/openapi/gong/metadata/main.go
```diff
@@ -63,12 +72,22 @@ func main() {
...
 
-	objects, err := explorer.ReadObjectsGet(
+	getReadObjects, err := explorer.ReadObjectsGet(
 		api3.NewDenyPathStrategy(ignoreEndpoints),
 		nil, nil, api3.IdenticalObjectLocator,
 	)
+
+	goutils.MustBeNil(err)
+
+	postReadObjects, err := explorer.ReadObjectsPost(
```
Cobalt0s, line 82: Thanks for using this!

scripts/openapi/gong/metadata/main.go
```diff
@@ -48,6 +49,14 @@ var (
 		// Requires query parameters: flowOwnerEmail.
 		"/v2/flows",
 	}
+
+	postReadAllowedEndPoints = []string{ // nolint:gochecknoglobals
+		"/v2/calls/transcript",
+	}
+
+	postObjectEndpoints = map[string]string{ // nolint:gochecknoglobals
+		"/v2/calls/transcript": "transcripts",
```
Cobalt0s, line 58: @RajatPawar this is related to what we are discussing. Many deep connectors so far follow URL to objectName mapping. Frankly, `transcripts` is a better objectName than `calls/transcript`. We for sure will have some names made up names because URLs would differ for the same object/concept. Having some objects as last URI part, some other objects as multiple URI parts and some made up all together would be even more confusing.

Cobalt0s (approved):

providers/gong/read_test.go
Cobalt0s: @RajatPawar a follow up question. We lower the fields for all deep connectors. I forgot the initial reason why it was done. Could remind me? I think it would be good to add a comment to `ExtractLowercaseFieldsFromRaw` to explain the reason. Dipu wondered about this and I cannot recall.

PR #1388, opened 2025-02-21
feat: add schema component to fetch one object at a time / in parallel

RajatPawar:
Some providers only let us fetch metadata for one object at a time, and not multiple objects at once. This introduces a `SingleObjectSchemaProvider` that can let you run metadata requests for N objects in parallel, or serially. 

This is in contrast with the existing ObjectSchemaProvider, which I've renamed to `Batch` because that's what it does - it lets you fetch metadata for N objects using a single http request.

### DIFF
--- a/internal/components/schema/composite.go
+++ b/internal/components/schema/composite.go
@@ -6,23 +6,24 @@ import (
 	"log/slog"
 
 	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/internal/components"
 )
 
 var ErrUnableToGetMetadata = errors.New("unable to get metadata")
 
 // CompositeSchemaProvider gets metadata from multiple providers with fallback.
-type CompositeObjectSchemaProvider struct {
-	schemaProviders []HTTPObjectSchemaProvider
+type CompositeSchemaProvider struct {
+	schemaProviders []components.SchemaProvider
 }
 
-func NewCompositeObjectSchemaProvider(schemaProviders ...HTTPObjectSchemaProvider) *CompositeObjectSchemaProvider {
-	return &CompositeObjectSchemaProvider{
+func NewCompositeSchemaProvider(schemaProviders ...components.SchemaProvider) *CompositeSchemaProvider {
+	return &CompositeSchemaProvider{
 		schemaProviders: schemaProviders,
 	}
 }
 
 // GetMetadata tries each schema provider in order, and returns the best result with the least errors.
-func (c *CompositeObjectSchemaProvider) GetMetadata(
+func (c *CompositeSchemaProvider) ListObjectMetadata(
 	ctx context.Context,
 	objects []string,
 ) (*common.ListObjectMetadataResult, error) {
@@ -63,13 +64,15 @@ func (c *CompositeObjectSchemaProvider) GetMetadata(
 		return nil, ErrUnableToGetMetadata
 	}
 
+	// TODO: Do a better implementation of best effort
+
 	return bestResult, nil
 }
 
 // safeGetMetadata is a helper function that safely executes the provider's GetMetadata method
 // and recovers from panics.
 func safeGetMetadata(
-	schemaProvider HTTPObjectSchemaProvider,
+	schemaProvider components.SchemaProvider,
 	ctx context.Context,
 	objects []string,
 ) (*common.ListObjectMetadataResult, error) {
@@ -86,7 +89,7 @@ func safeGetMetadata(
 		}
 	}()
 
-	result, err = schemaProvider.GetMetadata(ctx, objects)
+	result, err = schemaProvider.ListObjectMetadata(ctx, objects)
 
 	return result, err
 }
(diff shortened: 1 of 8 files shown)


internal/components/schema/single.go
```diff
@@ -0,0 +1,149 @@
...
+	ObjectName string
+	Response   common.ObjectMetadata
+}
+
+type objectMetadataError struct {
+	ObjectName string
+	Error      error
+}
+
+func (p *IndividualSchemaProvider) fetchParallel(
```
Cobalt0s, line 71: Is it possible for one of the requests to block forever?
I think this thought is not particular to this method. Goroutines remind of this. We should utilize context cancelation for all http.clients even for serial executions.

internal/components/schema/composite.go
```diff
@@ -86,7 +89,7 @@ func safeGetMetadata(
 		}
 	}()
 
-	result, err = schemaProvider.GetMetadata(ctx, objects)
+	result, err = schemaProvider.ListObjectMetadata(ctx, objects)
```
Cobalt0s, line 92:
```
return schemaProvider.ListObjectMetadata(ctx, objects)
```

internal/components/schema/single.go
```diff
@@ -0,0 +1,149 @@
...
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/internal/components/operations"
+)
+
+var (
+	FetchTypeParallel = "parallel" // nolint:gochecknoglobals
```
Cobalt0s, line 13: What do you think about renaming types to modes: `FetchModeParallel`?

providers/smartleadv2/connector.go
```diff
@@ -59,6 +60,23 @@ func constructor(base *components.Connector) (*Connector, error) {
...
+	// Other examples:
+	// 	schema.NewBatchSchemaProvider(
+	// 		connector.HTTPClient().Client,
+	// 		operations.ListObjectMetadataHandlers{
+	// 			BuildRequest:  connector.buildListObjectMetadataRequest,
+	// 			ParseResponse: connector.parseListObjectMetadataResponse,
+	// 		},
+	// 	),
+
+	//  schema.NewIndividualSchemaProvider(
```
Cobalt0s, line 72: For some reason at the beginning of review I thought that these schema providers are parallel vs serial. But now I see it is Batch all vs individual with 2 modes to enable parallel processing.

What do you think about changing the name Individual. Using ChatGPT for insipiration(still hard to find vocab): 
* NewBatchingSchemaProvider
* NewSingleSchemaProvider / NewSingularSchemaProvider

Or maybe another idea, given that there is metadata from static files:
* NewBatchFetchingSchemaProvider
* NewSingleFetchingSchemaProvider

Cobalt0s (approved):
I understood the 2 strategies for metadata, this is good!

internal/components/schema/composite.go
```diff
@@ -6,23 +6,24 @@ import (
...
-	return &CompositeObjectSchemaProvider{
+func NewCompositeSchemaProvider(schemaProviders ...components.SchemaProvider) *CompositeSchemaProvider {
+	return &CompositeSchemaProvider{
 		schemaProviders: schemaProviders,
 	}
 }
 
 // GetMetadata tries each schema provider in order, and returns the best result with the least errors.
-func (c *CompositeObjectSchemaProvider) GetMetadata(
+func (c *CompositeSchemaProvider) ListObjectMetadata(
```
Jkarage, line 26: Let's update the comment here.

providers/smartleadv2/connector.go
Jkarage: I would vouch for the latter: NewBatchFetchingSchemaProvider and NewSingleFetchingSchemaProvider.

internal/components/schema/single.go
RajatPawar: Good point! ExecuteRequest will send an HTTP request underneath the hood, which has a default timeout of 60 seconds. Anything over that and we will see a timeout error propogating over the channel. This is the exact code being used in hubspot's mechanism and it's been working, so I don't see any other issues there.

providers/smartleadv2/connector.go
RajatPawar: How about NewObjectSchemaProvider, which has two fetch modes - parallel & serial, and then NewAggregateSchemaProvider?

Cobalt0s: Maybe instead of calling SchemaProvider it is rather SchemaFetcher. Something to indicate that the object makes an API call. In golang I found out recently that it is prefered to name something fetch vs ordinary get when making API calls.

This object makes API calls to fetch object in batch/aggregate and singular with 2 modes. I prefer IndividualSchema over ObjectSchema. I am not sure what Provider is in this case, I assume it is refering to API fetching, right?

PR #1392, opened 2025-02-24
[ENG-1941] feat: add netsuite proxy & metadata connector 

RajatPawar:
## Catalog variables
- Account ID (workspace)

### Installation
Screenshot 2025-02-24 at 4 01 51 PM
Screenshot 2025-02-24 at 4 18 45 PM
Screenshot 2025-02-24 at 4 18 56 PM

### Object metadata (script in `test/netsuite/metadata`)
Screenshot 2025-02-24 at 6 52 49 PM

### GET
Screenshot 2025-02-24 at 6 43 35 PM

### POST
Screenshot 2025-02-24 at 6 44 01 PM

### DELETE
Screenshot 2025-02-24 at 6 45 20 PM

## Pagination
Screenshot 2025-02-24 at 6 47 14 PM

## Console operations
Screenshot 2025-02-24 at 6 56 13 PM

### DIFF
--- a/common/json.go
+++ b/common/json.go
@@ -195,11 +195,11 @@ var ErrMissingContentType = errors.New("missing content type")
 // If errOnMissing is true, an error is returned if the content type is missing.
 // Otherwise, the function returns nil.
 func EnsureContentType(pattern string, res *http.Response, errOnMissing bool) error {
-	ct := res.Header.Get("Content-Type")
+	ctype := res.Header.Get("Content-Type")
 
 	// If the content type is missing, return an error if errOnMissing is true,
 	// otherwise return nil.
-	if len(ct) == 0 {
+	if len(ctype) == 0 {
 		if errOnMissing {
 			return fmt.Errorf("%w: expected content type to be application/*json", ErrMissingContentType)
 		}
@@ -207,7 +207,7 @@ func EnsureContentType(pattern string, res *http.Response, errOnMissing bool) er
 		return nil
 	}
 
-	mimeType, _, err := mime.ParseMediaType(ct)
+	mimeType, _, err := mime.ParseMediaType(ctype)
 	if err != nil {
 		return fmt.Errorf("failed to parse content type: %w", err)
 	}
--- a/common/json.go
+++ b/common/json.go
@@ -136,9 +136,10 @@ func ParseJSONResponse(res *http.Response, body []byte) (*JSONHTTPResponse, erro
 			return nil, fmt.Errorf("failed to parse content type: %w", err)
 		}
 
-		// Providers implementing JSONAPISpeicifcations returns application/vnd.api+json
-		if mimeType != "application/json" && mimeType != "application/vnd.api+json" {
-			return nil, fmt.Errorf("%w: expected content type to be application/json or application/vnd.api+json , got %s",
+		// Check if the mimeType follows the pattern application/*json (e.g.
+		// application/vnd.api+json, application/schema+json, etc.)
+		if !strings.HasPrefix(mimeType, "application/") || !strings.HasSuffix(mimeType, "json") {
+			return nil, fmt.Errorf("%w: expected content type to be application/*json, got %s",
 				ErrNotJSON, mimeType,
 			)
 		}
@@ -163,6 +164,7 @@ func UnmarshalJSON[T any](rsp *JSONHTTPResponse) (*T, error) {
 	var data T
 
 	if len(rsp.bodyBytes) == 0 {
+
 		// Empty struct.
 		return &data, nil
 	}
--- a/common/json.go
+++ b/common/json.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"mime"
 	"net/http"
+	"regexp"
 	"strings"
 
 	"github.com/spyzhov/ajson"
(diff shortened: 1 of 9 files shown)


common/json.go
```diff
@@ -136,9 +136,10 @@ func ParseJSONResponse(res *http.Response, body []byte) (*JSONHTTPResponse, erro
 			return nil, fmt.Errorf("failed to parse content type: %w", err)
 		}
 
-		// Providers implementing JSONAPISpeicifcations returns application/vnd.api+json
-		if mimeType != "application/json" && mimeType != "application/vnd.api+json" {
-			return nil, fmt.Errorf("%w: expected content type to be application/json or application/vnd.api+json , got %s",
+		// Check if the mimeType follows the pattern application/*json (e.g.
+		// application/vnd.api+json, application/schema+json, etc.)
+		if !strings.HasPrefix(mimeType, "application/") || !strings.HasSuffix(mimeType, "json") {
```
jlimatampersand, line 141: Just a suggestion, maybe we can use regex match for this? `application/*json` should work for regex. Not really a big deal though.

providers/netsuite/handlers.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("getting properties object: %w", err)
+	}
+
+	result := &common.ObjectMetadata{
+		Fields: make(map[string]common.FieldMetadata),
+	}
+
+	// Iterate through fields
+	for property, metadata := range fields {
```
jlimatampersand, line 61: can we have some example data so we can see the data looks? it will make this easier to read.

jlimatampersand: or doc for the schema

providers/netsuite/handlers.go
```diff
@@ -0,0 +1,98 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+	"github.com/amp-labs/connectors/internal/jsonquery"
+	"github.com/spyzhov/ajson"
+)
+
+func (c *Connector) buildObjectMetadataRequest(ctx context.Context, object string) (*http.Request, error) {
```
jlimatampersand, line 15: Can we have link to api reference URL in comment above the method?

providers/netsuite.go
```diff
@@ -0,0 +1,40 @@
+package providers
+
+const Netsuite Provider = "netsuite"
+
+func init() {
+	SetInfo(Netsuite, ProviderInfo{
```
jlimatampersand, line 6: Can we add doc url?

providers/netsuite/handlers.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+
+		// If the field is actually a related object and not a first class field, skip it.
+		// If it doesn't have a type, skip it.
+		ftype, err := metadata.GetKey("type")
+		if err != nil {
+			continue
+		}
+
+		if ftype.IsNull() || ftype.MustString() != "object" {
```
jlimatampersand, line 74: must string panics. Should avoid and replace with other nil check instead. Chances are low, but netsuite may have some bug, that the field may be missing. Then this will panic.

jlimatampersand (approved):
Approved, to unblock, but there is code that may panic. Please fix before merging!

providers/netsuite/handlers.go
Cobalt0s: I dont like using ajson methods. It complicates the code very fast and we need to worry about nulls, field types, logic for reaching out nested objects, etc. At the same time `jsonquery` package is only good for logic which is short and operates on few fields.

We own a reference to `resp`. This `parseObjectMetadataResponse` method would become simpler if we would use:
```
common.UnmarshalJSON[responseFormat](resp)
```

The `responseFormat` struct should have a document Jay is asking. It will be more aparant on what is going on in a central place.
You can create a member method that would check if the object's field satisfies certain criteria, if not continue the loop.

test/netsuite/connector.go
```diff
@@ -0,0 +1,62 @@
...
+	options := []common.OAuthOption{
+		common.WithOAuthClient(http.DefaultClient),
+		common.WithOAuthConfig(getConfig(nil)),
+		common.WithOAuthToken(&oauth2.Token{
+			AccessToken:  reader.Get(credscanning.Fields.AccessToken),
+			RefreshToken: reader.Get(credscanning.Fields.RefreshToken),
+		}),
+	}
+
+	client, err := common.NewOAuthHTTPClient(ctx, options...)
```
Cobalt0s, line 28: This method could be moved under `test/utils`. Ex: `utils.MustAuthClient(...common.OAuthOption)`
Connector can be build in one chunk:
```
conn, err := netsuite.NewConnector(common.Parameters{
	AuthenticatedClient: utils.MustAuthClient(
		common.WithOAuthClient(http.DefaultClient),
		common.WithOAuthConfig(getConfig(nil)),
		common.WithOAuthToken(&oauth2.Token{
			AccessToken:  reader.Get(credscanning.Fields.AccessToken),
			RefreshToken: reader.Get(credscanning.Fields.RefreshToken),
		})),
	Workspace:           reader.Get(credscanning.Fields.Workspace),
})
```

test/netsuite/metadata/main.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		utils.Fail("error listing metadata for Netsuite", "error", err)
+	}
+
+	for obj, metadata := range response.Result {
+		fmt.Printf("Metadata for %s: %+v\n\n\n", obj, metadata)
+	}
+}
+
+// TODO: Implement this when read is done.
```
Cobalt0s, line 39: I don't think that it is needed. I was thinking about moving common `templates` on how we conduct tests into a `test/utils`. Let me know if you think it is needed. (So far some connectors benefited from the procedure described below).

Note: the raw print sometimes is preferred for the manual testing, which you have above. There could be more than 1 test script.

Cobalt0s (changes requested):
I think `parseObjectMetadataResponse` can be improved. Let me know if my suggestion works in practice.

Cobalt0s (approved):
I see Jay mentioned: "approved to unblock". I will do the same.

providers/netsuite/handlers.go
RajatPawar: Adding an example response to the test dir

RajatPawar:
> I think `parseObjectMetadataResponse` can be improved. Let me know if my suggestion works in practice.
> 
> Question: why proxy is combined with metadata impl?

We want to get this connector out ASAP

PR #1393, opened 2025-02-24
set json as the default content-type for sending data

Jkarage:


RajatPawar (approved):
This is ideally a client behavior, and we should inject the client with this configuration, but this should be good for now.

PR #1399, opened 2025-02-25
feat: Update supported module validation 

Jkarage:
- Updated the metadata validation as well.

### DIFF
--- a/common/params.go
+++ b/common/params.go
@@ -100,13 +100,13 @@ type RequireMetadata struct {
 	ExpectedMetadataKeys []string
 }
 
-func (RequireMetadata) validateMetadata(parameters Parameters) error {
+func (m RequireMetadata) validateMetadata(parameters Parameters) error {
 	if parameters.Metadata == nil {
 		return ErrMissingMetadata
 	}
 
-	for _, key := range parameters.Metadata {
-		if key == "" {
+	for _, key := range m.ExpectedMetadataKeys {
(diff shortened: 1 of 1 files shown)


common/params.go
```diff
@@ -96,16 +97,16 @@ type metadataValidator interface {
...
 
-func (RequireMetadata) validateMetadata(parameters Parameters) error {
+func (m RequireMetadata) validateMetadata(parameters Parameters) error {
 	if parameters.Metadata == nil {
 		return ErrMissingMetadata
 	}
 
-	for _, key := range parameters.Metadata {
-		if key == "" {
+	for _, key := range m.ExpectedMetadataKeys {
```
RajatPawar, line 108: Thanks for catching this! Might be a good idea to make this case insensitive and add an comment saying that in the struct

common/params.go
```diff
@@ -119,12 +120,18 @@ type moduleValidator interface {
...
+}
 
-func (RequireModule) validateModule(parameters Parameters) error {
+func (r RequireModule) validateModule(parameters Parameters) error {
 	if parameters.Module == "" {
 		return ErrMissingModule
 	}
 
+	if !slices.Contains(r.ExpectedModules, parameters.Module) {
+		return ErrMissingModule
```
RajatPawar, line 133: This should ideally be unsupported module, not missing module

RajatPawar (approved):
two minor comments, LGTM

PR #1405, opened 2025-02-26
[CON-1070] feat: Add Front Read Connector

Jkarage:
Tests:
Screenshot 2025-02-26 at 12 57 10

### DIFF
--- a/providers/front/connector.go
+++ b/providers/front/connector.go
@@ -5,6 +5,8 @@ import (
 
 	"github.com/amp-labs/connectors/common"
 	"github.com/amp-labs/connectors/internal/components"
+	"github.com/amp-labs/connectors/internal/components/operations"
+	"github.com/amp-labs/connectors/internal/components/reader"
 	"github.com/amp-labs/connectors/internal/components/schema"
 	"github.com/amp-labs/connectors/internal/staticschema"
 	"github.com/amp-labs/connectors/providers"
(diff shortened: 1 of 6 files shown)


test/front/connector.go
```diff
@@ -15,8 +15,18 @@ func GetFrontConnector(ctx context.Context) *front.Connector {
...
+	}
+
+	headerName, headerValue, err := info.GetApiKeyHeader(reader.Get(credscanning.Fields.ApiKey))
+	if err != nil {
+		utils.Fail(err.Error())
+	}
+
 	client, err := common.NewApiKeyHeaderAuthHTTPClient(
-		ctx, "Authorization", reader.Get(credscanning.Fields.ApiKey),
+		ctx, headerName, headerValue,
```
Jkarage, line 29: Would we need to wrap this in it's own function?

providers/front/connector.go
```diff
@@ -49,5 +50,21 @@ func constructor(base *components.Connector) (*Connector, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	connector.Reader = reader.NewHTTPReader(
+		connector.HTTPClient().Client,
+		registry,
+		connector.ProviderContext.Module(),
+		operations.ReadHandlers{
+			BuildRequest:  connector.buildRequest,
```
RajatPawar, line 63: since these methods will also exist for other operations, can you name these `buildReadRequest` and `parseReadRequest`? The parseResponse method also seems to be named incorrectly

test/front/connector.go
RajatPawar: I believe `providers.NewClient` will help with this

RajatPawar (approved):
Some changes required, please resolve them before merging. Can you also make sure that unsupported object reads are rejected?

PR #1406, opened 2025-02-26
CON-1071 feat: Add Front Write Connector

Jkarage:
Test:
Screenshot 2025-02-26 at 13 21 08


providers/front/handlers.go
```diff
@@ -44,3 +47,55 @@ func (c *Connector) parseWriteResponse(
...
+
+	url, err := urlbuilder.New(c.ProviderInfo().BaseURL, params.ObjectName)
+	if err != nil {
+		return nil, err
+	}
+
+	if params.RecordId != "" {
+		url.AddPath(params.RecordId)
+
+		method = http.MethodPatch
```
RajatPawar, line 62: Is this for both create and update?

providers/front/supports.go
```diff
@@ -16,12 +16,24 @@ func supportedOperations() components.EndpointRegistryInput {
...
+	}
+
 	return components.EndpointRegistryInput{
 		staticschema.RootModuleID: {
 			{
 				Endpoint: fmt.Sprintf("{%s}", strings.Join(readSupport, ",")),
 				Support:  components.ReadSupport,
 			},
+			{
+				Endpoint: fmt.Sprintf("{%s}", strings.Join(writeSupports, ",")),
```
RajatPawar, line 34: You could add `components.ReadWriteSupport` to combine this, or a function that accepts two supports and merges it - i think the merge function already exists. But if it's more than a couple of minutes worth of effort, feel free to skip it

RajatPawar (approved):
LGTM, can you please check the error cases? (4XX codes on writes, or some other errors for read) and verify if things are as expected?

providers/front/handlers.go
Jkarage: I have declared post for create in LN52.

providers/front/supports.go
Jkarage: I will look into this.

Jkarage:
Yeah, I have tried different request scenarios, And the response is as expected.

PR #1408, opened 2025-02-26
[CON-1420] feat(blueshift): add ListMetadata

immdipu:
# Testing
image


providers/blueshift/metadata/scraping.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/amp-labs/connectors/tools/scrapper"
+)
+
+var (
+	// Static file containing a list of object metadata is embedded and can be served.
+	//
+	//go:embed schemas.json
+	schemas []byte
+
+	FileManager = scrapper.NewMetadataFileManager[staticschema.FieldMetadataMapV1]( //nolint:gochecknoglobals
```
Cobalt0s, line 17: Could you change usage of `staticschema.FieldMetadataMapV1` to `staticschema.FieldMetadataMapV2`?

There was a redesign on the FiledMetadata which holds more information.

Version 1 had only field display names, Version 2 has field type, field values for enums, etc.

providers/blueshift/objectNames.go
```diff
@@ -0,0 +1,19 @@
+package blueshift
+
+import (
+	"github.com/amp-labs/connectors/internal/datautils"
+)
+
+// var supportedObjectsByRead = metadata.Schemas.ObjectNames()
+
+var ObjectNametoResponseField = datautils.NewDefaultMap(map[string]string{ //nolint:gochecknoglobals
```
Cobalt0s, line 9: I think this should be moved to `scripts/openapi/blueshift/metadata/main.go`

providers/blueshift/connector.go
```diff
@@ -0,0 +1,48 @@
+package blueshift
+
+import (
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/paramsbuilder"
+	"github.com/amp-labs/connectors/providers"
+)
+
+type Connector struct {
```
Cobalt0s, line 9: There is a new approach for building deep connectors. Before this connector gets too large could you try it out.
For inspiration check `smartleadv2`.

providers/blueshift/objectNames.go
Cobalt0s: It seems that with your special edge case there needs to be an update to the structure of `schema.json`.
You see, for now it holds `responseKey` assuming that nothing is nested.

But from the response samples you sent me for templates the responseKey would be an "array": first reach "templates" and then "results".
```
{
    "templates": {
        "total_count": 1.0,
        "total_pages": 0,
        "per_page": 2,
        "page": 0,
        "results": [
            {
                "name": "Updated Email Template",
                "created_at": "2024-06-15T07:39:15.000Z",
                "updated_at": "2024-08-07T10:04:13.000Z",
                "uuid": "7b8550aa-1d0d-4a9f-9495-fa967963e227",
                "archived": false,
                "author": "Integration User",
                "resource": {
                    "updated_at": "2024-06-15T07:39:15.000Z"
                }
            }
        ]
    }
}
```
vs 
```
{
    "templates": [
        {
            "name": "Updated Email Template",
            "created_at": "2024-06-15T07:39:15.000Z",
            "updated_at": "2024-08-07T10:04:13.000Z",
            "uuid": "7b8550aa-1d0d-4a9f-9495-fa967963e227",
            "archived": false,
            "author": "Integration User",
            "resource": {
                "updated_at": "2024-06-15T07:39:15.000Z"
            }
        }
    ]
}
```

Currently there are a few connectors that have consistent nesting of fields and it is hard coded in the read logic (Klaviyo, Atlassian, Outreach):
https://github.com/amp-labs/connectors/blob/7b2b16882ee0c36886646b6563c589d0195e18cf/providers/klaviyo/parse.go#L40
and OpenAPI script:
https://github.com/amp-labs/connectors/blob/7b2b16882ee0c36886646b6563c589d0195e18cf/scripts/openapi/klaviyo/metadata/main.go#L36-L39

For your case it is true for only several objects. There are 2 solutions:
First:
* Use `WithPropertyFlattening` for the specified objects
* Read would get from `schema.json` "responseKey": "templates" and then hard code the rest to reach for nested "results".
Second:
* The `WithPropertyFlattening` should be modified such that `schema.json` would be self-sufficient and record the whole JSON path from the root. Ex: `responseKey: string` + `optional nestedPath: []string`.

When designing the solution we need to consider that not all deep connectors will use `schema.json`.

@jlimatampersand @RajatPawar do you have opinions on this?
Did I understand the problem correctly @immdipu?

Cobalt0s (changes requested):

providers/blueshift/objectNames.go
immdipu: going with the first option as it is the most time-efficient to implement and won’t affect any other deep connectors, unlike the second option.

providers/blueshift/connector.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/amp-labs/connectors/tools/fileconv"
+	"github.com/amp-labs/connectors/tools/scrapper"
+)
+
+// nolint:gochecknoglobals
+var (
+	//go:embed schemas.json
+	schemaContent []byte
+	FileManager   = scrapper.NewMetadataFileManager[staticschema.FieldMetadataMapV2](
+		schemaContent, fileconv.NewSiblingFileLocator())
```
Cobalt0s, line 20: `fileconv.NewSiblingFileLocator()` assumes that schema.json is on the same directory level so it will be under `providers/blueshift/schema.json` when performing write. This code should live under metadata package to seperate `schemas.json` from this repo. The fewer files in this repo the better.

Note: that metadata package could have additional wrappers for schemas, example:
https://github.com/amp-labs/connectors/blob/769148865056e292ee4ac5a530e737fa9f4837a1/providers/zendesksupport/metadata/metadata.go#L36-L38

providers/kit/metadata/schemas.json
```diff
@@ -1,144 +1,144 @@
 {
-    "modules": {
```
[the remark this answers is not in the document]
immdipu: I accidentally imported the schemas file for the kit instead of Blueshift. I made a few commits and then had to revert the change.

Cobalt0s (approved):
Add `metadata_test.go` and it is good to go.

scripts/openapi/blueshift/metadata/main.go
```diff
@@ -0,0 +1,102 @@
...
+	for _, object := range readObjects { //nolint:gochecknoglobals
+		if object.Problem != nil {
+			slog.Error("Schema not extracted",
+				"objectName", object.ObjectName,
+				"error", object.Problem,
+			)
+		}
+
+		for _, field := range object.Fields {
+			schemas.Add("", object.ObjectName, object.DisplayName, object.URLPath, object.ResponseKey,
```
Cobalt0s, line 89: Ok, before saving the object.URLPath, you should cut prefix "api".
This way it will not conflict with BaseURL which already includes "api".

jlimatampersand (approved):

PR #1409, opened 2025-02-27
[ENG-1889] Salesforce Subscribe Add EmptyResult method to connector interface

jlimatampersand:
Problem:
- data is saved in the database as `byte[]`, server needs to convert it to the correct format to pass to the connector for updating a subscription. 

### Updated Solution
When we want to json unmarshal, we basically need the container or shell for the data. 

Instead of the connector interface to have json unmarshal interface methods, we instead have NewSubscribeParams, NewRegistrationParams, NewSubscribeResult NewRegistrationResult methods that will return shells for provider specific data structure we can json unmarshal into. This will make the interface easier to understand and having NewOOO(). 

### Previous Discussion

The RegistrationResult struct is defined as below

```
// in common/types.go
type RegistrationResult struct {
	RegistrationRef string
	Result          any  // this is stored as []byte in database
	Status          RegistrationStatus
}

type Regu


// in salesfoce/subscribe.go
// below struct is stored Result any
type ResultData struct {
	EventChannel     *EventChannel
	NamedCredential  *NamedCredential
	EventRelayConfig *EventRelayConfig
}

// in otherProvider/subscribe.go
type ResultData strunct {
   SomeOtherProviderSpecificField1
   SomeOtherProviderSpecificField2
   ...
}

```

The data stored in the database corresponds to `Request any` field. But at the time of deserializing the Request field from []byte to any, we need to know each one of ResultData structure imported in order to deserialize. Which is a lot of importing and switch mapping that won't scale.

Instead, we have the connector to define deserialization(json unmarshal) for each provider, then the model importing and complications will be abstracted away. 

Open to other suggestions.

A previously explored solution was to make `Result any` field to `Result []byte`. Then server does not need to worry about serialization. But the connector will do serialization and deserialization. 

The other solution could be making it ajson.Node. But this has draw back that we need to mode serialization and deserialization like `[]byte -> ajson.Node -> []byte -> salesforce.ResultData` that loses performance.

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -510,6 +510,7 @@ type SubscribeConnector interface {
 		ctx context.Context,
 		previousResult RegistrationResult,
 	) error
+	GetRegistrationResultUnMarshalFunc() UnmarshalFunc
 
 	Subscribe(
 		ctx context.Context,
@@ -529,3 +530,9 @@ type SubscribeConnector interface {
 }
 
 type UnmarshalFunc func(data []byte) (any, error)
+
+func Unmarshal[T any](data []byte) (T, error) {
+	var v T
+	err := json.Unmarshal(data, &v)
+	return v, err
+}
--- a/common/types.go
+++ b/common/types.go
@@ -531,8 +531,10 @@ type SubscribeConnector interface {
 
 type UnmarshalFunc func(data []byte) (any, error)
 
+//nolint:ireturn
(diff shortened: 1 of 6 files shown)


eberle1080 (approved):

laurenzlong (approved):
As discussed offline - please rename these to `EmptyXYZ`

PR #1411, opened 2025-02-27
[CON-1527]feat(zendesk): Human-readable custom fields

Cobalt0s:
# Description
`Requests` and `Tickets` objects share the same `ticket_fields` as its custom fields.

The id is mapped to `ticket_fields` to figure out the human readable name. However, there is no good JSON type field name. Ex: "customer type", which contains a space because field name is "Customer Type".
image

Custom fields have interesting value options which are either stored under one of the following:
image

There is no good documentation on the types of custom fields. Here is the approach I took to map field types to Ampersand's `ValueType`:
image

# Test
## Metadata(Tickets)
image
image
### Before
image
### After
Customer Type is a custom field and appears only 25 times nested under `fields`. If it was under `raw` it would be 50. Therefore, raw is uneffected. This is achieved by using `common.MakeMarshaledDataFunc` which is used by connectors where `fields =/= raw`.
image
### Mapping
Only one API call to get custom fields is made and then is applied for each record, respecting performence.
image


Cobalt0s:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1411?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1411** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1411?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1411?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1410** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1410?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1321** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1321?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 5 other dependent PRs ([#1320](https://github.com/amp-labs/connectors/pull/1320) <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1320?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1324](https://github.com/amp-labs/connectors/pull/1324) <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1324?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1325](https://github.com/amp-labs/connectors/pull/1325) <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1325?utm_source=stack-comment-icon" target="_blank">Graphite</a> and 2 others)
* `main`


<h2></h2>

RajatPawar:
@greptileai

providers/zendesksupport/metadata.go
```diff
@@ -10,5 +10,31 @@ import (
...
 	ctx context.Context, objectNames []string,
 ) (*common.ListObjectMetadataResult, error) {
-	return metadata.Schemas.Select(c.Module.ID, objectNames)
+	metadataResult, err := metadata.Schemas.Select(c.Module.ID, objectNames)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, objectName := range objectNames {
+		fields, err := c.requestCustomFields(ctx, objectName)
```
RajatPawar, line 19: nit: might be good to spawn a goroutine per object for this (but I'd make sure that the API supports concurrency & has limits supporting this)

providers/zendesksupport/metadata.go
```diff
@@ -10,5 +10,31 @@ import (
...
+		if err != nil {
+			metadataResult.Errors[objectName] = err
+
+			continue
+		}
+
+		// Attach fields to the object metadata.
+		objectMetadata := metadataResult.Result[objectName]
+		for _, field := range fields {
+			objectMetadata.AddFieldMetadata(field.Title, common.FieldMetadata{
```
RajatPawar, line 29: can we refactor this function to return a new metadata object after mutating? Easy to run into issues with functions that mutate

providers/zendesksupport/parse.go
```diff
@@ -49,3 +51,11 @@ func getNextRecordsURL(node *ajson.Node) (string, error) {
 
 	return jsonquery.New(node).StrWithDefault("after_url", "")
 }
+
+func makeGetRecords(moduleID common.ModuleID, objectName string) common.NodeRecordsFunc {
```
RajatPawar, line 55: I understand the reasoning behind `make` as a prefix, but can we please just call this `getRecords` to make it easier to read?

providers/zendesksupport/read.go
```diff
@@ -89,3 +95,190 @@ func formatStartTime(config common.ReadParams) string {
 
 	return strconv.FormatInt(unixTime, 10)
 }
+
+// requestCustomFields makes and API call to get model describing custom fields.
+// For not applicable objects the empty mapping is returned.
+// The mapping is between "custom field id" and struct containing "human-readable field name".
+//
+// Custom fields are always associated with "ticket_fields" regardless of the object type.
```
RajatPawar, line 103: might make it easier to read if we separate this out into a `customFields.go` file

RajatPawar (commented):
@Cobalt0s does this also alter the raw response? can you please add examples of how it looks currently and how it would look after the change?

providers/zendesksupport/metadata.go
Cobalt0s: Good idea! Done!

providers/zendesksupport/metadata.go
Cobalt0s: Initial object is not empty, it is a deep copy of `metadata.Schemas`.
We could copy it again but each go routine will be responsible for one part of map, one object-metadata. For distributed processing this is an appropriate scenario, because map is logically partitioned and there is no overlap for side effects to collide.

common/types.go
```diff
@@ -333,6 +336,12 @@ type ObjectMetadata struct {
 	FieldsMap map[string]string
 }
 
+// AddFieldMetadata updates Fields and FieldsMap fields ensuring data consistency.
+func (m ObjectMetadata) AddFieldMetadata(fieldName string, fieldMetadata FieldMetadata) {
+	m.Fields[fieldName] = fieldMetadata
+	m.FieldsMap[fieldName] = fieldMetadata.DisplayName
+}
```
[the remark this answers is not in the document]
Cobalt0s: `greptile-apps` is incorrect in many cases. The result behaves as expected. Since the value is passed by value, the address field is copied into the Fields map. We can have multiple `ObjectMetadata` instances that are copies of each other, but since slices and maps reference shared underlying data, changes will still be reflected across instances.

The suggestion from `greptile-apps` only applies to immutable variables that reside directly in `ObjectMetadata`, which we do not modify.

Additionally, there's a convention that all struct receivers should be consistent — if one method uses a pointer receiver, then all should, to indicate that the struct is `modifiable`. This isn't perfect, though, since structs can still be effectively modifiable due to `internal references` like maps or slices.

To avoid confusion, I may switch to using a pointer receiver, though the change isn't functionally necessary.

Cobalt0s:
@RajatPawar 
You know, I noticed that since only 2 objects support custom fields and both of them would make the same API call I think we do not need the go routines, any parallel processing and better to make only one API call.

Objects: tickets, requests --> request custom fields can share an API call instead of:
image

RajatPawar (approved):
LGTM. Nice work!

PR #1427, opened 2025-03-03
[CON-1421] feat: Add Blueshift Read Connector

immdipu:
# Testing

image
image


internal/staticschema/core.go
```diff
@@ -292,6 +292,19 @@ func (m *Metadata[F, C]) LookupURLPath(moduleID common.ModuleID, objectName stri
...
+	moduleID = moduleIdentifier(moduleID)
+
+	path := m.Modules[moduleID].Objects[objectName].URLPath
+	if len(path) == 0 {
+		return "", common.ErrResolvingURLPathForObject
+	}
+
+	return path, nil
+}
+
```
immdipu, line 307: Added a new method to retrieve the object URL path. The existing methods either return the full URL or the module path.

providers/blueshift/handlers.go
```diff
@@ -0,0 +1,102 @@
...
+		url *urlbuilder.URL
+		err error
+	)
+
+	path, err := metadata.Schemas.LookupObjectURLPath(c.Module(), params.ObjectName)
+	if err != nil {
+		return nil, err
+	}
+
+	fullPath := fmt.Sprintf("%s%s", "v", path)
```
Cobalt0s, line 25: Once you address comments in ListObjectMetadata PR, this line should be removed.

providers/blueshift/handlers.go
```diff
@@ -0,0 +1,102 @@
...
+
+	baseURL, err := urlbuilder.New(c.ProviderInfo().BaseURL, fullPath)
+	if err != nil {
+		return nil, err
+	}
+
+	if nestedObjects.Has(params.ObjectName) {
+		body, ok := response.Body()
```
Cobalt0s, line 65: Could you move the content of the if block into a dedicated method. Maybe add a documentaion to `templates` object.
This is to keep exceptions seperate from the main happy path.

providers/blueshift/parse.go
```diff
@@ -0,0 +1,53 @@
...
+const (
+	pageSizeKey = "per_page"
+	pageSize    = "200"
+	pageKey     = "page"
+	pageNumber  = "0"
+)
+
+func getRecords(objectName string) common.RecordsFunc {
+	return func(node *ajson.Node) ([]map[string]any, error) {
+		responseKey := metadata.Schemas.LookupArrayFieldName(staticschema.RootModuleID, objectName)
```
Cobalt0s, line 22: It may not always be the root module. ModuleID should be an argument to `getRecords`.

internal/staticschema/core.go
Cobalt0s: Use this: https://github.com/amp-labs/connectors/blob/e0afab840c86ced1344998a18eab9607f85f259d/internal/staticschema/core.go#L282

Given that you address the comments in ListObjectMetadata the URLPath will be correct in `schema.json`.

providers/blueshift/connector.go
```diff
@@ -33,5 +36,21 @@ func constructor(base *components.Connector) (*Connector, error) {
...
+	}
+
+	connector.Reader = reader.NewHTTPReader(
+		connector.HTTPClient().Client,
+		registry,
+		connector.ProviderContext.Module(),
+		operations.ReadHandlers{
+			BuildRequest:  connector.buildReadRequest,
+			ParseResponse: connector.parseReadResponse,
+			ErrorHandler:  common.InterpretError,
```
Cobalt0s, line 51: Is there a custom error format that we could benefit from instead of common?
See: `interpreter.ErrorHandler{}.Handle` method can replace `common.InterpretError`.

test/blueshift/read/main.go
```diff
@@ -0,0 +1,58 @@
...
+		Fields:     datautils.NewStringSet("uuid", "name"),
+	})
+	if err != nil {
+		return err
+	}
+
+	utils.DumpJSON(res, os.Stdout)
+
+	slog.Info("Reading email templates")
+	res2, err := connector.Read(ctx, common.ReadParams{
```
Cobalt0s, line 36: Instead of `res2, err :=` you could just reasign new values `res, err =`. Just FYI.

providers/blueshift/parse.go
immdipu: will do it

providers/blueshift/connector.go
immdipu: i don' t think there is custom error format.

Cobalt0s (approved):

PR #1428, opened 2025-03-03
[CON-1422] feat: Add Blueshift Write Connector

immdipu:
# Testing 
image


providers/blueshift/handlers.go
```diff
@@ -100,3 +105,63 @@ func (c *Connector) parseReadResponse(
...
+	if writeObjectWithSuffix.Has(params.ObjectName) {
+		params.ObjectName = fmt.Sprintf("%s.json", params.ObjectName) //nolint:perfsprint
+	}
+
+	url, err = urlbuilder.New(c.ProviderInfo().BaseURL, writeVersion, params.ObjectName)
+	if err != nil {
+		return nil, err
+	}
+
+	jsonData, errr := json.Marshal(params.RecordData)
```
Cobalt0s, line 125: errr is roaring.

providers/blueshift/handlers.go
```diff
@@ -100,3 +105,63 @@ func (c *Connector) parseReadResponse(
...
+) (*common.WriteResult, error) {
+	node, ok := response.Body()
+	if !ok {
+		return &common.WriteResult{
+			Success: false,
+		}, nil
+	}
+
+	rawID, err := jsonquery.New(node).IntegerOptional("id")
+	if err != nil { //nolint:nilerr
```
Cobalt0s, line 147: The only err type that can be thrown here is if the expected "id" field is not an integer. I think returning write success is appropriate, linter also makes sense.

Maybe to combine the block below with this one the if could be:
`if err != nil || rawID == nil {`

Cobalt0s (approved):
Also need write_tests.go

PR #1429, opened 2025-03-03
[CON-1437] feat: Add Brevo Read Connector

immdipu:
# Testing 
image


providers/brevo/parse.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+func nextRecordsURL(previousURL *url.URL) common.NextPageFunc {
+	return func(node *ajson.Node) (string, error) {
+		totalCount, err := jsonquery.New(node).IntegerOptional("count")
+		if err != nil {
+			return "", err
+		}
+
+		if totalCount == nil || *totalCount == pageSize {
```
Cobalt0s, line 24: Could you link documentation for totalCount?
Some providers have unusual behaviour with these page related values.

Is total count a general count of that object? Is it just how many records were returned. If it is the first couple of pages then pageSize will always be the same as number of records returned. But there still can be a next page.

It would be nice to have tests for this function. Review the docs and create tests. Then implementation should conform to it.

common/json.go
```diff
@@ -39,6 +39,9 @@ type JSONHTTPResponse struct {
 	// that it's JSON-unmarshalled, it's identical to bodyBytes.
 	// If there were no bytes this will be nil.
 	body *ajson.Node
+
+	// Request is the original http.Request that generated this response
+	Request *http.Request
```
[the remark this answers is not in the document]
Cobalt0s: The change seems fine. Seperate PR would be nice.

I think we might have a workaround in future for this problem. But regardles, `JSONHTTPResponse` can still hold a reference to a request, its very logical.

Cobalt0s (commented):
Need tests/docs-reference for the pagination.

providers/brevo/parse.go
immdipu: It refers to the total number of records of that specific object, not the number of records in the response.

immdipu: https://developers.brevo.com/reference/getting-started-1

Cobalt0s: If `totalCount` is 56 and `pageSize` is 100 then the next page will be generated but shouldn't.

Cobalt0s: Oh, so you attempt to create an offset and then check if it is in range. That makes sense.
Then it is sufficient of an itself. Is there a value in checking `*totalCount == pageSize`? Any edge case that I am missing? The totalCount chould be declared near ` if nextOffset >= int(*totalCount) { `, because that is the main goal of its existence.

Cobalt0s (approved):

providers/brevo/parse.go
immdipu: Yeah, you’re right. We can remove the check if *totalCount == pageSize. The nextOffset >= int(*totalCount) handles both scenarios.

PR #1430, opened 2025-03-03
[CON-1438] feat: Add Brevo Write Connector

immdipu:
#Testing

image

### DIFF
--- a/providers/brevo/parse.go
+++ b/providers/brevo/parse.go
@@ -37,7 +37,6 @@ func nextRecordsURL(previousURL *url.URL) common.NextPageFunc {
 
 		// calcuate the next offset value
 		nextOffset := currentOffset + pageSize
-
 		// if nextOffset would exceed total count, we've fetched all records
 		if nextOffset >= int(*totalCount) {
 			return "", nil


providers/brevo/handlers.go
```diff
@@ -57,3 +61,89 @@ func (c *Connector) parseReadResponse(
...
+) (*common.WriteResult, error) {
+	node, ok := response.Body()
+	if !ok {
+		// Handle empty response
+		return &common.WriteResult{
+			Success: true,
+		}, nil
+	}
+
+	recordIDPaths := map[string]string{ //nolint:unconvert
```
Cobalt0s, line 104: This can be a DefaultMap where the fallback method would return "id".

providers/brevo/handlers.go
```diff
@@ -57,3 +61,89 @@ func (c *Connector) parseReadResponse(
...
+
+	if exists {
+		idPath = mappedPath
+	}
+
+	// Try string first
+	rawID, err := jsonquery.New(node).StrWithDefault(idPath, "")
+	if err != nil {
+		// Try integer
+		IntID, err := jsonquery.New(node).IntegerOptional(idPath)
```
Cobalt0s, line 133: Capitalized variable name is unconventional.

providers/brevo/support.go
```diff
@@ -37,13 +37,39 @@ var supportLimitAndOffset = datautils.NewSet( //nolint:gochecknoglobals
...
 	readSupport := metadata.Schemas.ObjectNames().GetList(staticschema.RootModuleID)
+	writeSupport := []string{
+		"smtp/email", "smtp/templates", "smtp/blockedDomains",
+		"smtp/deleteHardbounces", "transactionalSMS/sms", "whatsapp/sendMessage",
+		"emailCampaigns", "emailCampaigns/images", "smsCampaigns", "whatsappCampaigns",
+		"whatsappCampaigns/template", "contacts", "contacts/doubleOptinConfirmation", "contacts/folders",
+		"contacts/export", "contacts/import", "events", "senders", "senders/domains", "webhooks", "webhooks/export",
+		"corporate/subAccount", "corporate/ssoToken", "corporate/subAccount/ssoToken", "corporate/subAccount/key",
+
+		// Split these long lines
```
Cobalt0s, line 48: The comment is not needed, or doesn't make sense. It is about file formatting as I understand it.

Cobalt0s (commented):
I will come back to it once you add mock tests as discussed in Slack.

providers/brevo/support.go
immdipu: yeah  — to fix lint error. will remove the comment.

providers/brevo/handlers.go
```diff
@@ -57,3 +62,87 @@ func (c *Connector) parseReadResponse(
...
+	},
+		func(objectName string) string {
+			return "id"
+		},
+	)
+
+	idPath := recordIDPaths.Get(params.ObjectName)
+
+	// Try string first
+	rawID, err := jsonquery.New(node).StrWithDefault(idPath, "")
```
Cobalt0s, line 129: There is a method for this exact case. This makes it two.
https://github.com/amp-labs/connectors/blob/4fc8c331b0adf86b87d15a905e6d2448fe4f36e2/internal/jsonquery/defaults.go#L36

immdipu: Thank you 🙏. will update it.

PR #1437, opened 2025-03-04
[CON-1590]feat(Salesforce): Subscription Event

Cobalt0s:


Cobalt0s:
* **#1437** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1437?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1437?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

providers/salesforce/subscribe.go
```diff
@@ -13,11 +13,11 @@ type SubscribeResult struct {
 	EventChannelMembers map[common.ObjectName]*EventChannelMember
 }
 
-func (conn *Connector) EmptySubscritpionParams() *common.SubscribeParams {
```
Cobalt0s, line 16: IDE suggestion for identical reciever name

providers/salesforce/subscribe.go
```diff
@@ -48,11 +48,11 @@ func (conn *Connector) Subscribe(
 		return nil, fmt.Errorf("invalid registration result: %w", err)
 	}
 
-	regstrationParams, ok := params.RegistrationResult.Result.(*ResultData)
+	registrationParams, ok := params.RegistrationResult.Result.(*ResultData)
```
Cobalt0s, line 51: IDE spelling.

providers/salesforce/test/subscription/new_account.json
```diff
@@ -0,0 +1,60 @@
+{
+  "ChangeEventHeader": {
```
Cobalt0s, line 2: From the documention (end of webpage): https://developer.salesforce.com/docs/atlas.en-us.change_data_capture.meta/change_data_capture/cdc_message_structure.htm

providers/salesforce/subscriptionEvent.go
```diff
@@ -0,0 +1,126 @@
...
+// A single event may contain multiple record identifiers and can be expanded into multiple SubscriptionEvent instances.
+//
+// Structure reference:
+// https://developer.salesforce.com/docs/atlas.en-us.change_data_capture.meta/change_data_capture/cdc_event_fields_header.htm.
+type ChangeEvent struct {
+	ChangeEventHeader ChangeEventHeader `json:"ChangeEventHeader"`
+	// ...
+	// Additional fields represent changes to the record and vary based on the record's model type.
+	// They are situated on the same level as ChangeEventHeader.
+	// ...
```
Cobalt0s, line 24: @jlimatampersand, are the fields related to the model necessary?

The interface does not include a method to return a map representing the record.

providers/salesforce/subscriptionEvent.go
```diff
@@ -0,0 +1,126 @@
...
+	CommitTimestamp int64    `json:"commitTimestamp"`
+	CommitNumber    int64    `json:"commitNumber"`
+	CommitUser      string   `json:"commitUser"`
+	NulledFields    []any    `json:"nulledFields"`
+	DiffFields      []any    `json:"diffFields"`
+	ChangedFields   []any    `json:"changedFields"`
+}
+
+// Unwrap splits bundled event into per record event.
+func (e ChangeEvent) Unwrap() []SubscriptionEvent {
```
Cobalt0s, line 47: @jlimatampersand, Salesforce ChangeEvent may include many record identifiers if they have similiar change and happend in 1 second window.
Given that webhook data is marshalled into `ChangeEvent`, I created `Unwrap` method to spread event into multiple `SubscriptionEvent`. Let me know what do you think about this approach of constructing these structs, what do you think about naming?

providers/salesforce/subscriptionEvent.go
```diff
@@ -0,0 +1,126 @@
...
+	return s.EntityName, nil
+}
+
+func (s SubscriptionEvent) Workspace() (string, error) {
+	// Not applicable
+	return "", nil
+}
+
+func (s SubscriptionEvent) RecordId() (string, error) {
+	return s.RecordIdentifier, nil
```
Cobalt0s, line 119: @jlimatampersand, record identifier may have a wildcard. I linked to the docs here and at the top of the struct.
Just heads up, so that there will be no surprises when using this method from the server.

providers/salesforce/subscriptionEvent.go
```diff
@@ -0,0 +1,126 @@
...
+package salesforce
+
+import (
+	"context"
+
+	"github.com/amp-labs/connectors/common"
+)
+
+func (*Connector) VerifyWebhookMessage(context.Context, *common.WebhookVerificationParameters) (bool, error) {
+	return true, nil
```
Cobalt0s, line 11: @jlimatampersand , my guess is that this method is not applicable for salesforce.

providers/salesforce/subscriptionEvent.go
```diff
@@ -0,0 +1,128 @@
...
+	SequenceNumber   int
+	CommitTimestamp  int64
+	CommitNumber     int64
+	CommitUser       string
+	NulledFields     []any
+	DiffFields       []any
+	ChangedFields    []any
+}
+
+func (s SubscriptionEvent) EventType() (common.SubscriptionEventType, error) {
```
jlimatampersand, line 88: Can we have some dock linked about the change types?

providers/salesforce/subscriptionEvent.go
```diff
@@ -0,0 +1,128 @@
...
+	// Additional fields represent changes to the record and vary based on the record's model type.
+	// They are situated on the same level as ChangeEventHeader.
+	// ...
+}
+
+// ChangeEventHeader holds event data.
+//
+// Record ID can have wildcard symbols:
+// https://developer.salesforce.com/docs/atlas.en-us.change_data_capture.meta/change_data_capture/cdc_field_conversion_single_event.htm.
+type ChangeEventHeader struct {
```
jlimatampersand, line 31: type SubscriptionEvent map[string]any

providers/salesforce/subscriptionEvent.go
```diff
@@ -0,0 +1,128 @@
...
+func (*Connector) VerifyWebhookMessage(context.Context, *common.WebhookVerificationParameters) (bool, error) {
+	return true, nil
+}
+
+// ChangeEvent represents data received from a subscription.
+// A single event may contain multiple record identifiers and can be expanded into multiple SubscriptionEvent instances.
+//
+// Structure reference:
+// https://developer.salesforce.com/docs/atlas.en-us.change_data_capture.meta/change_data_capture/cdc_event_fields_header.htm.
+type ChangeEvent struct {
```
jlimatampersand, line 19: We should change this to string map how we do in hubspot. Please refer to this https://github.com/amp-labs/connectors/blob/aee70e84fdd1eb3984016e8755a860971aaa8bca/providers/hubspot/subscriptionEvent.go#L33

```
type SubscriptionEvent map[string]any
```

The reason is that we don't know how the webhook message structure will change. They may not remove unused fields, but they may add new fields that will be dropped if we use a strictly typed struct. 

Also, a strongly typed struct won't be able to capture custom fields either. 

So in that regard we have decided to use string map. Sorry I missed this point in the task ticket.

jlimatampersand (changes requested):
Please take refer to the comment about the message type. Sorry again I missed the concern about the type in the ticket. Thank you for working on this!

providers/salesforce/subscriptionEvent.go
jlimatampersand: I think `ToRecordList`, sounds more clear here, but I like the idea of this unwrapping function.

providers/salesforce/subscriptionEvent.go
jlimatampersand: good call

jlimatampersand (approved):

PR #1441, opened 2025-03-05
[CON-1576] feat: Add Github Read Connector

immdipu:
# Testing

image
image


providers/github/handlers.go
```diff
@@ -0,0 +1,89 @@
...
+		makeNextRecordsURL(response.Headers),
+		common.GetMarshaledData,
+		params.Fields,
+	)
+}
+
+func makeNextRecordsURL(responseHeaders http.Header) common.NextPageFunc {
+	return func(node *ajson.Node) (string, error) {
+		// GitHub uses Link header for pagination
+		links := responseHeaders["Link"]
```
Cobalt0s, line 69: Just in time :smile: I built a utility for this last week. I am telling you simliar connectors come in batches ahahah.
Capsule also uses `Link` header. I delegate the parsing to the library.

Refere to: 
https://github.com/amp-labs/connectors/blob/0bf2127c698a143b003182424f402b72c4cabc71/internal/httpkit/response.go#L9-L15

This is the first method opening up the `httpkit` package :laughing:. HTTP utils may go there.

providers/github/supports.go
```diff
@@ -0,0 +1,43 @@
...
+		"gists/starred", "gpg_keys", "installation-requests", "installation/repositories",
+		"issues", "keys", "licenses", "marketplace_listing/plans",
+		"marketplace_listing/stubbed/plans", "marketplace_purchases", "migrations",
+		"notifications", "orgs", "packages", "gists/public", "public_emails", "repos",
+		"repository_invitations", "secrets", "social_accounts", "ssh_signing_keys",
+		"stubbed", "subscriptions", "teams", "user/installations", "user/issues",
+		"user/memberships/orgs", "user/starred",
+	)
+
+	supportSince = datautils.NewSet( //nolint: gochecknoglobals
```
Cobalt0s, line 26: In the README, you mention timestamp-based vs. ID-based since querying. Which category does this one fall into?

The ID-as-since parameter approach doesn’t work for us. Since is always a time, and we never pass IDs. The fact that IDs are ordered by time is just a provider-specific perk.

Cobalt0s (approved):
The build is broken due to `ireturn`. This happens with the `golangci-lint` tool. The issue comes and goes. You need to trigger a build again by changing the code in any way, commit and push. Just doing it from the github dashboard won't do it.

Honestly, this is the linter check that I find not helpful @RajatPawar do you think we could disable it by adding `ireturn` to `.golangci.yml`? Returning generics or using factory pattern to return interface is the only usages we have and it is appropriate. However, for other cases the check does make sense.

providers/github/supports.go
immdipu: It falls into timestamp-based.

providers/github/handlers.go
immdipu: Awesome! I’ll update the code with the utility function.

PR #1444, opened 2025-03-05
[CON-1591]feat(Salesforce): GetRecordsWithIds

Cobalt0s:
# Changes

* Extended existing SOQL builder to accept list of identifiers.
  * Added mock test to ensure query is built as expected.
* Added test script to query Accounts (see below).

# Tests
`salesforce/record/main.go`:
image
image

# Notes
Associations argument at `GetRecordsWithIds` is preserved to **match method signature** of other deep connectors. However, it is not used by the Salesforce implementation.

### DIFF
--- a/providers/salesforce/soql.go
+++ b/providers/salesforce/soql.go
@@ -3,6 +3,16 @@ package salesforce
 import (
 	"fmt"
 	"strings"
+
+	"github.com/amp-labs/connectors/common"
+)
+
+// nolint:lll
+const (
+	// See `Limiting Result Rows` section on this webpage:
+	// https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql_select_fields.htm
+	identifiersLimit    = 200
+	identifiersLimitStr = "200"
 )
 
 // soqlBuilder builder of Salesforce Object Query Language.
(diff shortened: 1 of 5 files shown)


Cobalt0s:
* **#1444** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1444?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1444?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

providers/salesforce/soql.go
```diff
@@ -47,6 +57,25 @@ func (s *soqlBuilder) Where(condition string) *soqlBuilder {
 	return s
 }
 
+func (s *soqlBuilder) WithIDs(identifiers []string) error {
+	if len(identifiers) > identifiersLimit {
```
jlimatampersand, line 61: is this the hard limit from Salesforce?  200?

jlimatampersand (approved):

providers/salesforce/soql.go
Cobalt0s: Looks like the limit is not hard for the WHERE close. The LIMIT must be 200 if `all fields` are requested (it is only a particular scenario).
I have quickly tested quering WHERE with more than 200 ids and LIMIT with over 200.
image
image

I will remove the hard limit.

Cobalt0s: We could build a more sophisticated `SOQLBuilder` struct that if wildcard `*` is used to request all fields then ensure that number of record identifiers cannot exceed 200. Only in that case the error would make sense.

Unlikely that we need it. I doubt we use the `fields=[]string{"*"}`.

PR #1448, opened 2025-03-06
[CON-1590] `collapsedSubscriptionEvent` interface for Salesforce webhook event message

jlimatampersand:
## This PR does the following
- Defines new interface type `CollapsedSubscriptionEvent` which is a type of webhook message that has multiple record changes defined in 1 message. This type separates records into multiple list of `SubscriptionEvent`s, which is existing interface
- This PR also moves around some Connector interface types from common tothe  connectors package since that is where all other connectors are defined
- Lastly, as `GetMarshalledDataWithId` to common package which is needed for salesforce webhook message parsing. 

## Note
This pr should be merged in order for related PR in server to merge. https://github.com/amp-labs/server/pull/2755

## Test Results

Screenshot 2025-03-06 at 3 22 57 PM


common/types.go
```diff
@@ -486,55 +490,3 @@ const (
 	// registration returned error, and failed to rollback some intermittent steps.
 	SubscriptionStatusFailedToRollback SubscriptionStatus = "failed_to_rollback"
 )
-
```
jlimatampersand, line 489: moved to connectors package since that is where all other connector interfaces are defined

connectors.go
```diff
@@ -88,11 +88,74 @@ type AuthMetadataConnector interface {
...
+		recordIds []string,
+		fields []string,
+		associations []string) ([]common.ReadResultRow, error)
 
 	// VerifyWebhookMessage verifies the signature of a webhook message.
 	VerifyWebhookMessage(ctx context.Context, params *common.WebhookVerificationParameters) (bool, error)
 }
 
+//nolint:interfacebloat
+type SubscribeConnector interface {
```
jlimatampersand, line 104: just moved from another file

providers/salesforce/subscriptionEvent.go
```diff
@@ -20,20 +20,61 @@ var (
...
 //
 // Structure reference:
 // https://developer.salesforce.com/docs/atlas.en-us.change_data_capture.meta/change_data_capture/cdc_event_fields_header.htm.
-type ChangeEvent map[string]any
+type CollapsedSubscriptionEvent map[string]any
 
 // ToRecordList splits bundled event into per record event.
 // Every property is duplicated across SubscriptionEvent. RecordIds is spread as RecordId.
-func (e ChangeEvent) ToRecordList() ([]SubscriptionEvent, error) {
+func (e CollapsedSubscriptionEvent) SubscriptionEventList() ([]common.SubscriptionEvent, error) {
```
jlimatampersand, line 77: the `[]SubscriptionEvent` looks similar to new `[]common.SubscriptionEvent` 

but `[]SubscriptionEvent` is referring to Salesforce Specific struct where as `[]common.SubscriptionEvent` is referring to an interface defined in common package.

eberle1080 (approved):

PR #1450, opened 2025-03-06
[CON-1592] fix(outreach): panic when there is no response

Jkarage:
Tests:
Screenshot 2025-03-06 at 17 55 39

### DIFF
--- a/providers/outreach/metadata.go
+++ b/providers/outreach/metadata.go
@@ -2,6 +2,7 @@ package outreach
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/amp-labs/connectors/common"
 )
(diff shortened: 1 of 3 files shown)


providers/outreach/metadata.go
```diff
@@ -65,6 +65,10 @@ func metadataMapper(resp *common.JSONHTTPResponse) (*common.ObjectMetadata, erro
 		FieldsMap: make(map[string]string),
 	}
 
+	if len(response.Data) == 0 {
+		return nil, common.ErrMissingExpectedValues
```
RajatPawar, line 69:
```suggestion
		return nil, fmt.Errorf("%w: could not find a record to sample fields from", common.ErrMissingExpectedValues)
```

PR #1451, opened 2025-03-06
[CON-1577] feat: Add GitHub Write + Delete Connector

immdipu:
# Testing 

## Write 
image
image

## Delete
image


providers/github/handlers.go
```diff
@@ -87,3 +91,95 @@ func makeNextRecordsURL(responseHeaders http.Header) common.NextPageFunc {
 		return nextLink[start+1 : end], nil
 	}
 }
+
+func (c *Connector) buildWriteRequest(ctx context.Context, params common.WriteParams) (*http.Request, error) { //nolint:lll
+	method := http.MethodPost
```
Cobalt0s, line 96: Variable declaration should be as close to the operations that modify and use it. I was reading the `MethodPatch` line and then thought where is the `method` variable itself, for a second I thought that it could be null.

providers/github/handlers.go
```diff
@@ -87,3 +91,95 @@ func makeNextRecordsURL(responseHeaders http.Header) common.NextPageFunc {
...
+		return &common.WriteResult{
+			Success: true,
+			Errors:  nil,
+			Data:    data,
+		}, nil
+	}
+
+	return &common.WriteResult{
+		Success:  true,
+		RecordId: *recordID,
```
Cobalt0s, line 154: You could have a null pointer dereference. You have used `StringOptional`. To be safe use `StrWithDefault`.

providers/github/handlers.go
```diff
@@ -87,3 +91,95 @@ func makeNextRecordsURL(responseHeaders http.Header) common.NextPageFunc {
...
+		Errors:   nil,
+		Data:     data,
+	}, nil
+}
+
+func (c *Connector) buildDeleteRequest(ctx context.Context, params common.DeleteParams) (*http.Request, error) {
+	url, err := urlbuilder.New(c.ProviderInfo().BaseURL, params.ObjectName)
```
Cobalt0s, line 161: `params.RecordId` could be added as one of the variadic arguments.
```
 urlbuilder.New(c.ProviderInfo().BaseURL, params.ObjectName, params.RecordId)
```

test/github/write/main.go
```diff
@@ -0,0 +1,90 @@
...
+	conn := hs.GetGithubConnector(ctx)
+
+	slog.Info("> TEST Add User Emails")
+
+	if err := addUserEmails(ctx, conn); err != nil {
+		slog.Error(err.Error())
+	}
+
+	slog.Info("> TEST Create gists")
+	if err := createGist(ctx, conn); err != nil {
```
Cobalt0s, line 34: Could you add the update test case?

Cobalt0s (changes requested):
There is a potential `nil pointer dereference`. See comments.

providers/github/handlers.go
immdipu: Good catch! I'll move the variable declaration.

providers/github/handlers.go
immdipu: Thanks! i'll update it. 🙂

providers/github/handlers.go
```diff
@@ -65,11 +69,115 @@ func (c *Connector) parseReadResponse(
 
 func makeNextRecordsURL(responseHeaders http.Header) common.NextPageFunc {
 	return func(node *ajson.Node) (string, error) {
-		nextURL := httpkit.HeaderLink(&common.JSONHTTPResponse{Headers: responseHeaders}, "next")
```
Cobalt0s, line 68: I think you had issue with git rebasing/merging. This shouldn't be removed. I think git conflicts slipped through.

Cobalt0s (approved):
There are git issues! 

The file had correct changes coming from `Github Read PR`. Please fix `handlers.go`. I will approve this and you can merge it after you address the issue.

PR #1454, opened 2025-03-10
[CON-1096] feat: Add Dixa Write Connector

Jkarage:
Tests:
Screenshot 2025-03-10 at 11 39 16


providers/dixa/handlers.go
```diff
@@ -42,3 +45,54 @@ func (c *Connector) parseReadResponse(
...
+	response *common.JSONHTTPResponse,
+) (*common.WriteResult, error) {
+	body, ok := response.Body()
+	if !ok {
+		return &common.WriteResult{
+			Success: true,
+		}, nil
+	}
+
+	resp, err := jsonquery.New(body).ObjectRequired("") // reading the current node.
```
RajatPawar, line 84: Can we just skip `.ObjectRequired("")`?

RajatPawar (approved):

RajatPawar (commented):
I see that `WriteResult` contains data under a `data` key, is this something that is consistent across the API? If it is, we could remove one level to avoid repetition, but if it is inconsistently structured, feel free to merge as is

Jkarage:
Yeah, it's consistent for all endpoints except the bulky ones, which I have commented currently. So I have updated the write to check the data field.

providers/dixa/handlers.go
Jkarage: This is now checking the `data` field.

PR #1456, opened 2025-03-10
[CON-1445] feat: Add Clari Read Connector

Jkarage:
Tests:
Screenshot 2025-03-10 at 15 37 28

### DIFF
--- a/providers/clari/supports.go
+++ b/providers/clari/supports.go
@@ -1,5 +1,13 @@
 package clari
 
+import (
+	"fmt"
+	"strings"
+
+	"github.com/amp-labs/connectors/internal/components"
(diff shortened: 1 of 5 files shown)


providers/clari/supports.go
```diff
@@ -10,3 +18,18 @@ func responseField(objectName string) string {
 		return ""
 	}
 }
+
+func supportedOperations() components.EndpointRegistryInput {
+	readSupport := []string{
+		"export/jobs", "audit/events", "admin/limits",
```
jlimatampersand, line 24: I looked deeper into their API doc and found that the correct objects we want to read from or write to are `opportunity`

Jkarage:
Closing this, Blocked

PR #1457, opened 2025-03-10
[CON-1446] feat: Add Clari Write Connector

Jkarage:
Tests:
Screenshot 2025-03-10 at 16 16 47


jlimatampersand:
The correct object to write or read is opportunity and the opportunity object does not support write operation other than bulk write. I left a comment in the task and tagged Rajat and Lauren, so let's wait till we hear from them.

Jkarage:
Closing this, Blocked

PR #1459, opened 2025-03-10
[CON-958] feat(monday): add `ListObjectMetadata`

jatinsandilya:
Screenshot 2025-04-03 at 2 20 13 PM

(updated after using introspection endpoint)


jatinsandilya:
@Cobalt0s sorry about that forgot to do the cleanup before asking for a review, resolved all your comments

providers/monday/metadata/schemas.json
```diff
@@ -0,0 +1,3 @@
+{
```
Cobalt0s, line 1: This file is not used, it should be removed.

providers/monday/handlers.go
```diff
@@ -0,0 +1,121 @@
...
+		"query": query,
+	}
+
+	// Marshal the request body to JSON
+	jsonBody, err := json.Marshal(requestBody)
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url.String(), bytes.NewReader(jsonBody))
```
Cobalt0s, line 70: This approach queries `boards/users` directly for reading records, which later **sample** the fields.

In previous review I pointed out to [Graphql introspection.
](https://graphql.org/learn/introspection/)
It is directly related to metadata. Does Monday offer this? This approach is preferred to `read sampling`, which could return no records, therefore no metadata to acquire.

Ideally API gives us the schema which we parse, for example:
```
{
  "data": {
    "__type": {
      "name": "Board",
      "fields": [
        {
          "name": "id",
          "type": {
            "name": null,
            "kind": "NON_NULL",
            "ofType": {
              "name": "ID",
              "kind": "SCALAR"
            }
          }
        }
        ....
        .... other fields
        ....
      ]
    }
  }
}
```
Please, let me know if this approach is achievable for Monday. If not current code will be good to go.

Cobalt0s (commented):
Thank you for splitting! Looks good, but I wonder about the API which can offer object schema directly, see comments.

providers/monday/handlers.go
Cobalt0s: Also found this: https://api.monday.com/v2/get_schema. But it includes all objects.

jatinsandilya: @Cobalt0s done

providers/monday/handlers.go
```diff
@@ -0,0 +1,133 @@
...
+	objectName string,
+	request *http.Request,
+	response *common.JSONHTTPResponse,
+) (*common.ObjectMetadata, error) {
+	objectMetadata := common.ObjectMetadata{
+		FieldsMap:   make(map[string]string),
+		DisplayName: naming.CapitalizeFirstLetterEveryWord(objectName),
+	}
+
+	data, err := common.UnmarshalJSON[map[string]any](response)
```
Cobalt0s, line 89: Instead of a general "JSON" (map of objects) you could declare a concrete **struct** and parse it.
You know the response format of the introspection beforehand.

This would remove `.(map[string]any)` or `.([]any)` or `.(string)` conversions, its presence implies that you secretly know the resoponse **structure** but thats not aparrent from the code.

On the other hand previous read sampling must use `map[string]any` which is format agnostic, making it the only way to collect "struct properties", justifying usage.

Cobalt0s (approved):
Awesome! :rocket:

providers/monday/handlers.go
RajatPawar: +1, a struct would make it very obvious to understand the expected response

providers/monday/connector.go
```diff
@@ -0,0 +1,53 @@
...
+type Connector struct {
+	// Basic connector
+	*components.Connector
+
+	// Require authenticated client
+	common.RequireAuthenticatedClient
+
+	// Supported operations
+	components.SchemaProvider
+	components.Reader
```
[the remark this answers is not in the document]
Cobalt0s: Ok, then it should be considered as a strict no, unless implemented.

providers/monday/handlers.go
```diff
@@ -0,0 +1,133 @@
...
+)
+
+func (c *Connector) buildSingleObjectMetadataRequest(ctx context.Context, objectName string) (*http.Request, error) {
+	url, err := urlbuilder.New(c.ProviderInfo().BaseURL, apiVersion)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build URL: %w", err)
+	}
+
+	// Map object names to their GraphQL type names
+	typeNameMap := map[string]string{
```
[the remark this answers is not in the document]
Cobalt0s: Naming package delegates. This utility is used in some places.
```
naming.NewSingularString(text).String()
```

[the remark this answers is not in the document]
jatinsandilya: @Cobalt0s thanks for this! Using this now

PR #1460, opened 2025-03-10
[CON-1594] feat : Add Ashby ListMetadata

immdipu:
# Testing
image


providers/ashby/connector.go
```diff
@@ -0,0 +1,37 @@
+package ashby
+
+import (
+	_ "embed"
```
eberle1080, line 4: I don't believe you need this line in this file

eberle1080 (approved):

PR #1461, opened 2025-03-10
[CON-1595] feat: add Ashby Read Connector

immdipu:
# Testing

image
image


providers/ashby/handlers.go
```diff
@@ -0,0 +1,70 @@
...
+func buildRequestbody(params common.ReadParams) map[string]any {
+	body := make(map[string]any)
+
+	if supportSince.Has(params.ObjectName) && !params.Since.IsZero() {
+		body[sinceKey] = params.Since.UnixMilli()
+	}
+
+	if supportPagination.Has(params.ObjectName) && params.NextPage != "" {
+		body[pageKey] = params.NextPage
+		body[pageSizeKey] = pageSize
```
Cobalt0s, line 51: I think constant variables introduce unnecessary indirection since they are not used outside this method. PageSize is more of a general connector concept at makes sense to be a dedicated const.

providers/ashby/parse.go
```diff
@@ -0,0 +1,35 @@
+package ashby
+
+import (
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/internal/jsonquery"
+	"github.com/amp-labs/connectors/providers/ashby/metadata"
+	"github.com/spyzhov/ajson"
+)
+
+func getRecords(objectName string, moduleID common.ModuleID) common.RecordsFunc {
```
Cobalt0s, line 10: This method can be replaced with: `common.GetRecordsUnderJSONPath`.

Cobalt0s (approved):

providers/ashby/handlers.go
immdipu: Got it, I’ll adjust. Thanks!

PR #1462, opened 2025-03-10
[CON-1596] feat: add Ashby Write Connector

immdipu:
# Testing

image
image


providers/ashby/handlers.go
```diff
@@ -74,3 +70,89 @@ func (c *Connector) parseReadResponse(
...
+		Success:  true,
+		RecordId: *rawID,
+		Data:     data,
+	}, nil
+}
+
+func handleErrorResponse(node *ajson.Node) (*common.WriteResult, error) {
+	errors, err := jsonquery.New(node).ArrayOptional("errors")
+	if err != nil {
+		return &common.WriteResult{Success: false}, nil //nolint:nilerr
```
Cobalt0s, line 145: There are just few connectors that return `Success: false`. This is nice to see!

providers/ashby/handlers.go
```diff
@@ -74,3 +70,89 @@ func (c *Connector) parseReadResponse(
...
+	params common.WriteParams,
+	request *http.Request,
+	response *common.JSONHTTPResponse,
+) (*common.WriteResult, error) {
+	node, ok := response.Body()
+	if !ok {
+		return &common.WriteResult{Success: true}, nil
+	}
+
+	success, err := jsonquery.New(node).BoolRequired("success")
```
Cobalt0s, line 105: Does it always return this field?
Would it be safe to assume that success can default to true unless this field is present with a `false` value?

Cobalt0s (approved):

providers/ashby/handlers.go
immdipu: It always returns a status of 200, even in case of an error, so it makes sense to include a `success` flag.

PR #1474, opened 2025-03-12
[CON-1567] feat: Add Clickup ListMetadata

immdipu:
# Testing
image


providers/clickup/metadata/scraping.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/amp-labs/connectors/internal/staticschema"
+	"github.com/amp-labs/connectors/tools/fileconv"
+	"github.com/amp-labs/connectors/tools/scrapper"
+)
+
+// nolint:gochecknoglobals
+var (
+	//go:embed schemas.json
+	schemaContent []byte
+	FileManager   = scrapper.NewMetadataFileManager[staticschema.FieldMetadataMapV1](
```
Cobalt0s, line 15: Use FieldMetadataMapV2 instead.

providers/clickup/metadata/schemas.json
```diff
@@ -0,0 +1,22 @@
+{
+  "modules": {
+    "root": {
+      "id": "",
+      "path": "",
+      "objects": {
+        "team": {
```
Cobalt0s, line 7: Are you sure there is only one object. The OpenAPI file you sent me is very large.

providers/clickup/metadata/scraping.go
immdipu: there is only one object and schema is generated manually so thought to just use V1.

Cobalt0s: Makes sense. I think it would be better to fully migrate to version 2 even though in your situation V1 would carry the same information but less verbose. This would create less technical debt for the future.

Cobalt0s (approved):

PR #1475, opened 2025-03-12
[CON-1568] feat: Add ClickUp Read Connector

immdipu:
# Testing
image


providers/clickup/parse.go
```diff
@@ -0,0 +1,44 @@
...
+			currentPage, err = strconv.Atoi(currentPageStr)
+			if err != nil {
+				return "", nil //nolint:nilerr
+			}
+		}
+
+		nextPage := currentPage + 1
+
+		query.Set("page", strconv.Itoa(nextPage))
+		nextURL.RawQuery = query.Encode()
```
Cobalt0s, line 40: I will be adding some utility methods that would help in using urlbuilder instead of url.URL where query parameter operations are simpler

Cobalt0s (approved):

PR #1477, opened 2025-03-12
feat(urlbuilder): Construct from url.URL

Cobalt0s:
# Background

With the newly refactored deep connectors, we often work directly with raw `url.URL`. This bypasses the query parameter utilities provided by `urlbuilder`, making simple query parameter modifications cumbersome.

# Change
`urlbuilder.URL` can now be created from a raw `url.URL`.

### DIFF
--- a/common/urlbuilder/url_test.go
+++ b/common/urlbuilder/url_test.go
@@ -2,9 +2,12 @@ package urlbuilder
 
 import (
 	"errors"
+	"net/url"
 	"reflect"
 	"strings"
 	"testing"
+
+	"github.com/amp-labs/connectors/test/utils/testutils"
 )
 
 func TestNewURL(t *testing.T) { // nolint:funlen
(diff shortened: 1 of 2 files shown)


Cobalt0s:
* **#1477** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1477?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1477?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

common/urlbuilder/url_test.go
```diff
@@ -189,3 +192,45 @@ func TestAddPath(t *testing.T) { // nolint:funlen
...
+			name:     "Trailing slash is preserved",
+			input:    "http://video.google.co.uk:80/",
+			expected: "http://video.google.co.uk:80/",
+		},
+		{
+			name:     "URL with query parameters",
+			input:    "https://video.google.co.uk:80/videoplay?docid=-7246927612831078230&hl=en",
+			expected: "https://video.google.co.uk:80/videoplay?docid=-7246927612831078230&hl=en",
+		},
+	}
```
RajatPawar, line 219: can you please add a test case for urls with encoded values like `%20%` and non-http protocols / urls without a protocol?

RajatPawar (approved):

common/urlbuilder/url_test.go
Cobalt0s: Ok, so what I have found out is that golang url pacakge encodes spaces in query params as `+`. This means `info=Hello%20World` -> `info=Hello+World`. This is not an error and permitted for query params. Let me know if we need to infer econding exceptions and preserve them when dealing with query params.

EncodingExceptions are part of `urlbuilder.URL`:
image

@RajatPawar also question about the package name `urlbuilder.URL` is a wrapper of `url.URL`. What do you think of package being called `urlx.URL` then constructor `urlx.New` vs `urlbuilder.New` where later kind of suggests that we created a builder.

RajatPawar: 1. I'll defer to you as long as the behavior is consistent across both packages! 
2. As for the naming, `urlbuilder` is very clear to me whereas `urlx` doesn't communicate what it does, but sure, feel free to take a call on it

PR #1479, opened 2025-03-13
chore(blueshift): update ListMetadata schema and write support

immdipu:
1.  Fix `onsite_slots` response key mapping in Blueshift 

2. Remove unsupported `campaigns` write object  in Blueshift

### DIFF
--- a/providers/blueshift/metadata/openapi/api.yaml
+++ b/providers/blueshift/metadata/openapi/api.yaml
@@ -9245,6 +9245,9 @@
             "content": {
               "application/json": {
                 "schema": {
+                   "type": "object",
+                    "properties": {
+                      "results": {
                   "type": "array",
                   "description": "The list of live content slots.",
                   "items": {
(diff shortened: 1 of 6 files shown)


providers/blueshift/metadata/openapi/api.yaml
```diff
@@ -9245,6 +9245,9 @@
             "content": {
               "application/json": {
                 "schema": {
+                   "type": "object",
```
Cobalt0s, line 9248: Did you add this manually or uploaded a new OpenAPI version?

Cobalt0s (approved):

providers/blueshift/metadata/openapi/api.yaml
immdipu: the OpenAPI doesn't have this. there are a lot mistakes in OpenAPI so i added this manually.

Cobalt0s: It would be interesting to try and feed this file into ChatGPT to fix any issues. And then see if you can extract data from it and of course check if it matches the docs.

PR #1482, opened 2025-03-14
[CON-1610] feat(attio): added support to standard & custom object

sanjay-a01:
### Notes
[https://github.com/amp-labs/connectors/pull/1046](url)
**Added** - standard and custom objects 
**Removed** - objects and webhooks module

### Functionalities
Handle both Attio APIs and standard/custom objects.
image

Fetch the metadata from the api_slug for standard/custom objects.
image

### Metadata result
image


providers/attio/metadata.go
```diff
@@ -24,14 +31,24 @@ func (c *Connector) ListObjectMetadata(ctx context.Context,
...
+	)
+
 	for _, obj := range objectNames {
+		isAttioStandardOrCustomObj = false
+
 		// Constructing the request url.
-		url, err := c.getApiURL(obj)
-		if err != nil {
-			return nil, err
+		if supportAttioApiObj.Has(obj) {
```
Cobalt0s, line 43: The for loop is quite large and consists of three parts:
1. Determining the URL and object category.
2. Making the API call and deciding how to extract metadata (via read sampling or attributes endpoint).
3. Fetching the object display name (custom objects get a fetched name, others default to the object name).

### Refactoring Proposal:
The first and second sections could be moved into a dedicated method, `getObjectAttributes`, and merged with `parseMetadataFromResponse`:
* If `isAttioStandardOrCustomObj`, select the read URL and parse using response sampling.
* Otherwise, select the attribute URL and parse/extract `api_slug`.

Keeping URL selection close to response parsing improves readability. The linter flagged `funlen` here, so it is worth addressing.

providers/attio/metadata_test.go
```diff
@@ -127,6 +108,48 @@ func TestListObjectMetadata(t *testing.T) { // nolint:funlen,gocognit,cyclop
 							"linked_records":    "linked_records",
 						},
 					},
+					"companies": {
+						DisplayName: "Companies",
```
Cobalt0s, line 112: This is a great test!

Cobalt0s (commented):
Thank you for adding a mock test! I am totally sold that it works :wink: 

I have added several suggestions. Please ping me for a review again once you respond, resolve the comments, and fix the linter build.

providers/attio/read.go
```diff
@@ -13,7 +13,7 @@ func (c *Connector) Read(ctx context.Context, config common.ReadParams) (*common
 		return nil, err
 	}
 
-	if !supportedObjectsByRead.Has(config.ObjectName) {
+	if !supportAttioApiObj.Has(config.ObjectName) {
```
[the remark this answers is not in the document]
sanjay-a01: I'll handle this when I work in read functionality and with a separate PR

providers/attio/url.go
```diff
@@ -0,0 +1,19 @@
...
+// Relative URL for retrieving metadata for standard and custom objects in Attio.
+func (c *Connector) getObjectsURL(objName string) string {
+	relativeURL := strings.Join([]string{"objects", objName, "attributes"}, "/")
+
+	return relativeURL
+}
+
+// Relative URL for retrieving display Name for standard and custom objects in Attio.
+func (c *Connector) getObjects(objName string) string {
+	relativeURL := strings.Join([]string{"objects", objName}, "/")
```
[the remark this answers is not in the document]
Cobalt0s: Just leave a comment stating that trailing slash is required. If you know the reason or there is a doc for that to link that would be even better.
If somebody will be tempted to remove it or not, the comment would help.

providers/attio/metadata.go
```diff
@@ -65,14 +174,74 @@ func parseMetadataFromResponse(resp *common.JSONHTTPResponse) (*common.ObjectMet
...
-	// Using the first result data to generate the metadata.
-	for k := range response.Data[0] {
-		metadata.FieldsMap[k] = k
+		for _, value := range response.Data {
+			apiSlug := value.APISlug
+			metadata[apiSlug] = common.FieldMetadata{
+				DisplayName:  apiSlug,
+				ValueType:    getFieldValueType(value.Type),
+				ProviderType: value.Type,
+				ReadOnly:     false,
```
Cobalt0s, line 192: JSON response has `is_writable` property. You should set it here. `ReadOnly: !IsWritable`

providers/attio/metadata.go
```diff
@@ -2,60 +2,169 @@ package attio
...
+		)
+	}
+
+	return metadataResult, nil
+}
+
+func (c *Connector) getObjectAttributes(
+	ctx context.Context, obj string,
+) (map[string]common.FieldMetadata, bool, error) {
+	isAttioStandardOrCustomObj := !supportAttioApiObj.Has(obj)
```
Cobalt0s, line 118: As I understand your variable name it should be negated. `isAttioStandardOrCustomObj := supportAttioApiObj.Has(obj)`. Am I understanding this?

providers/attio/metadata.go
```diff
@@ -65,14 +174,74 @@ func parseMetadataFromResponse(resp *common.JSONHTTPResponse) (*common.ObjectMet
 		return nil, common.ErrMissingExpectedValues
 	}
 
-	metadata := &common.ObjectMetadata{
-		FieldsMap: make(map[string]string),
-	}
+	metadata := make(map[string]common.FieldMetadata)
+
+	// Retrieving metadata for standard and custom objects in Attio using the api_slug field.
+	if isAttioStandardOrCustomObj {
```
Cobalt0s, line 180: All other methods make sense. I understand this one too but I think it would be more clear if this method would be split into 2 parts. Even though you would need to repeat the check `len(response.Data) == 0` I would argue it will become simpler to understand. One method would be concerned about `isAttioStandardOrCustomObj=true` and the other with `isAttioStandardOrCustomObj=false`.

One is `Metadata Discovery` the other is `Read Sampling`. I think drawing this distinction is important to tell the reader what methods you use to acquire metadata.

```golang
// Just an example. A pseudo code summary for this method.
func parseMetadataFromResponse(resp *common.JSONHTTPResponse,
	isAttioStandardOrCustomObj bool,
) (map[string]common.FieldMetadata, error) {
  if isAttioStandardOrCustomObj {
    return method1(...)
  }
  return method2(...)
}
```

Also write down docs to explain in what cases we do `Read sampling` and why we cannot use `Metadata Discovery` for better context.

providers/attio/metadata.go
```diff
@@ -65,14 +174,74 @@ func parseMetadataFromResponse(resp *common.JSONHTTPResponse) (*common.ObjectMet
...
+	return res, nil
+}
+
+func getFieldValueType(field string) common.ValueType {
+	switch field {
+	case "number":
+		return common.ValueTypeInt
+	case "text":
+		return common.ValueTypeString
+	case "select":
```
Cobalt0s, line 237: Nice!

providers/attio/metadata_test.go
```diff
@@ -50,25 +47,77 @@ func TestListObjectMetadata(t *testing.T) { // nolint:funlen,gocognit,cyclop
...
-							"api_slug":      "api_slug",
-							"created_at":    "created_at",
-							"id":            "id",
-							"plural_noun":   "plural_noun",
-							"singular_noun": "singular_noun",
-						},
-					},
 					"lists": {
 						DisplayName: "lists",
+						Fields: map[string]common.FieldMetadata{
```
Cobalt0s, line 63: Don't specify all the possible fields. Just a few of different types would be sufficient.

Cobalt0s (changes requested):
I left comments about `parseMetadataFromResponse`. I want the reason of using `Raw Read Sampling` vs `Metadata Discovery URL` to be more aparent in the methods structure and comments. 

For example a code reader might wonder why not use `Attio's Metadata Discovery URL` for all objects? When do we fallback to sampling read records?

I am sorry about this, but Attio is a complex connector. This should be the last push 🚀

providers/attio/metadata.go
sanjay-a01: You can assume the logic like this:

`isNotSupportedByAttioApi := !supportAttioApiObj.Has(obj)`

- If obj is not in supportAttioApiObj, then isNotSupportedByAttioApi becomes true, so the if block executes.

- If obj is in supportAttioApiObj, then isNotSupportedByAttioApi becomes false, so the else block executes.

This negation is intentional to satisfy the if-else condition. Let me know if you see any issue with this logic.

providers/attio/metadata_test.go
sanjay-a01: Specified the possible fields for testing metadata

image

providers/attio/metadata.go
Cobalt0s: I think the confusing bit is: if the object is not supported by attio why do we make api calls?

From last call I rememeber you told me about standard objects:
image
You keep them calling standard which is good, you should link to provider docs for those getting to know Attio for the first time. We cannot assume that Ampersand customers are fully familiar with this API the same for connector maintainers.

Then there are custom objects covered by `objects` and `records` endpoints: 
image

Does this `isAttioStandardOrCustomObj` variable represent the above? And for example the `tasks` would be covered by the `else` clause?
image

Judging from the mock tests, scripts output it is doing the right thing. If you could write down explanation which objects/endpoints fall under which category as method signature or/and README file this would be great. Like digested notes.

Cobalt0s (approved):
Please review the comments. Once you address them, I believe it's good to go.

Great job on figuring out this hard provider!

providers/attio/metadata.go
Cobalt0s: Maybe `isAttioStandardOrCustomObj` could be `discoverableObjects`. This sort of a word that unites both sets.

sanjay-a01: I think the naming is confusing here. I used the name `Attio API` which represent the below section(check the image), not all the APIs which in Attio connector.

Lists, Meta, Entries, Workspace  members, Notes, Tasks etc.. are somes under the section `Attio API` and the `supportAttioApiObj` variable represent this.
image

In case of people, companies, deals, workspaces, users, comes under the section` StandardOrCustom` and the `isAttioStandardOrCustomObj` variable represent this.
Ref- 
image

So, to differentiate between them I used the naming `supportAttioApiObj` and `isAttioStandardOrCustomObj`

Cobalt0s: Makes sense. Write down somewhere in code that standard is not a word that we use but a concept by Attio themselves. I think the variable name of `isAttioStandardOrCustomObj` makes sense in this context.

Cobalt0s (approved):
Since the last review everything is good.

Please, merge the PR after you add multiselect support.

PR #1493, opened 2025-03-18
[CON-1539] feat(Aha): add ListMetadata

immdipu:
#Testing

image
image
image

### DIFF
--- a/providers/aha/metadata/scraping.go
+++ b/providers/aha/metadata/scraping.go
@@ -1,0 +1,19 @@
+package metadata
+
+import (
+	_ "embed"
+
+	"github.com/amp-labs/connectors/internal/staticschema"
+	"github.com/amp-labs/connectors/tools/fileconv"
+	"github.com/amp-labs/connectors/tools/scrapper"
+)
+
+// nolint:gochecknoglobals
+var (
+	//go:embed schemas.json
+	schemaContent []byte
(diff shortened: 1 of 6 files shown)


providers/aha/metadata/scraping.go
```diff
@@ -0,0 +1,19 @@
+package metadata
```
Cobalt0s, line 1: Rename the file `scraping` to `metadata`. It was a mistake I was naming all these files like that. Several connectors started out with metadata scraped.

Cobalt0s (approved):
Good!

providers/aha/metadata/scraping.go
```diff
@@ -0,0 +1,19 @@
...
+	_ "embed"
+
+	"github.com/amp-labs/connectors/internal/staticschema"
+	"github.com/amp-labs/connectors/tools/fileconv"
+	"github.com/amp-labs/connectors/tools/scrapper"
+)
+
+// nolint:gochecknoglobals
+var (
+	//go:embed schemas.json
```
Cobalt0s, line 13: By the way, was the file autogenerated or manualy created? I think we should add a new property to file format to include source for better clarity.

immdipu: it was manually created. there is no OpenAPI available. for now i will add comment for better clarity.

PR #1496, opened 2025-03-19
[CON-1468] feat: Add Drift Metadata Connector

Jkarage:
Tests:
Screenshot 2025-03-19 at 17 29 36


RajatPawar (commented):
@Cobalt0s how do you suggest we gather metadata for this API? Trying to read a single object is okay, but can have it's downsides - any suggestions?

Cobalt0s:
> Pulling in @Cobalt0s - how do you suggest we gather metadata for this API? Trying to read a single object is okay, but can have it's downsides - any suggestions?

The API offers several models: [Drift API Documentation](https://devdocs.drift.com/docs/platform-apis).

We could consider scraping for this task. I have a brief summary [here](https://ampersand.slab.com/posts/sources-for-object-metadata-sl2kdcpi#h2zir-scraper).


# Approach

Given a list of **model documentation URLs** stored in `index.json`, the script would focus on locating the table that contains the metadata. For example, for the [User model](https://devdocs.drift.com/docs/user-model), you could run document.querySelectorAll(".rdmd-table tr") to get the rows. The Capsule script would be helpful since it works directly with tables. See the [code reference](https://github.com/amp-labs/connectors/blob/9c8000ea539cb9597d3d01138ac136f2faeca712/scripts/scraper/capsule/metadata/main.go#L152).

Note: `index.json` could be created **manually**. It will include the URLs that have **tables** which you want to turn into object metadata. Capsule and Salesloft used query selectors to create an index to act uppon.

In Capsule, I was able to extract **field values** using regex, and the same can be done in Drift:
image
Both the Availability and Role fields can be stored as singleSelect types.
Capsule `schema.json`:
image


## Pros
* Query selectors are straightforward to locate model's **table**.
* This approach offers more object metadata. V2 instead of V1.

## Cons
* Extracting the data may require some experimentation with string regex and formats. ([Capsule example](https://github.com/amp-labs/connectors/blob/9c8000ea539cb9597d3d01138ac136f2faeca712/scripts/scraper/capsule/metadata/main.go#L206))
* Using scraping for me is always a disadvantage, because in future the script may break.

providers/drift/handlers.go
```diff
@@ -0,0 +1,96 @@
...
+	request *http.Request,
+	response *common.JSONHTTPResponse,
+) (*common.ObjectMetadata, error) {
+	var (
+		firstRecord map[string]any
+		err         error
+	)
+
+	objectMetadata := common.ObjectMetadata{
+		FieldsMap:   make(map[string]string),
```
Cobalt0s, line 34: `ObjectMetadata.FieldsMap` is deprecated. Populate `ObjectMetadata.Fields`.
Invoking `common.NewObjectMetadata()` would infer `FieldsMap` which is a subset of `Fields`.

I don't know how to enforce data consistency in `ObjectMetadata`. On one side values should be set a certain way via designated constructor which means we should make the fields private to encourage that. But this would break JSON marshalling, because fields must be exported.

@RajatPawar from the usage perspective, is it something to worry about. What if `FieldsMap` is populated and `Fields` isn't? What if one data structure **divereged** from the other because only one was updated?

Jkarage:
Looking into this, seems like a better option.

providers/drift/supports.go
```diff
@@ -0,0 +1 @@
+package drift
```
Cobalt0s, line 1: Is this file needed?

providers/drift/schemas.json
```diff
@@ -0,0 +1,162 @@
+{
+  "modules": {
+    "root": {
+      "id": "root",
+      "path": "",
+      "objects": {
+        "users/list": {
+          "displayName": "Users",
+          "path": "/users/list",
+          "docs": "https://devdocs.drift.com/docs/listing-users",
```
Cobalt0s, line 10: Did you mange to write a scraper? I think these objects are manual. If you have some troubles at some stage let me know I can help you.

Cobalt0s (approved):

providers/drift/schemas.json
Jkarage: Yeah, we only support a few objects for Read, That's why I choose Manual approach. The scrapper might come in handy when updating this schema to be of MetadataV2. I will work on it in another PR.

providers/drift/supports.go
Jkarage: Removing

providers/drift/schemas.json
Cobalt0s: Sure, that works :smile:
Feel free to merge it as is.

providers/drift/handlers.go
Cobalt0s: @Jkarage as I mention below this PR can be merged and this is not really blocking. The question I have here applies to any connector and how we handle V2 and V1 and keep integrity.

PR #1497, opened 2025-03-19
[CON-1540] feat(Aha): Add Read Operation

immdipu:
# Testing 

image
image
image
image

### DIFF
--- a/providers/aha/handlers.go
+++ b/providers/aha/handlers.go
@@ -26,17 +26,12 @@ func (c *Connector) buildReadRequest(ctx context.Context, params common.ReadPara
 		return http.NewRequestWithContext(ctx, http.MethodGet, params.NextPage.String(), nil)
 	}
 
-	var (
-		url *urlbuilder.URL
-		err error
-	)
-
 	path, err := metadata.Schemas.LookupURLPath(c.Module(), params.ObjectName)
 	if err != nil {
 		return nil, err
 	}
 
-	url, err = urlbuilder.New(c.ProviderInfo().BaseURL, apiVersion, path)
+	url, err := urlbuilder.New(c.ProviderInfo().BaseURL, apiVersion, path)
 	if err != nil {
 		return nil, err
 	}
--- a/providers/aha/handlers.go
+++ b/providers/aha/handlers.go
@@ -77,16 +77,16 @@ func makeNextRecordsURL(baseURL *urlbuilder.URL, params common.ReadParams) commo
 
 		totalPage, err := jsonquery.New(pagination).IntegerRequired("total_pages")
 		if err != nil {
-			return "", err
+			return "", nil //nolint:nilerr
 		}
 
 		currentPage, err := jsonquery.New(pagination).IntegerRequired("current_page")
 		if err != nil {
-			return "", err
+			return "", nil //nolint:nilerr
 		}
 
 		if currentPage == totalPage {
-			return "", nil
+			return "", nil //nolint:nilerr
 		}
 
 		nextPage := currentPage + 1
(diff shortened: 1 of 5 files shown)


providers/aha/handlers.go
```diff
@@ -0,0 +1,107 @@
...
+		return baseURL.String(), nil
+	}
+}
+
+func addQueryParams(url *urlbuilder.URL, params common.ReadParams, page int64) {
+	url.WithQueryParam(pageSizeKey, pageSize)
+	url.WithQueryParam(pageKey, strconv.FormatInt(page, 10))
+
+	if supportSince.Has(params.ObjectName) {
+		url.WithQueryParam(sinceKey, datautils.Time.FormatRFC3339inUTC(params.Since))
```
Cobalt0s, line 105: If you could add a doc for since query parameter either for particular endpoint or general, whatever the provider has would be good fo quick access.

test/aha/connector.go
```diff
@@ -13,11 +13,12 @@ import (
 
 func GetAhaConnector(ctx context.Context) *aha.Connector {
 	filePath := credscanning.LoadPath(providers.Aha)
-	reader := utils.MustCreateProvCredJSON(filePath, false, false)
+	reader := utils.MustCreateProvCredJSON(filePath, true, true)
 
 	conn, err := aha.NewConnector(
 		common.Parameters{
 			AuthenticatedClient: utils.NewOauth2Client(ctx, reader, getConfig),
+			Workspace:           "ampersand2",
```
Cobalt0s, line 21: This value should be coming from creds file:
```
reader.Get(credscanning.Fields.Workspace)
```

providers/aha/supports.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+const (
+	objectNameAudits            = "audits"
+	objectNameHistoricalAudits  = "historical_audits"
+	objectNameIdeasEndorsements = "ideas/endorsements"
+	objectNameIdeas             = "ideas"
+)
+
+var supportSince = datautils.NewSet( //nolint:gochecknoglobals
```
Cobalt0s, line 20: So out of 29 objects only 4 support incremental read. Is it correct? (It is quite normal).

providers/aha/handlers.go
```diff
@@ -0,0 +1,107 @@
...
+func makeNextRecordsURL(baseURL *urlbuilder.URL, params common.ReadParams) common.NextPageFunc {
+	return func(node *ajson.Node) (string, error) {
+		pagination, err := jsonquery.New(node).ObjectRequired("pagination")
+		if err != nil {
+			return "", nil //nolint:nilerr
+		}
+
+		totalPage, err := jsonquery.New(pagination).IntegerRequired("total_pages")
+		if err != nil {
+			return "", err
```
Cobalt0s, line 80: I think similiar to the way you handle ObjectRequired you silently return no next page token. I think the same applies for `total_pages` and `current_page`.

`Required` methods would return error in 2 cases:
* Key not found - in our case this is fine to hanle as empty next page.
* Value type mismatch - you expected integer but for some reason it returns a string. If that would ever happen (unlikely), this error would indicate invalid implementation.

Of course one could argue that key not found also means provider API changed and implementation didn't adapt.

Cobalt0s (approved):

providers/aha/handlers.go
immdipu: If there’s a pagination object, it would definitely have `total_pages` and `current_pages`. That’s why I didn’t feel the need to add it, but doing same for `total_pages` and `current_pages` is also a good thing. i'll update it.

providers/aha/supports.go
immdipu: yes

providers/aha/handlers.go
Cobalt0s: Some APIs have different pagination format. Which is crazy. The way you enforced the format works as well. We would need to run into the issue of provider changing pagination contract and how to handle it. Being very strict could be of benefit for majority of connectors.

PR #1501, opened 2025-03-20
Fix UpdateSubscription method for Salesforce to delete correct objects and return correct ressult

jlimatampersand:
The Update method had bugs that the it was deleting all objects. This PR fixes the issue. 

Test Result
- The lookups are not deleted after update
Screenshot 2025-03-19 at 11 20 48 PM


RajatPawar (approved):
LGTM. Could you please add a comment on the deletion to clarify how objects are being picked for deletion?

PR #1504, opened 2025-03-20
[CON-1610] feat(attio): read : To support standard & custom objects

sanjay-a01:
### For Lists object
image

### For workspace_members object
image

### For Tasks object
image

### For Notes object
image

### For Companies object
image
image

## Test Case Result
image


providers/attio/read_test.go
```diff
@@ -130,7 +126,47 @@ func TestRead(t *testing.T) { // nolint:funlen,gocognit,cyclop
...
+			Name:  "Read list of all companies",
+			Input: common.ReadParams{ObjectName: "companies", Fields: connectors.Fields(""), Since: time.Date(2025, 3, 1, 0, 0, 0, 0, time.UTC), NextPage: "test?limit=50"},
+			Server: mockserver.Fixed{
+				Setup:  mockserver.ContentJSON(),
+				Always: mockserver.Response(http.StatusOK, responseCompanies),
+			}.Server(),
+			Expected: &common.ReadResult{
+				Rows: 1,
+				Data: []common.ReadResultRow{{
+					Fields: map[string]any{},
```
Cobalt0s, line 144: Ok, so you need to change this test to request `name` field. And it must be returned here despite being nested under `values`.

As we have discovered all the important fields are under `values` response property.

Check similiar flattenning but of `attrbitues` in [Klaviyo](https://github.com/amp-labs/connectors/blob/51f5ee84fb611a05c19454e6369bec54ebeddcb0/providers/klaviyo/parse.go#L36).

Note: The original `Raw` data should be left untouched. Some existing connectors have this mistake.
There is a PR that would simplify flatenning https://github.com/amp-labs/connectors/pull/1321 and for that there is a [document](https://ampersand.slab.com/posts/parse-read-response-lfrf3rru#hd5xy-proposal).

@RajatPawar this PR would benefit from utilities/refactoring (links above).

Cobalt0s (changes requested):
Everything is good!

But, there is one missing feature:
* Nested properties of `values` should be returned inside `common.ReadResultRow.Fields`.

providers/attio/params.go
```diff
@@ -18,7 +18,7 @@ type parameters struct {
 
 const (
 	// DefaultPageSize is number of elements per page.
-	DefaultPageSize = 10
+	DefaultPageSize = 50
```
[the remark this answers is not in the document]
RajatPawar: Please make sure this is as high as it can be to avoid hitting the API too many times

providers/attio/read_test.go
RajatPawar: Thanks @Cobalt0s, and yes, we should have a test case requesting actual fields that get populated inside `ReadResultRow.Fields`. Once done, can you please add a screenshot here? And as @Cobalt0s mentioned, the `Raw` field should be left untouched

providers/attio/read.go
```diff
@@ -13,10 +13,14 @@ func (c *Connector) Read(ctx context.Context, config common.ReadParams) (*common
 		return nil, err
 	}
 
-	if !supportAttioGeneralApi.Has(config.ObjectName) {
-		return nil, common.ErrOperationNotSupportedForObject
+	if supportAttioGeneralApi.Has(config.ObjectName) {
+		return c.readGeneralAPI(ctx, config)
```
RajatPawar, line 17: Just `readAPI` would be great since `general` is not a term that attio uses

providers/attio/read.go
```diff
@@ -36,6 +40,31 @@ func (c *Connector) Read(ctx context.Context, config common.ReadParams) (*common
 	)
 }
 
+func (c *Connector) readStandardOrCustomObject(
+	ctx context.Context, config common.ReadParams,
+) (*common.ReadResult, error) {
+	// To handle standarad/custom objects
```
RajatPawar, line 46:
```suggestion
	// To handle standard/custom objects
```

providers/attio/url.go
```diff
@@ -21,3 +21,10 @@ func (c *Connector) getObjectsURL(objName string) (*urlbuilder.URL, error) {
 
 	return urlbuilder.New(c.BaseURL, apiVersion, relativeURL)
 }
+
+// Relative URL for retrieving standard and custom object read URL.
+func (c *Connector) getObjectReadURL(objName string) (*urlbuilder.URL, error) {
+	relativeURL := strings.Join([]string{"objects", objName, "records", "query"}, "/")
+
+	return urlbuilder.New(c.BaseURL, apiVersion, relativeURL)
```
RajatPawar, line 29: minor: I believe urlbuilder.New already takes care of joining, so this can be 
```
return urlbuilder.New(c.BaseURL, apiVersion, []string{"objects", objName, "records", "query"}...)
```

RajatPawar (approved):
Approving this to unblock, but please address all the comments & add a new screenshot for the updated test cases before merging this

providers/attio/url.go
Cobalt0s: Thats right, the slice can be unwrapped further.
```
return urlbuilder.New(c.BaseURL, apiVersion, "objects", objName, "records", "query")
```

providers/attio/README.md
```diff
@@ -30,3 +30,8 @@ Used endpoint to get the display name: https://developers.attio.com/reference/ge
 Used endpoint to get fields: https://developers.attio.com/reference/get_v2-target-identifier-attributes
 
 For non-standard objects, directly used the object's endpoint.
+
+## Read Functions
+For standard/custom objects, a separate read function like "readStandardOrCustomObject" is used, which calls the endpoint to get the response: https://developers.attio.com/reference/post_v2-objects-object-records-query
+
+For non-standard object, a separate read function like "readGeneralAPI" is used and directly used the object's endpoint to get the response.
```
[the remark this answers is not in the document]
sanjay-a01: image

providers/attio/params.go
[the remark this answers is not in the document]
sanjay-a01: The pagination is supported for notes, tasks and standard/custom objects

For notes - Default page size is 50
For tasks and standard /custom objects - Default page size is 500

So I used two constant
image

Implemented new function as **seLimit** to handle the above thing
image

image

providers/attio/read.go
```diff
@@ -54,3 +83,31 @@ func (c *Connector) buildURL(config common.ReadParams) (*urlbuilder.URL, error)
...
+		if err != nil {
+			return nil
+		}
+
+		body["offset"] = offset
+	}
+
+	if !config.Since.IsZero() {
+		filter["created_at"] = map[string]string{
+			"$gte": config.Since.Format("2006-01-02T15:04:05.999999999Z"),
```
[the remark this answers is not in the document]
sanjay-a01: Tested wtih unix time and got the same result. Below I attached the screenshots of code
 is it correct?
image

Response
image

providers/attio/read.go
```diff
@@ -54,3 +83,31 @@ func (c *Connector) buildURL(config common.ReadParams) (*urlbuilder.URL, error)
...
+		}
+	}
+
+	body["limit"] = DefaultPageSize
+
+	body["filter"] = filter
```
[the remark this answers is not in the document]
sanjay-a01: No, Filter is not required fields in the POST Method.

If we give the filter value, based on the filter value the response will shown. If not given all records of the object is shown.
below, I attached code screenshots.
image

providers/attio/read_test.go
sanjay-a01: ## Read Response (companies object)
image

## Test Cases Result:
image

providers/attio/parse.go
```diff
@@ -40,3 +41,117 @@ func makeNextRecordsURL(reqLink *urlbuilder.URL) common.NextPageFunc {
...
+//	       .... (more response data will be there)
+//
+// The resulting fields for the above will be: id, created_at, record_id.
+func getStandardOrCustomObjRecords(node *ajson.Node) ([]map[string]any, error) {
+	arr, err := jsonquery.New(node).ArrayOptional("data")
+	if err != nil {
+		return nil, err
+	}
+
+	return flattenRecords(arr)
```
Cobalt0s, line 121: [major]
Sadly, this is a bug. Several connectors do this. I made this mistake because I thought it is intutive to work with records here. But the returned `[]map[string]any` will be used by `Fields` AND by the `Raw` inside `common.GetMarshaledData`.

This means we are changing both Fields (good) and Raw (bad). Raw response should be exactly what we get from provider API. Fields is the made up Ampersand data structure. And flatten is only relevant there. You should change the tests to reflect this requirement.

I have a pending [proposal](https://ampersand.slab.com/posts/parse-read-response-lfrf3rru#hd5xy-proposal) to simplify it. But it is not yet approved.

Sadly, this means you will have to fix it the hard way. 
* `getStandardOrCustomObjRecords` should be `common.GetRecordsUnderJSONPath`
* `GetMarshaledData` would do all the heavy lifting. Which will use flattenRecords only for the `Fields`. Raw will be unchanged.

PS: To simplify working with `[]map[string]any` which is hard because of all the type conversions, pass original response directly to acquire `*ajson.Node` which you rely on by `flattenRecords`. Thats one way of doing this (it kind of defeats the purpose of `common.ParseResult`, but its ok):
image

providers/attio/read.go
```diff
@@ -48,9 +78,42 @@ func (c *Connector) buildURL(config common.ReadParams) (*urlbuilder.URL, error)
...
+
 		url.WithQueryParam("offset", "0")
 	}
 
 	return url, nil
 }
+
+// construct body params for filter the data using since field.
+func constructBody(config common.ReadParams) map[string]any {
+	filter := make(map[string]any)
```
Cobalt0s, line 95: [minor]
This variable is only used within `!config.Since.IsZero()`, not in the outter scope. You should move it there. Right now, its kind of a local-global variable.

providers/attio/read_test.go
```diff
@@ -130,7 +126,58 @@ func TestRead(t *testing.T) { // nolint:funlen,gocognit,cyclop
...
+						},
+					},
+					Raw: map[string]any{
+						"id": map[string]any{
+							"workspace_id": "63d34516-b287-4c27-9d28-fe2adbebcd50",
+							"object_id":    "1fa986a6-952e-4e92-ba01-acca61a7b616",
+							"record_id":    "2db97cee-6c6b-4486-ae52-db8e4b6f44e9",
+						},
+						"created_at": "2025-03-12T07:55:38.327000000Z",
+						"name": []any{
```
Cobalt0s, line 165: [major]
The raw must have "values" object. That would be the correct implementation of a requirement. 

Currently the fields is correct, but now the Raw is broken.

Cobalt0s (changes requested):
The Raw response is broken. That is the main concern. I started those comments with `[major]` tag.

Other comments are just out there as notes/suggestions.

providers/attio/read.go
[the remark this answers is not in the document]
Cobalt0s: I meant if the API would accept raw number or it has a strict query parameter format.

To clarify, I noticed some providers have bad documentation on the time format. And sometimes they return time in different format in response. I thought that it is ok to use it as query parameter but it was rejected. I had to convert the time for API to succeed :cry: 

As soon as the RFC3339 with milliseconds works for all its good. An API could even complain if it is 3339 but without milliseconds(I had one such provider).

Don't check every object, if you got a feeling that timing query should work for most if not all objects thats good. I told you this for you to be aware of the fun providers, edge cases. Their documentation could lie or really wouldn't specify the time format well enough.

providers/attio/read.go
```diff
@@ -48,9 +78,45 @@ func (c *Connector) buildURL(config common.ReadParams) (*urlbuilder.URL, error)
...
+
+		url.WithQueryParam("limit", strconv.Itoa(pageSize))
 		url.WithQueryParam("offset", "0")
 	}
 
 	return url, nil
 }
+
+// construct body params for filter the data using since field.
+func constructBody(config common.ReadParams) map[string]any {
```
Cobalt0s, line 97: Alternative names: `constructRequestBody, constructSearchPayload`, etc.

providers/attio/parse.go
```diff
@@ -40,3 +41,177 @@ func makeNextRecordsURL(reqLink *urlbuilder.URL) common.NextPageFunc {
 		return "", nil
 	}
 }
+
+// To determine the next page records for the standard/custom objects.
+func makeNextRecordStandardObj(body map[string]any) common.NextPageFunc {
```
Cobalt0s, line 46: I was a little bit confused due to the naming. Body would usually refer to the response body. But in this case this means a search request payload which you used. You construct it, in the outter `readStandardOrCustomObject` which then passes it here.
To avoid confusion, I think we can directly pass the offset. Thats the goal of your next record function. You depend on the offset coming from the `request` you made. Unmarshalling and marshalling kind of suggests that you are treating it as an opaque data, but in fact it is not. You fully constructed this JSON map, so you know the offset.

providers/attio/parse.go
```diff
@@ -40,3 +41,177 @@ func makeNextRecordsURL(reqLink *urlbuilder.URL) common.NextPageFunc {
...
+		if !ok {
+			return nil, common.ErrEmptyRecordIdResponse
+		}
+
+		recordID, ok := id["record_id"].(string)
+		if !ok {
+			return nil, common.ErrEmptyRecordIdResponse
+		}
+
+		FeildRecord := flattenRecords[recordID]
```
Cobalt0s, line 167: Should start with the lowercase and fix spelling.

providers/attio/parse.go
```diff
@@ -40,3 +41,177 @@ func makeNextRecordsURL(reqLink *urlbuilder.URL) common.NextPageFunc {
...
+// To determine the next page records for the standard/custom objects.
+func makeNextRecordStandardObj(body map[string]any) common.NextPageFunc {
+	return func(node *ajson.Node) (string, error) {
+		// Extract the data key value from the response.
+		value, err := jsonquery.New(node).ArrayRequired("data")
+		if err != nil {
+			return "", err
+		}
+
+		if len(value) != 0 {
```
Cobalt0s, line 54: To shorten the if block. Use inverted condition `len(value) == 0` to `return "", nil`.

Reference: https://danp.net/posts/reducing-go-nesting/

providers/attio/parse.go
```diff
@@ -40,3 +41,177 @@ func makeNextRecordsURL(reqLink *urlbuilder.URL) common.NextPageFunc {
...
+
+		FeildRecord := flattenRecords[recordID]
+		data[i].Raw = record
+		data[i].Fields = common.ExtractLowercaseFieldsFromRaw(fields, FeildRecord)
+	}
+
+	return data, nil
+}
+
+func flattenRecords(arr []*ajson.Node) (map[string]map[string]any, error) {
```
Cobalt0s, line 175: Just an idea: instead of accepting an array of `*ajson.Node` flattener method could focus on one node at a time. Then you could instead of returning a map of JSON (which is `map[string]map[string]any`) just return single JSON (map[string]any). This could simplify the outter callers, but I imagine this means some reshuffling, you could try it for fun, but this works!

Cobalt0s (approved):
Awesome! The task is done. :partying_face:

providers/attio/parse.go
sanjay-a01: Based on the suggestion, I directly passed the offset value.
I've attached the screenshots below
image

image

providers/attio/parse.go
sanjay-a01: I modified the code like below because flattener method could focus on one node at a time.

image

image

PR #1515, opened 2025-03-24
[CON-1485] feat: Add Podium Read Connector

Jkarage:
Tests:
Screenshot 2025-03-24 at 12 11 15

### DIFF
--- a/providers/podium/handlers.go
+++ b/providers/podium/handlers.go
@@ -3,6 +3,7 @@ package podium
 import (
 	"context"
 	"net/http"
+	"strconv"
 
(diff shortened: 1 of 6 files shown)


providers/podium/handlers.go
```diff
@@ -26,9 +29,6 @@ func (c *Connector) buildSingleObjectMetadataRequest(ctx context.Context, object
 		return nil, err
 	}
 
-	// Limit response to 1 record data.
```
Jkarage, line 29: Removed this as there are resource that will error when this is provided.

eberle1080 (approved):

PR #1532, opened 2025-03-28
[CON-1585] feat: Add GitLab Write Connector

Jkarage:
Tests:
Screenshot 2025-03-28 at 18 22 15


providers/gitlab/supports.go
```diff
@@ -78,12 +78,54 @@ func supportedOperations() components.EndpointRegistryInput {
 		"web_commits/public_key",
 	}
 
+	writeSupport := []string{
```
jlimatampersand, line 81: can we have this list alphabetically sorted

jlimatampersand (approved):

PR #1544, opened 2025-04-04
[CON-1428] feat(salesforceMarketing): Add ListMetadata

immdipu:
# Test
image
image
image


Cobalt0s (approved):
Approved!

How did you create `schema.json`, was it manual?

RajatPawar (commented):
We are nearly finishing up module work, could this be a module on the salesforce connector? If so, it might be better to wait for that and then merge this in

@Cobalt0s

Cobalt0s:
> We are nearly finishing up module work, could this be a module on the salesforce connector? If so, it might be better to wait for that and then merge this in
> 
> @Cobalt0s

I don't think that it can be. Different BaseURLs are supported but not the Auth, and that has different `TokenURLs`:
image

VS Salesforce:
image

Authentication is not taken into account with current modules architecture.

RajatPawar (approved):

PR #1549, opened 2025-04-07
fix: generating Object Names using static schema

Jkarage:
In the current implementation, we only support object names without path separators — for example, `/privileges`. If the object name includes a path like `/account/user/privileges`, we incorrectly extract only the last segment (`privileges`), which might lead to errors.

This update adds support for path-separated object names. I’ve updated one provider (Dixa) as part of this PR. If approved, similar updates will be required for several other providers.

Screenshot 2025-04-07 at 14 39 23


tools/fileconv/api3/parameters.go
```diff
@@ -230,3 +231,15 @@ func WithDuplicatesResolver(duplicatesResolver DuplicatesResolver) Option {
...
 	}
 }
+
+// WithVersionPrefix sets a version prefix to strip from API paths
+// when generating object names.
+//
+// For example, if an API path is "/v1/customers/list", using
+// WithVersionPrefix("/v1/") will result in "customers/list" as
+// the object name.
+func WithVersionPrefix(prefix string) Option {
```
Cobalt0s, line 241: This change is related to the PR that was only recently merged: https://github.com/amp-labs/connectors/pull/1400.
However, for most apis where there is a common prefix path for all endpoints, this solution is shorter.
I think it can coexist. However, the line below `parts := strings.Split(path, e.versionPrefix)` enforces all objectNames to be everything that follows the version prefix. I don't think that all object names are going to turn into relative endpoint paths. @RajatPawar correct me if I am wrong.

Since `WithVersionPrefix` is a special case of `WithDuplicatesResolver` one should be defined in terms of the other. I created duplicates resolver for the Stripe connector. Which does the prefix cutting: https://github.com/amp-labs/connectors/blob/5d26d8f9dabf28d0edb0abc6ba9a66e0bddf4b98/scripts/openapi/stripe/metadata/main.go#L68

Whatever the change you will make I will update my Stripe PR to use your version of object name formation. In Stripe I assume if not colliding name then last uri part is the object name, otherwise full path without version prefix

PR #1555, opened 2025-04-09
Read heyreach connector

sanjay-a01:
## Sample Read Response
image
image

image
image

## Test case Result
image


providers/heyreach/handlers.go
```diff
@@ -0,0 +1,68 @@
...
+	return http.NewRequestWithContext(ctx, http.MethodPost, url.String(), bytes.NewReader(jsonData))
+}
+
+func (c *Connector) parseReadResponse(
+	ctx context.Context,
+	params common.ReadParams,
+	request *http.Request,
+	response *common.JSONHTTPResponse,
+) (*common.ReadResult, error) {
+	offset, _ := strconv.Atoi(params.NextPage.String())
```
Cobalt0s, line 41: I think it is beneficial to propagate an error to notify the user that NextPage their specified is invalid, otherwise we silently return the first page which could be not the intention.

providers/heyreach/handlers.go
```diff
@@ -0,0 +1,68 @@
...
+
+// construct body params for giving pagination value.
+func constructRequestBody(config common.ReadParams) map[string]any {
+	body := map[string]any{}
+
+	if len(config.NextPage) != 0 {
+		offset, err := strconv.Atoi(config.NextPage.String())
+		if err != nil {
+			return nil
+		}
```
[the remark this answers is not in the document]
Cobalt0s: Similiar to the above comment, should surface an error. Maybe even wrap it in `common.ErrNextPageInvalid`.

This function could have calls to `json.Marshal` and `bytes.NewReader`, they are kind of related to `payload` construction. It's up to you, if you want to move those bits here and there, partially or none.

providers/heyreach/handlers.go
```diff
@@ -0,0 +1,68 @@
...
+	return common.ParseResult(
+		response,
+		common.ExtractRecordsFromPath("items"),
+		makeNextRecord(offset),
+		common.GetMarshaledData,
+		params.Fields,
+	)
+}
+
+// construct body params for giving pagination value.
```
Cobalt0s, line 52: The link to some doc to reference why are you doing this would be helpful.

providers/heyreach/utils.go
```diff
@@ -0,0 +1,48 @@
...
+
+const (
+	DefaultPageSize = 100
+)
+
+func matchReadObjectNameToEndpointPath(objectName string) (urlPath string, err error) {
+	switch objectName {
+	case objectNameCampaign:
+		// https://documenter.getpostman.com/view/23808049/2sA2xb5F75#4aaf461e-54c4-4447-beb6-eb5ccca53fb3
+		return "campaign/GetAll", nil
```
Cobalt0s, line 19: You could hard code these values into `schema.json` since the file is manually created anyway.

Cobalt0s (approved):
LGTM

providers/heyreach/metadata/schemas.json
```diff
@@ -0,0 +1,147 @@
...
+              "displayName": "excludeNoProfilePicture",
+              "valueType": "other"
+            },
+            "excludeListId": {
+              "displayName": "excludeListId",
+              "valueType": "other"
+            }
+          }
+        },
+        "li_account/GetAll": {
```
Cobalt0s, line 70: The object could be named `li_account` while path is preserved as `li_account/GetAll`. That could count as endpoint name alternation.

@RajatPawar what is the correct way to call an object in this case? The last URI part signifies a verb. Using such objectName for Write/Delete could confuse.

RajatPawar: Let's keep it as the provider name (although I agree it doesn't read very well) to avoid having to document the mapping everywhere

eberle1080 (approved):

PR #1559, opened 2025-04-11
feat: Add a Mode to common.Header to control how headers get applied

eberle1080:
This adds a mode field to common.Header so we can have a little more control over how headers get applied to a request. They can be appended (use add), overwritten (use set), or added if unset (check for empty then add). This gives us the ability on the server side to specify how clashing headers are handled.

This also adds a generic modifier so that we can inject header removal logic, if necessary.

This is possibly necessary to fix the Asana and Gong connectors -- I think the Authorization header is being added more than once.


jlimatampersand (changes requested):
This mode seems to be getting applied to many providers which seems natural, but I am afraid that this may introduce regression. 

Can you separate the pr into 2 PRs
1. Just adding `Mode` field and header applying methods. Then add tests around `getRequestModifier` and `ApplyToRequest` without actually using the modifiers. 

2. Next PR, actually apply only to Asana and Gong as this is prerequisite to those connectors. Then test with Asana and Gong for that PR. 

I think once we have that, we can be confident applying to other providers. Applying to multiple providers at a time seems risky.

eberle1080:
@jlimatampersand that's a fair point about the regressions, I'll remove mode from the providers themselves and add some tests. However, the actual fixes for Asana and Gong will happen in the server repo, not here -- the issue has to do with oauth stuff, not core provider calls.

jlimatampersand (approved):

PR #1564, opened 2025-04-15
fix: Github objectName and schemas

immdipu:
# Context
The same object had different names for the Write and Read operations. Therefore, this is a fix to use the same object naming convention for both Read and Write operations.

### DIFF
--- a/providers/github/metadata/schemas.json
+++ b/providers/github/metadata/schemas.json
@@ -338,261 +338,6 @@
             }
           }
         },
-        "codespaces": {
-          "displayName": "Codespaces",
-          "path": "/user/codespaces",
-          "responseKey": "codespaces",
-          "fields": {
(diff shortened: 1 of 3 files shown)


providers/github/metadata/schemas.json
```diff
@@ -338,261 +338,6 @@
             }
           }
         },
-        "codespaces": {
```
Cobalt0s, line 341: Should this object be removed?

Cobalt0s (approved):

providers/github/metadata/schemas.json
Cobalt0s: Good! I was thinking that it is intended, but at last moment wrote a comment just in case. Asking never hurts :pray: .

Cobalt0s (approved):

PR #1570, opened 2025-04-17
chore: add Require workspace for the Aha connector

immdipu:


Cobalt0s (approved):

Cobalt0s:
The branch is calles ashby but the change is for Aha. It makes sense for the aha.

PR #1574, opened 2025-04-18
feat: Add the AllNames function

eberle1080:
There's a problem I'm working on right now where I need a list of all of the provider names. We don't really expose this, and I'd prefer not to build a custom catalog (this is happening inside of an init function, so I really don't want to deal with errors, which the catalog stuff has). I'm just trying to build a map, and need the names for keys.

### DIFF
--- a/providers/catalog.go
+++ b/providers/catalog.go
@@ -25,3 +25,14 @@ func getCatalog() *CatalogWrapper {
 		Timestamp: generated.Timestamp,
 	}
 }
+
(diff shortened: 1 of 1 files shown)


laurenzlong (approved):
This works. Alternatively, it might be easier to load https://github.com/amp-labs/connectors/blob/main/internal/generated/catalog.json into memory and return the keys as opposed to looping through the catalog

PR #1590, opened 2025-04-24
fix: (Salesforce) Normalize compound field names

eberle1080:
This adds logic to normalize field names, so that updates generated by Salesforce compound fields will strip out the compound field prefix.

### DIFF
--- a/providers/salesforce/subscriptionEvent.go
+++ b/providers/salesforce/subscriptionEvent.go
@@ -300,20 +300,22 @@ func (s SubscriptionEvent) normalizeUpdatedFieldName(name string) (string, error
 		return name, nil
 	}
 
+	// Compound fields look like "Field.Subfield"
+	// We're interested in the rightmost part, but to validate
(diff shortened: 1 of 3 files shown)


jlimatampersand:
the logic itself makes sense. Just to confirm. 

This is assuming that we are adding `firstname` or `lastname` in `requiredWatchFields` right? So that we can make the amp.yaml to free from concern to add compound field anywhere? eg) read mapping, and subscribe watch fields?

jlimatampersand (approved):

eberle1080:
> the logic itself makes sense. Just to confirm.
> 
> This is assuming that we are adding `firstname` or `lastname` in `requiredWatchFields` right? So that we can make the amp.yaml to free from concern to add compound field anywhere? eg) read mapping, and subscribe watch fields?

That's right. And they can still refer to compound fields if they want, but only the top-level field (e.g. "Name"), not the sub-fields (e.g. "Name.FirstName")

PR #1592, opened 2025-04-24
fix: Salesforce field normalization bug

eberle1080:


providers/salesforce/subscriptionEvent.go
```diff
@@ -272,7 +272,7 @@ func init() {
 		obj := strings.ToLower(field.Object)
 		fld := strings.ToLower(field.Field)
 
-		if _, ok := compositePrefixMap[field.Object]; !ok {
+		if _, ok := compositePrefixMap[obj]; !ok {
```
eberle1080, line 275: In case there's any confusion, the bug was that I wasn't using obj consistently

dionlow (approved):

PR #1594, opened 2025-04-25
fix: Reading Next-Page in Marketo Incremental Leads

Jkarage:
Issue:
Reading Leads, using the Incremental read approach, should be treated differently in the first call.  The next-page call should be delegated to the full Lead read, as it takes care of adding subsequent ids, provided the starting Id. 

Reading Next-Page Leads:
Screenshot 2025-04-25 at 11 59 40

### DIFF
--- a/providers/marketo/url.go
+++ b/providers/marketo/url.go
@@ -45,15 +45,11 @@ func (c *Connector) constructReadURL(ctx context.Context, params common.ReadPara
 		return nil, err
 	}
 
-	if err := constructURLQueries(url, params); err != nil {
+	if err := c.constructURLQueries(ctx, url, params); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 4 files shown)


providers/marketo/url.go
```diff
@@ -67,6 +63,33 @@ func (c *Connector) constructReadURL(ctx context.Context, params common.ReadPara
 	return url, nil
 }
 
+func (c *Connector) constructURLQueries(ctx context.Context, url *urlbuilder.URL, params common.ReadParams) error {
+	// We don't handle this scenario here, this is handled in the handleLeadsAPI
+	// function, this indicates first call for the Incremental Lead Read.
+	if params.ObjectName == leads && !params.Since.IsZero() && params.NextPage == "" {
```
Jkarage, line 69: Apart from restructuring, the change made is add next-page check here.

jatinsandilya (approved):

PR #1595, opened 2025-04-25
[CON-1602] feat: Instantly V2 connector metadata

sanjay-a01:
### Sample Response

image

### TestCase Result

image


providers/instantlyai/README.md
```diff
@@ -0,0 +1,33 @@
...
+| Subsequences              | subsequences               | read   |
+| Lead                      | leads/list                 | read   |
+| Campaigns analytics       | campaigns/analytics        | read   |
+| Campaigns analytics daily | campaigns/analytics/daily  | read   |
+| Campaigns analytics steps | campaigns/analytics/steps  | read   |
+| Email service provider    | inbox-placement-tests/email| read   |
+                              -service-provider-options  
+------------------------------------------------------------------
+
+The "Lead" object supports the POST API method instead of GET, while the Campaigns Analytics, Daily, Steps, and Email Service Provider objects return a direct response, meaning the response data is not nested within an object like "items" or "data." The remaining objects' responses are nested within an "items" object.
```
Cobalt0s, line 33: I think this sentence would be clearer if it's split into three parts:
* First, mention the use of the POST API — as opposed to GET, which is used for read actions like with the Lead object.
* Second, point out that objects like Analytics, Daily, and Steps return data directly at the root level of the JSON.
* Third, note that other objects return their data nested under an "items" field.

providers/instantlyai/metadata_test.go
```diff
@@ -0,0 +1,252 @@
...
+	"github.com/amp-labs/connectors/test/utils/mockutils/mockcond"
+	"github.com/amp-labs/connectors/test/utils/mockutils/mockserver"
+	"github.com/amp-labs/connectors/test/utils/testroutines"
+	"github.com/amp-labs/connectors/test/utils/testutils"
+)
+
+func TestListObjectMetadata(t *testing.T) { // nolint:funlen,gocognit,cyclop,maintidx
+	t.Parallel()
+
+	apiKeysResponse := testutils.DataFromFile(t, "api-keys.json")
```
Cobalt0s, line 18: I don't think API Keys should be considered a valid object for the connector, similar to webhooks and custom fields.

Reason: They don't seem to be "natural" data representations.
* API Keys are technical in nature.
* Webhooks are used in a specific way by the connector to enable subscriptions.
* Custom objects and fields are integrated in a special way by the connector to mimic built-in objects.

test/instantlyai/connector.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	client, err := common.NewApiKeyHeaderAuthHTTPClient(
+		ctx, headerName, headerValue)
+	if err != nil {
+		utils.Fail(err.Error())
+	}
+
+	conn, err := instantlyai.NewConnector(common.Parameters{
+		AuthenticatedClient: client,
```
Cobalt0s, line 34: I created a utility function to simplify the initialization code for the client:
```
"github.com/amp-labs/connectors/test/utils"

AuthenticatedClient: utils.NewAPIKeyClient(ctx, reader, providers.Front),
```

Internally it does the same thing.

Cobalt0s (approved):
Good!

PR #1597, opened 2025-04-25
[CON-1667] SPIKE: AWS Identity Center connector

Cobalt0s:


Cobalt0s:
* **#1597** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1597?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1597?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

common/awsClient.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	return AWSClient{
+		client: client,
+		cfg:    cfg,
+		region: region,
+	}, nil
+}
+
+func (c AWSClient) Do(req *http.Request) (*http.Response, error) {
```
Cobalt0s, line 53: Expectations:
* The request maker must pass service name via context.Context

Algorithm:
* Payload is sha256 encoded
* HTTP request is signed using AWS signer.

common/json.go
```diff
@@ -130,8 +130,9 @@ func ParseJSONResponse(res *http.Response, body []byte) (*JSONHTTPResponse, erro
 		}, nil
 	}
 
-	// Ensure the response is JSON
-	if err := EnsureContentType(`^application/.*json$`, res, false); err != nil {
+	// Ensure the response is JSON.
+	// Starts with application ends with JSON which may be followed by the version.
+	if err := EnsureContentType(`^application/.*json([-0-9.])*$`, res, false); err != nil {
```
Cobalt0s, line 135: AWS had version at the end: `application/x-amz-json-1.1`

providers/awsIdentityCenter.go
```diff
@@ -0,0 +1,33 @@
+package providers
+
+const AWSIdentityCenter Provider = "awsIdentityCenter"
+
+func init() {
+	SetInfo(AWSIdentityCenter, ProviderInfo{
+		DisplayName: "AWS Identity Center",
+		AuthType:    Basic,
+		BaseURL:     "https://AWS_SERVICE_PLACEHOLDER.{{.region}}.amazonaws.com",
```
Cobalt0s, line 9: There are 2 domains which are `sso` and `identitystore`. They are determined at runtime based on the `objectName`. This doesn't align well with proxy.

providers/awsic/commands.go
```diff
@@ -0,0 +1,81 @@
...
+		ServiceName:   "AWSIdentityStore",
+		Command:       "ListGroups",
+		PayloadBuilder: func(values map[string]string) []byte {
+			return servicePayload{
+				MaxResults:      10,
+				IdentityStoreId: goutils.Pointer(values["identityStoreID"]),
+			}.toBytes()
+		},
+	},
+	"Users": {
```
Cobalt0s, line 58: This registry captures what data is needed to read objects.
Payload differs:
* `identitystore` service needs `IdentityStoreID` arg
* `sso` service needs `InstanceArn` arg

providers/awsic/toBeRemoved.go
```diff
@@ -0,0 +1,58 @@
...
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/credentials"
+	"github.com/aws/aws-sdk-go-v2/service/ssoadmin"
+)
+
+// TODO this file is to be removed
```
Cobalt0s, line 13: Disregard this file. This uses SDK directly. I kept it for the reference.

test/awsic/connector.go
```diff
@@ -0,0 +1,48 @@
...
+	"net/http"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/scanning/credscanning"
+	"github.com/amp-labs/connectors/providers"
+	"github.com/amp-labs/connectors/providers/awsic"
+	testUtils "github.com/amp-labs/connectors/test/utils"
+)
+
+func GetAWSIdentityCenterConnector(ctx context.Context) *awsic.Connector {
```
Cobalt0s, line 14: `username` would hold AWS `accessKeyID` and
`password` would hold AWS `accessKeySecret`.

I assume for both services that connector is to be initialized with `IdentityStoreID` to scope `identitystore` service requests and `InstanceArn` for `sso` service.

test/awsic/read/users/main.go
```diff
@@ -0,0 +1,36 @@
...
+	res, err := conn.Read(ctx, common.ReadParams{
+		ObjectName: "Users",
+		Fields:     connectors.Fields("id"),
+	})
+	if err != nil {
+		utils.Fail("error reading from provider", "error", err)
+	}
+
+	slog.Info("Reading")
+	utils.DumpJSON(res, os.Stdout)
```
Cobalt0s, line 35: image

test/awsic/read/instances/main.go
```diff
@@ -0,0 +1,36 @@
...
+	res, err := conn.Read(ctx, common.ReadParams{
+		ObjectName: "Instances",
+		Fields:     connectors.Fields("id"),
+	})
+	if err != nil {
+		utils.Fail("error reading from provider", "error", err)
+	}
+
+	slog.Info("Reading")
+	utils.DumpJSON(res, os.Stdout)
```
Cobalt0s, line 35: image

test/awsic/read/groups/main.go
```diff
@@ -0,0 +1,36 @@
...
+	res, err := conn.Read(ctx, common.ReadParams{
+		ObjectName: "Groups",
+		Fields:     connectors.Fields("id"),
+	})
+	if err != nil {
+		utils.Fail("error reading from provider", "error", err)
+	}
+
+	slog.Info("Reading")
+	utils.DumpJSON(res, os.Stdout)
```
Cobalt0s, line 35: image

Cobalt0s (commented):
@eberle1080 

I left high level overview notes on the PR explaning the Spike.

common/awsClient.go
```diff
@@ -0,0 +1,98 @@
...
+// AWS service name into the context, therefore the request cannot be constructed for sending.
+var ErrRequestAWSMissingService = errors.New("AWS request is missing Service name, supplied via context")
+
+type AWSClient struct {
+	client *http.Client
+	cfg    aws.Config
+	region string
+}
+
+func NewAWSClient(ctx context.Context, client *http.Client, accessKeyID, accessKeySecret, region string) (AuthenticatedHTTPClient, error) {
```
eberle1080, line 33: You could pass "service" in at this point, rather than via context. It would be clearer here (less implied logic), but if there's a good reason to defer / delay that decision, that might be a good reason to use the context approach.

providers/awsIdentityCenter.go
eberle1080: we *might* be able to solve that by having the proxy caller include an extra header (e.g. X-Aws-Service). And then use that to populate the context. It's not great though.

Cobalt0s: This is a good idea. I have thought about the design of the AWS deep connector and how to go from here.
I think we could have 2 modules where each would have the service domain hard coded. But that would mean there will be no Proxy to AWS unless it is done via module.

@RajatPawar I found a use case for the module proxy. But this one is more of an exception than a trend among all connectors.

common/awsClient.go
Cobalt0s: We could, but I don't want the person which will instantiate `AWSClient` to require more context.

Also in current setup a deep connector may have calls with multiple services, in other words it is scoped to single API request rather than the Client. If we split deep connector by modules then the client may own this information.

Connector module <-- constitutes --> AWS Service.

test/awsic/connector.go
```diff
@@ -0,0 +1,48 @@
...
+	filePath := credscanning.LoadPath(providers.AWSIdentityCenter)
+	reader := testUtils.MustCreateProvCredJSON(filePath, false, false)
+
+	awsRegion := "us-east-2"
+
+	client, err := common.NewAWSClient(ctx,
+		http.DefaultClient,
+		reader.Get(credscanning.Fields.Username),
+		reader.Get(credscanning.Fields.Password),
```
eberle1080, line 23: I'd create new fields. I don't want developers getting confused by the terminology. It's not really a user nor a password.

eberle1080: For a spike this is fine, but for some future maintainer-friendly version, I'd change it.

Cobalt0s: Cool, I will address this one.

Cobalt0s:
Implementation is addressed here: https://github.com/amp-labs/connectors/pull/1603.

test/awsic/connector.go
Cobalt0s: The credscanning requires to comply with ProviderInfo. So I must utilize username/password for AuthType basic. This was done by design.

I am not sure if it was a good decision on my side. But it seems to be logical to respect ProviderInfo definition which "shapes" what creds should be.

PR #1601, opened 2025-04-28
[CON-1659] feat: Added metadata for fireflies connector

sanjay-a01:
### Sample Response

image


RajatPawar (approved):

providers/fireflies/types.go
```diff
@@ -0,0 +1,37 @@
+package fireflies
+
+type TypeKind string
+
+const (
+	KindScalar  TypeKind = "SCALAR"
+	KindObject  TypeKind = "OBJECT"
+	KindEnum    TypeKind = "ENUM"
```
Cobalt0s, line 8: I see that some fields could be Enums.
Are there additional fields returned as part of `type Field struct` that would indicate the list of options?
This would be useful in populating `FieldMetadata.Values` which is of type `[]common.FieldValue`.

providers/fireflies/handlers.go
```diff
@@ -0,0 +1,84 @@
...
+		return nil, fmt.Errorf(
+			"missing or empty fields for object: %s, error: %w",
+			objectName,
+			common.ErrMissingExpectedValues,
+		)
+	}
+
+	// Process each field from the introspection result
+	for _, field := range metadataResp.Data.Type.Fields {
+		objectMetadata.FieldsMap[field.Name] = field.Name
```
Cobalt0s, line 80: FieldsMap is deprecated and Fields should be a priority.
Method `AddFieldMetadata` populates both with focus on `map[string]FieldMetadata`.
If provider can give us field metadata that is not just limited to the display name of the field but its type, possible values that should be "harvested".
image

test/fireflies/connector.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	client, err := common.NewApiKeyHeaderAuthHTTPClient(
+		ctx, headerName, headerValue)
+	if err != nil {
+		utils.Fail(err.Error())
+	}
+
+	conn, err := fireflies.NewConnector(common.Parameters{
+		AuthenticatedClient: client,
```
Cobalt0s, line 34: Shorter version:
```
AuthenticatedClient: utils.NewAPIKeyClient(ctx, reader, providers.Fireflies)
```

test/fireflies/metadata/metadata.go
```diff
@@ -0,0 +1,25 @@
...
+	conn := fireflies.GetFirefliesConnector(ctx)
+
+	m, err := conn.ListObjectMetadata(ctx, []string{"users", "transcripts", "bites"})
+
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Print the results
+	fmt.Println("Results: ", m.Result)
```
Cobalt0s, line 23: Standard `Println` does not respect identation well.
Use:
```
utils.DumpJSON(res, os.Stdout)
```
This idents the JSON when printing.

Cobalt0s (commented):
Field metadata should be returned via `map[string]FieldMetadata`. This is possible with Fireflies.

providers/fireflies/types.go
sanjay-a01: Currently this connector supports three objects for getting metadata which is bite, transcript and user.  In that objects, there is no Enum values field.

providers/fireflies/handlers.go
```diff
@@ -0,0 +1,117 @@
...
+
+	return &objectMetadata, nil
+}
+
+func getFieldValueType(field string) common.ValueType {
+	if field == "" {
+		return ""
+	}
+
+	switch field {
```
Cobalt0s, line 107: Just to be safe this could be converted to lowercase strings. I don't think that we do this everywhere. But looking at String/ID case I just thought about this.

Cobalt0s (approved):
I am not sure why `test/fireflies/metadata/response.json` file is considered to be a BIN by github. I am assuming this is because of the file permissions.

sanjay-a01:
To resolve the binary (BIN) display issue for the test/fireflies/metadata/response.json file on GitHub, I changed its encoding format from UTF-16 LE to UTF-8. The JSON file is now shown correctly on GitHub.

PR #1603, opened 2025-04-29
[CON-1667] feat(aws): Connector

Cobalt0s:
# Features

* Defined AWS provider
  * Do we need to create new AuthType?
  * Provider BaseURL is Not Applicable.
  * Module BaseURL acts as a URL template, where region is known at connector init, service domain scoped to each API request.
* Added `AWSClient`. This depends on `aws-sdk-go-v2` package.
* Connector requires metadata: "region", "identityStoreID", "instanceArn".
* Internal `awscore` can be used by `aws` package only. It is intended for making HTTP calls to AWS.
* GetAWSConnector is an example on how to init a connector. It will be used by tests.

# Logos
image


Cobalt0s:
* **#1603** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1603?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1603?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

providers/aws.go
```diff
@@ -0,0 +1,49 @@
...
+				Support: Support{
+					Read:      false,
+					Subscribe: false,
+					Write:     false,
+				},
+			},
+		},
+		Media: &Media{
+			DarkMode: &MediaTypeDarkMode{
+				IconURL: "",
```
laurenzlong, line 40: can you add a TODO to add these logos later? or add them in this PR?

providers/aws/connector.go
```diff
@@ -0,0 +1,78 @@
...
+	// Require authenticated client
+	common.RequireAuthenticatedClient
+	common.RequireModule
+	common.RequireMetadata
+
+	// supported operations
+	components.Reader
+
+	region          string
+	identityStoreID string
```
laurenzlong, line 24: can we actually rename these to: `identityStoreId` and `instanceARN`?

laurenzlong: this is because "ARN" is a real acronym, and "Id" is an abbreviation.

common/awsClient.go
```diff
@@ -0,0 +1,102 @@
...
+// AWS service name into the context, therefore the request cannot be constructed for sending.
+var ErrRequestAWSMissingService = errors.New("AWS request is missing Service name, supplied via context")
+
+type AWSClient struct {
+	client *http.Client
+	cfg    aws.Config
+	region string
+}
+
+func NewAWSClient(ctx context.Context, client *http.Client,
```
laurenzlong, line 34: can you add some comments to explain why AWS is special and requires its own client? I know from reading the Slab doc, but it would be good to have it here a comment also .

laurenzlong (approved):

providers/aws/connector.go
Cobalt0s: Yeah, I was thinking about capitalizing both. I will rename ID to Id too :cry:

providers/aws.go
Cobalt0s: Uploaded icons. Attached screenshot to the PR description.

PR #1604, opened 2025-04-29
[CON-1667] feat(aws): Write

Cobalt0s:
# Changes

* AWS Core package defines object registry structs. 
* Each AWS package maps 
  * `object` to AWS commands we support,
  * `object` to `resource identifier` for request payload or id location in response.
image
* Create/Update injects IdentityStoreId or InstanceARN into payload. This is coming from connector metadata. Each service relies on these identifiers for majority of objects.
* Mock tests capture AWS headers sent and payload which is enhanced using connector-metadata variables and record id in case of update/delete operation.
* Added 2 tests to create/update/delete `User`, `Groups` and `TrustedTokenIssuers`, see testing section below.

# Testing
## IdentityStore
image
image
## SSO-Admin
image

### DIFF
--- a/providers/aws/internal/identitystore/read.go
+++ b/providers/aws/internal/identitystore/read.go
@@ -17,7 +17,9 @@ type readPayload struct {
 func ReadRequest(
 	ctx context.Context, params common.ReadParams, baseURL, identityStoreID string,
 ) (*http.Request, error) {
-	command := core.FormatCommand(ServiceName, ReadObjectCommands, params.ObjectName)
+	objectProps := Registry[params.ObjectName]
+
+	command := core.FormatCommand(ServiceName, objectProps.Commands.Read)
 
 	return core.BuildRequest(ctx, baseURL, ServiceDomain, ServiceSigningName, command, readPayload{
 		ReadPayload:     core.NewReadPayload(params),
--- a/providers/aws/internal/identitystore/write.go
+++ b/providers/aws/internal/identitystore/write.go
@@ -1,0 +1,35 @@
+package identitystore
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/providers/aws/internal/core"
+)
+
+func WriteRequest(
+	ctx context.Context, params common.WriteParams, baseURL, identityStoreID string,
+) (*http.Request, error) {
+	recordData, err := common.RecordDataToMap(params.RecordData)
+	if err != nil {
+		return nil, err
+	}
+
+	recordData["IdentityStoreId"] = identityStoreID
+
+	var command core.Command
+
+	if len(params.RecordId) == 0 {
+		objectProps := Registry[params.ObjectName]
+		command = core.FormatCommand(ServiceName, objectProps.Commands.Create)
+	} else {
+		objectProps := Registry[params.ObjectName]
+		command = core.FormatCommand(ServiceName, objectProps.Commands.Update)
+
+		identifierKey := objectProps.InputRecordID.Update
+		recordData[identifierKey] = params.RecordId
+	}
+
+	return core.BuildRequest(ctx, baseURL, ServiceDomain, ServiceSigningName, command, recordData)
+}
--- a/providers/aws/internal/ssoadmin/delete.go
+++ b/providers/aws/internal/ssoadmin/delete.go
@@ -1,0 +1,25 @@
+package ssoadmin
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/providers/aws/internal/core"
+)
+
+func DeleteRequest(
+	ctx context.Context, params common.DeleteParams, baseURL, instanceArn string,
+) (*http.Request, error) {
+	payload := map[string]any{
+		"InstanceArn": instanceArn,
+	}
+
+	objectProps := Registry[params.ObjectName]
+	identifierKey := objectProps.InputRecordID.Delete
+	payload[identifierKey] = params.RecordId
+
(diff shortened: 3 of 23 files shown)


Cobalt0s:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1604?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1604** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1604?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1604?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1603** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1603?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

providers/aws/internal/identitystore/read.go
```diff
@@ -17,7 +17,9 @@ type readPayload struct {
 func ReadRequest(
 	ctx context.Context, params common.ReadParams, baseURL, identityStoreID string,
 ) (*http.Request, error) {
-	command := core.FormatCommand(ServiceName, ReadObjectCommands, params.ObjectName)
+	objectProps := Registry[params.ObjectName]
```
eberle1080, line 20: You need to check to see if this key exists. If it doesn't exist, the value will be nil, and the next line will panic. And our users frequently get object names wrong.

providers/aws/internal/identitystore/write.go
```diff
@@ -0,0 +1,35 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	recordData["IdentityStoreId"] = identityStoreID
+
+	var command core.Command
+
+	if len(params.RecordId) == 0 {
+		objectProps := Registry[params.ObjectName]
```
eberle1080, line 24: Same note about checking that the key exists

providers/aws/internal/identitystore/write.go
```diff
@@ -0,0 +1,35 @@
...
+		command = core.FormatCommand(ServiceName, objectProps.Commands.Create)
+	} else {
+		objectProps := Registry[params.ObjectName]
```
eberle1080, line 27: Same note about checking that the key exists

providers/aws/internal/ssoadmin/delete.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func DeleteRequest(
+	ctx context.Context, params common.DeleteParams, baseURL, instanceArn string,
+) (*http.Request, error) {
+	payload := map[string]any{
+		"InstanceArn": instanceArn,
+	}
+
+	objectProps := Registry[params.ObjectName]
```
eberle1080, line 18: Same note about checking that the key exists

providers/aws/internal/ssoadmin/read.go
```diff
@@ -17,7 +17,9 @@ type readPayload struct {
 func ReadRequest(
 	ctx context.Context, params common.ReadParams, baseURL, instanceArn string,
 ) (*http.Request, error) {
-	command := core.FormatCommand(ServiceName, ReadObjectCommands, params.ObjectName)
+	objectProps := Registry[params.ObjectName]
```
eberle1080, line 20: Same note about checking that the key exists

providers/aws/internal/ssoadmin/write.go
```diff
@@ -0,0 +1,35 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	recordData["InstanceArn"] = instanceArn
+
+	var command core.Command
+
+	if len(params.RecordId) == 0 {
+		objectProps := Registry[params.ObjectName]
```
eberle1080, line 24: Same note about checking that the key exists

providers/aws/internal/ssoadmin/write.go
```diff
@@ -0,0 +1,35 @@
...
+		command = core.FormatCommand(ServiceName, objectProps.Commands.Create)
+	} else {
+		objectProps := Registry[params.ObjectName]
```
eberle1080, line 27: Same note about checking that the key exists

eberle1080 (approved):
looks good mostly, however there's a recurring bug having to do with checking the key for a map before using it. If you merge this right now it'll lead to a panic. We have `recover()` logic but it alerts whoever is on-call, and I'd prefer not to freak them out. So please fix those bugs, and then feel free to merge.

providers/aws/internal/identitystore/read.go
Cobalt0s: Initially I had the "protection" and checked the presence. Then I remembered that we use overall guarding mechanism (true for Read/Write/Delete operations): https://github.com/amp-labs/connectors/blob/7cba7915cda900dbcda16548373dca7717c17caa/internal/components/reader/reader.go#L47
The list of supported objects is supplied here:
https://github.com/amp-labs/connectors/blob/7cba7915cda900dbcda16548373dca7717c17caa/providers/aws/supports.go#L14-L23

Maybe we should do someting to make it explicit. For example make a copy of `common.ReadParams` which would signify validation passed and that ObjectName is valid within that connector domain. This could give special meaning to this **type**"
https://github.com/amp-labs/connectors/blob/7cba7915cda900dbcda16548373dca7717c17caa/common/types.go#L481 


Side note: The registry is a map where the value cannot be a pointer: `type Registry datautils.Map[string, ObjectProperties]`.
When the value is not present it will return instance of `ObjectProperties` with blank fields. So we would be safe from panic. The worst case scenario -- the eventual overall error returned would be misguiding of the true problem.

PR #1611, opened 2025-04-30
Read fireflies connector

sanjay-a01:
**Note:** Attached the sample response in the separate json file in this commit.

image

The response struct was defined with an errors field because the transcripts endpoint includes fields like audio_url and video_url, which require a paid plan to access. Without a paid plan, the API returns a response with these fields empty and includes an error message. This is why the response struct was implemented as shown above.

image


providers/fireflies/handlers.go
```diff
@@ -82,3 +82,289 @@ func (c *Connector) parseSingleObjectMetadataResponse(
...
+		skip  = 0
+		limit int
+		query string
+	)
+
+	if params.NextPage != "" {
+		// Parse the page number from NextPage
+		var pageNum int
+
+		_, err := fmt.Sscanf(string(params.NextPage), "%d", &pageNum)
```
Cobalt0s, line 102: Alternative:
```
strconv.Atoi(config.NextPage.String())
```
Or
```
strconv.ParseInt(config.NextPage.String(), 10, 64)
```

providers/fireflies/handlers.go
```diff
@@ -82,3 +82,289 @@ func (c *Connector) parseSingleObjectMetadataResponse(
...
+		}
+
+		skip = pageNum
+	}
+
+	limit = defaultPageSize
+
+	switch params.ObjectName {
+	case transcriptsObjectName, bitesObjectName:
+		query = getQueryForPaginationedObject(params.ObjectName, skip, limit)
```
Cobalt0s, line 114: Typo: Pagination or Paginated.

providers/fireflies/handlers.go
```diff
@@ -82,3 +82,289 @@ func (c *Connector) parseSingleObjectMetadataResponse(
...
+
+func getBiteQery(limit, skip int) string {
+	var arguments string
+	if skip == 0 {
+		arguments = fmt.Sprintf("(mine: false, my_team: true, limit: %d)", limit)
+	} else {
+		arguments = fmt.Sprintf("(mine: false, my_team: true, limit: %d, skip: %d)", limit, skip)
+	}
+
+	return fmt.Sprintf(`query {
```
Cobalt0s, line 266: The queryies are good.
Maybe it is worth having a directory of files each having a query for some object. They look significant so could have a dedicated place. Up to you.
The files could use golang `template` varriables `{{.var}}` and also it has support for conditionals that could be utilised for `skip` query above. Thats just an idea, could be overengineering on my side.

providers/fireflies/handlers.go
```diff
@@ -82,3 +82,289 @@ func (c *Connector) parseSingleObjectMetadataResponse(
...
+	data, err := common.UnmarshalJSON[Response](resp)
+	if err != nil {
+		return nil, common.ErrFailedToUnmarshalBody
+	}
+
+	var records []any
+
+	switch params.ObjectName {
+	case usersObjectName:
+		if len(data.Data.Users) == 0 {
```
Cobalt0s, line 327: Each and every `case` block is concerned with choosing `data.Data.Users` or `data.Data.Transcripts` or `data.Data.Bites`. The rest could be refactored and be moved after the switch operating on the selected array of records.

The type Users, Transcripts and Bites is not important for the code flow because it is converted immidietly to var records []any.

I know that Monday is similar in that regard, those changes would apply to Monday too (you don't need to change them, just telling you that I am aware of similar connectors having this "issue").

providers/fireflies/types.go
```diff
@@ -35,3 +36,153 @@ type MetadataResponse struct {
...
 	} `json:"data"`
 }
+
+type Response struct {
+	Errors any          `json:"errors"`
+	Data   ResponseData `json:"data"`
+}
+
+type ResponseData struct {
+	Users       []User       `json:"users,omitempty"`
```
Cobalt0s, line 46: Type could be `[]map[string]any` (same for transcripts and bites). Which arguably could be better because during JSON parsing will grab each field instead of those limited to the struct.

providers/fireflies/utils.go
```diff
@@ -0,0 +1,81 @@
...
+func getRecords(objectName string) func(*ajson.Node) ([]map[string]any, error) {
+	return func(node *ajson.Node) ([]map[string]any, error) {
+		// First get the data object
+		dataNode, err := node.GetKey("data")
+		if err != nil {
+			return nil, err
+		}
+
+		// Then get the array under the object name (e.g., "boards" or "users")
+		records, err := jsonquery.New(dataNode).ArrayOptional(objectName)
```
Cobalt0s, line 36: Equivalent: variadic arguments accept nested path:
```
records, err := jsonquery.New(node, "data").ArrayOptional(objectName)
```

providers/fireflies/utils.go
```diff
@@ -0,0 +1,81 @@
...
+			return nil, err
+		}
+
+		return jsonquery.Convertor.ArrayToMap(records)
+	}
+}
+
+func makeNextRecordsURL(params common.ReadParams, count int) func(*ajson.Node) (string, error) {
+	return func(node *ajson.Node) (string, error) {
+		if supportLimitAndSkip.Has(params.ObjectName) {
```
Cobalt0s, line 47: The if condition is larger then statements following it. This means you should reverse the if statement to shrink idents.
```
if !supportLimitAndSkip.Has(params.ObjectName) {
  return "", nil
}

... everything else follows.
```

providers/fireflies/utils.go
```diff
@@ -0,0 +1,81 @@
...
+			if count < defaultPageSize {
+				return "", nil
+			}
+
+			var currentPage int
+			if params.NextPage != "" {
+				_, err := fmt.Sscanf(string(params.NextPage), "%d", &currentPage)
```
Cobalt0s, line 54: Ok, looks like the reason to use Sscanf is if you need sophisticated "format" when extracting data. 
See post:
https://golang.cafe/blog/golang-string-to-int-conversion-example.html

Cobalt0s (approved):
Please address some comments. After that it is good to merge.

providers/fireflies/handlers.go
sanjay-a01: image

image

Is the above implementation is correct?

providers/fireflies/handlers.go
```diff
@@ -115,3 +116,263 @@ func getFieldValueType(field string) common.ValueType {
...
+						prompt
+						response
+					}
+				}
+				meeting_link
+			}
+	}`, paginationParams)
+}
+
+func getBiteQery(limit, skip int) string {
```
Cobalt0s, line 279: Query.

providers/fireflies/handlers.go
Cobalt0s: That sounds right!

Of course if the query cannot be built then it should error out and surface it.
Maybe even log to indicate something is wrong, because that should never happen under normal circumstances.

providers/fireflies/utils.go
```diff
@@ -0,0 +1,80 @@
...
+	transcriptsObjectName = "transcripts"
+	bitesObjectName       = "bites"
+)
+
+var supportLimitAndSkip = datautils.NewSet( //nolint:gochecknoglobals
+	transcriptsObjectName,
+	bitesObjectName,
+)
+
+func getRecords(objectName string) func(*ajson.Node) ([]map[string]any, error) {
```
Cobalt0s, line 27:
```
common.ExtractRecordsFromPath("data", objectName)
```

providers/fireflies/handlers.go
```diff
@@ -3,16 +3,22 @@ package fireflies
...
+	"strconv"
 	"strings"
 
 	"github.com/amp-labs/connectors/common"
 	"github.com/amp-labs/connectors/common/naming"
 	"github.com/amp-labs/connectors/common/urlbuilder"
 )
 
+//go:embed *.graphql
+var queryFS embed.FS
```
Cobalt0s, line 20: This is interesting.

Thanks for using the embed. Relative loading of path is only ok for local testing, etc. When the library is used it shouldn't assume the file paths. I like the idea of collective embeding! Very nice.

Cobalt0s (approved):
Interesting and cool!

RajatPawar (approved):
minor change, but good to go!

PR #1615, opened 2025-05-01
[ENG-2192] feat: update catalog to define required metadata

RajatPawar:
* Updates all catalogs to state the metadata that the connector needs. Most connectors need a workspace, so this PR adds that in as well.
  * **Exceptions:**
    * Hubspot ownerId: fetched in the server due to some limitations
    * AWS serverDomain: this is a variable that needs to be figured out at runtime based on the kind of request being made. It is not really metadata that is an input to the connector.
* Removes all `PostAuthInfoNeeded: false` because that is assumed
* For all `PostAuthInfoNeeded: true`, defines equivalent `Metadata.PostAuthentication` that we can start using in the server, but does not remove it in this PR.

### DIFF
--- a/providers/atlassian.go
+++ b/providers/atlassian.go
@@ -68,5 +68,18 @@ func init() {
 			Subscribe: false,
 			Write:     true,
 		},
+		Metadata: &ProviderMetadata{
+			PostAuthentication: []MetadataItemPostAuthentication{
+				{
+					Name: "cloudId",
+				},
+			},
+			Input: []MetadataItemInput{
+				{
+					Name:        "workspace",
+					DisplayName: "Site URL (Your domain)",
+				},
+			},
+		},
 	})
 }
--- a/providers/atlassian.go
+++ b/providers/atlassian.go
@@ -11,6 +11,7 @@ const (
 	ModuleAtlassianJiraConnect common.ModuleID = "atlassian-connect"
 )
 
+// nolint:funlen
 func init() {
 	// Atlassian Configuration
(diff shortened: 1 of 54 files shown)


providers/atlassian.go
```diff
@@ -68,5 +69,18 @@ func init() {
...
+		Metadata: &ProviderMetadata{
+			PostAuthentication: []MetadataItemPostAuthentication{
+				{
+					Name: "cloudId",
+				},
+			},
+			Input: []MetadataItemInput{
+				{
+					Name:        "workspace",
+					DisplayName: "Site URL (Your domain)",
```
Cobalt0s, line 81: This is great!

providers/aws.go
```diff
@@ -45,5 +45,20 @@ func init() {
...
+				{
+					Name: "region",
+				},
+				{
+					Name: "identityStoreId",
+				},
+				{
+					Name: "instanceArn",
+				},
+				// IMPORTANT: The 'serviceDomain' variable is figured out in the connector,
```
Cobalt0s, line 59: `serviceDomain` is part of "URL Template".

Maybe we should split values which are "catalog variables" -- true connector arguments for init.
And those that are templates for the implementation to plug in and out.

Cobalt0s (approved):

providers/aws.go
RajatPawar: I agree, this is not really metadata, but we'll come back to this when this goes live - this is definitely more of an internal state

PR #1625, opened 2025-05-05
chore: Add health check endpoints for hubspot and salesforce

eberle1080:
This just adds a few endpoints for hubspot and salesforce so that we can check the health of credentials


RajatPawar (approved):
Looks good!

cc @Cobalt0s @sanjay-a01 @Jkarage @immdipu FYI

PR #1634, opened 2025-05-06
[CON-1618] fix(catalog): Fallbacak on missing module

Cobalt0s:


Cobalt0s:
* **#1634** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1634?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1634?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

providers/utils.go
```diff
@@ -167,25 +167,31 @@ func (i *ProviderInfo) GetOption(key string) (string, bool) {
 	return val, ok
 }
 
-func (i *ProviderInfo) ReadModuleInfo(moduleID common.ModuleID) (ModuleInfo, error) {
+func (i *ProviderInfo) ReadModuleInfo(moduleID common.ModuleID) (*ModuleInfo, error) {
```
RajatPawar, line 170: On more thought, let's default to the root module even if `i.Modules` is empty, and log it instead. I'll set up a alert based on the log line to alert us, and we'll know if the case is being hit. If we don't receive an alert for a couple of weeks, we can then modify this logic to start returning errors on it.

providers/utils_test.go
```diff
@@ -189,3 +184,498 @@ func TestReadInfo(t *testing.T) { // nolint:funlen
...
+
+	tests := []struct {
+		name        string
+		input       inType
+		expected    *ModuleInfo
+		expectedErr error
+	}{
+		// Root for providers that have one module.
+		{
+			name: "Dynamics root module",
```
RajatPawar, line 205: Can you please add test cases for -
1. Providers that have modules, but no module input?
2. Providers that have no modules, and no module input?
3. Providers with no modules with a module input

providers/utils_test.go
```diff
@@ -189,3 +184,498 @@ func TestReadInfo(t *testing.T) { // nolint:funlen
...
+				Support: Support{
+					Proxy: true,
+					Read:  true,
+					Write: true,
+				},
+			},
+			expectedErr: nil,
+		},
+		{
+			name: "Atlassian root module",
```
RajatPawar, line 256: seems like this case is repeated

RajatPawar (approved):
Good to go once all the comments have been resolved.

providers/utils.go
Cobalt0s: Thats a great idea!

providers/utils_test.go
Cobalt0s: I think the third is covered by several cases. I will rename `providers with one module` to `providers without module`. I kind of thought that by default there is always 1 module hence the name.

1-module vs multi-module.

RajatPawar (approved):

PR #1642, opened 2025-05-07
[CON-1603] feat: Added read functionality for instantly v2 connector

sanjay-a01:
### Sample Response

**For api-keys object**
image

**For custom-tags** object
image

**For lead-lists** object
image

### TestCase Result
image


providers/instantlyai/handlers.go
```diff
@@ -0,0 +1,66 @@
...
+	"context"
+	"net/http"
+	"strconv"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+)
+
+func (c *Connector) buildReadRequest(ctx context.Context, params common.ReadParams) (*http.Request, error) {
+	var (
```
Cobalt0s, line 14: This variable declaration is redundant — you can simplify it by using the := shorthand:
```
url, err := urlbuilder.New(c.ProviderInfo().BaseURL, apiVersion, params.ObjectName)
```

providers/instantlyai/handlers.go
```diff
@@ -0,0 +1,66 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if postEndpointsOfRead.Has(params.ObjectName) {
+		return http.NewRequestWithContext(ctx, http.MethodPost, url.String(), bytes.NewReader([]byte("{}")))
+	}
+
+	return http.NewRequestWithContext(ctx, http.MethodGet, url.String(), nil)
```
Cobalt0s, line 40: Could we support incremental reading, or does the API not expose any time-based fields for querying?

providers/instantlyai/handlers.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func (c *Connector) parseReadResponse(
+	ctx context.Context,
+	params common.ReadParams,
+	request *http.Request,
+	response *common.JSONHTTPResponse,
+) (*common.ReadResult, error) {
+	if directResponseEndpoints.Has(params.ObjectName) {
```
Cobalt0s, line 49: The core difference in the if statement is whether `common.ExtractRecordsFromPath` uses `""` or `"items"`. Everything else remains the same. This could be refactored to more clearly reflect that some objects return data directly at the JSON root, while others nest their data under "items".

providers/instantlyai/utils.go
```diff
@@ -0,0 +1,49 @@
...
+		pagination, err := jsonquery.New(node).StringRequired("next_starting_after")
+		if err != nil {
+			return "", err
+		}
+
+		if pagination != "" {
+			nextLink := *reqLink
+			query := nextLink.Query()
+			query.Set("starting_after", pagination)
+			nextLink.RawQuery = query.Encode()
```
Cobalt0s, line 42: Query params are a bit hard to manage via `url.URL`. Maybe use `urlbuilder.URL`.
I think it would be nice later to change components such that it would support `urlbuilder.URL` natively which is a wrapper of `url.URL` with focus on query params and path building.

providers/instantlyai/read_test.go
```diff
@@ -0,0 +1,128 @@
...
+				Data: []common.ReadResultRow{
+					{
+						Fields: map[string]any{},
+						Raw: map[string]any{
+							"id":   "01966b52-73b2-7f5f-b87f-2799588cf4b5",
+							"name": "My API Key",
+							"scopes": []any{
+								"campaigns:create",
+							},
+							"key":               "a1b2c3d4e5f6g7h8i9j0",
```
Cobalt0s, line 46: If the keys are valid and not just random data, this could be problematic. I’d recommend avoiding including API keys in tests.

providers/instantlyai/read_test.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+func TestRead(t *testing.T) { // nolint:funlen,gocognit,cyclop
+	t.Parallel()
+
+	apiKeysResponse := testutils.DataFromFile(t, "api-keys.json")
+	customTagsResponse := testutils.DataFromFile(t, "custom-tags.json")
+	leadListsResponse := testutils.DataFromFile(t, "lead-lists.json")
+
+	tests := []testroutines.Read{
```
Cobalt0s, line 21: Could you add one test which will check the error format you specified in the `errors.go`?
```
{
  "message": "...",
  "error": "...",
  "statusCode": "..."
}
```

Cobalt0s (approved):
Several comments. But looks good.

providers/instantlyai/read_test.go
Cobalt0s: I see you did this for the delete functionality in the next PR, effectively it is tested, so no longer needed.

providers/instantlyai/handlers.go
sanjay-a01: No Incremental reading is supported for two endpoints: `campaigns/analytics/daily` and `campaigns/analytics/steps`. I have implemented the logic accordingly in the code.

providers/instantlyai/utils.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+func makeNextRecordsURL(reqLink *url.URL, objName string) common.NextPageFunc {
+	return func(node *ajson.Node) (string, error) {
+		url, err := urlbuilder.FromRawURL(reqLink)
+		if err != nil {
+			return "", err
+		}
+
+		if directResponseEndpoints.Has(objName) {
```
Cobalt0s, line 39: This if could be the first check in the function to avoid URL parsing.

Cobalt0s (approved):
Cool!

PR #1643, opened 2025-05-07
[CON-1604] feat: Added write functionality for the instantly V2 connector

sanjay-a01:
### Sample Response

**For lead-labels**
image

**For custom-tags**
image

**For lead-lists**
image

### Testcase result
image


providers/instantlyai/handlers.go
```diff
@@ -64,3 +67,96 @@ func (c *Connector) parseReadResponse(
 		params.Fields,
 	)
 }
+
+func (c *Connector) buildWriteRequest(ctx context.Context, params common.WriteParams) (*http.Request, error) {
+	var (
```
Cobalt0s, line 72: As in previous PRs, the var statement isn’t required for the function flow. While there are cases where var is necessary, using `:=` here is more concise and avoids unnecessary verbosity — unlike in languages like JavaScript, Java, or C++.

`method = http.MethodPost` would be the only exception. But it should be as close to the code block were it is used as possible. Variable proximity.

providers/instantlyai/handlers.go
```diff
@@ -64,3 +67,96 @@ func (c *Connector) parseReadResponse(
...
+	}, nil
+}
+
+func (c *Connector) buildDeleteRequest(ctx context.Context, params common.DeleteParams) (*http.Request, error) {
+	url, err := urlbuilder.New(c.ProviderInfo().BaseURL, apiVersion, params.ObjectName, params.RecordId)
+	if err != nil {
+		return nil, err
+	}
+
+	body := []byte(`null`)
```
Cobalt0s, line 134: Just like you explained in the PR description about the need to pass "null" as a string in the payload, it would be helpful to mention that here in the code as well. If the API documentation specifies this behavior, adding a reference link would be great — though that’s a nice-to-have.

providers/instantlyai/write_test.go
```diff
@@ -0,0 +1,110 @@
...
+			Name:         "Write object must be included",
+			Server:       mockserver.Dummy(),
+			ExpectedErrs: []error{common.ErrMissingObjects},
+		},
+		{
+			Name:  "creating the custom tags",
+			Input: common.WriteParams{ObjectName: "custom-tags", RecordData: "dummy"},
+			Server: mockserver.Conditional{
+				Setup: mockserver.ContentJSON(),
+				If:    mockcond.MethodPOST(),
```
Cobalt0s, line 32: Include `PathSuffix` check.

providers/instantlyai/write_test.go
```diff
@@ -0,0 +1,110 @@
...
+		{
+			Name: "Update custom tags as PATCH",
+			Input: common.WriteParams{
+				ObjectName: "custom-tags",
+				RecordId:   "0196837f-d659-7496-854c-84187bb6f708",
+				RecordData: "dummy",
+			},
+			Server: mockserver.Conditional{
+				Setup: mockserver.ContentJSON(),
+				If:    mockcond.MethodPATCH(),
```
Cobalt0s, line 59: Include `PathSuffix` check.

providers/instantlyai/delete_test.go
```diff
@@ -0,0 +1,80 @@
...
+			},
+		},
+		{
+			Name:  "Successful delete",
+			Input: common.DeleteParams{ObjectName: "custom-tags", RecordId: "8cba2d22-cdd0-453f-ac69-4c43f71942e5"},
+			Server: mockserver.Conditional{
+				Setup: mockserver.ContentJSON(),
+				If: mockcond.And{
+					mockcond.PathSuffix("/v2/custom-tags/8cba2d22-cdd0-453f-ac69-4c43f71942e5"),
+					mockcond.MethodDELETE(),
```
Cobalt0s, line 48: Since you mentioned that DELETE expects either `"null"` in the body or `{}`, it would be good to include this case in the test.

The appropriate checks to use are:
```
mockcond.Body
mockcond.BodyBytes
```

Explanation: each of these checks compares two things:
1. An exact string match of the request body. (relevant for raw text/data)
2. A parsed JSON comparison that checks for structural equality — which is often more reliable than raw string matching, as it handles differences in field order and formatting. (but relevant for JSON data)

If either check passes, the body is considered a match.

test/instantlyai/write/write.go
```diff
@@ -0,0 +1,205 @@
...
+)
+
+func main() {
+	os.Exit(MainFn())
+}
+
+func MainFn() int {
+	ctx := context.Background()
+
+	err := testLeadLabel(ctx)
```
Cobalt0s, line 22: These tests look good.

Just as an FYI — I recently refactored some utility methods that consolidate the flow for Create, Update, and Delete operations. They're not required here, but worth keeping in mind for future tests where they might fit.

You can check them out here:
https://github.com/amp-labs/connectors/blob/0df78df77f91fbd84eb441cc39164692d0153b15/test/dynamicscrm/write-delete/leads/main.go#L39

Cobalt0s (approved):
Well done!

Cobalt0s (approved):
Great! Good to go.

PR #1651, opened 2025-05-12
[CON-1309] feat(dynamicsbusiness): Read

Cobalt0s:
# Details
Incremental reads rely on the `lastModifiedDateTime` field.

For objects that lack this field, we send a probing request. If filtering by `lastModifiedDateTime` returns a `400 Bad Request`, we mark the object as unsupported for incremental reads.

To avoid repeated API calls, this result is cached in a concurrency-safe map. This map is defined as utility under `datautils` package.

# Testing
## Customers
image
## Balance Sheets
image


providers/dynamicsbusiness/requests.go
```diff
@@ -21,3 +27,124 @@ func (c *Connector) buildSingleObjectMetadataRequest(ctx context.Context, object
 
 	return http.NewRequestWithContext(ctx, http.MethodGet, url.String(), nil)
 }
+
+const defaultPageSize = 100
+
+func (c *Connector) buildReadRequest(ctx context.Context, params common.ReadParams) (*http.Request, error) {
```
jlimatampersand, line 33: I think it would be great to have API reference in comment.

providers/dynamicsbusiness/requests.go
```diff
@@ -21,3 +27,124 @@ func (c *Connector) buildSingleObjectMetadataRequest(ctx context.Context, object
...
+	req.Header.Set("Prefer", fmt.Sprintf("odata.maxpagesize=%v", defaultPageSize))
+
+	return req, nil
+}
+
+// buildReadURL constructs a URL for a read operation.
+// If params.NextPage is set, it returns the URL for the next page.
+// Otherwise, it builds a fresh URL for the initial request, adding field selection
+// and an optional incremental filter if applicable.
+func (c *Connector) buildReadURL(ctx context.Context, params common.ReadParams) (*urlbuilder.URL, error) {
```
jlimatampersand, line 53: Same here, API reference in comment.

providers/dynamicsbusiness/requests.go
```diff
@@ -21,3 +27,124 @@ func (c *Connector) buildSingleObjectMetadataRequest(ctx context.Context, object
...
+		c.incrementalRegistry.Set(params.ObjectName, isIncremental)
+	}
+
+	return isIncremental, nil
+}
+
+// fetchIsIncrementObject makes a probe request to check if the object supports time-based filtering.
+// This is determined by requesting one item and checking if the response is successful.
+// Errors in the 5xx range indicate temporary failures; 4xx indicates lack of support.
+func (c *Connector) fetchIsIncrementObject(ctx context.Context, params common.ReadParams) (bool, error) {
```
jlimatampersand, line 114: I don't quite follow why we need this logic here. Can we just list out the objects that are incremental? Could there possibly be smarter way to do this rather than making 1 more API call?

How do objects differ in supporting incrementals? Do we have documentation explaining some behaviors that led you to decide that we need this logic here?

jlimatampersand, line 114: it seems like this is just checking if it is incremental. So we can rename to something like `fetchObjectAndCheckIncrementalSupport`

providers/dynamicsbusiness/requests.go
```diff
@@ -21,3 +27,124 @@ func (c *Connector) buildSingleObjectMetadataRequest(ctx context.Context, object
...
+func applyIncrementalQuery(url *urlbuilder.URL, since time.Time) {
+	sinceValue := datautils.Time.FormatRFC3339inUTCWithMilliseconds(since)
+	url.WithQueryParam("$filter", fmt.Sprintf("lastModifiedDateTime ge %v", sinceValue))
+}
+
+// isIncrementObject determines whether the given object supports incremental (time-based) reads.
+//
+// First checks the local registry cache. If no entry is found, it performs a one-time probe via API,
+// then caches the result. Assumes provider behavior is stable and does not change support dynamically.
+func (c *Connector) isIncrementObject(ctx context.Context, params common.ReadParams) (bool, error) {
```
jlimatampersand, line 95: I am confused about this logic. Can you explain

jlimatampersand: What is the technical reason and behavior you are assuming to do this caching logic?

jlimatampersand (changes requested):

providers/dynamicsbusiness/requests.go
Cobalt0s: Described behaviour in buildReadRequest. Also Microsoft doesn't have good documentation to describe one functionality. Some of it is due to experience of working with MS and similarity with DynamicsCRM

providers/dynamicsbusiness/requests.go
Cobalt0s: As one of the solutions we could request metadata for our instance and hard code the objects that have `lastModifiedDateTime` field. This ignores custom objects that a client may have with `lastModifiedDateTime` field.

The real reason why I went down this path is making a query on a field which doesn't exist will return an error. So the API assumes that we know `lastModifiedDateTime` exists beforehand.

Therefore, I am doing a probing request to see if it succeeds. If so then we can proceed with incremental reading. Otherwise, we cannot filter by the `params.Since` argument.

Cobalt0s: We are not interested in the object and to the caller of the method it doesn't matter if it fetches the object. But it matters that it `fetches`. Fetch implies costly operation, so I wanted to preserve that intent. `IsIncrementalObject` is the ultimate check and main purpose of the function.

providers/dynamicsbusiness/requests.go
Cobalt0s: My assumption is if the connector performed read for object `customers` using `params.Since` once, then all consequent calls can assume that they can do it. And therefore no need to waste 1 more API call to check if the object is incremental. We did it once, we should keep note of it.

This implementation could be removed entirely, and if there is performence repercussions in the future we can address it later. Of course if the connector is created to perform Read only once, then this caching will never be used.

PR #1656, opened 2025-05-15
fix: Gracefully handle unsupported objects via Errors map

Jkarage:
In scenarios where an object is not supported (i.e., it does not have a static schema), we currently fail the entire call. This PR adds the error to the Errors map.
This change will enable future flexibility—for example, if we implement a fallback schema provider, we can later use it for the failed objects.


Cobalt0s (approved):
This is a very good change!
Majority of conenctors didn't utilize the Error field correctly. This approach existed for a long time.

PR #1660, opened 2025-05-16
fix: Resolve semgrep findings

Jkarage:


test/gong/write/main.go
```diff
@@ -82,10 +89,23 @@ func createCalls(ctx context.Context, conn *gong.Connector, payload *CallsPayloa
 	return res
 }
 
-func createUniqueID() string {
-	minV := 1
-	maxV := 10000
-	uniqueID := strconv.Itoa(rand.Intn(maxV-minV+1) + minV)
+func uniqueId(x, y int) (int, error) {
```
Jkarage, line 92: updated the random id generator to use the crypto/rand instead of math/rand.

test/gong/write/main.go
```diff
@@ -82,10 +89,23 @@ func createCalls(ctx context.Context, conn *gong.Connector, payload *CallsPayloa
...
+	}
+
+	rangeSize := y - x
+
+	// Read random bytes (8 bytes for 64-bit value)
+	var buf [8]byte
+	_, err := rand.Read(buf[:])
+	if err != nil {
+		return 0, err
+	}
```
eberle1080, line 104: This works but for a test it feels like overkill. I don't really see a problem with a PRNG. Your call, the work is already done so it's fine, but this isn't for production.

eberle1080 (approved):

PR #1672, opened 2025-05-21
feat: Add incremental read contacts Front Connector

Jkarage:
With this we can support incremental reads for contacts object in Front Connector.
Updated the pageSize as well (it was 1 by mistake)

Tests:
Screenshot 2025-05-21 at 14 49 14


Cobalt0s (approved):
Good!

As a follow-up PR, could you add support for `ReadParams.Until`? You have fresh context on this connector, so it would be a good opportunity to include it.

PR #1690, opened 2025-05-26
chore(tests): Use full URL path

Cobalt0s:
# Goal

Validation of explicit `full` URL paths.
This should catch most issues with the growing number of BaseURLs due to modules.

# Commits

1. Replace PathSuffix check with tests verifying full path constructed by deep connector. (Path is everything that comes after the URL origin.)
2. Each connector used by the mock tests replaces the base URL with target mock server. This is wrong because some BaseURLs may include extra path. I would rather the test be as explicit as possible. Ex: 
image
(therefore, the test case must specify the path starting with `/api`)
3. Each test case which didn't specify the full URL path and only the last URI parts is updated

# Note

Existing mock tests do not validate origin. This has to be done in addition to the usual tests.
Example: AWS already has a test specificly to validate the domain is handled by deep connector.


Cobalt0s:
* **#1690** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1690?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1690?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

RajatPawar (approved):

PR #1691, opened 2025-05-26
tidy(zoom): Combine modules

Cobalt0s:


Cobalt0s:
* **#1691** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1691?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1691?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

providers/zoom/metadata/schemas.json
```diff
@@ -56,6 +56,17 @@
             "type": "type"
           }
         },
+        "contacts_groups": {
```
RajatPawar, line 59: it seems that object names are also changed, can you please retain the original? i.e. in this case it should be `contacts/groups`

I believe at the time of writing the zoom connector, the server may not have supported objects with special characters, but now we do so we need to fix this. can you please also do the same for any other connectors? Thank you.

Cobalt0s: Turns out there was no mistake and it is consistent with what we had. But I understand that objects even in the former version didn't follow our "new approach". 

Here is a dedicated [follow up PR](https://github.com/amp-labs/connectors/pull/1697).

PR #1704, opened 2025-05-29
[CON-1605] feat: Add initialization to connector for instantlyai.

sanjay-a01:


RajatPawar (approved):
@sanjay-a01 in the future, for PRs like this which only add initialization logic, feel free to get a review from Joseph / Constantin / Dipu to unblock faster

PR #1707, opened 2025-05-30
fix: Backfill for Reading Leads Marketo Connector

Jkarage:
-  Removes the Sequential numeric IDs Logic on Leads Ids
-  Uses the activities next page token when reading Leads.
-  I have tested this by deleting lead ids from 5..300
-  Reading works fine the returned Leads are (1,2,3,4, 301...619)

Screenshot 2025-05-30 at 11 59 06

- I have tested Reading Incremental Leads as well (reading with timestamp other than 1970)


providers/marketo/types.go
```diff
@@ -23,7 +23,7 @@ var (
 	marketoGUIDResponseObjects = []string{"namedAccountLists", "namedaccounts", "opportunities"}
 
 	// idFilteringObjects represents objects that uses id as filtering values in read connector.
-	idFilteringObjects = []string{"leads", "salespersons", "companies"}
```
RajatPawar, line 26: post this PR, can you please evaluate if salespersons and companies also face the same issue?

RajatPawar (approved):

providers/marketo/types.go
Jkarage: Noted.

PR #1708, opened 2025-05-30
[CON-1719] feat: Add Clari Copilot Connector

immdipu:
# Conventions

## Notes
Currently, the `AuthType` is set to `ApiKey`, but since this is a special case where two headers are required for authentication, the existing `ApiKey` AuthType does not work.

As discussed with @eberle1080 , we need a separate `AuthType` to handle this use case. For testing purposes, I have created and added a new `AuthType`(in a separate branch) to the connector. The changes can be viewed in this commit: [d8a4d33](https://github.com/amp-labs/connectors/commit/d8a4d33555a801019d4fa526d14c892569342aac).

We may need to finalize and add the new `AuthType` before merging this PR.

### GET
URL: http://localhost:4444/calls

image

---
### POST
URL: http://localhost:4444/create-contact

image

---
### PUT
URL: http://localhost:4444/update-contact
image

## Pagination
The pagination in Clari Copilot works by using `limit` and `skip`. `limit` represents the number of records to return, and `skip` represents the number of records to skip. The response contains a `pagination` object, which includes `hasMore` (indicating whether another page is available) and `nextPageSkip` (the number of records to skip in the next request).

First page has `skip` 0 and `limit` 2
image

Second page has `skip` and `limit` 2
image

### Errors

Invalid payload
image

invalid url path
image


providers/claricopilot/httpClient.go
```diff
@@ -0,0 +1,28 @@
...
+	"context"
+
+	"github.com/amp-labs/connectors/common"
+)
+
+// NewClariCopilotAuthHTTPClient returns a new http client with
+// dual header authentication.
+func NewClariCopilotAuthHTTPClient( //nolint:ireturn
+	ctx context.Context,
+	header1Name, header1Value, header2Name, header2Value string,
```
eberle1080, line 13: I think since this is for a single connector, you can safely assume the header names and just had-code them. That will make calling this more obvious.

immdipu: sure i'll update it.

providers/clariCopilot.go
```diff
@@ -0,0 +1,40 @@
+package providers
+
+const ClariCopilot Provider = "clariCopilot"
+
+func init() {
+	SetInfo(ClariCopilot, ProviderInfo{
+		DisplayName: "ClariCopilot",
```
RajatPawar, line 7:
```suggestion
		DisplayName: "Clari Copilot",
```

RajatPawar (commented):
the checklist i was referring to: https://github.com/amp-labs/connectors/blob/main/.github/PULL_REQUEST_TEMPLATE/new_connector.md

looking at it, it's making me realize that this could be a clari module - can you run through the list and evaluate if it makes sense?

immdipu:
As mentioned in the PR notes, it has a different `AuthType` (which is yet to be defined) than Clari connector. So, I believe it can't be considered a Clari module.

PR #1719, opened 2025-06-04
fix: Add logging to write response parse failings ServiceNow Connector

Jkarage:
The write response schema for 200 is not standard for all objects. 

- This updates the response to the caller to never fail, it's not their issue.
 - Adds logging when we fail to parse the response.


providers/servicenow/handlers.go
```diff
@@ -82,6 +83,8 @@ func (c *Connector) parseReadResponse(
 }
 
 func (c *Connector) buildWriteRequest(ctx context.Context, params common.WriteParams) (*http.Request, error) {
+	logging.With(ctx, "connector", "serviceNow")
```
RajatPawar, line 86: would be better to use `providers.ServiceNow`

RajatPawar (approved):

PR #1721, opened 2025-06-04
feat: Mutation generalize logic for GraphQL

sanjay-a01:


providers/fireflies/handlers.go
```diff
@@ -173,15 +167,17 @@ func (c *Connector) buildReadRequest(ctx context.Context, params common.ReadPara
 	return req, nil
 }
 
-func getQuery(limit, skip int, filePath, queryName string) string {
+func getQuery(limit, skip int, queryName string) (string, error) {
```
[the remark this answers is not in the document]
sanjay-a01: Sure, will do this. Thanks for the suggestion.

internal/graphql/graphqlOperation.go
```diff
@@ -0,0 +1,42 @@
...
+	Before, After, HasNextPage, HasPreviousPage      bool
+	StartCursor, EndCursor                           string
+}
+
+// GraphQLOperation loads and renders a GraphQL operation (query or mutation) from a template file.
+// This function supports dynamic GraphQL operations by loading a .graphql template from the embedded
+// filesystem and executing it with provided data. It supports injecting pagination parameters for queries
+// (such as limit, offset, or cursor-based pagination) as well as input payloads for mutations
+// through data parameter.
+func GraphQLOperation(queryFS embed.FS, queryType, queryName string, data any) (string, error) {
```
Cobalt0s, line 21: The package name already starts with `graphql`. In golang you can "remove" this. So the function will be named `Operation`. It seems "colliding" but it is in fact a good naming because within the package thats the only `Operation` so it is unique within the package and for the outside users it reads as `graphql.Operation`.

providers/fireflies/handlers.go
```diff
@@ -246,3 +218,123 @@ func (c *Connector) parseReadResponse(
...
+	node, ok := resp.Body()
+	if !ok {
+		return &common.WriteResult{Success: true}, nil
+	}
+
+	objectResponse, err := jsonquery.New(node).ObjectRequired("data")
+	if err != nil {
+		return nil, err
+	}
+
```
Cobalt0s, line 275:
```
var recordID string
```
line should be here instead of above.
`var err error` is already declared by `objectResponse, err :=`.

So `bite` is the only object to have recordID? Is it so?

Cobalt0s (approved):

Cobalt0s (approved):

providers/fireflies/graphql/mutation_audio.graphql
```diff
@@ -1,5 +1,12 @@
 mutation {
-    uploadAudio(input: { url:"{{.url}}"{{if .title}}, title:"{{.title}}"{{end}}{{if .attendees}}, attendees:[{{range $i, $a := .attendees}}{{if $i}},{{end}}{displayName:"{{$a.displayName}}", email:"{{$a.email}}", phoneNumber:"{{$a.phoneNumber}}"}{{end}}]{{end}}{{if .custom_language}}, custom_language:"{{.custom_language}}"{{end}}{{if .client_reference_id}}, client_reference_id:"{{.client_reference_id}}"{{end}}{{if ne .save_video nil}}, save_video:{{.save_video}}{{end}}}){
+    uploadAudio(input: {
+         url:"{{.url}}"{{if .title}},
```
RajatPawar, line 3: I think it's probably easier to read if it's formatted like: 
```
uploadAudio(input: {
  url: "{{.url}}",
  {{if .title}}
  title: "{{.title}}"
  {{end}}
```

and so on.

RajatPawar (approved):
Looks cool!

PR #1739, opened 2025-06-10
fix: reading next page records serviceNow connector

Jkarage:
Tests:
- The last read is reading next-page of the previous read activity.
Screenshot 2025-06-10 at 13 03 59


Cobalt0s (changes requested):
Should use `httpkit.HeaderLink()` when creating `ReadParams.NextPage`. The `Link` header can have next, previous page, it has sophisticated pattern which is better to outsource to the library we wrap.

Cobalt0s (approved):

PR #1755, opened 2025-06-11
[CON-1380] feat: Added the metadata functionality for the facebook connector

sanjay-a01:
## Metadata sample response
image
image
image


RajatPawar (changes requested):
We cannot support this properly because all the endpoints require an ID in the path which can be the ad account or the business account. We can fetch this information from the API, but there can be multiple accounts and the user needs to make a choice about which account they're interested in. Blocking this from merging

PR #1764, opened 2025-06-16
[CON-1709] feat: Added the read functionality for the avoma connector

sanjay-a01:
### Sample Read Results
image

image


providers/avoma/utils.go
```diff
@@ -29,18 +29,22 @@ var EndpointsWithResultsPath = datautils.NewSet( //nolint:gochecknoglobals
...
-		}
-
-		url, err := urlbuilder.New(nextPage)
+		nextPage, err := jsonquery.New(node).StringOptional("next")
 		if err != nil {
 			return "", err
 		}
 
-		url.AddEncodingExceptions(avomaQueryEncodingExceptions)
+		if nextPage != nil {
```
Cobalt0s, line 37: It should be shorter to reverse the if statement:
`if nextPage == nil {return "", nil}`
Also keeps nested blocks as short as possible, even if it is the "first level" of if nesting

PR #1768, opened 2025-06-16
Feat/zohocrm subscribe

jlimatampersand:


common/types.go
```diff
@@ -476,6 +476,7 @@ type WebhookVerificationParameters struct {
 	URL          string
 	ClientSecret string
 	Method       string
+	MatcherRefs  map[string]any // optional, depends on the provider
```
laurenzlong, line 479: can you explain what this is in a comment? Our types should be well documented, especially when the field name doesn't obviously communicate what it is.

laurenzlong (changes requested):

common/types.go
```diff
@@ -474,8 +474,14 @@ type WebhookVerificationParameters struct {
...
 	URL          string
-	ClientSecret string
+	ClientSecret string // TODO: deprecate this field and add to client secret
 	Method       string
+	// MatcherRefs is a map of matcher references to the provider.
+	// Some providers do not send signing mechanism,
+	// but instead, they ask us to provide tokens of our choice that they attach to webhook messages
+	// Instead of simple token field, it is better to just to have map for backward compatibility
+	// so that we can add custom data to the webhook message as needed.
+	MatcherRefs map[string]any
```
laurenzlong, line 484: Thanks for adding the comment, can you actually add a link to the Zoho docs that talk about these tokens? I couldn't find it, https://help.zoho.com/portal/en/kb/crm/automate-business-processes/actions/articles/webhooks-workflow#Scenarios doesn't really mention it. 

As for the name, I still find MatcherRefs to be unintuitive, here are some options that ChatGPT suggested, I like the words"challenge" or "echo". If we want to avoid using the word token (to not be confused with OAuth tokens), maybe "Code" could be the suffix (e.g. "echoCodes", "challengeCodes").

Screenshot 2025-06-26 at 4 11 09 PM

common/types.go
```diff
@@ -474,8 +474,14 @@ type WebhookVerificationParameters struct {
 	Headers      http.Header
```
laurenzlong, line 474: can you expand on the comment? Is the WebhookVerificationParameters initialized per webhook we receive?

common/types.go
laurenzlong: add link: https://www.zoho.com/crm/developer/docs/api/v8/notifications/enable.html

laurenzlong: capturing offline discussion here. for now we can go with something like this, down the road, we figure out how to make the construction of the verification params live in the connectors repo and not the server.

```
func (*Connector) VerifyWebhookMessage(
	_ context.Context, message *common.WebhookRequest, params *verificationParams,
) (bool, error) {



struct webhookRequest {
	Header map[string]string
	Body string
	...
}

struct hubSpotVerificationParams {
  clientSecret string // for Hubspot
}	

struct zohoVerificationParams {
    echoChallengeKey string // for Zoho
    echoChallengeValue string // for Zoho
}

type verificationParams struct {
	Param any
}
```

laurenzlong (approved):

jlimatampersand:
* **#1768** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1768?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1768?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1770, opened 2025-06-17
fix: Resolved the panic issue on getting the metadata

sanjay-a01:


providers/pinterest/handlers.go
```diff
@@ -46,6 +46,10 @@ func (c *Connector) parseSingleObjectMetadataResponse(
 		return nil, common.ErrFailedToUnmarshalBody
 	}
 
+	if len(data.Items) == 0 {
+		return nil, common.ErrEmptyJSONHTTPResponse
```
jlimatampersand, line 50: I think this is the wrong error type to return. 

This function is called after the response has been returned, just with items empty. 

I think we should have a new error type, something like, no objectMetadata returned or something like that can clearly tell what is happening.

PR #1772, opened 2025-06-17
[CON-1710] feat: Added the write functionality of the avoma connector.

sanjay-a01:
### Sample write response
For **smart_categories** object
image

For **calls** object
image


providers/avoma/handlers.go
```diff
@@ -0,0 +1,113 @@
...
+func (c *Connector) buildReadRequest(ctx context.Context, params common.ReadParams) (*http.Request, error) {
+	url, err := urlbuilder.New(c.ProviderInfo().BaseURL, apiVersion, params.ObjectName, "/")
+	if err != nil {
+		return nil, err
+	}
+
+	if EndpointsWithResultsPath.Has(params.ObjectName) {
+		url.WithQueryParam("page_size", pageSize)
+
+		if !params.Since.IsZero() && !params.Until.IsZero() {
```
RajatPawar, line 24: Is it not possible to only request Since or Until?

providers/avoma/handlers.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	jsonData, err := json.Marshal(params.RecordData)
+	if err != nil {
+		return nil, err
+	}
+
+	return http.NewRequestWithContext(ctx, http.MethodPost, url.String()+"/", bytes.NewReader(jsonData))
```
RajatPawar, line 72: Is the '/' needed? If so, can you please introduce a helper here so that no one accidentally removes it? (common.WithTrailingSlash, etc)

RajatPawar (dismissed):
Could you update the PR description with the checklist from the deep connector PR template?

providers/avoma/handlers.go
sanjay-a01: No, it is not possible because some of the endpoints need both from_date and to_date query params. If any one of the params is missing it will show an issue. Form date is passed via since field and to data is passed via until field.

Cobalt0s: You should write down a comment stating that Provider API requires it. Next time looking at the conditional the motivation will be aparent

scripts/openapi/avoma/metadata/main.go
```diff
@@ -0,0 +1,99 @@
...
+		"/v1/recordings/{uuid}/",
+		"/v1/scorecards/{uuid}/",
+		"/v1/smart_categories/{uuid}/",
+		"/v1/template/{uuid}/",
+		"/v1/transcriptions/{uuid}/",
+		"/v1/users/{uuid}/",
+	}
+
+	ObjectNameToResponseField = datautils.NewDefaultMap(map[string]string{ //nolint:gochecknoglobals
+		"v1/notes/":                 "results",
```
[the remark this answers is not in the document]
sanjay-a01: Got it, but I have trimmed the verison along with objectName before adding the metadata. Is this what you are saying?
image

Cobalt0s (approved):

PR #1794, opened 2025-06-20
[CON-1908] feat: Added lever read functionality.

sanjay-a01:
## Checklist

## Sample read response
image
image
image
image

## Test case result
image


providers/lever/handlers.go
```diff
@@ -50,3 +49,56 @@ func (c *Connector) parseSingleObjectMetadataResponse(
...
 	return &objectMetadata, nil
 }
+
+func (c *Connector) buildReadRequest(ctx context.Context, params common.ReadParams) (*http.Request, error) {
+	url, err := urlbuilder.New(c.ProviderInfo().BaseURL, apiVersion, params.ObjectName)
+	if err != nil {
+		return nil, err
+	}
+
+	url.WithQueryParam("limit", strconv.Itoa(DefaultPageSize))
```
RajatPawar, line 59: `DefaultPageSize` doesn't need to be exported, can be `defaultPageSize`

providers/lever/handlers.go
```diff
@@ -50,3 +49,56 @@ func (c *Connector) parseSingleObjectMetadataResponse(
...
+			url.WithQueryParam(prefix+"_start", strconv.Itoa(int(params.Since.UnixMilli())))
+		}
+
+		if !params.Until.IsZero() {
+			url.WithQueryParam(prefix+"_end", strconv.Itoa(int(params.Until.UnixMilli())))
+		}
+	}
+
+	// Apply timestamp parameters for each endpoint type
+	addTimeParams("created_at", EndpointWithCreatedAtRange.Has(params.ObjectName))
```
RajatPawar, line 77: same for EndpointWithCreatedAtRange, doesn't need to be exported

providers/lever/handlers.go
```diff
@@ -50,3 +49,56 @@ func (c *Connector) parseSingleObjectMetadataResponse(
...
+func (c *Connector) buildReadRequest(ctx context.Context, params common.ReadParams) (*http.Request, error) {
+	url, err := urlbuilder.New(c.ProviderInfo().BaseURL, apiVersion, params.ObjectName)
+	if err != nil {
+		return nil, err
+	}
+
+	url.WithQueryParam("limit", strconv.Itoa(DefaultPageSize))
+
+	// variadic function to add timestamp query parameters
+	addTimeParams := func(prefix string, hasEndpoint bool) {
```
RajatPawar, line 62: can this be a function outside of read?

providers/lever/supports.go
```diff
@@ -0,0 +1,35 @@
+package lever
```
RajatPawar, line 1: file name should be `support.go`

RajatPawar (approved):
Some changes required, good to go after that

PR #1797, opened 2025-06-20
[CON-1937] feat: Added Lever write functionality

sanjay-a01:
## Checklist

## Sample write response
image
image
image
image

## Testcase Result
image

## Sample Delete Response
image

## Testcase Result
image

**Notes:**
- Below delete endpoints cannot be delete that were created within the Lever application. Only endpoints that were created via API can be deleted via API.
    - feedback_templates
    - form_templates


providers/lever/README.md
```diff
@@ -6,36 +6,27 @@ Below is an exhaustive list of objects & methods supported on the objects
...
+| requisitions            | requisitions       | read, write  |
+| requisition_fields      | requisition_fields | read, write  |
+| uploads                 | uploads            | write        |
+| contacts                | contacts           | write        |
 
 
 Notes:
 - Excluded the endpoints /eeo/responses/pii and /eeo/responses because they are not direct endpoints, and their responses are embedded within their respective objectName under data. Other endpoints follow a consistent structure where responses are contained under data.
 - Excluded the endpoint /surveys/diversity/:posting because it includes a posting ID in the URL path, only one endpoints with posting in the connector.
+- Below delete endpoints cannot be delete that were created within the Lever application. Only endpoints that were created via API can be deleted via API.
```
RajatPawar, line 30:
```suggestion
- The delete endpoints below cannot be removed if they were created within the Lever application. Only endpoints created through the API are eligible for deletion via API.
```

providers/lever/handlers.go
```diff
@@ -105,3 +106,89 @@ func (c *Connector) parseReadResponse(
...
+	if err != nil {
+		return nil, err
+	}
+
+	method := http.MethodPost
+
+	if len(params.RecordId) > 0 {
+		url.AddPath(params.RecordId)
+
+		// Except postings object(use POST method), other objects use PUT method.
```
RajatPawar, line 121:
```suggestion
		// All objects use the PUT method except for the postings object, which uses the POST method.
```

providers/lever/handlers.go
```diff
@@ -105,3 +106,89 @@ func (c *Connector) parseReadResponse(
...
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodDelete, url.String(), nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return req, nil
```
RajatPawar, line 177:
```suggestion
	return http.NewRequestWithContext(ctx, http.MethodDelete, url.String(), nil)
```

providers/lever/supports.go
```diff
@@ -8,17 +8,8 @@ import (
 	"github.com/amp-labs/connectors/internal/components"
 )
```
RajatPawar, line 10: Please ensure that all future PRs with this file name it `support.go`. If any other connectors currently call it `supports.go`, please update that in a future PR as well, thanks

PR #1798, opened 2025-06-20
[CON-1891] feat: Add Braze Read connector

Jkarage:
Tests:
Screenshot 2025-06-20 at 12 43 30

### DIFF
--- a/providers/braze/parse.go
+++ b/providers/braze/parse.go
@@ -1,31 +1,150 @@
 package braze
 
-import "github.com/amp-labs/connectors/internal/datautils"
+import (
+	"errors"
+	"strconv"
+	"time"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+	"github.com/amp-labs/connectors/internal/datautils"
+	"github.com/amp-labs/connectors/internal/httpkit"
+	"github.com/amp-labs/connectors/internal/jsonquery"
+	"github.com/spyzhov/ajson"
+)
+
+const (
+	defaultLimit   = 100
+	nextPageOffset = 1
+	offset         = "offset"
+	page           = "page"
+	next           = "next"
+)
+
+var ErrMissingUntilTimestamp = errors.New("messages/scheduled_broadcasts requires an 'until' timestamp parameter")
 
 var dataFields = datautils.NewDefaultMap(datautils.Map[string, string]{ //nolint:gochecknoglobals
-	"catalogs":                      "catalogs",
-	"cdi/integrations":              "results",
-	"email/hard_bounces":            "emails",
-	"email/unsubscribes":            "emails",
-	"campaigns/list":                "campaigns",
-	"canvas/list":                   "canvases",
-	"events/list":                   "events",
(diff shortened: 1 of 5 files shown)


providers/braze/parse.go
```diff
@@ -29,3 +41,108 @@ var dataFields = datautils.NewDefaultMap(datautils.Map[string, string]{ //nolint
...
+	}
+
+	nxtPg := prvPg + nextPageOffset
+	url.WithQueryParam(page, strconv.Itoa(nxtPg))
+
+	return url.String(), nil
+}
+
+func filterBySince(params common.ReadParams, url *urlbuilder.URL) error {
+	if params.ObjectName == "campaigns/list" || params.ObjectName == "canvas/list" {
```
jlimatampersand, line 133: We should make this a switch case instead of a long list of if statements.

providers/braze/parse.go
```diff
@@ -29,3 +41,108 @@ var dataFields = datautils.NewDefaultMap(datautils.Map[string, string]{ //nolint
...
+		url.WithQueryParam("last_edit.time[gt]", params.Since.Format(time.RFC3339))
```
jlimatampersand, line 134: add link to doc that shows these query params for reference.

providers/braze/parse.go
```diff
@@ -29,3 +41,117 @@ var dataFields = datautils.NewDefaultMap(datautils.Map[string, string]{ //nolint
 }, func(key string) string {
 	return "data"
 })
+
+// cursorPaginatedObjects represents a set of objects that follows cursor pagination
+// approach in braze API.
+var cursorPaginatedObjects = datautils.NewSet( //nolint:gochecknoglobals
```
jlimatampersand, line 47: can you add docks that shows this list or API spec? same for below set.

providers/braze/parse.go
```diff
@@ -1,25 +1,37 @@
...
+	defaultLimit   = 100
+	nextPageOffset = 1
+	offset         = "offset"
+	page           = "page"
+	next           = "next"
+)
+
+var ErrMissingUntilTimestamp = errors.New("messages/scheduled_broadcasts requires an 'until' timestamp parameter")
 
 var dataFields = datautils.NewDefaultMap(datautils.Map[string, string]{ //nolint:gochecknoglobals
```
jlimatampersand, line 26: Please add reference links here as well.

jlimatampersand (changes requested):

providers/braze/parse.go
Jkarage: There is no one specific link for getting the cursor paginated objects (similar to the offset objects). Adding the object links.

providers/braze/parse.go
Jkarage: same as above.

jlimatampersand: Hey, I saw your comment, and there is no one API that shows every object. In this kind of case, I think just linking `https://www.braze.com/docs/api/endpoints` above this dataFields map will be sufficient. Same for all other object lists below. 

But I really appreciate you actually looking into each and every one of them.

PR #1806, opened 2025-06-23
[CON-1929] feat(linear): add ListObjectMetadata

immdipu:
## Checklist

## Sample Metadata Response
image
image
image

### DIFF
--- a/common/naming/formatting.go
+++ b/common/naming/formatting.go
@@ -20,3 +20,13 @@ func CapitalizeFirstLetterEveryWord(text string) string {
 
 	return text
 }
+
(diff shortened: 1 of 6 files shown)


common/naming/formatting.go
```diff
@@ -20,3 +20,13 @@ func CapitalizeFirstLetterEveryWord(text string) string {
 
 	return text
 }
+
+func CapitalizeFirstLetter(text string) string {
```
immdipu, line 24: Added a new function to capitalize the first letter of the word.

eberle1080 (approved):

PR #1811, opened 2025-06-25
[CON-1620,CON-1621] feat(atlassian): getModuleURL

Cobalt0s:
# Live tests

- [x] auth-metadata
image
- [x] metadata
image
- [x] read
image
- [x] write-delete
image

### DIFF
--- a/common/urlbuilder/url.go
+++ b/common/urlbuilder/url.go
@@ -100,6 +100,14 @@ func (u *URL) Path() string {
 	return u.delegate.Path
 }
 
+func (u *URL) Origin() string {
+	if u.delegate.Scheme == "" || u.delegate.Host == "" {
+		return ""
+	}
+
+	return u.delegate.Scheme + "://" + u.delegate.Host
+}
+
 func (u *URL) String() string {
 	// Everything stays the same
 	// The only thing that we alter in the delegate's query params
--- a/providers/atlassian/authmetamodel.go
+++ b/providers/atlassian/authmetamodel.go
@@ -1,5 +1,9 @@
 package atlassian
 
+import (
+	"github.com/amp-labs/connectors/common/substitutions/catalogreplacer"
+)
+
 const cloudIdKey = "cloudId"
 
 // AuthMetadataVars is a complete list of authentication metadata associated with connector.
@@ -19,5 +23,12 @@ func NewAuthMetadataVars(dictionary map[string]string) *AuthMetadataVars {
 func (v AuthMetadataVars) AsMap() *map[string]string {
 	return &map[string]string{
 		cloudIdKey: v.CloudId,
+	}
+}
+
+func (v AuthMetadataVars) GetSubstitutionPlan() catalogreplacer.SubstitutionPlan {
+	return catalogreplacer.SubstitutionPlan{
+		From: "cloudId",
+		To:   v.CloudId,
 	}
 }
--- a/providers/atlassian/connector.go
+++ b/providers/atlassian/connector.go
@@ -1,31 +1,26 @@
 package atlassian
 
 import (
-	"errors"
 	"fmt"
-	"strings"
 
 	"github.com/amp-labs/connectors/common"
 	"github.com/amp-labs/connectors/common/interpreter"
 	"github.com/amp-labs/connectors/common/paramsbuilder"
-	"github.com/amp-labs/connectors/common/substitutions/catalogreplacer"
 	"github.com/amp-labs/connectors/common/urlbuilder"
 	"github.com/amp-labs/connectors/providers"
 )
 
-// ErrMissingCloudId happens when cloud id was not provided via WithMetadata.
-var ErrMissingCloudId = errors.New("connector missing cloud id")
+const apiVersion = "3"
 
 type Connector struct {
-	Client     *common.JSONHTTPClient
-	moduleInfo *providers.ModuleInfo
-	moduleID   common.ModuleID
+	Client *common.JSONHTTPClient
 
 	// workspace is used to find cloud ID.
 	workspace string
-	cloudId   string
 
-	*providers.ProviderInfo
+	providerInfo *providers.ProviderInfo
+	moduleInfo   *providers.ModuleInfo
+	moduleID     common.ModuleID
 }
 
 func NewConnector(opts ...Option) (conn *Connector, outErr error) {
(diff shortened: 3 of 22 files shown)


Cobalt0s:
* **#1811** <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1811?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/amp-labs/connectors/1811?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

providers/atlassian/authmetamodel.go
```diff
@@ -21,3 +25,10 @@ func (v AuthMetadataVars) AsMap() *map[string]string {
 		cloudIdKey: v.CloudId,
 	}
 }
+
+func (v AuthMetadataVars) GetSubstitutionPlan() catalogreplacer.SubstitutionPlan {
+	return catalogreplacer.SubstitutionPlan{
+		From: "cloudId",
+		To:   v.CloudId,
```
Cobalt0s, line 32: The Cloud ID is correctly handled through substitutions, and ModuleInfo will contain the populated data.

providers/atlassian/connector.go
```diff
@@ -71,69 +66,24 @@ func (c *Connector) String() string {
 	return fmt.Sprintf("%s.Connector[%s]", c.Provider(), c.moduleID)
 }
 
-// URL format follows structure applicable to Oauth2 Atlassian apps.
-// https://developer.atlassian.com/cloud/jira/platform/rest/v2/intro/#other-integrations
-func (c *Connector) getJiraRestApiURL(arg string) (*urlbuilder.URL, error) {
```
Cobalt0s, line 76: This method is now part of catalog:
image

providers/atlassian/connector.go
```diff
@@ -71,69 +66,24 @@ func (c *Connector) String() string {
...
-	// When the module is Atlassian Connect, the base URL is different, so we need to override it.
-	// TODO: Replace options with substitution map in the future to avoid having to know
-	// which values need to be substituted.
-	if c.moduleID == providers.ModuleAtlassianJiraConnect {
-		vars := []catalogreplacer.CatalogVariable{
-			&paramsbuilder.Workspace{Name: c.workspace},
-		}
-
-		override := &providers.ProviderInfo{
-			BaseURL: "https://{{.workspace}}.atlassian.net",
```
Cobalt0s, line 129: image

providers/atlassian/connector.go
```diff
@@ -71,69 +66,24 @@ func (c *Connector) String() string {
...
 func (c *Connector) getAccessibleSitesURL() (*urlbuilder.URL, error) {
-	return urlbuilder.New(c.BaseURL, "oauth/token/accessible-resources")
-}
-
-func (c *Connector) setBaseURL(newURL string) {
-	c.BaseURL = newURL
-	c.Client.HTTPClient.Base = newURL
-}
-
-func (c *Connector) getCloudId() (string, error) {
```
Cobalt0s, line 103: This data will be embedded into the ModuleInfo and is no longer needed at the connector level. In other words, it is scoped only to the catalog data.

providers/atlassian/delete_test.go
```diff
@@ -69,21 +66,3 @@ func TestDelete(t *testing.T) { // nolint:funlen,cyclop
...
-		WithAuthenticatedClient(mockutils.NewClient()),
-		WithWorkspace("test-workspace"),
-		WithModule(providers.ModuleAtlassianJira),
-	)
-	if err != nil {
-		t.Fatal("failed to create connector")
-	}
-
-	_, err = connector.Delete(context.Background(), common.DeleteParams{ObjectName: "issues", RecordId: "123"})
-	if !errors.Is(err, ErrMissingCloudId) {
```
Cobalt0s, line 86: These tests are obsolete, there is one for each connector operation.

providers/atlassian/errors.go
```diff
@@ -78,3 +81,18 @@ func (r ResponseStatusError) CombineErr(base error) error {
 
 	return fmt.Errorf("%w: %v - %v", base, r.Error, r.Message)
 }
+
+func (c *Connector) interpretHTMLError(res *http.Response, body []byte) error {
```
Cobalt0s, line 85: To run the JWT live test, I would need to create the app, but I don't think it's worth the hassle since the functionality is the same and the correct base URL is already selected. While making requests, I noticed that the error response is a full HTML file, so this change adds an HTTP response parser.

providers/utils.go
```diff
@@ -159,14 +159,21 @@ func (c CustomCatalog) ReadInfo(provider Provider, vars ...catalogreplacer.Catal
...
 				vars.AddDefaults(catalogreplacer.CustomCatalogVariable{Plan: catalogreplacer.SubstitutionPlan{
-					From: metadatInput.Name,
-					To:   metadatInput.DefaultValue,
+					From: metadataInput.Name,
+					To:   metadataInput.DefaultValue,
 				}})
 			}
 		}
+
+		for _, authData := range i.Metadata.PostAuthentication {
```
Cobalt0s, line 171: Default dummy values are added when Cloud ID is not provided. This enables the Proxy and OAuth2 scripts to function correctly. Otherwise, we would need to accept an arbitrary list of metadata, which should be embedded through catalog substitutions.

test/atlassian/connector.go
```diff
@@ -12,11 +12,15 @@ import (
 	"golang.org/x/oauth2"
 )
 
-const cloudId = "35745fff-f0de-466c-b08e-a63f69888611"
+var fieldCloudID = credscanning.Field{
+	Name:      "cloudId",
+	PathJSON:  "metadata.cloudId",
```
Cobalt0s, line 17: Live tests no longer need to hardcode the values. Everything can be inside the `creds.json` file.

test/atlassian/write-delete/main.go
```diff
@@ -44,120 +41,43 @@ func main() {
...
-	slog.Info("> TEST Create/Update/Delete issue")
-	slog.Info("Creating issue")
+	oldTitle := "The very new title of Jira issue"
+	newTitle := "Fix button dropdown in main menu"
 
-	view := createIssue(ctx, conn, &issuePayload{
-		Fields: issueFields{
-			Project: identifier{
-				Id: projectID,
+	testscenario.ValidateCreateUpdateDelete(ctx, conn,
```
Cobalt0s, line 47: Functionally the same change but just much shorter. Functions just like before as expected.

providers/atlassian/connector.go
RajatPawar: Base URLs still shouldn't have the version # in the URL even if it is a module base URL, can you please remove that and put it in a const?

providers/atlassian/errors.go
RajatPawar: this shouldn't be required now?

providers/utils.go
```diff
@@ -159,22 +159,22 @@ func (c CustomCatalog) ReadInfo(provider Provider, vars ...catalogreplacer.Catal
...
-					To:   metadatInput.DefaultValue,
+					From: metadataInput.Name,
+					To:   metadataInput.DefaultValue,
 				}})
 			}
 		}
 
 		// To prevent OAuthConnect from erroring out due to missing PostAuthentication variables,
-		// we add a default value of " " for each PostAuthentication variable.
+		// we add a default value of "UNDEFINED_PRE_AUTH" for each PostAuthentication variable.
```
RajatPawar, line 172: I am not sure if we need this sentinel value, "" should be fine

common/urlbuilder/url.go
```diff
@@ -100,6 +100,10 @@ func (u *URL) Path() string {
 	return u.delegate.Path
 }
 
+func (u *URL) Origin() string {
```
RajatPawar, line 103: Can you please add a test case for this method?

RajatPawar (approved):

providers/utils.go
Cobalt0s: I thought in case we debug or view we immediately would know whats up. Thought it was nice to have.

PR #1814, opened 2025-06-26
feat: Removed the instantly V1 support and replaced instantlyai into instantly

sanjay-a01:
**Notes:**
  - Removed the files related to old instantly due to deprecated V1 version.
  - Renamed instantlyai into instantly.


providers/instantly/metadata/schemas.json
```diff
@@ -2,72 +2,1128 @@
...
-      "path": "",
+      "path": "/v2",
       "objects": {
-        "campaigns": {
-          "displayName": "Campaigns",
-          "path": "/campaigns",
-          "docs": "https://developer.instantly.ai/campaign-1/list-campaigns",
+        "accounts": {
+          "displayName": "Accounts",
+          "path": "v2/accounts",
```
Cobalt0s, line 9: The schema path shouldn't include version. This should be added via code.

Cobalt0s (approved):

laurenzlong (changes requested):
This is a breaking change, we cannot rename connectors like this.

PR #1815, opened 2025-06-26
[CON-1373] feat: Added campaign monitor read functionality

sanjay-a01:
## Checklist
  
## Sample Read Response
image
image
image
image

## Test case Result
image


jlimatampersand (commented):
Added a minor comment

I have a question, is there a reason this Provider is not using v2 connector?

sanjay-a01:
> Added a minor comment
> 
> I have a question, is there a reason this Provider is not using v2 connector?

Yes, there is no way to extract values from the endpoint response and assign them to the FieldMetadata struct, as the response does not contain any relevant values for its fields.
image

sanjay-a01:
FYI: Rajat asked me to remove the clientId input requirement, so I refactored the related logic accordingly.

PR #1816, opened 2025-06-26
[CON-1374] feat: Added campaign monitor write functionality.

sanjay-a01:
## Checklist

## Sample Write response
image
image

## Test cases Result
image

## Sample Delete Response
image

## Test cases Result
image


providers/campaignmonitor/utils.go
```diff
@@ -56,6 +69,29 @@ func (c *Connector) constructURL(objName string) (*urlbuilder.URL, error) {
 	return url, nil
 }
 
+func (c *Connector) constructWriteURL(objName string) (*urlbuilder.URL, error) {
+	// Endpoint with client id in the url
+	if endpointsWithClientIdAfterObjName.Has(objName) {
```
jlimatampersand, line 74: Can we make this a switch statement?

providers/campaignmonitor/supports.go
```diff
@@ -24,12 +24,39 @@ func supportedOperations() components.EndpointRegistryInput {
 		"journeys",
 	}
 
+	writeSupport := []string{
```
jlimatampersand, line 27: Can we have some links to the docs that show this list in comment?

providers/campaignmonitor/handlers.go
```diff
@@ -124,3 +128,87 @@ func (c *Connector) parseReadResponse(
...
+			Success: true,
+		}, nil
+	}
+
+	value, err := body.Value()
+	if err != nil {
+		return nil, err
+	}
+
+	switch v := value.(type) {
```
jlimatampersand, line 164: Can we have documentation link here?

jlimatampersand: This is strange behavior. Can you add an explanation when each case happens? Were you able to trigger both paths?

jlimatampersand (changes requested):
Left some comments

jlimatampersand (approved):

PR #1822, opened 2025-06-27
[CON-1405] feat: Added snapchatads connector read functionality

sanjay-a01:
## Checklist

## Sample Read Response
image
image
image
image

## Testcase Results
image


providers/snapchatads/utils.go
```diff
@@ -1,9 +1,15 @@
...
 
 import (
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/naming"
 	"github.com/amp-labs/connectors/common/urlbuilder"
+	"github.com/amp-labs/connectors/internal/jsonquery"
+	"github.com/spyzhov/ajson"
 )
 
+const defaultPageSize = 1
```
RajatPawar: why are we reading only 1 record?

RajatPawar (approved):

PR #1829, opened 2025-06-30
[CON-1348] feat(teamleader): add ListObjectMetadata

immdipu:
## Checklist

## Sample read response
image
image

### DIFF
--- a/providers/teamleader/handlers.go
+++ b/providers/teamleader/handlers.go
@@ -11,11 +11,11 @@ import (
 )
 
 const (
-	objectNameSuffix = ".list"
+	apiListSuffix = ".list"
 )
 
 func (c *Connector) buildSingleObjectMetadataRequest(ctx context.Context, objectName string) (*http.Request, error) {
-	fullObjectName := objectName + objectNameSuffix
+	fullObjectName := objectName + apiListSuffix
 
 	url, err := urlbuilder.New(c.ProviderInfo().BaseURL, fullObjectName)
 	if err != nil {
--- a/providers/teamleader/handlers.go
+++ b/providers/teamleader/handlers.go
@@ -1,0 +1,67 @@
+package teamleader
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/naming"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+)
+
+const (
+	objectNameSuffix = ".list"
+)
+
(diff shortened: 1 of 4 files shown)


providers/teamleader/handlers.go
```diff
@@ -0,0 +1,67 @@
...
+	return http.NewRequestWithContext(ctx, http.MethodGet, url.String(), nil)
+}
+
+func (c *Connector) parseSingleObjectMetadataResponse(
+	ctx context.Context,
+	objectName string,
+	request *http.Request,
+	response *common.JSONHTTPResponse,
+) (*common.ObjectMetadata, error) {
+	objectMetadata := common.ObjectMetadata{
```
RajatPawar, line 34: please confirm if this is Metadata v2

providers/teamleader/handlers.go
```diff
@@ -0,0 +1,67 @@
...
+		FieldsMap:   make(map[string]string),
+		DisplayName: naming.CapitalizeFirstLetterEveryWord(objectName),
+	}
+
+	res, err := common.UnmarshalJSON[map[string]any](response)
+	if err != nil {
+		return nil, common.ErrFailedToUnmarshalBody
+	}
+
+	if len(*res) == 0 {
```
RajatPawar, line 44: this could panic if res is nil

providers/teamleader/handlers.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (c *Connector) buildSingleObjectMetadataRequest(ctx context.Context, objectName string) (*http.Request, error) {
+	fullObjectName := objectName + objectNameSuffix
+
+	url, err := urlbuilder.New(c.ProviderInfo().BaseURL, fullObjectName)
+	if err != nil {
+		return nil, err
+	}
+
+	return http.NewRequestWithContext(ctx, http.MethodGet, url.String(), nil)
```
RajatPawar, line 25: can we limit response to just one page or one record?

providers/teamleader/handlers.go
```diff
@@ -0,0 +1,67 @@
...
+	"fmt"
+	"net/http"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/naming"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+)
+
+const (
+	objectNameSuffix = ".list"
```
RajatPawar, line 14: would be more readable if we call this something like `apiListSuffix` (can be anything else too, just that it needs to communicate that this is specific to listing, not the object itself)

RajatPawar (approved):

providers/teamleader/handlers.go
immdipu: yes i'll do it.

PR #1833, opened 2025-07-01
[CON-1349] feat(teamleader): add Read

immdipu:
## Checklist

## Sample Read Response
image
image
image

## Test cases Result
image

### DIFF
--- a/providers/teamleader/connector.go
+++ b/providers/teamleader/connector.go
@@ -4,6 +4,7 @@ import (
 	"github.com/amp-labs/connectors/common"
 	"github.com/amp-labs/connectors/internal/components"
 	"github.com/amp-labs/connectors/internal/components/operations"
+	"github.com/amp-labs/connectors/internal/components/reader"
 	"github.com/amp-labs/connectors/internal/components/schema"
 	"github.com/amp-labs/connectors/providers"
 )
@@ -17,6 +18,7 @@ type Connector struct {
 
 	// Supported operations
 	components.SchemaProvider
+	components.Reader
 }
 
 func NewConnector(params common.ConnectorParams) (*Connector, error) {
@@ -33,6 +35,23 @@ func constructor(base *components.Connector) (*Connector, error) {
 		operations.SingleObjectMetadataHandlers{
 			BuildRequest:  connector.buildSingleObjectMetadataRequest,
 			ParseResponse: connector.parseSingleObjectMetadataResponse,
+		},
+	)
+
+	registry, err := components.NewEndpointRegistry(supportedOperations())
+	if err != nil {
+		return nil, err
+	}
+
+	// Set the read provider for the connector
+	connector.Reader = reader.NewHTTPReader(
+		connector.HTTPClient().Client,
+		registry,
+		common.ModuleRoot,
+		operations.ReadHandlers{
+			BuildRequest:  connector.buildReadRequest,
+			ParseResponse: connector.parseReadResponse,
+			ErrorHandler:  common.InterpretError,
 		},
 	)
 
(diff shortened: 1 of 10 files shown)


providers/teamleader/connector.go
```diff
@@ -36,5 +38,22 @@ func constructor(base *components.Connector) (*Connector, error) {
...
+	registry, err := components.NewEndpointRegistry(supportedOperations())
+	if err != nil {
+		return nil, err
+	}
+
+	// Set the read provider for the connector
+	connector.Reader = reader.NewHTTPReader(
+		connector.HTTPClient().Client,
+		registry,
+		common.ModuleRoot,
```
RajatPawar, line 50: Can you please update this to be ProviderContext.Module() for consistency?

providers/teamleader/handlers.go
```diff
@@ -65,3 +69,65 @@ func (c *Connector) parseSingleObjectMetadataResponse(
...
+		nextRecordsURL(params),
+		common.GetMarshaledData,
+		params.Fields,
+	)
+}
+
+func buildRequestBody(params *common.ReadParams) map[string]any {
+	body := make(map[string]any)
+
+	if !params.Since.IsZero() {
```
RajatPawar, line 114: can this API support Until?

providers/teamleader/handlers.go
```diff
@@ -65,3 +69,65 @@ func (c *Connector) parseSingleObjectMetadataResponse(
...
+
+	if params.NextPage != "" {
+		body["page"] = map[string]any{
+			"size":   pageSize,
+			"number": params.NextPage,
+		}
+	} else {
+		body["page"] = map[string]any{
+			"size":   pageSize,
+			"number": "1",
```
RajatPawar, line 128: what happens when page = 0?

providers/teamleader/parse.go
```diff
@@ -0,0 +1,48 @@
...
+		}
+
+		return jsonquery.Convertor.ArrayToMap(records)
+	}
+}
+
+func nextRecordsURL(params common.ReadParams) common.NextPageFunc {
+	return func(node *ajson.Node) (string, error) {
+		res, err := jsonquery.New(node).ArrayRequired("data")
+		if err != nil {
```
RajatPawar, line 25: can be grouped as `err != nil || res == nil`

providers/teamleader/parse.go
```diff
@@ -0,0 +1,48 @@
...
+		}
+
+		if len(res) < pageSize {
+			return "", nil
+		}
+
+		currentPage := 1
+
+		if params.NextPage != "" {
+			page, err := strconv.Atoi(string(params.NextPage))
```
RajatPawar, line 40: we should error out if Atio fails

providers/teamleader/supports.go
RajatPawar, line 1: the file should be called `support.go`

providers/teamleader/supports.go
```diff
@@ -0,0 +1,30 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/internal/components"
+)
+
+func supportedOperations() components.EndpointRegistryInput {
+	//nolint:lll
+	readSupport := []string{
```
RajatPawar, line 13: can you please list one object per line?

RajatPawar (approved):

providers/teamleader/parse.go
immdipu: Thank you. i'll do it

providers/teamleader/supports.go
immdipu: i'll update it

PR #1841, opened 2025-07-07
[CON-1406] feat: Added snapchatads write functionality

sanjay-a01:
## Checklist

## Sample Write Response
image
image
image

## Test Cases Result
image

## Sample Delete Response
image

## Test Cases Result
image


providers/snapchatads/README.md
```diff
@@ -8,18 +8,20 @@ SnapchatAds API version : v1
...
+| adaccounts              | adaccounts             | read, write  |
+| members                 | members                | read, write  |
+| roles                   | roles                  | read, write  |
+-------------------------------------------------------------------
  
 Notes:
 - The organization_id is retrieved using the postAuthentication method.
 - Currently, we only support objects that include the organization_id in the URL path.
-
+- The API documentation specify using the POST method to update billingcenters, it turns out the PUT method works just as well.
```
RajatPawar, line 25:
```suggestion
- The API documentation mentions that we must use the POST method to update billingcenters, but it turns out that the PUT method works too.
```

providers/snapchatads/handlers.go
```diff
@@ -98,3 +101,110 @@ func (c *Connector) parseReadResponse(
...
+	if err != nil {
+		return nil, err
+	}
+
+	method := http.MethodPost
+
+	// When updating an object, the record ID is not included in the URL path; instead,
+	// it must be provided in the body parameters.
+	// EX. Refer https://developers.snap.com/api/marketing-api/Ads-API/billing-centers#update-a-billing-center.
+	if params.RecordId != "" {
```
RajatPawar, line 116: can you make sure that the json body contains the record ID in case this is set?

providers/snapchatads/supports.go
RajatPawar, line 1: the file should be called `support.go`

RajatPawar (approved):

providers/snapchatads/errors.go
```diff
@@ -16,7 +19,10 @@ var errorFormats = interpreter.NewFormatSwitch( // nolint:gochecknoglobals
 	}...,
 )
 
-var ErrObjNotFound = errors.New("object not found")
+var (
+	ErrObjNotFound    = errors.New("object not found")
```
RajatPawar: seems like both of these can be unexported

PR #1848, opened 2025-07-08
[CON-1985] feat(fathom): add write

immdipu:
## Checklist
- [ ] Read supports pagination

## Sample Read Response
image

## Test cases Result
image


providers/fathom/support.go
```diff
@@ -12,13 +12,20 @@ func supportedOperations() components.EndpointRegistryInput {
 	readSupport := []string{
 		"meetings", "teams", "team_members",
 	}
+	writeSupport := []string{
```
RajatPawar, line 15: are there any other objects that are available for writing, or is it just the one? If there are limitations, can yoi please document them here?

RajatPawar (approved):

PR #1871, opened 2025-07-15
created calendly connector (read and subscribe functionalities)

sramesh64:
Deep Calendly connector (Read and Subscribe implementations). Read implementation focuses on /scheduled_events.

### DIFF
--- a/providers/calendly.go
+++ b/providers/calendly.go
@@ -15,6 +15,7 @@ func init() {
 			SuccessStatusCodes: []int{http.StatusOK},
 			Url:                "https://api.calendly.com/users/me",
 		},
+		PostAuthInfoNeeded: true,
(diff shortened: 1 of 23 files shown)


Jkarage:
@sramesh64 Thanks for putting this together — great work! 🙌
Before we send it over to the core team, I have a few suggestions that could help streamline the review process.

Let's split this into three separate PRs?

Metadata PR: This would include the schemas file and the logic for assigning the schemaProvider in the connector + tests.

Read PR: This would focus on the Read implementation. From what I’ve seen with `usrURI` and `orgURI`, it looks like we could potentially support more objects beyond just scheduledEvents — really promising! (I can help add the rest)

Subscribe Action PR: We can handle the subscription logic here. 

For linting the repo is using: golangci-lint v1.62.0, `make lint` should show the issues.

PR #1881, opened 2025-07-18
[CON-2087] feat: [snapchatads] Extend the supported read objects.

sanjay-a01:
## Sample Response
image

### DIFF
--- a/providers/snapchatads/handlers.go
+++ b/providers/snapchatads/handlers.go
@@ -51,7 +51,7 @@ func (c *Connector) parseSingleObjectMetadataResponse(
 
 	// Determine key for inner map
 	objKey := objectName
-	if endpointsWithSharedId.Has(objectName) {
+	if endpointsRequiringOrganizationMetadata.Has(objectName) {
 		objKey = naming.NewSingularString(objectName).String()
 	}
 
--- a/providers/snapchatads/utils.go
+++ b/providers/snapchatads/utils.go
@@ -5,8 +5,8 @@ import (
 	"github.com/amp-labs/connectors/internal/datautils"
 )
 
-// The endpoints shared id(organization Id) in the url path.
-var endpointsWithSharedId = datautils.NewSet( //nolint:gochecknoglobals
+// Endpoints that require an organization ID (provided via metadata) as shared ID in the URL path.
+var endpointsRequiringOrganizationMetadata = datautils.NewSet( //nolint:gochecknoglobals
 	"fundingsources",
 	"billingcenters",
 	"transactions",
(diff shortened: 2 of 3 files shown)


providers/snapchatads/utils.go
```diff
@@ -2,13 +2,44 @@ package snapchatads
...
+		prefix = "targeting/interests/"
+	case "categories_loi":
+		prefix = "targeting/location/"
+	default:
+		prefix = ""
+	}
+
+	fullPath := prefix + objName
+
+	if endpointsWithSharedId.Has(objName) {
```
RajatPawar, line 38: are we defaulting to an org ID here?

RajatPawar:
@sanjay-a01 please request a review again when you reply to the comments

providers/snapchatads/handlers.go
```diff
@@ -34,7 +34,14 @@ func (c *Connector) parseSingleObjectMetadataResponse(
 		return nil, common.ErrEmptyJSONHTTPResponse
 	}
 
-	objectResponse, err := jsonquery.New(node).ArrayRequired(objectName)
+	// Determine correct node path
+	nodePath := "targeting_dimensions"
```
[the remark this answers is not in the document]
sanjay-a01: > is this a special key name that applies for all objects?

No, this special key name is for object with no sharedID in the URL. The nodepath will change if the conditon below is stasified which means object with sharedID in the URL.
image

providers/snapchatads/utils.go
sanjay-a01: No, we’re not defaulting to an org ID because it’s only included when the object name is listed in endpointsWithSharedId. For other endpoints, the URL does not include the org ID.

providers/snapchatads/utils.go
```diff
@@ -2,13 +2,44 @@ package snapchatads
...
 
 func (c *Connector) constructURL(objName string) (*urlbuilder.URL, error) {
-	url, err := urlbuilder.New(c.ProviderInfo().BaseURL, apiVersion, "organizations", c.organizationId, objName)
-	if err != nil {
-		return nil, err
+	var prefix string
+
+	switch objName {
+	case "age_group", "gender", "languages", "advanced_demographics":
+		prefix = "targeting/demographics/"
```
RajatPawar: we shouldn't be mapping names (as we've recently discussed in other PRs), let's have these as `targeting/demographics/languages` and so on

providers/snapchatads/utils.go
```diff
@@ -2,13 +2,37 @@ package snapchatads
 
 import (
 	"github.com/amp-labs/connectors/common/urlbuilder"
+	"github.com/amp-labs/connectors/internal/datautils"
+)
+
+// The endpoints shared id(organization Id) in the url path.
+var endpointsWithSharedId = datautils.NewSet( //nolint:gochecknoglobals
```
RajatPawar: `endpointsRequiringOrganizationMetadata`

PR #1883, opened 2025-07-18
fix: Map ObjectName to endpoints in Braze connector

Jkarage:
With this builders can use `campaigns` instead of `campaigns/list` and similar objects while reading.

# Checklist

- [ ] Read supports incremental sync

Metadata Live Tests:
Screenshot 2025-07-20 at 12 10 20
Screenshot 2025-07-20 at 12 10 41

Read Live Tests:
Screenshot 2025-07-20 at 12 11 41
Screenshot 2025-07-20 at 12 12 01

Write Live Tests:
Screenshot 2025-07-20 at 12 13 49

### DIFF
--- a/providers/braze/handlers.go
+++ b/providers/braze/handlers.go
@@ -20,7 +20,15 @@ type catalogPayload struct {
 var ErrInvalidData = errors.New("invalid request data provided")
 
 func (c *Connector) metadataRequest(ctx context.Context, objectName string) (*http.Request, error) {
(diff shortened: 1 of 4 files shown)


jlimatampersand:
We are now requiring everyone to use templates to when making PRs. 

There is `.github` folder where you can select PRs from. Please fill out the template and the checklist in the template. The re-request review.

jlimatampersand (changes requested):
please use the PR templates and re-request review

PR #1891, opened 2025-07-22
[CON-2008] feat(google): Contacts Read

Cobalt0s:
- [ ] Read supports incremental sync

# Live Test
### contactGroups
image

### myConnections
image

### otherContacts
image


providers/google/internal/contacts/read.go
```diff
@@ -0,0 +1,245 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	url.WithQueryParam("pageSize", strconv.Itoa(defaultPageSize))
+
+	switch params.ObjectName {
+	case objectNameContactGroups:
+		// https://developers.google.com/people/api/rest/v1/contactGroups/list
+		readURLForContactGroups(params, url)
```
RajatPawar, line 51: as a general style, please prefer functions that return values explictly over functions that mutate params in place, because it's easy to introduce subtle bugs with functions that mutate in place

providers/google/internal/contacts/read.go
```diff
@@ -0,0 +1,245 @@
...
+		return url.String(), nil
+	}
+}
+
+// readURLForContactGroups constructs the query parameters for the request.
+// - Filters out fields that are not permitted in groupFields.
+// https://developers.google.com/people/api/rest/v1/contactGroups/list
+func readURLForContactGroups(params common.ReadParams, url *urlbuilder.URL) {
+	groupFields := datautils.NewSetFromList([]string{
+		"clientData", "groupType", "memberCount", "metadata", "name",
```
RajatPawar, line 151: if these are the only fields we allow, can this not be a part of the Schemas file?

RajatPawar (approved):

RajatPawar:
> # Live Test
> ### contactGroups
> image
> ### myConnections
> image
> ### otherContacts
> image

Please update the description to have the checklist from the PR template

providers/google/internal/contacts/read.go
Cobalt0s: There is an issue with this. One one side of the spectrum there is C style where everything is a pointer and you pass pointers to methods which create side effects and then there is pure funcitonal where everything is a copy.
`urlbuilder.URL` uses a golang map to keep track of query params. So by design this data structure is modified by reference. This means 2 things:
* `urlbuilder.URL` should be rewritten in pure funcitonal paradigms
* Side effects still would happen.

I think a method that would accept `url` and return `url` would be misleading because it will tell that it is a "new copy/instance". I have the same worry (I like Lisp/Clojure, functional ideas) but I don't know how that can be expressed in pure terms with golang map.

providers/google/internal/contacts/read.go
Cobalt0s: Hmm, I took a look at this, I think we should keep it seperate. For example client may request `memberCount` fields and `etag`. First must be mentioned in `groupFields` query parameter, while the second would produce an error (I have a unit test that "omits" etag when invoking provider). However, the etag will still be transfered from `raw` to `fields` properties on ReadResult.

Google contacts (People API) is one of the most unique in the way it requires weird query params which all depend on several field conditionals.

PR #1900, opened 2025-07-23
[ENG-2405] feat: add salesforce associations

RajatPawar:
- Subqueries can be used to fetch associations when reading. Source in comments.
- Added a test for associations

### Reading without associations
Screenshot 2025-07-22 at 5 16 45 PM

### Reading with associations
Screenshot 2025-07-22 at 5 17 25 PM

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -264,7 +264,6 @@ type Association struct {
 	// AssociationType is the type of association.
 	AssociationType string         `json:"associationType,omitempty"`
 	Raw             map[string]any `json:"raw,omitempty"`
-	Fields          map[string]any `json:"fields,omitempty"`
 }
 
 // WriteResult is what's returned from writing data via the Write call.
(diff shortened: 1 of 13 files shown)


common/types.go
```diff
@@ -264,7 +264,6 @@ type Association struct {
 	// AssociationType is the type of association.
 	AssociationType string         `json:"associationType,omitempty"`
 	Raw             map[string]any `json:"raw,omitempty"`
-	Fields          map[string]any `json:"fields,omitempty"`
```
RajatPawar, line 267: I added this in a previous PR before realizing that we don't support specifying fields for associated objects. We don't need this for now.

providers/salesforce/read.go
```diff
@@ -33,7 +33,7 @@ func (c *Connector) Read(ctx context.Context, config common.ReadParams) (*common
 		rsp,
 		getRecords,
 		getNextRecordsURL,
-		common.GetMarshaledData,
+		getSalesforceDataMarshaller(config.AssociatedObjects),
```
RajatPawar, line 36: Custom marshal func that helps extract associations

PR #1902, opened 2025-07-23
[CON-2027] chore: [HighLevel] Support extra OAuth2 fields in proxy connector

sanjay-a01:
## Note:
In the HighLevel connector, most of the endpoints require a `locationId` as a query parameter. We can get the `locationId` from the token response, so I added the necessary field in the OtherFields section of the TokenMetadataFields struct.

FYI:- There are two types of access tokens — one is the Agency Access Token and the other is the Sub-Account Access Token. When a sub-account is selected during installation, the `locationId` will be included in the token response. If the agency account is selected instead, the `locationId` will not be returned in the response, because it is available only for the Sub-Account Access Token. The locationId is unique since only one sub-account is selected during installation.
image

### DIFF
--- a/providers/highlevel.go
+++ b/providers/highlevel.go
@@ -21,6 +21,28 @@ func init() {
 			TokenMetadataFields: TokenMetadataFields{
 				ScopesField:      "scope",
 				ConsumerRefField: "userId",
+				OtherFields: &TokenMetadataFieldsOtherFields{
+					{
+						DisplayName: "Location Id",
+						Name:        "locationId",
+						Path:        "",
+					},
+					{
+						DisplayName: "User Type",
+						Name:        "userType",
+						Path:        "",
+					},
+					{
+						DisplayName: "Plan Id",
+						Name:        "planId",
+						Path:        "",
+					},
+					{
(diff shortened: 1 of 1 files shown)


jlimatampersand:
@sanjay-a01 Can you add some documentation link showing what the requirements are? and some token response example. 

Also this code does not show how the are being used. Can you point us to where we are using this?

providers/highlevel.go
```diff
@@ -21,6 +21,28 @@ func init() {
 			TokenMetadataFields: TokenMetadataFields{
 				ScopesField:      "scope",
 				ConsumerRefField: "userId",
+				OtherFields: &TokenMetadataFieldsOtherFields{
```
jlimatampersand, line 24: Thanks for providing detailed explanation in the PR thread. But I meant to ask you to leave comment in code with docs url 
```
// docs: http://...... 
``
so that later when other people find this code, they understand better. 

Sorry if I have confused you on this.

jlimatampersand (approved):
Looks good, feel free to merge after resolving my comment.

PR #1906, opened 2025-07-23
[ENG-2414] Modifty event interface to return raw map

jlimatampersand:
This PR enables event message interface to return underlying message with RawMap() method.


common/types.go
```diff
@@ -456,6 +456,7 @@ type SubscriptionEvent interface {
 	Workspace() (string, error)
 	RecordId() (string, error)
 	EventTimeStampNano() (int64, error)
+	RawMap() (map[string]any, error)
```
jlimatampersand, line 459: returning error for backward compatibility in case we see a new event type that cannot be converted to a map. But currently we are not using it.

PR #1907, opened 2025-07-23
[ENG-2415] fix: Fix logger error where body gets swapped for wrong content

eberle1080:


Cobalt0s (approved):

common/http.go
```diff
@@ -302,25 +302,29 @@ func (h *HTTPClient) httpGet(ctx context.Context, //nolint:dupl
...
 
 	correlationId := uuid.Must(uuid.NewRandom()).String()
 
 	if logging.IsVerboseLogging(ctx) {
-		logRequestWithBody(logging.VerboseLogger(ctx), req, "POST", correlationId, url, body)
+		// body is nil because makePostRequest sometimes returns an altered body reader
+		// so we rely on req and not body for logging purposes. If we use body, it will
+		// not just log the wrong thing, it will swap the body out, which will lead to
+		// unexpected behavior in the request.
+		logRequestWithBody(logging.VerboseLogger(ctx), req, "POST", correlationId, url, nil)
```
laurenzlong, line 319: @eberle1080 would it be better to modify the signature of this function to not have the last parameter?

PR #1911, opened 2025-07-25
[CON-2026] feat: Add Outreach Connector Associations

Jkarage:
This adds support for querying record associated objects. 

Tests:

- Reading without associations:
Screenshot 2025-08-01 at 14 52 05

- Reading with associations
Screenshot 2025-08-01 at 14 51 18

### DIFF
--- a/providers/outreach/read.go
+++ b/providers/outreach/read.go
@@ -18,7 +18,7 @@ type Record struct {
 	Data DataItem `json:"data"`
 }
 
-type Associations struct {
+type RecordAssociations struct {
 	ObjectId          string                          // ObjectId represents the id of the object we are reading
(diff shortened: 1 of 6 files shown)


providers/outreach/read.go
```diff
@@ -2,18 +2,35 @@ package outreach
...
+// Record represents a single object record returned by the Outreach connector.
+type Record struct {
+	Data DataItem `json:"data"`
+}
+
+type RecordAssociations struct {
+	ObjectId          string                          // ObjectId represents the id of the object we are reading
+	AssociatedObjects map[string][]common.Association // Associated objects
+}
+
```
Jkarage, line 25: Do you think this could be in common. I wanted recordId with their associated objects.

PR #1913, opened 2025-07-26
fix: improve remainingObjects tracking and preserve original object names slice

Jkarage:
Updates the compositeSchema logic to never alter the objectName slice param.

Tests:
Screenshot 2025-07-26 at 21 50 26

### DIFF
--- a/internal/components/schema/composite.go
+++ b/internal/components/schema/composite.go
@@ -3,7 +3,6 @@ package schema
 import (
 	"context"
 	"errors"
-	"fmt"
 	"log/slog"
 	"maps"
 
@@ -40,7 +39,7 @@ func (c *CompositeSchemaProvider) ListObjectMetadata(
 	remainingObjects := make([]string, len(objects))
 	copy(remainingObjects, objects)
 
-	for _, schemaProvider := range c.schemaProviders {
+	for idx, schemaProvider := range c.schemaProviders {
 		if len(remainingObjects) == 0 {
 			break
 		}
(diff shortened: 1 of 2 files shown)


internal/components/schema/composite.go
```diff
@@ -51,23 +51,34 @@ func (c *CompositeSchemaProvider) ListObjectMetadata(
 			continue
 		}
 
-		// Append successful object metadatas to the result metadata.
-		// do not replace this with map.Copy
```
laurenzlong, line 55: this comment says "do not replace this with map.Copy", can you explain why it is now safe to ignore this comment?

Jkarage: The compiler was suggesting this and I was worried, we might be overwriting the Result Map, but after looking that will never happen as the as soon we have metadata for a particular object name. we will never fetch metadata for that particular object, rather the failed ones.

laurenzlong (approved):
@Jkarage I modified this PR in a few ways:
- fixed the issue where the actual error message was getting swallowed and replaced by the generic error message "failed to get metadata for object from any provider". This makes it easier for us to debug customer issues, since the actual error will flow all the way to the server and get logged there, where as "failed to get metadata for object from any provider" doesn't really help us when debugging issues.
- used a set instead of a slice to track unprocessed objects, since it is a lot more performant for removing elements
- re-write logic to be easier to follow
- added unit tests

I verified that the test script still works and now prints the appropriate errors, see output: 

Screenshot 2025-07-26 at 20 16 46

PR #1916, opened 2025-07-27
feat: Extend merge_requests, projects supports Gitlab connector

Jkarage:
- Adds support for reading user owned projects + membership projects. 
- Adds since parameter implementation, works for both merge_requests and projects.
- Adds live tests for `merge_requests` and `projects`

With this we can now:
-  read: `merge_requests`, `projects`
- write `projects`

Tests:
Screenshot 2025-07-27 at 21 09 11
Screenshot 2025-07-27 at 21 10 12

### DIFF
--- a/providers/gitlab/handlers.go
+++ b/providers/gitlab/handlers.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/amp-labs/connectors/common"
 	"github.com/amp-labs/connectors/common/naming"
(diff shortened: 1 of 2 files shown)


providers/gitlab/handlers.go
```diff
@@ -84,25 +89,45 @@ func (c *Connector) parseSingleHandlerResponse(
...
 }
 
+func (c *Connector) constructReadURL(params common.ReadParams) (*urlbuilder.URL, error) {
+	url, err := urlbuilder.New(c.ProviderInfo().BaseURL, restAPIVersion, params.ObjectName)
+	if err != nil {
+		return nil, err
+	}
+
+	if params.ObjectName == projects {
+		// sets owned=true&membership=true
```
laurenzlong, line 111:
```suggestion
		// sets owned=true&membership=true so that we are fetching the user's projects
		// instead of all public projects on GitLab
```

providers/gitlab/handlers.go
```diff
@@ -84,25 +89,45 @@ func (c *Connector) parseSingleHandlerResponse(
...
+		url.WithQueryParam(ownedQuery, "true")
+		url.WithQueryParam(membershipQuery, "true")
+	}
+
+	if !params.Since.IsZero() {
+		url.WithQueryParam(updatedAfterQuery, params.Since.Format(time.RFC3339))
+
+		if params.ObjectName == projects {
+			// This is required for reading Projects, if sice is provided.
```
laurenzlong, line 120:
```suggestion
			// This is required for reading Projects, if since is provided.
```

laurenzlong (approved):

PR #1917, opened 2025-07-28
feat: update Gitlab provider OAuth scheme with PATs

Jkarage:
Tests:
Screenshot 2025-07-28 at 07 37 01

### DIFF
--- a/providers/gitlab.go
+++ b/providers/gitlab.go
@@ -7,12 +7,20 @@ func init() {
 		DisplayName: "GitLab",
 		AuthType:    Oauth2,
 		BaseURL:     "https://gitlab.com",
-		Oauth2Opts: &Oauth2Opts{
-			GrantType:                 AuthorizationCode,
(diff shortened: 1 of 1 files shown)


providers/gitlab.go
```diff
@@ -7,12 +7,21 @@ func init() {
...
-		Oauth2Opts: &Oauth2Opts{
-			GrantType:                 AuthorizationCode,
-			AuthURL:                   "https://gitlab.com/oauth/authorize",
-			TokenURL:                  "https://gitlab.com/oauth/token",
-			ExplicitScopesRequired:    false,
-			ExplicitWorkspaceRequired: false,
+		CustomOpts: &CustomAuthOpts{
+			Headers: []CustomAuthHeader{
+				{
+					Name:          "PRIVATE-TOKEN",
```
laurenzlong, line 13: @Jkarage previously you had "PRIVATE TOKEN" (no dash) which is not a valid header value

PR #1920, opened 2025-07-28
[ENG-2378] feat: add netsuite restapi module

RajatPawar:
#### Listing records (internally) which are then fetched 
Screenshot 2025-07-28 at 2 22 57 PM
Screenshot 2025-07-28 at 2 23 05 PM

### Read-write 
Screenshot 2025-07-28 at 3 22 06 PM

### Metadata 
Screenshot 2025-07-28 at 2 04 45 PM
Screenshot 2025-07-28 at 2 04 27 PM
Screenshot 2025-07-28 at 2 04 23 PM

### Delete
Screenshot 2025-07-28 at 3 44 22 PM

### Errors
Screenshot 2025-07-28 at 3 32 00 PM


providers/netsuite/internal/restapi/read.go
```diff
@@ -0,0 +1,198 @@
...
+	// Collect the results.
+	results := make([]map[string]any, len(recordsToFetch))
+
+	var allErrors []error
+
+	for res := range resultChan {
+		if res.err != nil {
+			allErrors = append(allErrors, res.err)
+		} else {
+			results[res.index] = res.data
```
Cobalt0s, line 191: Given the line:
```
results := make([]map[string]any, **len(recordsToFetch))**
```

When the error occurs the value won't be placed into the array "slot". Therefore, they will be set to `nil`. I checked downstream and no nil pointer can occur. Also the overall procedure would collect all errors and disregard rows data.

Predeclaring the slice with a size implies that each slot will be initialized. Otherwise, it is better to use append and for performence -- set capacity instead.

I think if `ReadResult` had both error and rows then the comment I have above would be more critical.

Cobalt0s (approved):
Looks good!
I’ve left some notes.

Also, it seems there are quite a few edge cases in how NetSuite works. Let me know if I can help add some mock tests in a follow-up PR. Also, ping me on Slack with your thoughts about the adapter “pattern.”

providers/netsuite.go
```diff
@@ -1,8 +1,16 @@
...
-const Netsuite Provider = "netsuite"
+const (
+	Netsuite Provider = "netsuite"
 
-// nolint:lll
+	// NetsuiteModuleSuiteQL is a read-only module that uses SuiteQL to read data.
+	NetsuiteModuleSuiteQL = "suiteql"
+
+	// NetsuiteModuleRESTAPI is a read-write module that uses the REST API to read and write data.
+	NetsuiteModuleRESTAPI = "restapi"
```
laurenzlong, line 10: what do you think of `rest` as the module name instead? it's simpler and reads better than `restapi`, which looks like it should be pronounced "res-tah-pee"

laurenzlong: similarly the constant name can be `NetsuiteModuleREST` for brevity

providers/netsuite.go
```diff
@@ -28,6 +36,24 @@ func init() {
 			Subscribe: false,
 			Write:     false,
 		},
+		DefaultModule: NetsuiteModuleRESTAPI,
+		Modules: &Modules{
+			NetsuiteModuleSuiteQL: {
+				DisplayName: "Netsuite (SuiteQL)",
+				BaseURL:     "https://{{.workspace}}.suitetalk.api.netsuite.com/services/rest/query",
+				Support: Support{
```
laurenzlong, line 44: isn't proxy supported for one or both of these modules?

providers/netsuite/connector.go
```diff
@@ -20,30 +20,85 @@ type Connector struct {
 	common.RequireAuthenticatedClient
 	common.RequireWorkspace
 
-	// Supported operations
-	components.SchemaProvider
+	// TODO: Expose concurrency knobs to the server.
```
laurenzlong, line 23: a little confused by this comment, can you elaborate a bit more?

providers/netsuite/internal/restapi/metadata.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+
+		// backward compatibility
+		result.FieldsMap[name] = displayName
+	}
+
+	return result, nil
+}
+
+func oneOf(candidates ...string) string {
```
laurenzlong, line 93: `oneOf` is a confusing name for this function, maybe `firstNonEmpty` or something

providers/netsuite/internal/suiteql/read.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/amp-labs/connectors/providers/netsuite/internal/shared"
+)
+
+const (
+	// SuiteQL timestamp format for TO_TIMESTAMP function. This seems to be a magic
+	// date format that is accepted by SuiteQL. We should verify it this actually
+	// works for all instances.
+	suiteQLTimestampFormat = "2006-01-02 15:04:05.000000000"
+
+	defaultLimit = 1000
```
laurenzlong, line 23: `maxRecords` or `pageSize` would be a clearer name, I had to look at how it's used in the code to understand what it actually means

laurenzlong (approved):

providers/netsuite/internal/suiteql/metadata.go
```diff
@@ -0,0 +1,75 @@
...
+
+	for field := range record {
+		fields[field] = common.FieldMetadata{DisplayName: field}
+		fieldsMap[field] = field
+	}
+
+	metadata := &common.ObjectMetadata{
+		DisplayName: objectName,
+		Fields:      fields,
+		FieldsMap:   fieldsMap,
```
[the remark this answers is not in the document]
RajatPawar: cool, fixed! Updated to 
```

	fields := common.FieldsMetadata{}

	for field := range response.Items[0] {
		fields.AddFieldWithDisplayOnly(field, field)
	}

	return common.NewObjectMetadata(objectName, fields), nil
```

PR #1943, opened 2025-08-04
[CON-2018] feat: [Highlevel] Read functionality

sanjay-a01:
# Conventions

# Sample Read Response
image

image

# Test case Result
image

### DIFF
--- a/providers/highlevelstandard/handlers.go
+++ b/providers/highlevelstandard/handlers.go
@@ -3,6 +3,7 @@ package highlevelstandard
 import (
 	"context"
 	"net/http"
+	"strconv"
 	"strings"
 
 	"github.com/amp-labs/connectors/common"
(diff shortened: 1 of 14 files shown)


providers/highlevelstandard/handlers.go
```diff
@@ -103,3 +104,100 @@ func (c *Connector) parseSingleObjectMetadataResponse(
...
+			url.WithQueryParam("offset", strconv.Itoa(nextPage))
+		}
+	}
+
+	// For single-segment paths (e.g., "businesses"), the URL must have a trailing slash at the end.
+	// Example: https://highlevel.stoplight.io/docs/integrations/a8db8afcbe0a3-get-businesses-by-location
+	//
+	// For multi-segment paths (e.g., "calendars/groups"), the URL does not require a trailing slash.
+	// Example: https://highlevel.stoplight.io/docs/integrations/89e47b6c05e67-get-groups
+	if !(strings.Contains(params.ObjectName, "/")) {
```
jlimatampersand, line 154: same question from metadata connector, please make sure that order of this slash at the end works with query.

jlimatampersand (approved):

jlimatampersand (approved):

PR #1976, opened 2025-08-15
[CON-1317] feat(nutshell): Read

Cobalt0s:
# Installation

# Conventions

# Read
For each read object:
Ex: image

# Write
For each write object:

# Examples


internal/datautils/timing.go
```diff
@@ -26,3 +27,8 @@ func (timing) FormatRFC3339inUTCWithMilliseconds(input time.Time) string {
 
 	return utcTime.Format("2006-01-02T15:04:05.000Z")
 }
+
+// Unix returns the string representing the number of seconds elapsed since January 1, 1970.
+func (timing) Unix(input time.Time) string {
```
RajatPawar, line 32: interesting!

providers/nutshell/connector.go
```diff
@@ -26,5 +33,34 @@ func constructor(base *components.Connector) (*Connector, error) {
...
 	return connector, nil
 }
+
+func (c *Connector) getReadURL(objectName string) (*urlbuilder.URL, error) {
+	objectPath, err := metadata.Schemas.FindURLPath(common.ModuleRoot, objectName)
+	if err != nil {
+		return nil, err
+	}
+
+	return urlbuilder.New(c.ModuleInfo().BaseURL, apiURIPart, objectPath)
```
RajatPawar, line 65: since there are no modules, it would read easier if this was c.ProviderInfo().BaseURL, but feel free to ignore

providers/nutshell/handlers.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/amp-labs/connectors/internal/datautils"
+	"github.com/amp-labs/connectors/internal/jsonquery"
+	"github.com/amp-labs/connectors/providers/nutshell/internal/metadata"
+	"github.com/spyzhov/ajson"
+)
+
+// The API accepts large page numbers without an error.
+// This value (5,000) was chosen arbitrarily.
+// The actual maximum limit is 10,000; anything higher will return an error.
+const defaultPageSize = "5000"
```
RajatPawar, line 21: if there are no performance implications, can we set this to the max? (9999 or 10000)

providers/nutshell/handlers.go
```diff
@@ -0,0 +1,144 @@
...
+		return urlbuilder.New(params.NextPage.String())
+	}
+
+	// First page
+	url, err := c.getReadURL(params.ObjectName)
+	if err != nil {
+		return nil, err
+	}
+
+	if params.ObjectName == objectNameEvents {
```
RajatPawar, line 50: can you please add context as to why this special case is needed?

RajatPawar (approved):

PR #1980, opened 2025-08-18
[CON-2079] feat: [blackbaud] Metadata functionality

sanjay-a01:
# Conventions

# Sample Metadata response
image

image

image

image

image

image

# Testcase Result
image


test/blackbaud/connector.go
```diff
@@ -0,0 +1,50 @@
...
+
+func GetBlackbaudConnector(ctx context.Context) *blackbaud.Connector {
+	filePath := credscanning.LoadPath(providers.Blackbaud)
+	reader := utils.MustCreateProvCredJSON(filePath, true, fieldSubscriptionKey)
+
+	conn, err := blackbaud.NewConnector(
+		common.ConnectorParams{
+			AuthenticatedClient: utils.NewOauth2Client(ctx, reader, getConfig),
+			Metadata: map[string]string{
+				"BbApiSubscriptionKey": reader.Get(fieldSubscriptionKey),
```
jlimatampersand, line 28: Okay, so this seems a little off from our expectations. Though it seems like they are describing as some sort of key, this really is more like a WorkspaceRef that is dynamic.  I think this logic should be written as the same way use our workspaceRef.

jlimatampersand: Please correct me if I am wrong. If not, change this accordingly.

jlimatampersand: Okay this is new behavior we have never experienced. 

@dionlow Do we have UI library support for this kind of case? cc @RajatPawar @laurenzlong 

This API key seems like from the user, so we should also have our backend to support this. Do we have such a scenario handled?

jlimatampersand (approved):
Thanks for working on this. LGTM!

test/blackbaud/connector.go
sanjay-a01: you may refer the below snippet
image

PR #1981, opened 2025-08-18
[CON-1477] feat: Add calendly read connector

Jkarage:
# Conventions

  Read Live Tests:
  
Screenshot 2025-08-18 at 16 07 05
Screenshot 2025-08-18 at 16 06 41


providers/calendly/handlers.go
```diff
@@ -1 +1,106 @@
...
+		case userBusyTimes:
+			urlBuilder.WithQueryParam("start_time", params.Since.Format(time.RFC3339))
+		case outgoingCommunications:
+			urlBuilder.WithQueryParam("min_created_at", params.Since.Format(time.RFC3339))
+		case scheduledEvents:
+			urlBuilder.WithQueryParam("min_start_time", params.Since.Format(time.RFC3339))
+		}
+	}
+
+	if !params.Until.IsZero() {
```
jlimatampersand, line 71: From our use case, we read schedule since the time specified until current time. So if params.Until.IsZero() returns true, then we should set those end times to time.Now().

jlimatampersand (changes requested):
Sorry, I know this PR has been waiting for a while. Please ping me after addressing the PR. I will review it immediately.

providers/calendly/handlers.go
Jkarage: In Calendly, some objects behave differently in this regard. For instance, reading `scheduled_events` without specifying an "Until" time returns all events, including future ones. This is more expected behavior than limiting the results to the current time (time.Now).

For other objects, the default is time.Now() even if not specified (I haven't tested this, however, as it requires an Enterprise account)

Jkarage: scheduled events: https://developer.calendly.com/api-docs/2d5ed9bbd2952-list-events
user busy times: https://developer.calendly.com/api-docs/5920076156501-list-user-busy-times

jlimatampersand: then we will run into this scenario, like. if some event is scheduled 3 years later today, then we will be reading this object for 3 years every 15 minutes. @RajatPawar

jlimatampersand: @laurenzlong I think it is your call now.

jlimatampersand: @laurenzlong I agree, and that seems like the best compromise, but we will end up receiving growing list of calendar which may hang or get cancelled, but hopefully this does not happen as often.  

@Jkarage That being said, what we should do is to do full fetch, and look through the full fetched records, and filter out records that was created/updated before the Since, so we only return the delta from since. 

@Jkarage Would you mind creating support ticket at Calendly and request query filtering by created/updated time? I think it won't hurt us to ask.

jlimatampersand (changes requested):
Refer to comments and implement full fetch, then programatically filter based on create/update timestamp.

providers/calendly/handlers.go
Jkarage: @jlimatampersand  I've removed the previous usage of start/end event timestamp keys. I've added `updated_at` filtering for `event_types`, as it's the only object that supports sorting by this field. While other objects have `updated_at` fields, they don't support sorting by them. Per @laurenzlong  guidance, we should avoid supporting filtering unsorted responses for now.

PR #1985, opened 2025-08-20
[CON-2080] feat: [blackbaud] Read functionality

sanjay-a01:
# Conventions

# Sample Read response
image

image

# Testcase Result
image

# Note:
In this connector, pagination is supported but is not working, so I have contacted the support team regarding the issue. Once I receive a response from them, I will implement the pagination logic in the read functionality.
image

I received a response from the Blackbaud support team regarding the pagination issue. They mentioned that pagination is not implemented as of now.
image


providers/blackbaud/support.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/internal/components"
+)
+
+func supportedOperations() components.EndpointRegistryInput {
```
jlimatampersand, line 11: can we have some link to docs that may show this list or something we can refer to.

providers/blackbaud/handlers.go
```diff
@@ -68,3 +69,56 @@ func (c *Connector) parseSingleObjectMetadataResponse(
...
+
+	if objectNameWithListResource.Has(params.ObjectName) {
+		url.AddPath("list")
+	}
+
+	url.WithQueryParam("limit", strconv.Itoa(defaultPageSize))
+
+	// offset := 0
+
+	// if params.NextPage != "" {
```
jlimatampersand, line 91: Please remove these comments unless we need them.

jlimatampersand (changes requested):

providers/blackbaud/errors.go
```diff
@@ -0,0 +1,47 @@
...
+	Detail     string `json:"detail,omitempty"`
+	TraceId    string `json:"trace_id,omitempty"` // nolint:tagliatelle,revive
+	SpanId     string `json:"span_id,omitempty"`  // nolint:tagliatelle,revive
+}
+
+func (r ResponseError) CombineErr(base error) error {
+	message := r.Message
+
+	if r.Detail != "" {
+		document, err := goquery.NewDocumentFromReader(bytes.NewReader([]byte(r.Detail)))
```
jlimatampersand, line 37: please add documentation in comment that includes error response schema.

jlimatampersand (changes requested):
Since this PR has been out for a while, feel free to ping me on Slack after adding the documentation. I will review asap.

sanjay-a01:
Hi @jlimatampersand,
          I have attached the information about pagination above.

jlimatampersand (approved):

PR #1986, opened 2025-08-20
[CON-2081] feat: [blackbaud] write functionality

sanjay-a01:
# Conventions

# Sample write response
image

image

# Testcase result
image

# Sample delete response
image

# Testcase result
image


providers/blackbaud/support.go
```diff
@@ -41,12 +42,115 @@ func supportedOperations() components.EndpointRegistryInput {
 		"crm-volmg/volunteers",
 	}
 
+	writeSupport := []string{
+		"crm-adnmg/batches",
```
jlimatampersand, line 46: can you sort this and list below in alphabetical order? It might be even better if you group them by prefixes as you organize these and add a comment about each group for clarity. Please add a comment in the documentation about these groups as well.

sanjay-a01: I added the information about groups in the README file.

PR #1995, opened 2025-08-22
fix: add recordId gitlab write connector

Jkarage:
# Conventions

  ## Context
  Adds recordId value in the write connector result.
  
  ## Live Test
Screenshot 2025-08-22 at 14 02 59

### DIFF
--- a/providers/gitlab/handlers.go
+++ b/providers/gitlab/handlers.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strconv"
 	"time"
 
 	"github.com/amp-labs/connectors/common"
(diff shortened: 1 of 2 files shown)


providers/gitlab/handlers.go
```diff
@@ -166,6 +167,19 @@ func (c *Connector) buildWriteRequest(ctx context.Context, params common.WritePa
 	return http.NewRequestWithContext(ctx, method, url.String(), bytes.NewReader(jsonData))
 }
 
+func retrieveRecordId(data map[string]any) string {
```
jlimatampersand, line 170: Better if we name it parseRecordId or extractRecordId.

jlimatampersand (approved):
Approved, but please resolve my comment before merging.

PR #1998, opened 2025-08-26
fix: Add recordId in dixa connector writeResults

Jkarage:
# Conventions

  Live Tests:
  
Screenshot 2025-08-26 at 13 15 27

### DIFF
--- a/providers/dixa/handlers.go
+++ b/providers/dixa/handlers.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"net/http"
 	"strconv"
+	"strings"
 
 	"github.com/amp-labs/connectors/common"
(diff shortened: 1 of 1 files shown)


providers/dixa/handlers.go
```diff
@@ -111,8 +112,24 @@ func (c *Connector) parseWriteResponse(
...
-		Success: true,
-		Data:    data,
+		Success:  true,
+		Data:     data,
+		RecordId: recordId,
 	}, nil
 }
+
+func parseRecordId(data map[string]any) string {
+	switch v := data["id"].(type) {
```
RajatPawar, line 125: can you please find this field case-insensitively?

RajatPawar (approved):

PR #1999, opened 2025-08-26
[CON-2032] feat: [loxo] Add loxo connector

sanjay-a01:
# Conventions
- [ ] Provider name is camelcase (`goTo` and not `goto`)

### GET
URL: http://localhost:4444/people
image

### GET by Id
URL: http://localhost:4444/people/1388693
image

### POST
URL: http://localhost:4444/people
image

### UPDATE
URL: http://localhost:4444/people/1388693
image

## Pagination
Request the list of people paginated results by using **per_page** to limit the response.
URL:  http://localhost:4444/people?per_page=1
image

Request the next paginated result by using **scroll_id**  in the response and use the value on the **scroll_id**  query param.
URL: http://localhost:4444/people?per_page=1&scroll_id=5B313735363230393730353637302C302E302C313338383337315D
image

# Logo & Icons
Logo for light mode
image

Logo for dark mode
image

Icon for light & dark mode
image

### DIFF
--- a/providers/loxo.go
+++ b/providers/loxo.go
@@ -1,0 +1,54 @@
+package providers
+
+const Loxo Provider = "loxo"
+
+//nolint:lll
+func init() {
+	// Loxo configuration
+	SetInfo(Loxo, ProviderInfo{
+		DisplayName: "Loxo",
+		AuthType:    ApiKey,
+		BaseURL:     "https://pod4.app.loxo.co/api/integration-user-loxo-withampersand-com",
+		ApiKeyOpts: &ApiKeyOpts{
+			AttachmentType: Header,
+			Header: &ApiKeyOptsHeader{
+				Name:        "Authorization",
+				ValuePrefix: "Bearer ",
+			},
+			DocsURL: "https://help.loxo.co/en/articles/446640-loxo-s-open-api#h_668904ffbd",
+		}, Support: Support{
+			BulkWrite: BulkWriteSupport{
+				Insert: false,
+				Update: false,
+				Upsert: false,
+				Delete: false,
+			},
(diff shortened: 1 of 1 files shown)


providers/loxo.go
```diff
@@ -0,0 +1,54 @@
...
+
+const Loxo Provider = "loxo"
+
+//nolint:lll
+func init() {
+	// Loxo configuration
+	SetInfo(Loxo, ProviderInfo{
+		DisplayName: "Loxo",
+		AuthType:    ApiKey,
+		BaseURL:     "https://pod4.app.loxo.co/api/integration-user-loxo-withampersand-com",
```
eberle1080, line 11: This looks wrong, too specific to ampersand. Should this be a template? Also `pod4` seems... off. If they literally all have that, that's fine, but this feels like an implementation detail that we shouldn't be exposing (and indeed one that might break for other users).

eberle1080: The openapi docs (or whatever schema this is from) seem to indicate that domain is something assigned by Loxo on a per-customer basis. I'd suggest you parameterize the entire domain.

See for example: https://loxo.readme.io/reference/disability_statusesindex

Looks like they also make reference to an "agency slug", which I think is where you have the ampersand-specific part. So you'd want to make that a template too.

eberle1080 (changes requested):

providers/loxo.go
sanjay-a01: Sorry @eberle1080, I checked endpoints with the exact base URL and forgot to remove it. Now I update correct implementation of base URL. Kindly check it.

eberle1080 (approved):

PR #2007, opened 2025-08-28
feat: Add MetadataUpsertConnector interface

eberle1080:


connectors.go
```diff
@@ -77,6 +77,14 @@ type ObjectMetadataConnector interface {
 	ListObjectMetadata(ctx context.Context, objectNames []string) (*ListObjectMetadataResult, error)
 }
 
+// MetadataUpsertConnector is an interface that extends the Connector interface with
+// the ability to create metadata.
+type MetadataUpsertConnector interface {
```
laurenzlong, line 82: can this be UpsertMetadataConnector? All the other interfaces are VerbNounConnector

connectors.go
```diff
@@ -77,6 +77,14 @@ type ObjectMetadataConnector interface {
 	ListObjectMetadata(ctx context.Context, objectNames []string) (*ListObjectMetadataResult, error)
 }
 
+// MetadataUpsertConnector is an interface that extends the Connector interface with
+// the ability to create metadata.
```
laurenzlong, line 81:
```suggestion
// the ability to create custom objects and fields in the SaaS instance.
```

common/metadata.go
```diff
@@ -0,0 +1,156 @@
...
+	// UpsertMetadataActionUpdate indicates that the object/field was updated.
+	UpsertMetadataActionUpdate UpsertMetadataAction = "update"
+	// UpsertMetadataActionNone indicates that the object/field was not changed.
+	UpsertMetadataActionNone UpsertMetadataAction = "none"
+)
+
+// UpsertMetadataParams represents parameters for upserting metadata.
+type UpsertMetadataParams struct {
+	// Maps object names to field definitions.
+	FieldsDefinitions map[string][]*FieldDefinition `json:"customFields"`
```
laurenzlong, line 17: Thinking about this some more, I think this should be `Fields` in order to match https://docs.withampersand.com/reference/objects-&-fields/get-object-metadata-for-connection, but the type can still be FieldDefinition, or FieldMetadata might make even more sense given that the method is called UpsertMetadata

laurenzlong: Also:
- I think `[string][]FieldMetadata` would be an easier type to deal with, and the json annotation still says `customFields`

laurenzlong: So this would become:

```
Fields map[string][]FieldMetadata `json:"Fields"`
```

common/metadata.go
```diff
@@ -0,0 +1,156 @@
...
+
+// FieldDefinition represents a field definition. Note that not all
+// providers will support all fields. This is a best-effort attempt
+// to create a common schema for custom fields across providers.
+//
+// In the event that a provider doesn't support a particular field,
+// and assuming that the field has a value, then the value should be
+// ignored, and a warning should be added to the UpsertMetadataResult.
+type FieldDefinition struct {
+	// FieldName is the short name of the field, e.g. "My_Custom_Field".
```
laurenzlong, line 60:
```suggestion
	// FieldName is the identifier of the field, e.g. "My_Custom_Field".
```

common/metadata.go
```diff
@@ -0,0 +1,156 @@
...
+	FieldName string `json:"fieldName"`
+	// DisplayName is the human-readable name of the field, e.g. "My Custom Field".
+	DisplayName string `json:"displayName"`
+	// Description is an optional description of the field.
+	Description string `json:"description,omitempty"`
+	// ValueType is the data type of the field.
+	ValueType FieldType `json:"valueType"`
+	// Length is the maximum length of the field (for string types).
+	Length *int `json:"length,omitempty"`
+	// Precision is the total number of digits (for decimal types).
```
laurenzlong, line 70: is this the number of digits after the decimal point?

common/metadata.go
```diff
@@ -0,0 +1,156 @@
...
+// aspects of the association. This is a best-effort attempt to create
+// a common schema for associations across providers.
+//
+// In the event that a provider doesn't support a particular aspect of
+// the association, and assuming that the field has a value, then the
+// value should be ignored, and a warning should be added to the
+// UpsertMetadataResult.
+type AssociationDefinition struct {
+	// Kind is the high-level association variety (e.g., 'foreignKey', 'lookup', 'ref').
+	// The provider determines the exact behavior.
```
laurenzlong, line 107: AssociationType might be more intuitive than Kind

common/metadata.go
```diff
@@ -0,0 +1,156 @@
...
+	Kind string `json:"kind"`
+	// TargetObject is the name of the referenced/parent object.
+	TargetObject string `json:"targetObject"`
+	// TargetField is the name of the referenced field on the target object.
+	// Defaults to the target's primary key when omitted.
+	TargetField string `json:"targetField,omitempty"`
+	// Relationship cardinality from the referencing field's perspective (e.g., 'many-to-one', 'one-to-one').
+	Cardinality string `json:"cardinality,omitempty"`
```
laurenzlong, line 115: this should be an enum of valid cardinality values, not a freeform string.

common/metadata.go
```diff
@@ -0,0 +1,156 @@
...
+	// OnDelete defines the behavior upon foreign object deletion, where supported.
+	// E.g., 'cascade', 'restrict', 'setNull'.
+	OnDelete string `json:"onDelete,omitempty"`
```
laurenzlong, line 118: same here, should be an enum.

common/metadata.go
```diff
@@ -0,0 +1,156 @@
...
+	// Required means that, if true, a referenced record must exist (i.e., NOT NULL foreign key).
+	Required bool `json:"required,omitempty"`
+	// InverseName is an optional inverse relationship/property name exposed on the target object.
+	InverseName string `json:"inverseName,omitempty"`
```
laurenzlong, line 122: I'm a bit confused by what this is still, is this like a field on the parent object that gets updated to reference the child object? If so, maybe "ReverseLookupFieldName"?

common/metadata.go
```diff
@@ -0,0 +1,156 @@
...
+	Plural   string `json:"plural"`
+}
+
+// UpsertMetadataResult contains results for all created/updated objects and fields.
+type UpsertMetadataResult struct {
+	// Indicates if the upsert operation was successful.
+	Success bool `json:"success"`
+
+	// Maps object names to field upsert results.
+	Fields map[string][]*FieldUpsertResult `json:"fields"`
```
laurenzlong, line 139: I think `map[string]map[string]FieldUpsertResult` might be easier to work with, so the API caller can easily look up the result of a specific field of a specific object, instead of having to loop through an array.

common/metadata.go
```diff
@@ -0,0 +1,156 @@
...
+	// Description is an optional description of the field.
+	Description string `json:"description,omitempty"`
+	// ValueType is the data type of the field.
+	ValueType FieldType `json:"valueType"`
+	// Length is the maximum length of the field (for string types).
+	Length *int `json:"length,omitempty"`
+	// Precision is the total number of digits (for decimal types).
+	Precision *int `json:"precision,omitempty"`
+	// Scale is the number of digits to the right of the decimal point (for decimal types).
+	Scale *int `json:"scale,omitempty"`
```
laurenzlong, line 73: Looking at this list, I wonder if it makes sense to nest certain options that only make sense for specific field types (similar to the [auth mode specific options](https://docs.withampersand.com/reference/connection/generate-a-new-connection) in the GenerateConnection endpoint). 

So this becomes:

```
	FieldName string `json:"fieldName"`
	// DisplayName is the human-readable name of the field, e.g. "My Custom Field".
	DisplayName string `json:"displayName"`
	// Description is an optional description of the field.
	Description string `json:"description,omitempty"`
	// ValueType is the data type of the field.
	ValueType FieldType `json:"valueType"`
	// Length is the maximum length of the field (for string types).
	StringOpts *StringFieldOpts
	FloatOpts *FloatFieldOpts
	SelectOpts *SelectFieldOpts // for singleSelect and multiSelect
	...
	}
```

laurenzlong (changes requested):

common/metadata.go
eberle1080: I believe if you think of associations as a DAG, then as we're pointing at some foreign object, this sets up the foreign object to point back to us (and specifies the name). So it turns it from a 1-way association to a 2-way association.

common/metadata.go
eberle1080: oops good catch. Sounds good.

common/metadata.go
eberle1080: From the salesforce docs:

The precision for number values. Precision is the number of digits in a number. For example, the number 256.99 has a precision value of 5.

common/metadata.go
eberle1080: sure, we can do that, I was trying to make it look like the input. But that works too.

common/metadata.go
eberle1080: Seems do-able, let me see what I can come up with

common/metadata.go
```diff
@@ -0,0 +1,235 @@
...
+const (
+	associationCardinalityNotSet    AssociationCardinality = ""
+	AssociationCardinalityManyToOne AssociationCardinality = "many-to-one"
+	AssociationCardinalityOneToOne  AssociationCardinality = "one-to-one"
+)
+
+func (ac AssociationCardinality) IsValid() bool {
+	switch ac {
+	case associationCardinalityNotSet,
+		AssociationCardinalityManyToOne,
```
laurenzlong, line 36: can you add some comments to explain these? Especially the directionality of ManyToOne

common/metadata.go
```diff
@@ -0,0 +1,235 @@
...
+		return true
+	default:
+		return false
+	}
+}
+
+type AssociationOnDeleteAction string
+
+const (
+	associationOnDeleteActionNotSet   AssociationOnDeleteAction = ""
```
laurenzlong, line 47: can you add some comments to explain these? setNull isn't obvious for example

common/metadata.go
laurenzlong: I think you missed my suggestion to turn this field name into `Fields`

common/metadata.go
```diff
@@ -0,0 +1,235 @@
...
+	ValueType FieldType `json:"valueType"`
+	// Required indicates if the field is required.
+	Required bool `json:"required,omitempty"`
+	// Unique indicates if the field must be unique across all records.
+	Unique bool `json:"unique,omitempty"`
+	// Indexed indicates if the field should be indexed for faster search.
+	Indexed bool `json:"indexed,omitempty"`
+	// StringOptions contains additional options for string fields (if any).
+	StringOptions *StringFieldOptions `json:"stringOptions,omitempty"`
+	// NumericOptions contains additional options for numeric fields (if any).
```
laurenzlong, line 127: Does this include int and float types? if so, can you specify in the comments?

common/metadata.go
```diff
@@ -0,0 +1,235 @@
...
+// In the event that a provider doesn't support a particular option,
+// and assuming that the option has a value, then the value should be
+// ignored, and a warning should be added to the UpsertMetadataResult.
+type StringFieldOptions struct {
+	// Length is the maximum length of the string field.
+	Length *int `json:"length,omitempty"`
+	// Pattern is a regex pattern that the string field value must match (if any).
+	Pattern string `json:"pattern,omitempty"`
+	// Values is a list of allowed values for enum fields (if any).
+	Values []string `json:"values,omitempty"`
```
laurenzlong, line 168: Would this make more sense as `*[]string`?

common/metadata.go
laurenzlong: got it, then I would name it "ReverseLookupFieldName" for clarity.

laurenzlong (approved):

PR #2015, opened 2025-08-29
[CON-2149] feat(Pylon): add Read

immdipu:
# Conventions

   * This is not covered in this PR for simplicity

# Live Tests
### accounts
image

### contacts
image

# Testcase Result
image


providers/pylon/utils.go
```diff
@@ -18,3 +24,29 @@ func inferValueTypeFromData(value any) common.ValueType {
 		return common.ValueTypeOther
 	}
 }
+
+func addIssuesTimeWindowQuery(url *urlbuilder.URL, params common.ReadParams) error {
+	var startTime, endTime time.Time
+
+	if params.Since.IsZero() {
+		// Default to last 29 days to be on safe side.
+		startTime = time.Now().UTC().AddDate(0, 0, -29)
```
eberle1080, line 33: are you doing this due to possible timezone issues? If so you might want to add that to the comment for clarity.

eberle1080 (approved):

providers/pylon/utils.go
immdipu: Yes, the API rejects requests at exactly 30 days despite the docs saying 'less than or equal to'. Using 29 days avoids timing precision issues.

PR #2026, opened 2025-09-01
[CON-2155] feat(salesforce): UpsertMetadata

Cobalt0s:
# Description

Field definitions created successfuly and it is evident in the dashboard.
(In my custom golang script I am able to retrieve the metadata as well).
image
The underlined fields are those that were created via `UpsertMetadata`.
The script covers all data types. With trial and error discovered the right payload for each. In case something is missed let me know. Likely `Associations` for external fields may need more work, it depends on the use cases if there are any!

# Changes
Former CreateMetadata was reused to make SOAP calls.
To simplify the already verbose `~/providers/salesforce`, the upsert functionality lives in its own package. Salesforce struct just delegates to this one-purpose struct. This is done to better isolate functionality, it should be more readable. You would look at this code only if you care about upsert.


providers/salesforce/internal/crm/custom/metadata.go
```diff
@@ -10,32 +11,57 @@ import (
 
 var ErrCreateMetadata = errors.New("error in CreateMetadata")
 
+// UpsertMetadata creates or updates definition of a custom field.
+// https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_upsertMetadata.htm
+func (a *Adapter) UpsertMetadata(
+	ctx context.Context, params *common.UpsertMetadataParams,
+) (*common.UpsertMetadataResult, error) {
+	customFields := NewCustomFieldsPayload(params)
```
Cobalt0s, line 19: Params are turned into the payload that Salesforce understands and SOAP API calls is reused from `CreateMetadata`. This method is the "bridge".

providers/salesforce/internal/crm/custom/metadata.go
```diff
@@ -10,32 +11,57 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+
+	accessToken, present := common.GetAuthToken(ctx)
+	if !present {
+		return nil, common.ErrMissingAccessToken
+	}
+
+	response, err := a.CreateMetadata(ctx, data, accessToken.String())
```
eberle1080, line 31: The naming is slightly confusing. CreateMetadata doesn't necessarily imply upsert behavior. Can you either change the name or add a comment explaining that the behavior is upsert?

providers/salesforce/internal/crm/custom/payload.go
```diff
@@ -0,0 +1,285 @@
...
+	longTextAreaLength = 256
+	fieldTypeText      = "Text"
+	fieldTypeLongText  = "LongTextArea"
+)
+
+type UpsertMetadataPayload struct {
+	XMLName xml.Name `xml:"upsertMetadata"`
+
+	// List of fields to create
+	FieldsMetadata []UpsertMetadataCustomField `xml:"metadata"`
```
Cobalt0s, line 22: We can have `ObjectsMetadata` in the payload. Object creation is very similar. However, the requests for `Field` upsert should be separate from `Object` upsert if both are supplied in params.

providers/salesforce/internal/crm/custom/response.go
```diff
@@ -0,0 +1,103 @@
...
+           <success>true</success>
+       </result>
+       <result>
+           <created>true</created>
+           <fullName>TestObject15__c.Hobby__c</fullName>
+           <success>true</success>
+       </result>
+   </upsertMetadataResponse>
+*/
+func parseResponse(response []byte) (*common.UpsertMetadataResult, error) {
```
Cobalt0s, line 29: `common.UpsertMetadataResult` is completely based on the XML response Salesforce sends.

providers/salesforce/connector.go
```diff
@@ -68,6 +65,8 @@ func NewConnector(opts ...Option) (conn *Connector, outErr error) {
 		XML:  &interpreter.DirectFaultyResponder{Callback: conn.interpretXMLError},
 	}.Handle
 
+	conn.customAdapter = custom.NewAdapter(httpClient, conn.moduleInfo)
```
Cobalt0s, line 68: Custom field creation adapter is always created. The XML Client ownership is moved to it.

test/salesforce/metadata/write/many/main.go
```diff
@@ -0,0 +1,130 @@
...
+	ctx = common.WithAuthToken(ctx, connTest.GetSalesforceAccessToken())
+
+	params := &common.UpsertMetadataParams{
+		Fields: map[string][]common.FieldDefinition{
+			"TestObject15__c": {
+				{
+					FieldName:   "Birthday__c",
+					DisplayName: "Birthday",
+					Description: "Story describing birthday",
+					ValueType:   common.ValueTypeString,
```
Cobalt0s, line 39: This string field will be a `Text` because the length is under 256.

test/salesforce/metadata/write/many/main.go
```diff
@@ -0,0 +1,130 @@
...
+					Indexed:     false,
+					StringOptions: &common.StringFieldOptions{
+						Length: goutils.Pointer(30),
+					},
+				},
+				{
+					FieldName:   "Hobby__c",
+					DisplayName: "Hobby",
+					Description: "Your hobby description",
+					ValueType:   common.ValueTypeString,
```
Cobalt0s, line 51: This will result into `LongTextArea` because the length of the field is larger than 256.

test/salesforce/metadata/write/many/main.go
```diff
@@ -0,0 +1,130 @@
...
+					Indexed:     false,
+					StringOptions: &common.StringFieldOptions{
+						Length: goutils.Pointer(444),
+					},
+				},
+				{
+					FieldName:   "Age__c",
+					DisplayName: "Age",
+					Description: "How many years you lived.",
+					ValueType:   common.ValueTypeInt,
```
Cobalt0s, line 63: Int creation requires default value, precision and scale properties. Precision must be larger than the scale.

test/salesforce/metadata/write/many/main.go
```diff
@@ -0,0 +1,130 @@
...
+						DefaultValue: goutils.Pointer(18.0),
+						Precision:    goutils.Pointer(3),
+						Scale:        goutils.Pointer(2),
+					},
+				},
+				{
+					FieldName:   "Interests__c",
+					DisplayName: "Interests",
+					Description: "Topics that are of interest.",
+					ValueType:   common.ValueTypeMultiSelect,
```
Cobalt0s, line 77: MutliSelect is similar to SingleSelect. You can even update one type into the other without a problem. Under the hood MutliSelect also requires VisibleLines property which is automatically supplied.

test/salesforce/metadata/write/many/main.go
```diff
@@ -0,0 +1,130 @@
...
+						Values:           []string{"art", "travel", "swimming"},
+						ValuesRestricted: true,
+						DefaultValue:     goutils.Pointer("art"),
+					},
+				},
+				{
+					FieldName:   "IsReady__c",
+					DisplayName: "IsReady",
+					Description: "Indicates the readiness for next steps.",
+					ValueType:   common.ValueTypeBoolean,
```
Cobalt0s, line 91: This is a checkbox. True or False value is read via string options -> default value.

test/salesforce/metadata/write/many/main.go
```diff
@@ -0,0 +1,130 @@
...
+					Indexed:     false,
+					StringOptions: &common.StringFieldOptions{
+						DefaultValue: goutils.Pointer("false"),
+					},
+				},
+				{
+					FieldName:   "Connection__c",
+					DisplayName: "Connection",
+					Description: "Connection to other objects.",
+					ValueType:   common.ValueTypeOther,
```
Cobalt0s, line 103: The value type doesn't matter. A present `association` option will create either `Lookup` field or `IndirectLookup`.

test/salesforce/metadata/write/multi-select/main.go
```diff
@@ -0,0 +1,88 @@
...
+	"syscall"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/internal/goutils"
+	"github.com/amp-labs/connectors/providers/salesforce"
+	connTest "github.com/amp-labs/connectors/test/salesforce"
+	"github.com/amp-labs/connectors/test/utils"
+)
+
+func main() {
```
Cobalt0s, line 18: Create Single select with 2 options then update to have 3 options. Works as expected.

Image

providers/salesforce/internal/crm/custom/payload.go
```diff
@@ -0,0 +1,290 @@
...
+
+	if *definition.StringOptions.Length < longTextAreaLength {
+		field.Type = fieldTypeText
+
+		return field
+	}
+
+	field.Type = fieldTypeLongText
+	// Must specify 'visibleLines' for a CustomField of type LongTextArea
+	field.VisibleLines = goutils.Pointer("10")
```
eberle1080, line 158: Might be nice if users could specify this. Could you add this to the interface type (one of the string options)?

eberle1080 (approved):
LGTM. I have a few suggestions but I don't see any blockers.

providers/salesforce/internal/crm/custom/metadata.go
Cobalt0s: Yeah, its not even upsert. This method is SOAP API call. It can accept anything. I was using this method to request all custom fields.
I will add a better comment and rename it.

providers/salesforce/internal/crm/custom/payload.go
Cobalt0s: Ok, I am using 10 as default to avoid the error. User may specify `StringFieldOptions.NumDisplayLines` to override this.

See the override result:
image

PR #2027, opened 2025-09-03
[CON-2157] feat(hubspot): UpsertMetadata

Cobalt0s:


providers/hubspot/internal/custom/concurrent.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+// concurrentProcessing executes the provided process function in parallel for each element in the list.
+//
+// All calls share the same context. If any invocation returns a non-nil error,
+// the context is cancelled and the function returns immediately with that error.
+//
+// The process function receives a context, the element, and a mutex to safely
+// synchronize access to shared resources (e.g., maps) across goroutines.
+func concurrentProcessing[T any, P func(context.Context, T, *sync.Mutex) error](
```
eberle1080, line 15: There's a better version of this in the server repo:

https://github.com/amp-labs/server/blob/main/shared/utils/simultaneously.go

It's better because it lets you specify an upper limit for concurrency. We have to be very careful about exploding goroutines - remember that we process requests in parallel, and that even our backend workers tend to have about 1000 threads which can all potentially run at the same time. So the multiplicity is a real problem for us.

You don't need to copy it verbatim - adapt it to your needs and copy-paste whatever makes sense. But the max parallelism is really important.

eberle1080: Also, I worry about the increased possibility here of being rate-limited.

providers/hubspot/internal/custom/payload.go
```diff
@@ -0,0 +1,221 @@
...
+	case common.FieldTypeDateTime:
+		return "datetime"
+	case common.FieldTypeSingleSelect:
+		return "radio"
+	case common.FieldTypeMultiSelect:
+		return "select"
+	case common.FieldTypeInt, common.FieldTypeFloat:
+		return "number"
+	default:
+		return "text"
```
eberle1080, line 75:
```suggestion
func matchFieldType(definition common.FieldDefinition) (string, error) {
	switch definition.ValueType {
	case common.FieldTypeString:
		return "text", nil
	case common.FieldTypeBoolean:
		return "booleancheckbox", nil
	case common.FieldTypeDate:
		return "date", nil
	case common.FieldTypeDateTime:
		return "datetime", nil
	case common.FieldTypeSingleSelect:
		return "radio", nil
	case common.FieldTypeMultiSelect:
		return "select", nil
	case common.FieldTypeInt, common.FieldTypeFloat:
		return "number", nil
	default:
		return "", fmt.Errorf("unrecognized value type: %s", definition.ValueType)
	}
}
```

One thing I noticed in my testing is that this needs to fail loudly. At one point I tried to use a type of "number" (obviously an error) and the system silently decided to use text instead, resulting in a very unintuitive error (Salesforce said something about how text fields weren't allowed to have a precision value - which is true but the root cause is non-obvious).

I'd prefer if this failed more loudly.

eberle1080 (changes requested):
Mostly looks good. My only concerns are the unbounded parallelism and the silent failure mode.

providers/hubspot/internal/custom/payload.go
Cobalt0s: Good point, I was not sure what to put here. Error will be the cleanest implementation. I will update Salesforce code too.

Cobalt0s: Addressed both connectors:

image

image

PR #2047, opened 2025-09-11
[CON-2212] feat: [jobber] Added Metadata functionality

sanjay-a01:
# Conventions

# Sample  Metadata result
image

image


providers/jobber/utils.go
```diff
@@ -0,0 +1,34 @@
+package jobber
+
+import "github.com/amp-labs/connectors/internal/datautils"
+
+const apiVersion = "2025-01-20"
+
+var objectNameMapping = datautils.NewDefaultMap(map[string]string{ //nolint:gochecknoglobals
```
jlimatampersand, line 7: can you add `https://developer.getjobber.com/docs` and say that thid soc will just give high level overview in code comments and mention that the full list can be retrieved after logging in.

jlimatampersand (approved):
Please make sure to resolve the comment before merging.

PR #2052, opened 2025-09-12
[CON-2022] feat: [Highlevel] Enable read and write

sanjay-a01:
# Installation
# For HighLevelStandard Connector
image

image

# For HighLevelWhiteLabel Connector
image

image

# Conventions

# For HighLevelStandard Connector
# Read
For each read object:
- [x] Insert screenshot of metadata fields from read object installation.
image

- [x] Insert screenshot of Svix destination.
- [x] Screenshot of operations on dashboard
image

image

image

image

# Write
For each write object:
- [x] Insert screenshot of dashboard.
- [x] Insert screenshot of HTTP call.

image
image

image
image

image
image

image
image

# For HighLevelWhiteLabel Connector
# Read
For each read object:
- [x] Insert screenshot of metadata fields from read object installation.
image

- [x] Insert screenshot of Svix destination.
- [x] Screenshot of operations on dashboard

image

image

image

image

# Write
For each write object:
- [x] Insert screenshot of dashboard.
- [x] Insert screenshot of HTTP call.

image
image

image
image

image
image

# Examples
Link pull request to testdata repo which contains installation file: https://github.com/amp-labs/testdata/pull/87

**Note:**
In this PR, I have added support for some read and write endpoints, and removed certain delete endpoints because they require query parameters or body parameters in the request. 
And also chekced all this and tested it.

### DIFF
--- a/providers/highlevelstandard/support.go
+++ b/providers/highlevelstandard/support.go
@@ -43,12 +43,12 @@ func supportedOperations() components.EndpointRegistryInput {
 		"store/shipping-zone",
 		"store/shipping-carrier",
 		"store/store-setting",
-		"snapshots",
 		"surveys",
 		"users",
 		"workflows",
 		"locations/search",
 		"custom-menus",
+		"marketplace/billing/charges",
 	}
 
 	writeSupport := []string{
@@ -86,6 +86,8 @@ func supportedOperations() components.EndpointRegistryInput {
 		"products",
 		"products/collections",
 		"store/shipping-zone",
+		"store/shipping-carrier",
+		"store/store-setting",
 	}
 
 	deleteSupport := []string{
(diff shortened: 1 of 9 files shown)


providers/highlevelstandard/support.go
```diff
@@ -96,21 +98,10 @@ func supportedOperations() components.EndpointRegistryInput {
 		"calendars/groups",
 		"contacts",
 		"associations",
-		"associations/relations",
```
jlimatampersand, line 99: So you removed these because the requests required query parameters and body parameters? Add documentation/api reference comment in code that explains these. 

Also ,rather than removing the objects, just comment them out and add explanation here.

jlimatampersand: And whay you wrote on the description is true, then we should try to accommodate those query/body parameters if needed. Can you revise this PR to implement that instead of removing the objects? If there is a strict limitation that you are not able to achieve it, please add that explanation in a code comment as well.

jlimatampersand (changes requested):

providers/highlevelstandard/support.go
sanjay-a01: The removed objects require locationId, altId, and altType in both the query parameters and the body parameters. Previously, I assumed that common.DeleteParams did not accept query and body parameter values, which is why I removed those endpoints. Now, I have re-enabled the removed objects by providing the required parameter values. For the query and body parameters **locationId and altId** have same vlaue, we obtain them from the user during installation, while altType is always set to the value **location**.

PR #2054, opened 2025-09-12
[CON-2118] feat(SageIntacct): add Read

immdipu:
# Conventions

   * This is not covered in this PR for simplicity

# Live Tests
### general-ledger/account
image

### company-config/contact
image

#### company-config/employee
image

# Testcase Result
image

### DIFF
--- a/providers/sageintacct/support.go
+++ b/providers/sageintacct/support.go
@@ -8,6 +8,7 @@ import (
 	"github.com/amp-labs/connectors/internal/components"
 )
 
+// API reference: https://developer.sage.com/intacct/docs/openapi/z-indexes/
 // nolint:funlen,maintidx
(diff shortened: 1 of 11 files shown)


providers/sageintacct/support.go
```diff
@@ -0,0 +1,409 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/internal/components"
+)
+
+// nolint:funlen,maintidx
+func supportedOperations() components.EndpointRegistryInput {
```
jlimatampersand, line 12: please add link to doc in code comment. thanks for ordering them!

jlimatampersand (approved):
Please make sure to resolve comment before merging.

PR #2061, opened 2025-09-16
[ENG-2553] update(pipeliner): Workspace,Region,OpenAPI

Cobalt0s:
# Description
When I was doing some changes to the workspace/region in ProviderInfo, I decided that it would be shorter and more consistent to reuse the code in `components`. There is no need to review those in details. The unit-tests work as expected and live tests too.

Here is a list of all the important changes:
* Connector is now using:
  * `common.ConnectorParams`
  * `components` lib is used for Read/Write/Delete/Metadata
  * Latest OpenAPI was downloaded, schemas regenerated (live test caught that it was outdated, so I had to fix it).
* Added Workspace as a metadata variable (main goal)
* Region can be specified in metadata too (important change)
image
image

# Live Tests
## Metadata
image

## Read
image

### Write
image


Cobalt0s (commented):
There are many things updated, a lot is removed. The gist, important bits of the changes are commented below:

RajatPawar (approved):
Awesome! Approving this since all the tests are passing & no builders are actively using this on Ampersand, but once this is merged in & live (I'll let you know when it is), could you please deploy an integration on Ampersand to make sure that metadata/read/write all work as expected?

PR #2073, opened 2025-09-17
[CON-2221] feat: Add ZohoDesk Metadata connector

Jkarage:
# Conventions

Adds the zoho Desk module, plus it's Metadata connector:

Live Tests:
ZohoCRM:
Screenshot 2025-09-17 at 18 44 10

Screenshot 2025-09-17 at 18 45 35

ZOhoDesk:
Screenshot 2025-09-17 at 18 45 01

Screenshot 2025-09-17 at 18 45 48

### DIFF
--- a/providers/zoho.go
+++ b/providers/zoho.go
@@ -8,6 +8,7 @@ const (
 	ZohoDeskV2 common.ModuleID = "zohodesk"
 )
 
+// nolint: funlen
 func init() {
 	// Zoho configuration
 	SetInfo(Zoho, ProviderInfo{
--- a/providers/zoho.go
+++ b/providers/zoho.go
@@ -3,9 +3,9 @@ package providers
 import "github.com/amp-labs/connectors/common"
 
 const (
-	Zoho       Provider        = "zoho"
-	ZohoCRM    common.ModuleID = "zohocrm"
-	ZohoDeskV2 common.ModuleID = "zohodesk"
+	Zoho     Provider        = "zoho"
+	ZohoCRM  common.ModuleID = "crm"
+	ZohoDesk common.ModuleID = "desk"
 )
 
 // nolint: funlen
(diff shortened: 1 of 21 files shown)


eberle1080 (approved):

providers/zoho.go
```diff
@@ -53,5 +60,38 @@ func init() {
 				LogoURL: "https://res.cloudinary.com/dycvts6vp/image/upload/v1722471890/media/zoho_1722471890.svg",
 			},
 		},
+		DefaultModule: ZohoCRM,
+		Modules: &Modules{
+			ZohoCRM: {
+				BaseURL:     "https://www.zohoapis.com",
```
eberle1080, line 66: This needs to be a template value due to multi-region support. See the non-module baseurl

providers/zoho.go
```diff
@@ -53,5 +60,38 @@ func init() {
...
+				DisplayName: "Zoho CRM",
+				Support: Support{
+					Read:      true,
+					Subscribe: false,
+					Write:     true,
+				},
+			},
+			ZohoDeskV2: {
+				BaseURL:     "https://desk.zoho.com",
```
eberle1080, line 75: Guessing this has similar multi-datacenter rules as the other stuff, please check

providers/zoho.go
eberle1080: We do have customers in EU regions as well as US regions, and they're not at all compatible (as in, tokens generated for one don't work with the other), which is why this matters.

eberle1080 (changes requested):
Was about to approve but the regional stuff is a blocker. 

See https://www.zoho.com/accounts/protocol/oauth/multi-dc.html

providers/zoho/location.go
```diff
@@ -13,6 +13,7 @@ var (
 
 type LocationDomains struct {
 	ApiDomain   string `json:"api_domain"`
+	DeskDomain  string `json:"desk_domain"`
 	TokenDomain string `json:"token_domain"`
 }
```
Jkarage, line 18: I assume this will be used i the server. I have added the `DeskDomain`.

PR #2076, opened 2025-09-17
fix: Add support for users object zoho connector

Jkarage:
This fixes the users object only, the rest (lowercased) don't use the same metadata retrieval approach as the rest. 

Live Tests:
Meatadata:
Screenshot 2025-09-17 at 21 54 30

Read:
Screenshot 2025-09-17 at 21 55 15

Screenshot 2025-09-17 at 22 03 15

### DIFF
--- a/providers/zohocrm/url.go
+++ b/providers/zohocrm/url.go
@@ -9,14 +9,20 @@ import (
 )
 
 func (c *Connector) buildReadURL(config common.ReadParams) (*urlbuilder.URL, error) {
+	var obj = config.ObjectName
+
 	// Check if we're reading the next-page.
 	if len(config.NextPage) > 0 {
 		return urlbuilder.New(config.NextPage.String())
 	}
 
-	// Object names in ZohoCRM API are case sensitive.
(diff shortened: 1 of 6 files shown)


providers/zohocrm/url.go
```diff
@@ -9,14 +9,20 @@ import (
...
 	if len(config.NextPage) > 0 {
 		return urlbuilder.New(config.NextPage.String())
 	}
 
-	// Object names in ZohoCRM API are case sensitive.
-	// Capitalizing the first character of object names to form correct URL.
-	obj := naming.CapitalizeFirstLetterEveryWord(config.ObjectName)
+	// objects like user, org, org/currencies, __features,
+	// uses lowecased object-names
+	if config.ObjectName != users {
```
RajatPawar, line 21: can you please add this for `org` too?

RajatPawar (approved):

providers/zohocrm/url.go
Jkarage: Adding, though org do not work with the current metadata collection approach.

PR #2080, opened 2025-09-18
[CON-2222] feat: Add zoho desk read connector

Jkarage:
# Conventions

  Live Tests:
  
Screenshot 2025-09-22 at 16 21 24


providers/zoho/read.go
```diff
@@ -48,3 +91,34 @@ func constructHeaders(config common.ReadParams) []common.Header {
 
 	return []common.Header{}
 }
+
+// Manual incremental synchronization implementation for Apollo
```
eberle1080, line 95: Shouldn't this be Zoho Desk and not Apollo?

test/apollo/write/write.go
```diff
@@ -107,7 +107,7 @@ func testCreatingAccounts(ctx context.Context) error {
 	params := common.WriteParams{
 		ObjectName: "accounts",
 		RecordData: map[string]any{
-			"name":         "Google",
+			"name":         "Googlettt",
```
eberle1080, line 110: Did you mean to check this in?

eberle1080 (changes requested):
Looks like some apollo-related changes made it in, please remove those

test/apollo/write/write.go
Jkarage: No, thanks my bad.

providers/zoho/read.go
Jkarage: copy and paste.

PR #2081, opened 2025-09-18
[CON-2231] feat(sellsy): Custom fields Read,Meatadata

Cobalt0s:
# Installation

# Conventions

# Read
For each read object:
Ex: image

# Write
For each write object:

# Examples


providers/sellsy/handlers.go
```diff
@@ -26,33 +26,55 @@ func (c *Connector) buildReadRequest(ctx context.Context, params common.ReadPara
...
+	readURL, err := c.getReadURL(params.ObjectName)
 	if err != nil {
 		return nil, err
 	}
 
-	req, err := http.NewRequestWithContext(ctx, method, endpointURL, bytes.NewReader(jsonData))
+	readURL.WithQueryParam("limit", defaultPageSize)
+
+	// Request custom field embedding.
+	definitions, err := c.fetchCustomFieldDefinitions(ctx, []string{params.ObjectName})
```
jlimatampersand, line 63: looking into this, I think this is going to become very expensive especially when we are going to make `connector.Read(...)` call in a loop. Then, this will make 2-3 additional calls on top of the api we are trying to hit. 

Knowing this definition is going to won't change much, I think it might be better to somehow cache it to save API quota.

providers/sellsy/custom.go
```diff
@@ -0,0 +1,239 @@
...
+	"item":        "items",
+	"opportunity": "opportunities",
+	"staff":       "staffs",
+	"task":        "tasks",
+}
+
+// https://docs.sellsy.com/api/v2/#operation/search-custom-fields
+func (c *Connector) fetchCustomFieldDefinitions( // nolint:cyclop
+	ctx context.Context, objectNames []string,
+) (map[string]customFieldDefinitions, error) {
```
jlimatampersand, line 28: can we actually make this customFieldDefinition part of the connector struct and cache it, so we only get to make api call once?

jlimatampersand (changes requested):

jlimatampersand (approved):

PR #2085, opened 2025-09-22
[CON-2247] feat(amplitude): add Write

immdipu:
## Checklist
- [ ] Read supports pagination

## Sample Write Response
image

The third API response returns raw text instead of JSON, so the connector throws an error even though the request was successful

### DIFF
--- a/providers/amplitude/handlers.go
+++ b/providers/amplitude/handlers.go
@@ -12,6 +12,7 @@ import (
 
 const (
 	apiV2 = "2"
+	apiV3 = "3"
 )
 
 func (c *Connector) buildSingleObjectMetadataRequest(ctx context.Context, objectName string) (*http.Request, error) {
--- a/providers/amplitude/handlers.go
+++ b/providers/amplitude/handlers.go
@@ -1,0 +1,114 @@
+package amplitude
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/naming"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+)
+
+const (
+	apiV2 = "2"
+	apiV3 = "3"
+)
+
+func (c *Connector) buildSingleObjectMetadataRequest(ctx context.Context, objectName string) (*http.Request, error) {
+	url, err := c.constructURL(objectName)
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url.String(), nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return req, nil
+}
+
+func (c *Connector) parseSingleObjectMetadataResponse(
+	ctx context.Context,
+	objectName string,
+	request *http.Request,
+	response *common.JSONHTTPResponse,
+) (*common.ObjectMetadata, error) {
+	objectMetadata := common.ObjectMetadata{
+		Fields:      make(map[string]common.FieldMetadata),
+		DisplayName: naming.CapitalizeFirstLetterEveryWord(objectName),
+	}
+
+	res, err := common.UnmarshalJSON[map[string]any](response)
+	if err != nil {
+		return nil, common.ErrFailedToUnmarshalBody
+	}
+
+	if res == nil || len(*res) == 0 {
+		return nil, common.ErrMissingExpectedValues
+	}
+
+	responseField := objectResponseField.Get(objectName)
+
+	data, ok := (*res)[responseField].([]any)
+	if !ok {
+		return nil, fmt.Errorf("couldn't convert the data response field data to an array: %w", common.ErrMissingExpectedValues) // nolint:lll
+	}
+
+	if len(data) == 0 {
+		return nil, fmt.Errorf("%w: could not find a record to sample fields from", common.ErrMissingExpectedValues)
+	}
+
+	firstRecord, ok := data[0].(map[string]any)
+	if !ok {
+		return nil, fmt.Errorf("couldn't convert the first record data to a map: %w", common.ErrMissingExpectedValues)
+	}
+
+	for field, value := range firstRecord {
+		objectMetadata.Fields[field] = common.FieldMetadata{
+			DisplayName:  field,
+			ValueType:    inferValueTypeFromData(value),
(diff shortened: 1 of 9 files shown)


providers/amplitude/handlers.go
```diff
@@ -0,0 +1,243 @@
...
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+
+	return req, nil
+}
+
+func createRequestForParamsPayload(ctx context.Context, url *urlbuilder.URL, params common.WriteParams,
+) (*http.Request, error) {
+	recordMap, ok := params.RecordData.(map[string]interface{})
+	if !ok {
+		return nil, errors.New("params.RecordData is not a map[string]interface{}") //nolint: err113
```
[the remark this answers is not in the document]
Cobalt0s: Yes use `any` instead of `interface{}`

providers/amplitude/support.go
```diff
@@ -33,6 +44,18 @@ var objectResponseField = datautils.NewDefaultMap(datautils.Map[string, string]{
...
+	return objectname
+})
+
+var payloadKeyMapping = datautils.NewDefaultMap(datautils.Map[string, string]{ //nolint:gochecknoglobals
+	objectNameAttribution: "event",
+}, func(key string) string {
+	return ""
+})
+
 func supportedOperations() components.EndpointRegistryInput {
```
Cobalt0s, line 59: `supportedOperations` can be omitted in the future. You can use `components.NewEmptyEndpointRegistry`.
There are more issues of supporting this feature for all connectors. It seems that we cannot have this feature consistently.

And of course, uou can preserve it for this PR, thats fine.

providers/amplitude/utils.go
Cobalt0s, line 41: Possible alternative:
```
if objectName == objectNameEvents {
  objectName = objectName + "/list"
}
return urlbuilder.New(c.ProviderInfo().BaseURL, path, "api", apiVersion, objectName)
```

Note you don't need these for sure:
```
if err != nil {
    return nil, err
}  
return url, nil
```
you can return immediately.

providers/amplitude/utils.go
```diff
@@ -40,3 +45,35 @@ func (c *Connector) constructURL(objectName string) (*urlbuilder.URL, error) {
 
 	return url, nil
 }
+
+// Helper to extract API key (username) from the auth client.
+func extractAPIKey(client common.AuthenticatedHTTPClient, ctx context.Context) (string, error) {
```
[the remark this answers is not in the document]
Cobalt0s: This method is no longer used, don't forget to remove it

test/amplitude/write/main.go
```diff
@@ -0,0 +1,134 @@
...
+func main() {
+	if err := run(); err != nil {
+		slog.Error(err.Error())
+	}
+}
+
+func run() error {
+	ctx := context.Background()
+
+	conn := amplitude.GetAmplitudeConnector(ctx)
```
Cobalt0s, line 25: I see that your live tests do not use API V2 which would rely on the access token.
It would be nice if you could add one scenario like that.
The access token will be injected into context like so: https://github.com/amp-labs/connectors/blob/c58a4bad834f4056d872a0d94a0567dfbd6b335f/test/salesforce/metadata/write/multi-select/main.go#L32
And then access token itself for the live tests comes from creds.json like so:
https://github.com/amp-labs/connectors/blob/c58a4bad834f4056d872a0d94a0567dfbd6b335f/test/salesforce/connector.go#L54-L58

Cobalt0s (commented):
Left some comments. Overall looks good!

providers/amplitude/handlers.go
```diff
@@ -112,3 +117,85 @@ func (c *Connector) parseReadResponse(
...
+	}
+
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+
+	return req, nil
+}
+
+func createRequestForParamsPayload(ctx context.Context, url *urlbuilder.URL, params common.WriteParams,
+) (*http.Request, error) {
+	recordMap, ok := params.RecordData.(map[string]any)
```
Cobalt0s, line 186: There is a utility that could come in handy:
https://github.com/amp-labs/connectors/blob/c58a4bad834f4056d872a0d94a0567dfbd6b335f/common/types.go#L193-L196

providers/amplitude/write.go
```diff
@@ -0,0 +1,110 @@
...
+
+	contentType := resp.Header.Get("Content-Type")
+	isJSON := strings.Contains(contentType, "application/json")
+
+	if isJSON || len(contentType) == 0 {
+		if result, err := c.tryParseAsJSON(params, bodyBytes); err == nil {
+			return result, nil
+		}
+
+		if isJSON {
```
Cobalt0s, line 50: Checking `isJSON` within the if block which already does it, is not easy to read. You should reorder the lines in some way. I understand you algortithm like this:
```
switch {
case isJSON:
	return c.mustParseJSON(params, bodyBytes)
case len(contentType) == 0:
	return c.tryParseJSONOrFallback(params, bodyBytes, resp.StatusCode) // a combination of default and JSON, hoping that language is JSON even though no media states that
default:
	return c.handleNonJSONSuccess(resp.StatusCode, bodyBytes), nil
}
```

test/amplitude/write/main.go
```diff
@@ -0,0 +1,136 @@
...
+func main() {
+	if err := run(); err != nil {
+		slog.Error(err.Error())
+	}
+}
+
+func run() error {
+	ctx := context.Background()
+
+	ctx = common.WithAuthToken(ctx, amplitude.GetAmplitudeAPIkey())
```
Cobalt0s, line 25: Thanks!

test/amplitude/connector.go
```diff
@@ -31,3 +30,14 @@ func GetAmplitudeConnector(ctx context.Context) *amplitude.Connector {
 
 	return conn
 }
+
+func GetAmplitudeAPIkey() common.AuthToken {
+	reader := getAmplitudeJSONReader()
+	return common.AuthToken(reader.Get(credscanning.Fields.Username))
```
Cobalt0s, line 36: I thought linter would complain that there is no space before the `return` 🤔

Cobalt0s (approved):
Looks good.

PR #2088, opened 2025-09-22
[CON-2223] feat: Add zoho desk write connector

Jkarage:
# Conventions

Live Tests:
Screenshot 2025-09-22 at 17 50 35


eberle1080 (approved):

providers/zoho/write.go
```diff
@@ -49,15 +50,70 @@ const dataKey = "data"
 func (c *Connector) Write(ctx context.Context, config common.WriteParams) (*common.WriteResult, error) {
 	ctx = logging.With(ctx, "connector", "zoho CRM")
```
eberle1080, line 51: Can you make this also aware of `moduleId`?

PR #2108, opened 2025-09-24
fix (Asana): add missing write objects

immdipu:
### Projects
image

### custom_fields
image

### portfolios
image

### tasks
image
image

### teams
image


Cobalt0s (approved):
Looks good

PR #2115, opened 2025-09-26
[CON-1332] feat(salesflare): Metadata

Cobalt0s:
# Installation

# Conventions

# Read
For each read object:
Ex: image

# Write
For each write object:

# Examples


scripts/openapi/salesflare/metadata/main.go
```diff
@@ -0,0 +1,105 @@
...
+
+func main() {
+	schemas := staticschema.NewMetadata[staticschema.FieldMetadataMapV2]()
+	registry := datautils.NamedLists[string]{}
+
+	for _, object := range Objects() {
+		objectName := object.ObjectName
+
+		if object.Problem != nil {
+			slog.Error("schema not extracted",
```
eberle1080, line 32: Avoid calling slog directory, use `logging.Logger(ctx)` - that way we can add attributes to the log messages (such as customer id) from the server.

eberle1080 (approved):

PR #2126, opened 2025-09-30
[CON-2040] feat: [loxo] Metadata functionality

sanjay-a01:
# Conventions

# Sample metadata response
image

image

image

# Testcase result
image

### DIFF
--- a/providers/loxo/README.md
+++ b/providers/loxo/README.md
@@ -1,0 +1,66 @@
+# Loxo connector
+
+
+## Supported Objects 
+Below is an exhaustive list of objects & methods supported on the objects
+
+
+| Object                                    | Resource                                 | Method     |
+| ----------------------------------------- | ---------------------------------------- | ---------- |
+| activity_types                            | activity_types                           | read       |
+| address_types                             | address_types                            | read       |
+| bonus_payment_types                       | bonus_payment_types                      | read       |
+| bonus_types                               | bonus_types                              | read       |
+| companies                                 | companies                                | read       |
+| company_global_statuses                   | company_global_statuses                  | read       |
+| company_types                             | company_types                            | read       |
+| compensation_types                        | compensation_types                       | read       |
+| countries                                 | countries                                | read       |
+| currencies                                | currencies                               | read       |
+| deal_workflows                            | deal_workflows                           | read       |
+| deals                                     | deals                                    | read       |
+| disability_statuses                       | disability_statuses                      | read       |
+| diversity_types                           | diversity_types                          | read       |
+| dynamic_fields                            | dynamic_fields                           | read       |
+| education_types                           | education_types                          | read       |
+| email_tracking                            | email_tracking                           | read       |
+| email_types                               | email_types                              | read       |
+| equity_types                              | equity_types                             | read       |
+| ethnicities                               | ethnicities                              | read       |
+| fee_types                                 | fee_types                                | read       |
+| form_templates                            | form_templates                           | read       |
+| forms                                     | forms                                    | read       |
+| genders                                   | genders                                  | read       |
+| job_categories                            | job_categories                           | read       |
+| job_contact_types                         | job_contact_types                        | read       |
+| job_owner_types                           | job_owner_types                          | read       |
+| job_statuses                              | job_statuses                             | read       |
+| job_types                                 | job_types                                | read       |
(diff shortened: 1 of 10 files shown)


providers/loxo/connector.go
```diff
@@ -0,0 +1,57 @@
...
+	*components.Connector
+
+	// Require authenticated client
+	common.RequireAuthenticatedClient
+	common.RequireMetadata
+
+	// Supported operations
+	components.SchemaProvider
+
+	AgencySlug string
```
RajatPawar: can you add a comment here / doc link that explains what this is and how to get it?

providers/loxo/errors.go
```diff
@@ -0,0 +1,26 @@
+package loxo
+
+import (
+	"fmt"
+
+	"github.com/amp-labs/connectors/common/interpreter"
+)
+
+var errorFormats = interpreter.NewFormatSwitch( // nolint:gochecknoglobals
```
RajatPawar: is this any different than the common error handler?

providers/loxo/README.md
```diff
@@ -0,0 +1,66 @@
...
+| job_categories                            | job_categories                           | read       |
+| job_contact_types                         | job_contact_types                        | read       |
+| job_owner_types                           | job_owner_types                          | read       |
+| job_statuses                              | job_statuses                             | read       |
+| job_types                                 | job_types                                | read       |
+| jobs                                      | jobs                                     | read       |
+| people                                    | people                                   | read       |
+| people/emails                             | people/emails                            | read       |
+| person_phones                             | person_phones                            | read       |
+| people/update_by_email                    | people/update_by_email                   | read       |
```
RajatPawar: Either we should merge this as an update with the people object, or not support this object, because this is an action/verb, not a resource

RajatPawar (changes requested):
Some clarifications required!

providers/loxo/connector.go
sanjay-a01: The agency slug is the path segment of the Loxo API URL. We must specify this in all endpoints.
image

To get this see the below snippet
image

providers/loxo/utils.go
```diff
@@ -0,0 +1,73 @@
+package loxo
+
+import (
+	"github.com/amp-labs/connectors/internal/datautils"
+)
+
+var objectsNodePath = datautils.NewDefaultMap(map[string]string{ //nolint:gochecknoglobals
+	"activity_types":                 "",
```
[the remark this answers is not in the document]
sanjay-a01: Some object responses include a node path, while others do not.
For ex. **activity_types** object have no node path
image

**companies** object have node path
image

For this reason, I have defined some object have values and empty values.

providers/loxo/handlers.go
```diff
@@ -0,0 +1,76 @@
...
+	"net/http"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/naming"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+	"github.com/amp-labs/connectors/internal/jsonquery"
+)
+
+func (c *Connector) buildSingleObjectMetadataRequest(ctx context.Context, objectName string) (*http.Request, error) {
+	if objectWithPrefixValue.Has(objectName) {
```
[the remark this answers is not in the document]
sanjay-a01: No, we are just checking whether the objectName is in the objectWithPrefixValue variable. There is no specific reason. As shown below, the scorecards name is constant, but only the suffix value changes. So, I decided to check the suffix of the objectName provided by the user, and if the object exists in the objectWithPrefixValue variable, add the prefix value scorecards/.
image

Also, these objects are actions/verbs, not resources. So I think we should remove them, right?

providers/loxo/connector.go
RajatPawar: Thanks! Can you please add this context as a comment in code too?

providers/loxo/utils.go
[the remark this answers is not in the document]
RajatPawar: makes sense! please add this as comments to the code too. It'll make the connector code easier to understand. Thanks

providers/loxo/handlers.go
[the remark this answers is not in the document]
sanjay-a01: Understood man

PR #2131, opened 2025-10-02
[CON-2277] gong: revise calls read 

immdipu:
## Description
This PR updates the calls object read endpoint to use /v2/calls/extensive to retrieve additional fields (context and parties).

The calls endpoint uses POST method which supports incremental read with both Since and Until parameters, so I added `Until` as well.

## Test result
image
image

### DIFF
--- a/providers/gong/metadata/schemas.json
+++ b/providers/gong/metadata/schemas.json
@@ -6,12 +6,11 @@
       "objects": {
         "calls": {
           "displayName": "Calls",
-          "path": "/calls/extensive",
+          "path": "/calls",
           "responseKey": "calls",
           "fields": {
             "calendarEventId": "calendarEventId",
             "clientUniqueId": "clientUniqueId",
-            "context": "context",
             "customData": "customData",
             "direction": "direction",
             "duration": "duration",
@@ -20,7 +19,6 @@
             "language": "language",
             "media": "media",
             "meetingUrl": "meetingUrl",
-            "parties": "parties",
             "primaryUserId": "primaryUserId",
             "purpose": "purpose",
             "scheduled": "scheduled",
--- a/providers/gong/metadata/schemas.json
+++ b/providers/gong/metadata/schemas.json
@@ -6,11 +6,12 @@
       "objects": {
         "calls": {
           "displayName": "Calls",
-          "path": "/calls",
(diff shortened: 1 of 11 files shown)


providers/gong/read.go
```diff
@@ -52,10 +55,48 @@ func (c *Connector) Read(ctx context.Context, config common.ReadParams) (*common
...
+
+	metaData, err := jsonquery.Convertor.ObjectToMap(metaDataNode)
+	if err != nil {
+		return nil, err
+	}
+
+	delete(record, "metaData")
+	maps.Copy(record, metaData)
+
+	return record, nil
```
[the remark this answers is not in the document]
immdipu: The record map is created fresh in `flattenMetaData` from `ObjectToMap(node)`. Nothing else is using it so i think safe to modify directly. Creating a new map would just add extra memory allocation.

providers/gong/metadata/schemas.json
```diff
@@ -6,11 +6,12 @@
       "objects": {
         "calls": {
           "displayName": "Calls",
-          "path": "/calls",
+          "path": "/calls/extensive",
           "responseKey": "calls",
           "fields": {
             "calendarEventId": "calendarEventId",
             "clientUniqueId": "clientUniqueId",
+            "context": "context",
```
laurenzlong, line 14: please remove `context` and `parties` from this list since they are not true fields.

laurenzlong (changes requested):

providers/gong/read.go
```diff
@@ -52,10 +54,35 @@ func (c *Connector) Read(ctx context.Context, config common.ReadParams) (*common
...
+
+// extractMetaDataFields extracts only the metaData fields for ReadResultRow.Fields.
+// ReadResultRow.Raw will contain the full response including context and parties.
+func extractMetaDataFields(node *ajson.Node) (map[string]any, error) {
+	metaDataNode, err := jsonquery.New(node).ObjectOptional("metaData")
+	if err != nil {
+		return nil, err
+	}
+
+	// if metaData is not present, return the record as is
```
laurenzlong, line 82:
```suggestion
	// if metaData is not present, return an empty record
```

laurenzlong (approved):

PR #2138, opened 2025-10-06
[CON-2165] feat: [chorus] Enable proxy

sanjay-a01:
# Conventions
- [ ] Provider name is camelcase (`goTo` and not `goto`)
- [ ] DocsURLs actually link to user-friendly documentation (do not link to very technical documentation)
- [ ] All required metadata variables are templated (`{{.var}}`) and defined in `ProviderInfo.Metadata`
- [ ] Basic smoke tests added (valid request succeeds, invalid request fails)
- [ ] Docs and logos attached or linked

# Note
 Testing is not possible.

### DIFF
--- a/providers/chorus.go
+++ b/providers/chorus.go
@@ -24,7 +24,7 @@ func init() {
 				Upsert: false,
 				Delete: false,
 			},
-			Proxy:     false,
+			Proxy:     true,
 			Read:      false,
 			Subscribe: false,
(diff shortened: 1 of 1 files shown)


jlimatampersand:
@RajatPawar Re-routing to you for review since I am OOO.

RajatPawar (approved):

PR #2140, opened 2025-10-06
[CON-1381] feat: [facebook] Read functionality

sanjay-a01:
# Conventions

# Sample read response
image

# Test case result
image


RajatPawar (changes requested):
We cannot support this properly because all the endpoints require an ID in the path which can be the ad account or the business account. We can fetch this information from the API, but there can be multiple accounts and the user needs to make a choice about which account they're interested in. Blocking this from merging

PR #2276, opened 2025-11-07
feat(codec): Dual marshal preserve RawJSON

Cobalt0s:
# Description

This PR introduces `codec.RawJSON[T any]`.

### Key changes

* **`codec.RawJSON[T any]`**
  A generic wrapper for JSON marshalling/unmarshalling that preserves data in two forms:
  * the original, unmodified `map[string]any`
  * the parsed concrete type `T`
  This enables lossless handling of responses. It’s particularly useful when multiple representations are possible—for example, when parsing provider responses where `T` represents a success model, but failures require access to the raw payload (e.g. to construct `common.NewHTTPError`).

image

* **JSON comparison utilities**
  Adds `mockutils.JSONComparator`, a lightweight helper for comparing Go values against JSON string representations (commonly used in tests).
  * `Equals` compares single values
  * `ListsEqual` compares slices

  This simplifies assertions where expected data is expressed as **JSON literals**.

# Live Tests

image

image

### DIFF
--- a/internal/codec/record.go
+++ b/internal/codec/record.go
@@ -8,38 +8,57 @@ import (
 	"github.com/amp-labs/connectors/internal/datautils"
 )
 
-// DecoratedRecord merges a dynamic record with a structured extension,
-// producing a single flattened JSON object when marshaled.
+// DecoratedRecord merges a dynamic record with a typed extension and marshals them as one flat JSON object.
 //
-// It embeds a [common.Record] containing user-defined key–value pairs
-// and a typed struct [T] representing schema-bound fields that must
-// coexist with those user-supplied values in the API payload.
+// Important: zero values in Extension are still marshaled unless the field is tagged with `omitempty`.
+// Those zero values can overwrite values already present in Record during the merge unless that is your intention.
 //
-// When marshaled, fields from both Record and Extension are serialized
-// at the same level in the resulting JSON. This allows connectors to
-// enrich arbitrary record data with well-defined metadata or attributes.
+// Merge rules:
+//   - Keys present only in Record are preserved in the output.
+//   - Fields from Extension are added to the output.
+//   - Matching keys favor Extension values.
+//   - Nested structs in Extension are supported.
+//   - Zero-value fields in Extension still override unless omitted from JSON.
 //
 // Example:
 //
-//	type MyPayloadForRecord = codec.DecoratedRecord[RecordExtension]
-//
 //	type RecordExtension struct {
-//		ObjectName string `json:"objectName"`
+//	    ObjectName string `json:"objectName"`
 //	}
 //
-//	record := common.Record{"name": "Bob"}
-//	extension := RecordExtension{ObjectName: "users"}
-//	item := codec.DecoratedRecord[RecordExtension]{Record: record, Extension: extension}
+//	item := codec.NewDecoratedRecord(
+//		map[string]any{
+//			"name": "Bob",
+//		},
+//		RecordExtension{
+//			ObjectName: "users",
+//		})
 //
-//	// Output:
-//	// {"name": "Bob", "objectName": "users"}
+//	// JSON:
+//	// {"name":"Bob","objectName":"users"}
 type DecoratedRecord[T any] struct {
 	common.Record
 
 	Extension T
 }
 
-func (d *DecoratedRecord[T]) MarshalJSON() ([]byte, error) {
+// NewDecoratedRecord creates a DecoratedRecord from a base record and a typed extension.
+//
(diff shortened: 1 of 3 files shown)


Cobalt0s:
* **#2277** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2277?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2276** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2276?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2276?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3009** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3009?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#3014](https://github.com/amp-labs/connectors/pull/3014) <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3014?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#2948** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2948?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2876** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2876?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

anushat05 (commented):
The build is failing for this should we address that first?

Cobalt0s:
@anushat05 The original reason to include the utility is gone with Salesforce using new create method. However, the utility could still come in handy. I think it should replace some of the `jsonquery` package in favour of concrete Golang data types and still have the raw `map[string]any` alongside. The concrete data type doesn't have to be exhaustive. All the "extra" fields not unmarshalled into concrete golang struct will still be in Raw.

Cobalt0s:
@anushat05 the PR is still relevant and was useful in the Microsoft Susbcribe implementation.

Cobalt0s:
@jlimatampersand 
The util is already used by Salesforce it enhances it a bit.

`codec.RawJSON[YourType]` allows to deal with Raw representation of data `map[string]any` **alongside** the concrete `Struct Type`.
* Conrete Type allows easy access to fields.
* The Raw map will always contain all data so nothing is "wasted". Useful when you want the original data before Marshalling happened into the struct.

internal/codec/raw.go
```diff
@@ -0,0 +1,99 @@
...
+//
+//	type Contact struct {
+//		codec.RawJSON[UserData]
+//	}
+//
+//	type UserData struct {
+//		ID   string `json:"id"`
+//		Name string `json:"name"`
+//	}
+type RawJSON[T any] struct {
```
jlimatampersand: I see what this util is trying to achieve, but is this a necessary util?

I guess the win is there when you edit records in such a way that you are strictly adding items and fields, but if this wrapper is used when removing or deleting for a purpose, this is going to be very buggy and will introduce friction. 

You will understand this utility, but many others will easily use it incorrectly unless they actually read this logic fully themselves, which will not be the case. Most likeley 90% of the contributors won't read the description as carefully as the logic itself.

jlimatampersand: Convince me or fix the logic so I can approve. I am open to suggestions.

jlimatampersand (changes requested):

internal/codec/raw.go
Cobalt0s: I understand that it can be misused. Probably it should be more dedicated, or some features stripped away towards more immutable design. I started small, then got some code generated and thought it fits the concept.

(1)
For Microsoft Batch I need Dual Unmarshal -- therefore, it will be part of that Microsoft impl and **not** a utility:
image
Usage:
image

(2)
I changed the code and added tests for `DecoratedRecord` which has only one responsibility and that won't be growing.

Summary:
* The existing `DecoratedRecord` is used by Salesforce. The [mock test](https://github.com/amp-labs/connectors/blob/main/providers/salesforce/batch-write_test.go#L156) validates that Salesforce write operation sends correct payload..
* The "DecoratedRecord" is doing only one thing -- enhancing any map with "sidecar data" to produce JSON for payload:
  *  `map[string]any` + `your extension,decoration struct` => `JSON`
Example:
image

PS: There are many write payloads that are "decorated", so this idea of `DecoratedRecord` makes payload construction automatic without any `map[string]any` assertions, key checks, etc -- it just combines 2 JSON representations. Let me know if it is an overkill :sweat_smile: .

jlimatampersand (approved):

PR #2504, opened 2025-12-28
[CON-2757] feat: add stripe subscription support

Lafetz:
# Stripe Webhook Subscription Support

## Summary
This PR implements comprehensive webhook subscription support for the Stripe connector, enabling users to subscribe to webhook events, update existing subscriptions, delete subscriptions, and verify webhook message signatures.

## Implementation Details

- **Single Endpoint Strategy**: Stripe allows multiple events per endpoint, so we create one endpoint with all requested events rather than multiple endpoints
- **Composite IDs**: Each object gets a copy of the endpoint response with a composite ID (`endpointID:objectName`) to track per-object subscriptions while sharing the actual endpoint

### 1. Subscribe 
Screenshot from 2026-01-20 21-26-57

Screenshot from 2026-01-20 21-27-20

### 2. Update Subscription and Delete Subscription

- Updates existing webhook subscriptions by merging new event requests with existing subscriptions
- Preserves subscriptions for objects not being updated
- Deletes webhook subscriptions for specified objects
- Supports partial deletion (removes events for specific objects from the endpoint)
- Handles full deletion (removes entire endpoint when no events remain)
- Extracts base endpoint ID from composite IDs for proper endpoint management

Screenshot from 2026-01-20 21-31-53

Screenshot from 2026-01-20 21-32-12

Screenshot from 2026-01-20 21-32-50

Screenshot from 2026-01-20 21-34-13

### 3. verify and process 
Screenshot from 2026-01-21 20-34-53


test/stripe/subscribe/webhook-verify/main.go
```diff
@@ -0,0 +1,167 @@
...
+
+func main() {
+	ctx, done := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
+	defer done()
+
+	utils.SetupLogging()
+
+	conn := connTest.GetStripeConnector(ctx)
+
+	webhookURL := ""
```
Cobalt0s: The validation breaks because this value is empty. You should get this variable from the command arguments or from the creds file. But it should also state how to get this value so that anybody can run this script.

test/stripe/subscribe/webhook-verify/main.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	slog.Info("Creating subscription...")
+	subscribeResult, err := conn.Subscribe(ctx, subscribeParams)
+	if err != nil {
+		logging.Logger(ctx).Error("Error subscribing", "error", err)
+		return
+	}
+
+	stripeResult := subscribeResult.Result.(*stripe.SubscriptionResult)
```
Cobalt0s: There is a condition below that check `if subscribeResult != nil`. This is therefore in wrong order.
The check below can never be false. Because current line would panic in that case before reaching the condition and executing the else branch.

test/stripe/subscribe/main.go
```diff
@@ -0,0 +1,89 @@
...
+	"syscall"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/logging"
+	"github.com/amp-labs/connectors/providers/stripe"
+	connTest "github.com/amp-labs/connectors/test/stripe"
+	"github.com/amp-labs/connectors/test/utils"
+)
+
+func main() {
```
Cobalt0s: The flow of the script is identical to `testscenario.ValidateCreateUpdateDelete`
You Subscribe/Update/Delete. Please create helpers. It will make the scripts shorter and we will focus only on the data. The `SubscriptionEvents` is the "core". The rest is mechanical and you can rely on `SubscribeConnector interface`.

PS: The scripts run for me locally no problem. Just like partial-delete too.

test/stripe/subscribe/partial-delete/main.go
```diff
@@ -0,0 +1,274 @@
...
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/logging"
+	"github.com/amp-labs/connectors/internal/datautils"
+	"github.com/amp-labs/connectors/providers/stripe"
+	connTest "github.com/amp-labs/connectors/test/stripe"
+	"github.com/amp-labs/connectors/test/utils"
+)
+
+func main() {
```
Cobalt0s: Do you think we can generalize it? Also it would be nice to describe what the script does there are lots of things going on and not everybody has time to read the code. Think of documenting as a thought process "cache" for a reader.

What I see is that you are trying to create a big set of subscriptions and then slowly prune it and validate the provider's data state at each stage.
```
Order of invoked methods:
--------------
Subscribe
DeleteSubscription
GetWebhookEndpoint // unique to stripe (Reading)
DeleteSubscription
GetWebhookEndpoint // unique to stripe (Reading)
DeleteSubscription
--------------
Subscribe
DeleteSubscription
GetWebhookEndpoint // unique to stripe (Reading)
--------------
Subscribe
GetWebhookEndpoint // unique to stripe (Reading)
DeleteSubscription
--------------
```

I think in general you are trying:
```
--------------
Subscribe
DeleteSubscription
GetWebhookEndpoint // unique to stripe (Reading)
(
  DeleteSubscription
  GetWebhookEndpoint // unique to stripe (Reading)
) --> repeat from [0 to n] times
--------------
```
`GetWebhookEndpoint` is a reading plus validation step. You could create generalized scenario where you can specify many (Delete+Validate) clauses. This way the script will focus on data only. Ex: initially I subscribe to A,B,C. Then when I remove A I expect B,C, etc. That should be the whole test. This would be perfect to scale and use by all the providers that do subscriptions.

providers/stripe/vars.go
```diff
@@ -0,0 +1,18 @@
...
+import "errors"
+
+var (
+	errInvalidRequestType     = errors.New("invalid request type")
+	errMissingParams          = errors.New("missing required parameters")
+	errUnsupportedEventType   = errors.New("unsupported event type")
+	errObjectTypeMismatch     = errors.New("object type mismatch")
+	errObjectFieldNotFound    = errors.New("object field not found in metadata")
+	errNoValuesDefined        = errors.New("no values defined for object field in metadata")
+	errNotYetImplemented      = errors.New("not yet implemented")
```
Cobalt0s: There is `common.ErrNotImplemented`.

providers/stripe/record.go
```diff
@@ -0,0 +1,24 @@
...
+//
+//nolint:revive
+func (c *Connector) GetRecordsByIds(
+	_ context.Context,
+	_ string,
+	_ []string,
+	_ []string,
+	_ []string,
+) ([]common.ReadResultRow, error) {
+	return nil, fmt.Errorf("%w", errNotYetImplemented)
```
Cobalt0s: I expect a follow up PR. For now you can leave a TODO comment.

providers/stripe/subscriptionEvent.go
```diff
@@ -0,0 +1,198 @@
...
+	if !isDataMap {
+		return "", fmt.Errorf("%w: expected map[string]any, got %T", errTypeMismatch, data)
+	}
+
+	obj, isObjMap := dataMap["object"].(map[string]any)
+	if !isObjMap {
+		return "", fmt.Errorf("%w: expected map[string]any for data.object, got %T", errTypeMismatch, dataMap["object"])
+	}
+
+	id, isString := obj["id"].(string)
```
Cobalt0s: Alternatively:
```
node, err := jsonquery.Convertor.NodeFromMap(m)
return jsonquery.New(node, "data", "object").StrWithDefault("id", "")
```
Many consecutive `.(map[string]any)` look verbose to me. So this comment is only a matter of personal taste.

Cobalt0s (commented):
This is the first batch, I will need to review other files later.

test/stripe/subscribe/partial-delete/main.go
Lafetz: yes that's the idea, since it's quite complex when doing partial delete

test/stripe/subscribe/partial-delete/main.go
```diff
@@ -0,0 +1,289 @@
...
+	WebhookURL string
+
+	BuildRequest func(webhookURL string) any
+
+	Steps []partialDeleteStep
+
+	OnSubscribe func(result *common.SubscriptionResult)
+}
+
+func validatePartialDelete(
```
Cobalt0s: Moves this to `testscenario` package.

test/stripe/subscribe/webhook-verify/main.go
```diff
@@ -0,0 +1,137 @@
...
+		slog.Info("Webhook verified and processed", "bodySize", len(body))
+		return nil
+	})
+
+	http.HandleFunc("/", handler)
+
+	slog.Info("Starting webhook server on :8080/")
+	slog.Info("Press Ctrl+C to stop")
+
+	server := &http.Server{Addr: ":8080"}
```
Cobalt0s: The "server" part would also be reused. With all of its logs, error checks.
```
handler := testscenario.CreateWebhookHandler(ctx, conn, verificationParams, func(body []byte) error {
	slog.Info("Webhook verified and processed", "bodySize", len(body))
	return nil
})

server := testwebhookserver.StartAndWait(ctx, 8080, handler)
```

test/stripe/subscribe/main.go
```diff
@@ -0,0 +1,77 @@
...
+	suite := testscenario.SubscribeTestSuite{
+		WebhookURL: webhookURL,
+		BuildRequest: func(url string) any {
+			return &stripe.SubscriptionRequest{
+				WebhookEndPoint: url,
+			}
+		},
+	}
+
+	testscenario.ValidateSubscribeUpdateDelete(ctx, conn, initialEvents, updateEvents, suite)
```
Cobalt0s: Perfect!

providers/stripe/subscribe.go
```diff
@@ -0,0 +1,467 @@
...
+	result, err := buildSubscriptionResult(response, params.SubscriptionEvents)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build subscription result: %w", err)
+	}
+
+	return result, nil
+}
+
+// buildRequestedEventSet builds a set of requested events from subscription events.
+func buildRequestedEventSet(subscriptionEvents map[common.ObjectName]common.ObjectEvents) (map[string]bool, error) {
```
Cobalt0s: This is used by `subscribeUpdate` so it should be moved there. No usages in this file.

Cobalt0s (approved):
Amazing work! 🚀

I have some ideas but that has nothing to do with Stripe itself and more about the structure of connectors, utils. 

The PR is correct and because it is big I would not modify it any further. This Stripe implementation should be an example for future Subscribe connectors!

Cobalt0s (commented):
Reviewed (`GetRecordsByIds`)
No action required on the comments below.

The implementation itself looks solid. I'd prefer not to increase the scope of this PR further.

Establishing patterns for future Subscribe implementations (structure, helpers, tests, etc.) feels like something we should tackle in a dedicated refactor, where we can be more deliberate about the design.

jlimatampersand (changes requested):
Thanks for working on this. There are a few comments I left. Please resolve and re-request for a review. 

Thanks!

jlimatampersand (approved):

PR #2579, opened 2026-01-21
[CON-2843] feat(superSend): superSend-metadata

ejoo:
## Description

  - [x] Unit tests cover metadata logic (placed in /providers/superSend)
  - [x] live tests (placed in /test/superSend)

  ## Tests

  ### Unit Test
Screenshot 2026-01-21 at 11 24 46 PM

  ### Live Test
Screenshot 2026-01-21 at 11 23 27 PM
Screenshot 2026-01-21 at 11 23 44 PM
Screenshot 2026-01-21 at 11 23 53 PM


providers/supersend/metadata/schemas.json
```diff
@@ -0,0 +1,929 @@
+{
```
Cobalt0s: How was this file created? Could you add a README here. As an alternative we could do the read sampling.

Cobalt0s (approved):
LGTM

PR #2639, opened 2026-02-11
[CON-2802] feat: update the Braze provider Base URL provisioning.

Jkarage:
- Braze have several regional data centers. The EU data center is unique as it uses `.eu` domain. The rest uses `.com`. 
- This lets the customer provide full base url domain.

Proxy Tests:
Screenshot 2026-02-11 at 10 45 49

### DIFF
--- a/providers/braze.go
+++ b/providers/braze.go
@@ -2,6 +2,9 @@ package providers
 
 const Braze Provider = "braze"
 
+// BrazeEU serves customers using braze's france data-center only, the rest should use Braze.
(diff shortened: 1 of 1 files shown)


RajatPawar (changes requested):
It would be better to have a separate EU connector and reuse the underlying connector, so that's we can treat the data differently if needed in the future

RajatPawar (approved):
Actually, approving this to unblock, but please make the changes before merging it in

PR #2668, opened 2026-02-19
[ENG-4138] [Attio] Restore event parsing methods

jlimatampersand:
From the previous PR, some logics were unexpectedly removed, so this PR restores those event parsing logics.

Plus, this PR is adding logic to get objectnames and field names from the ObjectMetadata object because the event itself does not provide object names and updated field names.

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -764,6 +764,10 @@ type FieldMetadata struct {
 	// True means the field must have a value, false means it is optional.
 	IsRequired *bool
 
+	// FieldId is the provider's unique identifier for this field.
+	// It is nil when the provider does not expose field identifiers.
+	FieldId *string //nolint:revive
+
 	// Values is a list of possible values for this field.
 	// It is applicable only if the type is either singleSelect or multiSelect, otherwise slice is nil.
 	Values []FieldValue
--- a/providers/attio/subscriptionEvent.go
+++ b/providers/attio/subscriptionEvent.go
@@ -5,7 +5,10 @@ import (
 	"crypto/hmac"
 	"crypto/sha256"
 	"encoding/hex"
+	"errors"
 	"fmt"
+	"maps"
+	"strings"
 
 	"github.com/amp-labs/connectors/common"
 )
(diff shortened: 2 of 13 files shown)


common/types.go
```diff
@@ -654,6 +654,10 @@ type FieldMetadata struct {
 	// True means the field must have a value, false means it is optional.
 	IsRequired *bool
 
+	// FieldId is the provider's unique identifier for this field.
+	// It is nil when the provider does not expose field identifiers.
+	FieldId *string //nolint:revive
```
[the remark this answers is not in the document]
jlimatampersand: Other lines are okay, but I wanted to confirm this field with you since this will be to the builders if available. For now, only for Attio as we need this to parse field names for update event.

providers/attio/subscriptionEvent.go
```diff
@@ -61,31 +70,228 @@ func (c *Connector) VerifyWebhookMessage(
 	return true, nil
 }
 
+func (evt SubscriptionEvent) PreLoadData(data *common.SubscriptionEventPreLoadData) error {
```
jlimatampersand: all these functions in this file are just being restored; they were already approved before. Don't waste time reviewing too much here.

providers/attio/subscriptionEvent.go
```diff
@@ -61,31 +70,228 @@ func (c *Connector) VerifyWebhookMessage(
...
+
+	name, ok := eventsMap["event_type"].(string)
+	if !ok {
+		return "", fmt.Errorf("%w: expected %T, got %T", errTypeMismatch, name, eventsMap["event_type"])
+	}
+
+	objectName := strings.Split(name, ".")[0]
+
+	return objectName, nil
+}
```
immdipu: I believe this will be updated later. Based on our last discussion, the webhook response doesn't always contain the `objectName`. I can see a new `GetObjectNameFromObjectMetadata` function. I'm guessing that will be used somewhere in this.

immdipu (approved):

PR #2677, opened 2026-02-21
feat(components): Constructor with params

Cobalt0s:
# Description

Connector constructors currently lack access to params, which forces verbose setup outside the constructor.
This PR allows constructors to receive params directly, making initialization simpler and more natural.

Before:
image

After:
image

### DIFF
--- a/internal/components/connector.go
+++ b/internal/components/connector.go
@@ -15,6 +15,27 @@ var ErrSupportNotConfigured = errors.New("support not configured")
 // TODO: Convert this to a type alias for easier usage when we go to go1.24: https://go.dev/doc/go1.24#language
 type ConnectorConstructor[T any] func(*Connector) (*T, error)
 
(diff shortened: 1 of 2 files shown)


internal/components/connector.go
```diff
@@ -31,10 +52,25 @@ var (
 // Initialize initializes a connector with the given provider and parameters
 // by using Connector as a base type. It runs the constructor with the connector
 // and returns the connector as the specified T type.
+//
+// Deprecated: use Init.
 func Initialize[T any](
```
RajatPawar, line 57: could you please file a follow-up task to use `Init` in place of Initialize?

RajatPawar (approved):
🎉

internal/components/connector.go
Cobalt0s: Will do!

PR #2680, opened 2026-02-23
[CON-2413] feat(Expensify): add Write

immdipu:
## Checklist

## Sample Write response 
```
    ~/Desktop/Ampersand/connectors    dipu/expensify-write *11  go run ./test/expensify/write                                                                                                                  ✔  system   11:32:04 PM  
2026/02/23 23:32:07 INFO Creating new policy...
{
  "success": true,
  "recordId": "02D88BBBFEE6EBED",
  "data": {
    "policyID": "02D88BBBFEE6EBED",
    "policyName": "My New Policy",
    "responseCode": 200
  }
}
2026/02/23 23:32:09 INFO Creating new report...
{
  "success": true,
  "recordId": "R00HKdQdAzza",
  "data": {
    "reportID": "R00HKdQdAzza",
    "reportName": "Name of the report",
    "responseCode": 200
  }
}
```

### DIFF
--- a/providers/expensify/write.go
+++ b/providers/expensify/write.go
@@ -25,14 +25,14 @@ func (c *Connector) Write(ctx context.Context, params common.WriteParams) (*comm
 	//nolint:bodyclose
 	resp, err := c.executeRequest(ctx, body)
 	if err != nil {
-		return nil, fmt.Errorf("error executing read request: %w", err)
+		return nil, fmt.Errorf("error executing write request: %w", err)
 	}
 
 	bodyBytes := common.GetResponseBodyOnce(resp)
 
 	var result map[string]any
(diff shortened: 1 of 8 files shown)


providers/expensify/write.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/amp-labs/connectors/common"
+)
+
+func (c *Connector) Write(ctx context.Context, params common.WriteParams) (*common.WriteResult, error) {
```
RajatPawar, line 11: Seems like no updates are supported - we should error out if we receive an update in that case

RajatPawar (approved):

PR #2688, opened 2026-02-24
[CON-2872] feat(Workday) : Workday read

ejoo:
## Installation                                                                                                                                                              
  Workday using OAuth2.                                                                                                                                                        
                                                                                                                                                                               
  ## Conventions                                                                                                                                                               
  - [x] Unit tests cover read/metadata logic (placed in /test/)

  ## Known Limitations
  - Workday REST API v1 does not natively support time-based filtering on most endpoints
  - Write/Delete support will follow in a separate PR

  ## Read

  ### Unit Test
  
<img width="1440" height="593" alt="Screenshot 2026-02-24 at 11 14 56 PM" src="https://github.com/user-attachments/assets/c81f21e2-ea44-43a1-a99f-b06f48b96bda" />


  ### Live Test
Live tests will be performed after getting the credentials.


jlimatampersand (approved):

providers/workday/handlers.go
```diff
@@ -0,0 +1,121 @@
...
+	if params.NextPage != "" {
+		url, err := urlbuilder.New(params.NextPage.String())
+		if err != nil {
+			return nil, err
+		}
+
+		return http.NewRequestWithContext(ctx, http.MethodGet, url.String(), nil)
+	}
+
+	url, err := urlbuilder.New(c.ProviderInfo().BaseURL, "ccx", "api", "v1", c.tenantName, params.ObjectName)
```
jlimatampersand, line 30: documentaion link in code comment please.

PR #2701, opened 2026-02-25
[CON-2872] feat(Workday): Add Workday write and delete connector

ejoo:
## Installation                                                                                                                                                              
  Workday using OAuth2.                                                                                                                                                        
                                                                                                                                                                               
  ## Conventions                                                                                                                                                               
  - [x] Provider errors are mapped if non-standard (404 → `ErrNotFound`)
  - [x] Unit tests cover read/write/delete/metadata logic (placed in /providers/workday)
  - [x] Appropriate object names are used. Objects need to be resources, not actions (`workers` and not `workers.list`).
     - [ ] Different base URLs
     - [ ] Object name collisions

  ## Write
  ### Unit Tests
<img width="1190" height="437" alt="Screenshot 2026-02-25 at 11 45 56 PM" src="https://github.com/user-attachments/assets/867e7e1a-22ab-4849-88e0-97c4bc2a0c98" />

  ### Live Tests
Live tests will be performed after getting the credentials.
  ## Delete
  ### Unit Tests
<img width="1190" height="316" alt="Screenshot 2026-02-25 at 11 46 22 PM" src="https://github.com/user-attachments/assets/6cc7756c-6a35-4d33-95ce-9fa8d0b2bf80" />

  ### Live Tests
Live tests will be performed after getting the credentials.

### DIFF
--- a/providers/workday/handlers.go
+++ b/providers/workday/handlers.go
@@ -1,7 +1,9 @@
 package workday
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"net/http"
 	"strconv"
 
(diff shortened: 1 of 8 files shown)


providers/workday/handlers.go
```diff
@@ -112,6 +114,80 @@ func makeNextRecordsURL(reqURL *urlbuilder.URL) common.NextPageFunc {
 	}
 }
 
+func (c *Connector) buildWriteRequest(ctx context.Context, params common.WriteParams) (*http.Request, error) {
+	method := http.MethodPost
+
+	url, err := urlbuilder.New(c.ProviderInfo().BaseURL, "ccx", "api", "v1", c.tenantName, params.ObjectName)
```
jlimatampersand, line 120: can we add documenation url describing this endpoint?

providers/workday/handlers.go
```diff
@@ -112,6 +114,80 @@ func makeNextRecordsURL(reqURL *urlbuilder.URL) common.NextPageFunc {
...
+	return &common.WriteResult{
+		Success:  true,
+		RecordId: recordID,
+		Data:     data,
+	}, nil
+}
+
+func (c *Connector) buildDeleteRequest(ctx context.Context, params common.DeleteParams) (*http.Request, error) {
+	url, err := urlbuilder.New(
+		c.ProviderInfo().BaseURL, "ccx", "api", "v1", c.tenantName, params.ObjectName, params.RecordId,
```
jlimatampersand, line 171: same here. docs url in comment please.

jlimatampersand (approved):
Approved, but please respond to my comment and add docs url before merging. thank you!

PR #2714, opened 2026-03-03
feat(pipeliner): Add incremental reading

Cobalt0s:
# Live Tests
Read contacts since some timestamp:
image

image

Newest records are returned first due to the sort query param. A portion of all existing records is return later than the since timestamp:
image

### DIFF
--- a/providers/pipeliner/handlers.go
+++ b/providers/pipeliner/handlers.go
@@ -6,16 +6,18 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"strconv"
+	"strings"
 
 	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/readhelper"
 	"github.com/amp-labs/connectors/common/urlbuilder"
+	"github.com/amp-labs/connectors/internal/datautils"
 	"github.com/amp-labs/connectors/internal/jsonquery"
 	"github.com/spyzhov/ajson"
 )
 
 // DefaultPageSize is number of elements per page.
-const DefaultPageSize = 100
+const DefaultPageSize = "100"
 
 func (c *Connector) buildReadRequest(ctx context.Context, params common.ReadParams) (*http.Request, error) {
 	url, err := c.buildReadURL(params)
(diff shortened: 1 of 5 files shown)


Cobalt0s:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2714?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2715** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2715?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2714** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2714?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2714?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2713** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2713?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

providers/pipeliner/handlers.go
```diff
@@ -150,3 +161,18 @@ func (c *Connector) parseDeleteResponse(ctx context.Context, params common.Delet
 		Success: true,
 	}, nil
 }
+
+// Almost all object support filtering by modified field.
+// There are a select few that have no usable timestamp field.
+func supportsFilterByTime(objectName string) bool {
```
RajatPawar, line 181: from my AI: the switch operates on `strings.ToLower(objectName)` but the case values are PascalCase (`"Activities"`, `"LeadOppties"`, `"Types"`). A lowercased string will never match these, so the function always returns `true`. This defeats the exclusion list entirely and could cause API errors for objects that don't support time-based filtering. Also, the `fallthrough` chain can be simplified to a single multi-value case: `case "activities", "leadoppties", "types"`

providers/pipeliner/handlers.go
```diff
@@ -45,17 +47,26 @@ func getNextRecordsURL(node *ajson.Node) (string, error) {
...
 
-	if len(config.NextPage) != 0 {
+	if len(params.NextPage) != 0 {
 		// Next page
-		url.WithQueryParam("after", config.NextPage.String())
+		url.WithQueryParam("after", params.NextPage.String())
+	}
+
+	// https://developers.pipelinersales.com/api-docs/core-api-concepts/api-parameters#filter-filter-op
+	if !params.Since.IsZero() && supportsFilterByTime(params.ObjectName) {
```
RajatPawar: is there `until` support?

RajatPawar (approved):

providers/pipeliner/handlers.go
Cobalt0s: There is support for until. But it is either or, so I chosen Since. I think I should add comment on this or better change logic if both Since & Until -> use Since, otherwise use the one given.

From the linked docs it says that we express the request using 2 queries one for the operation operator and one for the value itself. The operation cannot be both `gte` and `lte`. 
```
filter-op[modified]=gte
filter[modified]=<TIME>
```
As I recall I tried using the query list and it complains because duplicates are not allowed. Therefore the best usage of the API is just using Since. I am not sure if we have just Until without the Since, so I "short-circuited" the implementation.

PS: I need to apply for the trial, since it is not a high priority I will do it later to test `Coevera/Pipeliner` later.

providers/pipeliner/handlers.go
Cobalt0s: You are right, missed this entirely. The cases are "original" format and lower case is to be safe, but the cases where not updated...

PR #2773, opened 2026-03-19
[CON-3027] feat(connectWise): Metadata

Cobalt0s:
# Live Tests

Data comes from `schema.json` which is processed OpenAPI file.

image


scripts/openapi/connectwise/internal/files/All.json
eberle1080, line 1: I feel like at some point soon we should version this stuff with git+lfs. Files this big bloat the history and slow git clone if they change too often.

scripts/openapi/connectWise/metadata/main.go
```diff
@@ -0,0 +1,431 @@
...
+		"/sales/opportunities/default",
+		// Conceptually not needed.
+		"/system/info",
+		"/system/reports",
+		"/system/workflows/triggers/options",
+		"/system/audittrail",
+	}
+	displayNameOverride = map[string]string{
+		// "external": "External connections",
+	}
```
eberle1080: What's the purpose of this if it isn't being used? Can you add a comment or something

eberle1080 (approved):

scripts/openapi/connectWise/metadata/main.go
Cobalt0s: I will remove it.

scripts/openapi/connectwise/internal/files/All.json
Cobalt0s: Interesting. Yes, that would be beneficial. To be honest all the OpenAPI files could be moved into a new repo. The extraction tool can move there too. The `connectors` doesn't really care about it. The only things that we would preserve in this repo are the `schema.json` files.

With growing number of providers that could be one of the solutions.

PR #2805, opened 2026-03-26
[CON-3100,CON-3101,CON-3102] feat(fastSpring): Deep connector

Fullstack-Hero:
## Description

Gateway PR that contains:

- ListObjectMetadata
- Read
- Write
- Delete


providers/fastspring/write.go
```diff
@@ -0,0 +1,247 @@
...
+	case objectOrders:
+		return extractOrderWriteRecordID(body)
+	case objectSubscriptions:
+		return extractSubscriptionWriteRecordID(body)
+	default:
+		return "", false
+	}
+}
+
+func extractAccountWriteRecordID(body *ajson.Node) (string, bool) {
```
Cobalt0s: This method always return false as a second argument.
It think it would be nice to use named return values like above: `(recordID string, ambiguousMultiple bool)`.

Cobalt0s (approved):
Awesome!

Looks good!

PR #2858, opened 2026-04-10
[ENG-3807] refactor(salesforce): parameterize provider name via WithProvider option

RajatPawar:
## Summary
- Adds a `WithProvider` Option to the salesforce connector that threads a provider name through `Connector.Provider()` and the internal CRM + Pardot adapters
- Defaults to `providers.Salesforce` when unset, so existing callers are unaffected
- Pure refactor — zero behavior change for the standard salesforce provider

## Why
The Salesforce connector hardcoded `providers.Salesforce` in three places, blocking reuse of the same implementation under a twin provider name. This PR unblocks a follow-up that adds `salesforceJWT` — a twin that shares the CRM/Pardot implementation but authenticates via the OAuth 2.0 JWT Bearer Flow instead of Authorization Code.

## Changes
- `providers/salesforce/params.go` — new `provider` field on `parameters` + `WithProvider` option
- `providers/salesforce/connector.go` — `Connector.provider` field, `Provider()` returns it dynamically, default applied in `NewConnector`
- `providers/salesforce/internal/crm/adapter.go` — `NewAdapter(params, provider)` signature
- `providers/salesforce/internal/pardot/adapter.go` — same

## Test plan
- [x] `go build ./providers/salesforce/... ./connector/...`
- [x] `go test ./providers/salesforce/...` — existing tests pass unmodified
- [ ] Verify no regression in prod salesforce connections after deploy

### DIFF
--- a/providers/salesforce/connector.go
+++ b/providers/salesforce/connector.go
@@ -56,7 +56,7 @@ func NewConnector(opts ...Option) (*Connector, error) {
 		return nil, err
 	}
 
-	// Default the provider to Salesforce when not explicitly overridden.
+	// Default the provider to Salesforce (OAuth provider) when not explicitly overridden.
 	if params.provider == "" {
 		params.provider = providers.Salesforce
 	}
--- a/providers/salesforce/connector.go
+++ b/providers/salesforce/connector.go
@@ -27,6 +27,12 @@ type Connector struct {
 	moduleInfo   *providers.ModuleInfo
 	moduleID     common.ModuleID
 
+	// provider is the provider name this connector reports. It allows the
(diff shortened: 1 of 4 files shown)


providers/salesforce/connector.go
```diff
@@ -50,6 +56,11 @@ func NewConnector(opts ...Option) (*Connector, error) {
 		return nil, err
 	}
 
+	// Default the provider to Salesforce when not explicitly overridden.
```
laurenzlong:
```suggestion
	// Default the provider to Salesforce (OAuth provider) when not explicitly overridden.
```

laurenzlong (approved):

PR #2862, opened 2026-04-10
[CON-3034] Export request and response, clean up some logic

jlimatampersand:
Updated Internal Types

### Test Result
Screenshot 2026-04-13 at 2 07 30 PM

### DIFF
--- a/providers/google/internal/mail/subscribe.go
+++ b/providers/google/internal/mail/subscribe.go
@@ -109,6 +109,8 @@ func (a *Adapter) EmptySubscriptionResult() *common.SubscriptionResult {
 // It maps each object to a Gmail label ID (e.g. "messages" → "INBOX", "drafts" → "DRAFT"),
 // then issues a single watch API call with all labels combined. Only "messages" and "drafts"
 // are supported; any other object is rejected.
+//
+//nolint:cyclop
 func (a *Adapter) Subscribe(
 	ctx context.Context,
 	params common.SubscribeParams,
@@ -123,7 +125,15 @@ func (a *Adapter) Subscribe(
 		return nil, err
 	}
 
-	watchReq.LabelIDs = labels
+	// Use label IDs and filter behavior from the request if provided,
+	// otherwise default to the labels derived from the subscribed objects.
+	if len(watchReq.LabelIDs) == 0 {
+		watchReq.LabelIDs = labels
+	}
+
+	if watchReq.LabelFilterBehavior == "" {
+		watchReq.LabelFilterBehavior = "include"
+	}
 
 	watchURL, err := url.JoinPath(a.ModuleInfo().BaseURL, apiVersion, "users/me/watch")
 	if err != nil {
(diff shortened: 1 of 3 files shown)


providers/google/internal/mail/subscribe.go
```diff
@@ -2,114 +2,167 @@ package mail
...
 var (
-	errMissingParams      = errors.New("missing required parameters")
-	errInvalidRequestType = errors.New("invalid request type")
+	errMissingParams              = errors.New("missing required parameters")
+	errInvalidRequestType         = errors.New("invalid request type")
+	errUnsupportedSubscribeObject = errors.New("unsupported subscribe object")
 )
 
+const (
+	messagesObject common.ObjectName = "messages"
```
jlimatampersand, line 25: @Jkarage, I added drafts as well, as it was listed in our Gmail documentation. https://docs.withampersand.com/provider-guides/google#supported-gmail-objects

providers/google/internal/mail/subscribe.go
```diff
@@ -2,114 +2,167 @@ package mail
...
 	LabelFilterBehavior string `json:"labelFilterBehavior"`
 
-	// TopicName represents a fully qualified Google Cloud Pub/Sub API topic name to publish the events to
+	// TopicName is the fully qualified Google Cloud Pub/Sub topic to publish events to.
 	TopicName string `json:"topicName"`
 }
 
-type watchResponse struct {
+// WatchResponse represents the response from the Gmail watch API.
+type WatchResponse struct {
```
jlimatampersand, line 61: FYI, these request and response types should be exported so that we can json unmarshal. I saw you have map[string]any type as external type and internally json serialized and deserialized again to me the internal type, but we prefer to keep stricted typed struct better.

RajatPawar (approved):

providers/google/internal/mail/subscribe.go
jlimatampersand: @Jkarage

providers/google/internal/mail/subscribe.go
```diff
@@ -2,114 +2,167 @@ package mail
...
-	// LabelIDs is a list of labelIds to restrict notifications about.
-	// By default, if unspecified, all changes are pushed out.
-	// If specified then dictates which labels are required for a push notification to be generated.
+// WatchRequest represents the Gmail watch API request body.
+// A single watch call covers all subscribed objects; the LabelIDs field is
+// populated by mapping each requested object to its corresponding Gmail label.
+type WatchRequest struct {
+	// LabelIDs restricts which mailbox changes trigger push notifications.
+	// Built from the subscribed objects (e.g. "messages" → "INBOX", "drafts" → "DRAFT").
 	LabelIDs []string `json:"labelIds"`
```
jlimatampersand, line 51: @anushat05 I think we can get INBOX and DRAFT labels from Object types since we list messages and drafts as object names. 

But keep the label IDs from the request as override values in case of additional labels (email folders) are requested. 

Same for `LabelFilterBehavior`, we can default to include, but use it as an override.

PR #2863, opened 2026-04-12
[CON-3139] feat(clio): add proxy

Lafetz:
# Conventions
- [x] Should cover all modules within the connector (ex, `goTo` has modules `webinar` and `meeting` or `google` has modules `drive` and `calendar`)
- [x] If OAuth2 connector, if `workspace` is required, `Oauth2Opts.ExplicitWorkspaceRequired` is ALSO set to true
- [X] Modules are only being added because:
  - [ ]  They share the same authentication scheme
  - [X] Different base URLs (drive.google.com vs google.com)
  - [X] Object name collisions (same object name exists in two or more modules)
    E.g Contacts ( [Clio Manage](https://docs.developers.clio.com/clio-manage/api-reference/#tag/Contacts/operation/Contact#show),   [Clio Grow](https://docs.developers.clio.com/clio-grow/api-reference/#tag/Contacts/operation/getContact)), custom actions ( [Clio Manage](https://docs.developers.clio.com/clio-manage/api-reference/#tag/Custom-Fields/operation/CustomField#index),   [Clio Grow](https://docs.developers.clio.com/clio-grow/api-reference/#tag/Custom-Actions/operation/listCustomActions)) 

# TESTS

#### Me
Screenshot from 2026-04-13 02-18-13

#### Pagination
Screenshot from 2026-04-12 23-41-28

### POST

Screenshot from 2026-04-12 23-38-22

### DELETE
Screenshot from 2026-04-12 23-39-06

### PATCH
Screenshot from 2026-04-13 02-30-08

## Grow

No access yet

#### Icon
Icon

#### Logo
Logo

#### Icon
Regular Icon

#### Logo
Regular Logo


Lafetz:
@eberle1080  Should Clio Manage vs Grow stay one provider with modules or become two providers? They currently use separate OAuth endpoints and different API base URLs

eberle1080 (approved):

Cobalt0s:
> @eberle1080 Should Clio Manage vs Grow stay one provider with modules or become two providers? They currently use separate OAuth endpoints and different API base URLs

Different Base URLs means different modules. Different auth types means different connectors.

providers/clio.go
```diff
@@ -0,0 +1,136 @@
...
+						ModuleClioGrow:   ModuleDependency{},
+						ModuleClioManage: ModuleDependency{},
+					},
+				},
+				// Platform uses a distinct auth host (e.g. US auth.api.clio.com).
+				// Manage uses the same host as the API.
+				{
+					Name:         "clio_oauth_domain",
+					DisplayName:  "Clio OAuth domain",
+					DefaultValue: "app.clio.com",
```
Cobalt0s, line 123: So is the URL different for both modules? The default modules "point" to the same URL.

Lafetz:
> > @eberle1080 Should Clio Manage vs Grow stay one provider with modules or become two providers? They currently use separate OAuth endpoints and different API base URLs
> 
> Different Base URLs means different modules. Different auth types means different connectors.

the auth type is the same but the url is different

providers/clio.go
Lafetz: Yes, but since only one module can be active at a time, the user will use the appropriate URL.

Cobalt0s: Yes, I suppose that should be valid. And the module is specified during installation in the definitions file.
So the only difference is conceptually people should look for the Base URL with respect to module (kind of defeats the point of modules in some way, but given the nature of provider it makes sense).

providers/clio.go
```diff
@@ -14,7 +14,7 @@ func init() { //nolint:funlen
 	SetInfo(Clio, ProviderInfo{
 		DisplayName: "Clio",
 		AuthType:    Oauth2,
-		BaseURL:     "https://{{.clio_api_domain}}",
+		BaseURL:     "https://{{if and .region (ne .region \"us\") (ne .region \"US\")}}{{.region}}.{{end}}{{.workspace}}",
```
Cobalt0s: You should make a constant that will be reused in this file:
```
const clioRegionTemplate = "{{if and .region (ne .region \"us\") (ne .region \"US\")}}{{.region}}.{{end}}"
```

providers/clio.go
```diff
@@ -100,31 +109,25 @@ func init() { //nolint:funlen
...
-					Name:         "clio_oauth_domain",
-					DisplayName:  "Clio OAuth domain",
+					Name:         "workspace",
+					DisplayName:  "API host",
 					DefaultValue: "app.clio.com",
 					DocsURL: "https://docs.developers.clio.com/handbook/getting-started/" +
 						"regions",
-					Prompt: "Provide the Clio regional OAuth domain (Manage/Platform). " +
-						"For example, Manage US: app.clio.com, Platform US: auth.api.clio.com",
+					Prompt: "Clio Manage: app.clio.com. Clio Grow: api.clio.com.",
```
Cobalt0s: Wrap "app.clio.co" and "api.clio.com" in quotations.

PR #2868, opened 2026-04-13
[ENG-3793]feat: add HistoryList connector method for Gmail

anushat05:
## Summary
  - Adds `HistoryList()` on the Gmail adapter wrapping `users.history.list`, with pagination
  over `nextPageToken`
  - Re-exports `HistoryListParams` / `HistoryListResult` on `*google.Connector` so external
  (server-side) callers can use them without importing `internal/mail`
  - Extends the existing `test/google/mail/subscriber` harness to exercise `HistoryList` using
  the `historyId` returned from `users.watch()`

  ## Context
  Part of ENG-3798 (ProcessGmailEventWorkflow). The new Temporal workflow will call this method
   each time a Gmail Pub/Sub notification arrives, passing the last-processed checkpoint from
  the upcoming `subscription_states` table and persisting the response's root `historyId` as
  the next checkpoint.

  ## Test plan
  - [x] `go build ./providers/google/... ./test/google/mail/subscriber/...` passes
  - [ ] Live-test via `go run ./test/google/mail/subscriber` (requires
  `google-gmail-creds.json` — deferred; will be exercised end-to-end in the workflow PR)

### DIFF
--- a/providers/google/internal/mail/subscribe.go
+++ b/providers/google/internal/mail/subscribe.go
@@ -6,9 +6,11 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strconv"
 
 	"github.com/amp-labs/connectors"
 	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/urlbuilder"
 )
 
 // Errors for subscribe validation failures.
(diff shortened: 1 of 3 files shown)


providers/google/internal/mail/subscribe.go
jlimatampersand, line 257: @anushat05, as discussed earlier, we call the HistoryList method in here so that we can satisfy the interface type.

jlimatampersand (approved):
Left comment, but you can address it maybe in later pr, or here. up to you.

providers/google/internal/mail/subscribe.go
anushat05: I'll follow up with another PR.

PR #2876, opened 2026-04-14
feat(components): Subscription interfaces

Cobalt0s:
# Description

This PR refactors the subscription components to improve composability and testability, while working within the constraints of existing interfaces.

### Key changes

* **Interface decomposition for better testing**
  `SubscribeConnector` is split into smaller, focused interfaces (`SubscriptionCreator`, `SubscriptionUpdater`, etc.).
  This makes unit testing simpler and more targeted (single-method interfaces).
  A compile-time assertion via `dummySubscribeConnector` ensures that the composed interfaces remain fully compatible with `SubscribeConnector`.

* **Generic input/output helper**
  Introduces `SubscriptionInputOutput[I, O any]`, which provides out-of-the-box implementations of
  `EmptySubscriptionParams` and `EmptySubscriptionResult`.
  This avoids repetitive boilerplate while preserving type safety.
  Since the original interface cannot be changed to use generics, this acts as a lightweight workaround.

* **Event interface unification (optional/experimental)**
  A new `Event` interface extracts shared behavior from subscription event types and is embedded into each.
  This makes their relationship explicit, while keeping usage flexible. (Can be reverted if needed, it is a proposal)

### Benefits

* Easier and more granular unit testing
* Reduced boilerplate for subscription connectors
* Safer and more modular interface design without breaking existing contracts
* More flexible PRs: implementations can be delivered incrementally without requiring the full `SubscribeConnector` surface.

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -744,17 +744,34 @@ type SubscriptionEventPreLoadData struct {
 	RequestHeaders *http.Header
 }
 
+// Event represents a raw webhook payload and serves as the common abstraction across all event types in the system.
+// Implementations are expected to provide access to the underlying payload in a normalized map form.
+//
+// The following higher-level event interfaces build on top of Event
+// and should be considered during implementation:
+//   - SubscriptionEvent: core interface for all subscription-related events
+//   - SubscriptionUpdateEvent: specialization for update/change events
+//   - CollapsedSubscriptionEvent: for payloads containing multiple events
+//
+// A concrete provider event type may implement one or more of these
+// interfaces depending on the structure of the incoming webhook payload.
+type Event interface {
+	RawMap() (map[string]any, error)
+}
+
 // SubscriptionEvent is an interface for webhook events coming from the provider.
 // This interface defines methods to extract information from the webhook event.
 type SubscriptionEvent interface {
+	Event
+
 	EventType() (SubscriptionEventType, error)
 	RawEventName() (string, error)
 	ObjectName() (string, error)
 	Workspace() (string, error)
 	RecordId() (string, error)
 	EventTimeStampNano() (int64, error)
-	RawMap() (map[string]any, error)
-	// PreLoadData is used to pre-load data into the subscription event.
+
+	// PreLoadData is used to preload data into the subscription event.
 	// Assume that the data will include the entire request body in case it is needed for the event processing.
 	// This method will be called for every event as the first step in the event processing.
 	PreLoadData(data *SubscriptionEventPreLoadData) error
@@ -762,7 +779,8 @@ type SubscriptionEvent interface {
 
 type SubscriptionUpdateEvent interface {
 	SubscriptionEvent
-	// GetUpdatedFields returns the fields that were updated in the event.
+
+	// UpdatedFields returns the fields that were updated in the event.
 	UpdatedFields() ([]string, error)
 }
 
@@ -770,8 +788,11 @@ type SubscriptionUpdateEvent interface {
 // This interface is used to extract individual events to SubscriptionEvent type
 // from a collapsed event for webhook parsing and processing.
 type CollapsedSubscriptionEvent interface {
+	Event
+
+	// SubscriptionEventList collapses the event response into the list of messages.
+	// In case of an error returns ErrSubscriptionEventList.
 	SubscriptionEventList() ([]SubscriptionEvent, error)
-	RawMap() (map[string]any, error)
 }
 
 // WebhookRequest is a struct that contains the request parameters for a webhook.
--- a/internal/components/subscriptions.go
+++ b/internal/components/subscriptions.go
@@ -1,0 +1,92 @@
+package components
+
+import (
+	"context"
+
+	"github.com/amp-labs/connectors"
+	"github.com/amp-labs/connectors/common"
+)
+
+// SubscriptionInputOutput is a generic helper that provides type-safe implementations
+// of the EmptySubscriptionParams and EmptySubscriptionResult methods
+// required by connectors.SubscribeConnector.
+type SubscriptionInputOutput[I, O any] struct{}
+
+func (SubscriptionInputOutput[I, O]) EmptySubscriptionParams() *common.SubscribeParams {
+	var request I
+
+	return &common.SubscribeParams{
+		Request: &request,
+	}
+}
+
+func (SubscriptionInputOutput[I, O]) EmptySubscriptionResult() *common.SubscriptionResult {
+	var result O
+
+	return &common.SubscriptionResult{
+		Result: &result,
+	}
+}
+
+// WebhookMessageVerifier is the minimal interface for a connector that can verify webhook messages.
+type WebhookMessageVerifier interface {
+	VerifyWebhookMessage(
+		ctx context.Context,
+		request *common.WebhookRequest,
+		params *common.VerificationParams,
+	) (bool, error)
+}
+
+// SubscriptionCreator is the minimal interface for a connector that can create subscriptions.
+type SubscriptionCreator interface {
+	Subscribe(
+		ctx context.Context,
+		params common.SubscribeParams,
+	) (*common.SubscriptionResult, error)
+}
+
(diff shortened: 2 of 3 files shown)


Cobalt0s:
* **#3009** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3009?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#2276](https://github.com/amp-labs/connectors/pull/2276) <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2276?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#3014](https://github.com/amp-labs/connectors/pull/3014) <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3014?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#2948** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2948?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2876** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2876?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2876?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jlimatampersand (changes requested):
Temporarily marking as request changes. 

What kind of problems did you actually run into to consider this change? 

The description says to improve composability and testability, but I had no issue with existing interfaces and types. 

So if you can provide concrete examples of the problem you ran into working on the Subscribe action, then we can make a decision here.

jlimatampersand (commented):
I am open to naming those connectors to

SubscriptionCreateTester, SubscriptionUpdateTester, etc.

Seems like these are used for testing mostly, so if that is the friction point, we can rename the interface names so that they are safeguarded and we can prevent them from being used in production code.

Cobalt0s:
> I am open to naming those connectors to
> 
> SubscriptionCreateTester, SubscriptionUpdateTester, etc.
> 
> Seems like these are used for testing mostly, so if that is the friction point, we can rename the interface names so that they are safeguarded and we can prevent them from being used in production code.

(1)
Primary intention -- test each method of an interface. A Strategy/Adapter or any **struct** can implement **only one method** out of the whole interface. This doesn't make this struct a `connectors.Connector` or `connectors.SubscribeConnector` because it would lack all the rest of the methods.

This enables developing features bit by bit. For some connectors you would want to do it all at once. But if provider API is very complex dedicated PRs are possible.

I was thinking where could I put this and I found that it naturally fits the `components` package because it already does "decomposed" interfaces. We could name it `SubscriptionUpdateTester` but it will be prefixed with `components.SubscriptionUpdate` anyway.

Example: 
(embed interfaces)
image

(init with out of the box solutions)
image

(dedicated interfaces)
image

However, unlike the `Reader` example shown above -- the subscribe **won't have** out of the box abstract implementations like `reader.NewHTTPReader`. I don't have that as a goal to be clear. Maybe some algorithms could emerge but probably after 20+ implementations of subscribe but I doubt it. I am not planning on using the Interfaces directly on the structs, they are referenced only:
* For unit tests
* For live test scripts (a script may auto remove subscription if DeleteSubscription method impl is present, again this allows incremental development even for live tests)

Summary: decomposed interfaces help with testing; we could add suffix "Test" but I think there is a prefix of "components"; changing core interfaces is not my goal. Putting all the subscribe functionality (which is usually huge) into `<some-provider>/internal/subscribe` is the goal.

(2)
`SubscriptionInputOutput[I, O any]` fits into the idea of components library and gives implementation of `EmptySubscriptionParams` and `EmptySubscriptionResult` reducing boilerplate and any pointer type mistakes/doubts during development. That is the most that can be "simplified" across all subscribe implementors using a true embedable "component":
image

Note, as I understand the SubscribeConnector conceptually acts as follows (I am not changing anything or commiting this):
image

(3) 
`Event` is mostly symbolic: it documents the relationship between `SubscriptionEvent`, `SubscriptionUpdateEvent`, and `CollapsedSubscriptionEvent`. The unit-test template benefits from this: a single [test routine](https://github.com/amp-labs/connectors/blob/eeddc73f6d8f5d4cc9c8bc3e11396a01aab531c6/test/utils/testroutines/subscription-event.go#L43) can validate any of those three event types.

Cobalt0s:
@jlimatampersand let me know the next steps and I will do them.

jlimatampersand:
I see where you are going, and your intent, and it seems to be the right direction. But right now, the Subscribe action is still not fully mature in the sense that I am working on a major refactoring, and then also looking into updating the documentation on how to contribute and add a new provider for the Subscribe action. So I just don't have the bandwidth to deal with changes outside of my control at the moment. 

I think I have mentioned many times in your PRs that any new interfaces or conventions you introduce are huddles and learning curves for all the other contributors. 

You are the most knowledgeable person for this library, probably accounting for more than 70% of this repository, and you are that much of an expert for this repo. 

But what that also means is that not many people will understand the code and packages you introduce in this library. It is very possible that many people will misuse this library and will produce bugs in our production. 

You mentioned that the package is prefixed with `components`, but it nowehere in the prefix does it indicate that it is for testing. So either you need to move it to a package that clearly indicates what the package is for, and maybe call the package `testInterfaces` or something, or make the struct/interface indicate that it is for testing. We can revisit this later when I am done with refactoring and documentation, but today, adding `test` to either the package name or the struct/interface itself, otherwise, I cannot approve it. 

I can totally imagine some contributors using the components for production code, and someone other than myself reviewing and approving, then breaking production because even the reviewers won't understand the underlying intent of the packages unless you really make it obvious, and being under `component` is definitely not obvious. 

So either add a `Test` suffix somewhere or close this PR. I am only with Event Interface, though, because it is very obvious, but nothing else. 

I am happy to revisit this later when my refactoring and documentation are done. 

Again, please do remember that anytime you introduce a new thing, no one is as knowledgeable as you are, and any new introduction will be a learning curve and friction points for everyone for many contributors who come and go.

jlimatampersand (changes requested):

Cobalt0s:
I do suffer from overengineering once in a while I cannot deny that. I understand that if many things become interconnected it will raise chances of errors by devs.
 
I did more analysis on the codebase and on what you said and what I wrote.

I layed out some arguments for the components package where I've missed important thing -- it offers "abstract/template" solutions for the simple operations like Read/Write/Metadata. Therefore, probably `components` is not the right package :thinking:. Only `SubscriptionInputOutput` does fit into that idea. I already kind of prove it by saying that there won't be general solutions for subscribe and `components` is mostly about template solutions.

My main goal is to test that connector sends **correct payloads** and **invokes correct URLs** proving subscription feature (or any feature for that matter). Read/Write/Metadata connector mock tests are a common practice now. The subscription unit tests would look identical due to `testroutines.TestCase` and `mockserver/mockcond` packages. Without unit tests I am scared to touch and change any code. They regularly catch mistakes when I code.

Back to main topic. Since I am after testing only -- lets add interfaces to `testroutines` package. This would express the narrow purpose of the interface **without introducing the danger** of abusing and overusing it where it is not intended.

# Code Suggestion

* `testroutines.CreateSubscription` --> is the definition of a unit test struct (probably should call it :thinking: `testroutines.TestCaseCreateSubscription`)
* `testroutines.TestableSubscriptionCreator` --> would be the method that we are testing (represented by an interface). What do you think about adding `Testable` as a prefix besides the package? This would indicate that there is a coresponding `TestCase`.

image

### Result
* **Interfaces** and **TestCase** naturally go together: since testroutines is dedicated to unit‑testing connector methods, it makes sense that the interface being tested and the corresponding test struct **live in the same place**.
* `testroutines.TestCase<what-is-tested>` -- unit test struct
* `testroutines.Testable<what-is-tested>` -- interface under test
  * This way `testroutines` would also gain its distinct identity and IDE autocomplete will be more helpful. (even though often times I generate the base layout of unit tests with llms).
  * No need to redefine these in each connector it will be done once in `testroutines`.

image

jlimatampersand (approved):

PR #2875, opened 2026-04-14
[ENG-3815] feat(microsoft): Webhook Message format

Cobalt0s:
# Description

Implements Microsoft subscription events and refactors HubSpot event tests to use `testroutines.SubscriptionEventTestCase`. It is used as a template for all future unit tests of this kind.

### Key changes

* **Reusable test routine**
  `SubscriptionEventTestCase` accepts a generic `common.Event` and validates it by casting to:

  * `common.CollapsedSubscriptionEvent` (for list-based events)
  * `common.SubscriptionEvent` (core contract assertions)
  * `common.SubscriptionUpdateEvent` (adds updated-fields validation)

  Test inputs use a unified parameter structure, enabling universal validation across all event interfaces.

* **Microsoft event implementation**
  Implements the event interfaces for Microsoft subscription events.
  Responses always contain a single item, but are still modeled as a `common.CollapsedSubscriptionEvent`.

* **Exported event collection**
  `webhook.EventCollection` is exported as `microsoft.EventCollection` for external server use.

> NOTE: the `clientState` is used to carry the `objectName`. Microsoft returns the resource name but that would require mapping each and every string back to the object name we use for Read/Write, etc.

PR #2874, opened 2026-04-14
[ENG-3815] feat(microsoft): BatchRecordReader

Cobalt0s:
# Description

Adds `testroutines.ReadByIds`, enabling reusable unit tests for any `BatchRecordReaderConnector` implementation.

Also introduces a Microsoft implementation of `GetRecordsByIds`, built on top of the batch strategy for the Microsoft Graph Batch API.

### Testing

* Live test creates 3 Outlook messages
* Reads them in batch by ID
* Cleans up the created records afterward

# Live Tests
Script output is copied. Some object fields are truncated.
```
[SETUP] Creating a record for object 'me/messages'
Reading me/messages
Finding recently created me/messages
Object me/messages(AAMkAGY0YzAwY2ViLWQyODktNDI3NS1iNmY4LTE5YzU0MjI5ZTA4OQBGAAAAAABeMJSlO8qLToz2i2IQ1wsqBwB8hj1Rtd60SKTngNs3if9RAACsvo1TAAB8hj1Rtd60SKTngNs3if9RAAEuy3j8AAA=) is ready
[SETUP] Creating a record for object 'me/messages'
Reading me/messages
Finding recently created me/messages
Object me/messages(AAMkAGY0YzAwY2ViLWQyODktNDI3NS1iNmY4LTE5YzU0MjI5ZTA4OQBGAAAAAABeMJSlO8qLToz2i2IQ1wsqBwB8hj1Rtd60SKTngNs3if9RAACsvo1TAAB8hj1Rtd60SKTngNs3if9RAAEuy3j9AAA=) is ready
[SETUP] Creating a record for object 'me/messages'
Reading me/messages
Finding recently created me/messages
Object me/messages(AAMkAGY0YzAwY2ViLWQyODktNDI3NS1iNmY4LTE5YzU0MjI5ZTA4OQBGAAAAAABeMJSlO8qLToz2i2IQ1wsqBwB8hj1Rtd60SKTngNs3if9RAACsvo1TAAB8hj1Rtd60SKTngNs3if9RAAEuy3j_AAA=) is ready
========================
```

```
[
  {
    "Associations": {},
    "Fields": {
      "body": {
        "content": "Zoe Stamm",
        "contentType": "text"
      },
      "from": {
        "emailAddress": {
          "address": "Reilly5008@test.com",
          "name": "Reilly5008"
        }
      },
      "subject": "Nelson Dickens [1]",
      "torecipients": [
        {
          "emailAddress": {
            "address": "Rutherford6844@test.com",
[...]

PR #2877, opened 2026-04-14
[ENG-3815] feat(microsoft): WebhookVerifier

Cobalt0s:
# Description

Adds an implementation of `WebhookVerifierConnector` for Microsoft.

### Key points

* **Microsoft webhook verification**
  Implements `WebhookVerifierConnector` with a no-op verifier.
  Although Microsoft supports message encryption and `clientState`, neither is used here.
  `clientState` is instead leveraged to carry the original `objectName`.

* **Reusable test template**
  Introduces `testroutines.WebhookMessageVerification` for validating webhook verification logic.
  Not used by Microsoft due to the no-op implementation.

### DIFF
--- a/providers/microsoft/internal/webhook/verification.go
+++ b/providers/microsoft/internal/webhook/verification.go
@@ -1,0 +1,27 @@
+package webhook
+
+import (
+	"context"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/providers"
+)
+
+type Verifier struct {
+	client     *common.JSONHTTPClient
+	moduleInfo *providers.ModuleInfo
+}
+
+func NewVerifier(client *common.JSONHTTPClient, moduleInfo *providers.ModuleInfo) *Verifier {
+	return &Verifier{
+		client:     client,
+		moduleInfo: moduleInfo,
+	}
+}
+
+func (v Verifier) VerifyWebhookMessage(
+	ctx context.Context, request *common.WebhookRequest, params *common.VerificationParams,
+) (bool, error) {
+	//TODO implement me
+	panic("implement me")
+}
--- a/providers/microsoft/internal/webhook/verification.go
+++ b/providers/microsoft/internal/webhook/verification.go
@@ -1,0 +1,46 @@
+package webhook
+
+import (
+	"context"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/providers"
+)
+
+type NoopVerifier struct {
+	client       *common.JSONHTTPClient
+	providerInfo *providers.ProviderInfo
+}
+
+func NewVerifier(client *common.JSONHTTPClient, providerInfo *providers.ProviderInfo) *NoopVerifier {
+	return &NoopVerifier{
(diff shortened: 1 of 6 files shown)


Cobalt0s:
* **#2878** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2878?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3122** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3122?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2877** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2877?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2877?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

providers/microsoft/internal/webhook/verification.go
```diff
@@ -0,0 +1,30 @@
...
+		moduleInfo: moduleInfo,
+	}
+}
+
+// VerifyWebhookMessage allows all messages to pass.
+//
+// If we needed event message to include record data then validation would be required:
+// https://learn.microsoft.com/en-us/graph/change-notifications-with-resource-data?tabs=csharp#how-to-validate
+func (v Verifier) VerifyWebhookMessage(
+	ctx context.Context, request *common.WebhookRequest, params *common.VerificationParams,
```
jlimatampersand, line 43: request object has all the event body and headers you need.

jlimatampersand (changes requested):
you can use request body to verify the mesassge.

providers/microsoft/internal/webhook/verification.go
Cobalt0s: Microsoft Graph doesn’t use provider signing keys or signature headers for these notifications. For the events I’m working with, there’s no useful header or JWT available to validate the incoming payload, so the only Graph-specific fields we can rely on are `clientState` for regular notifications and, in some cases, a JWT for richer resource-data scenarios.

image

In my current implementation, I’m using `clientState` as metadata to carry business context. On subscription creation, I set `clientState` to the Ampersand object name, which makes the mapping simpler.

Example:
```
      "clientState": "me/messages",
      "resource": "Users/ae87552f-48fc-4dec-9322-65040bf9fdfd/Messages/message_654",
```

Microsoft already knows we are subscribing to messages, but the webhook event only returns the concrete resource path (`Users/{id}/Messages/{id}`). Without extra mapping, I would need to track every object explicitly.


@jlimatampersand. A possible improvement would be to use `common.SubscriptionResult.Result`, which already contains a `map[string]microsoft.SubscriptionResource`. Then, when handling an event and resolving `Event.ObjectName()`, we could look up the subscription by ID and recover the object name from that map. That would remove the need to **overload clientState** with business data.

The downside is that this would introduce a dependency between event handling and subscription creation state, so I'm not sure whether that coupling is desirable. I'm a bit wary of this coupling.

jlimatampersand (approved):

PR #2878, opened 2026-04-14
[ENG-3815] feat(microsoft): Impl Create Subscription

Cobalt0s:
## Description

Implements Microsoft create-subscription support. Also, using the new scripting infrastructure, adds integration tests and unit tests.

### Key changes

* **Subscription implementation**
  Implements `SubscriptionCreator` using the `subscriber` strategy (this strategy will implement update & delete in subsequent PRs).

* **Microsoft-specific types**
  Exports `SubscribeRequest` and `SubscribeResponse` for server integration.

* **Script integration**
  Uses the new scripts to validate real scenarios:

  * Outlook Calendar: create/update/delete → receives only subscribed events (create + delete)
  * Outlook Messages: create/update/delete → receives all events (subscribed to create + update + delete)

* **Unit testing**
  Adds `testroutines.CreateSubscription` for validating subscription creation logic.

* **Time abstraction**
  Introduces `components.Clock` (`RealClock` and `FixedClock`) to support deterministic testing.

# Live Tests

### Outlook Calendar Events
Calendar events

### Outlook Mail Messages
Outlook messages

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -147,6 +147,6 @@ output/
 # Salesforce test binaries
 /apex
 test/salesforce/apex/apex
-/subscribe
+./subscribe
 ./subscriber
 test/salesforce/subscribe/subscribe
--- a/internal/datautils/pair.go
+++ b/internal/datautils/pair.go
@@ -1,0 +1,6 @@
+package datautils
+
+type Pair[L, R any] struct {
+	Left  L
+	Right R
+}
--- a/providers/microsoft/internal/subscriber/create.go
+++ b/providers/microsoft/internal/subscriber/create.go
@@ -1,0 +1,196 @@
+package subscriber
+
+import (
+	"context"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/internal/datautils"
+	"github.com/amp-labs/connectors/providers/microsoft/internal/batch"
+)
+
+// Subscribe
+// https://learn.microsoft.com/en-us/graph/api/subscription-post-subscriptions?view=graph-rest-1.0&tabs=http#request-body
+func (s Strategy) Subscribe(
+	ctx context.Context,
+	params common.SubscribeParams,
+) (*common.SubscriptionResult, error) {
+	if err := params.ValidateParams(); err != nil {
+		return nil, err
+	}
+
+	batchParams, err := s.paramsForBatchCreateSubscriptions(params)
+	if err != nil {
+		return nil, err
+	}
+
+	bundledResponse, err := batch.Send[SubscriptionResource](ctx, s.batchStrategy, batchParams)
+	if err != nil {
+		return nil, err
+	}
+
+	return &common.SubscriptionResult{
+		Result:       bundledResponse.Registry,
+		ObjectEvents: objectEventsFrom(bundledResponse),
+		Status:       common.SubscriptionStatusSuccess,
+	}, nil
+}
+
+func objectEventsFrom(
+	response *batch.BundledResponse[SubscriptionResource],
+) map[common.ObjectName]common.ObjectEvents {
+	result := make(map[common.ObjectName]common.ObjectEvents)
+
+	for objectName, resource := range response.Registry {
+		result[common.ObjectName(objectName)] = common.ObjectEvents{
+			Events:            resource.ChangeType.EventTypes(),
+			WatchFields:       resource.Resource,
+			WatchFieldsAll:    false,
+			PassThroughEvents: nil,
+		}
+	}
+
+	return result
+}
+
+func (s Strategy) paramsForBatchCreateSubscriptions(params common.SubscribeParams) (*batch.Params, error) {
+	input, err := s.TypedSubscriptionRequest(params)
+	if err != nil {
+		return nil, err
+	}
+
+	url, err := s.getCreateSubscriptionURL()
+	if err != nil {
+		return nil, err
+	}
+
+	batchParams := &batch.Params{}
+	for objectName, events := range params.SubscriptionEvents {
+		// TODO what if nothing is specified
+		resource := objectName.String() + "?$select=" + strings.Join(events.WatchFields, ",")
+
+		// TODO this must be chosen based on the Resource/Object.
+		expirationDateTime := datautils.Time.FormatRFC3339inUTC(time.Now().Add(time.Hour))
+
+		certificate, certID, err := generateGraphEncryptionCert()
+		if err != nil {
+			return nil, err
+		}
+
+		body := SubscriptionResource{
+			ChangeType:          newChangeType(events.Events),
+			WebhookURL:          input.WebhookURL,
+			Resource:            resource,
+			ExpirationDateTime:  expirationDateTime,
+			IncludeResourceData: true,
+			SubscriptionVerification: SubscriptionVerification{
+				ClientState:              "123456_my_state",
+				EncryptionCertificateID:  certID,
+				EncryptionCertificate:    certificate,
+				LifecycleNotificationUrl: nil,
+				NotificationQueryOptions: nil,
+				NotificationUrlAppId:     nil,
+			},
+		}
+
+		batchParams.WithRequest(objectName.String(), http.MethodPost, url, body, map[string]any{
+			"Content-Type": "application/json",
+		})
+	}
+
+	return batchParams, nil
+}
+
+// SubscriptionResource represents a subscription that allows a client app
+// to receive change notifications about changes to data in Microsoft Graph.
+//
+// https://learn.microsoft.com/en-us/graph/api/resources/subscription?view=graph-rest-1.0#properties
+type SubscriptionResource struct {
+	// ChangeType indicates the mode of subscription, ex: created, updated, deleted.
+	ChangeType ChangeType `json:"changeType"`
+	// WebhookURL is the target URL where messages will be sent.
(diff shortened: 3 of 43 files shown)


Cobalt0s:
* **#2878** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2878?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2878?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3122** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3122?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2877** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2877?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

.gitignore
```diff
@@ -147,6 +147,6 @@ output/
 # Salesforce test binaries
 apex
 test/salesforce/apex/apex
-subscribe
-subscriber
+./subscribe
+./subscriber
```
Cobalt0s, line 151: I am not sure what is the reason for ignoring these and where exactly this data is stored. The directory names are good and shouldn't be excluded.

providers/microsoft/internal/subscriber/strategy.go
```diff
@@ -0,0 +1,91 @@
...
+// https://learn.microsoft.com/en-us/graph/change-notifications-delivery-webhooks?tabs=http#subscription-request
+func (s Strategy) getSubscriptionURL() (*urlbuilder.URL, error) {
+	return urlbuilder.New(s.providerInfo.BaseURL, apiVersion, "subscriptions")
+}
+
+// constructTestStrategy creates a Strategy configured for unit testing.
+//
+// It uses a mock HTTP client and overrides the base URL to point to a test server.
+// A fixed clock is injected to ensure deterministic behavior in tests.
+func constructTestStrategy(serverURL string) (*Strategy, error) {
```
jlimatampersand: this should be in test folder

providers/microsoft/internal/subscriber/create.go
```diff
@@ -0,0 +1,224 @@
...
+	bundledResponse := batch.Execute[SubscriptionResource](ctx, s.batchStrategy, batchParams)
+	state := getStateFromCreateResponse(bundledResponse)
+
+	if len(bundledResponse.Errors) != 0 {
+		// Some requests failed; initiate rollback.
+		return s.rollbackSubscriptionCreation(ctx, params, state, bundledResponse)
+	}
+
+	return &common.SubscriptionResult{
+		Result:       Output{},
```
jlimatampersand: There is no need to have an empty struct type as a placeholder. `Result` type is any, so we don't even need to populate it as it will default to nil.

providers/microsoft/internal/subscriber/state.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+type (
+	// State describes the set of object events that the connector should
+	// try to achieve (desired) and that are also reported back as the observed outcome.
+	//
+	// It is part of the connector's input/output contract: the caller expresses intent
+	// with State, and the connector returns the same type reflecting what actually
+	// happened (which may differ on failure or rollback).
+	State map[ObjectName]common.ObjectEvents
```
jlimatampersand: is this wrapper layer needed? Unless there is a strong reason, I would like to keep each provider addition to a minimal context. There is no new variable, method, defined for this State struct, so I would remove this.

internal/components/clock.go
```diff
@@ -0,0 +1,35 @@
+package components
+
+import "time"
+
+// Clock provides the current time, enabling deterministic testing of time-dependent logic.
+// Implementations include production (real time) and test (fixed time) variants.
+type Clock interface {
```
jlimatampersand: I am strongly against adding new components that are just wrappers of native packages. 

This only introduces onboarding complexity and load. 

Please remove this type and just the native package.

providers/microsoft/internal/subscriber/create.go
```diff
@@ -0,0 +1,224 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if len(bundledResponse.Errors) == 0 {
+		// Full rollback succeeded.
+		objectNames := datautils.FromMap(params.SubscriptionEvents).Keys()
+
+		return &common.SubscriptionResult{
+			Result:       Output{},
```
jlimatampersand: nil is sufficient. or just leave it empty.

internal/datautils/pair.go
```diff
@@ -4,3 +4,10 @@ type Pair[L, R any] struct {
 	Left  L
 	Right R
 }
+
+func NewPair[L, R any](left L, right R) *Pair[L, R] {
```
jlimatampersand: unnecessary util func. Please remove.

providers/microsoft/internal/subscriber/strategy.go
```diff
@@ -0,0 +1,91 @@
...
+//   - Clock for time.Now. (useful in tests for deterministic outcomes)
+type Strategy struct {
+	components.SubscriptionInputOutput[Input, Output]
+
+	client       *common.JSONHTTPClient
+	providerInfo *providers.ProviderInfo
+
+	// Dependent services.
+	batchStrategy *batch.Strategy
+	clock         components.Clock
```
jlimatampersand: is this necessary? if this is just useful for testing, I would just opt for the native time package.

jlimatampersand (changes requested):
There are too many unnecessary structs and functions that are beyond the scope of what is necessary to implement Subscribe. 

Please remember this is an open-source project that many different contributors with various experiences and familiarity contribute to. 

Any new introduction of internal packages and data utils becomes an immediate huddle and a learning curve for every contributor, making onboarding harder and adding complexity. 

Please try using native packages unless you really need to use wrapper packages.

jlimatampersand (changes requested):
Many comments unresolved. Please resolve or reason about your decisions.

providers/microsoft/internal/subscriber/state.go
Cobalt0s: I will be revisiting all the code and I will try to avoid this!

I will rely on the new utility `subscriptionhelper.SegmentSubscriptionEvents` which we have agreed on.

internal/components/clock.go
Cobalt0s: This is a wrapper to allow dependency injection. This technique is used to acquire control of non-deterministic code.

The payload that Microsoft connector must send according to API depends on `time.Now()`.
There is no any other way to test it deterministically, -- but to make it configurable and plug in our "time" version for testing.

I cannot use native packages directly. Alternatively we can remove unit tests, which is not good. Therefore, we should create some wrapper or employ **some technique** to make software testable. 

If you have a preferred alternative for testing time-dependent logic, I'm open to it without any hesitation. The problem itself is inherent to time, but the solution can vary.

internal/datautils/pair.go
Cobalt0s: I have used it a bit. And the Pair data type is just not good. I will be removing everything.

It doesn't help the readability and the intent gets lost.

providers/microsoft/internal/subscriber/strategy.go
Cobalt0s: See message above. Yes it is to make the code deterministic and therefore testable by unit tests.

internal/components/clock.go
Cobalt0s: I could modify the way we `validate` what Payloads the connector producers. The test utility would ignore some fields (ex: in our case `SubscriptionResource.ExpirationDateTime`) and the payload will still be considered as "valid".

Do you prefer this route more?

I would need to put some thought how to make the test utility easy to use.

Draft example (introducing `mockcond.IgnoreBodyField`): 
image

Maybe this would be a better approach? Let me know.

Cobalt0s: I think the IgnoreBodyField is an important feature for the tests to have.

Here is the PR: https://github.com/amp-labs/connectors/pull/3122.

I removed the `Clock` interface as the justification for it is now weak to none.

providers/microsoft/internal/subscriber/delete.go
```diff
@@ -0,0 +1,41 @@
+package subscriber
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/amp-labs/connectors/providers/microsoft/internal/batch"
+)
+
+func (s Strategy) removeSubscriptionsByIDs(
```
jlimatampersand: we use Id instead of ID for all our code bases for consistency. Please modify here and in other PRs. 

Also, please use ids, instead of identifiers as well.

jlimatampersand (approved):
Preapproved, but left some minor comments. Please resolve before merging.

providers/microsoft/internal/subscriber/delete.go
Cobalt0s: I will fix these naming cases. The issue is broader and relates to the whole `connectors` project. The go linter at the moment is enforcing the typical to golang pattern.

See linter output:
image

We would need a PR to deprecate `revive` linting rule and then fix ID, URL, API string occurrences and other such initialisms, acronyms, etc. (Or fix them as we go as there are too many of them).

Cobalt0s: The linter has some config related to ID but it is not working as intented:
image

PR #2891, opened 2026-04-16
feat (phoneburner): Support custom fields

Fullstack-Hero:
## Description
This PR adds native custom field support for the PhoneBurner connector on the contacts object: definitions are discovered in ListObjectMetadata, and read responses expose custom values in ReadResultRow.Fields while keeping ReadResultRow.Raw as the provider payload (nested custom_fields unchanged on Raw).

## What changed

- `ListObjectMetadata`: When contacts is requested, merges member-defined definitions from GET /rest/1/customfields (paginated, page_size=100) into ObjectMetadata.Fields with IsCustom: true, ValueType / ProviderType from PhoneBurner’s type_id / type_name.
- `Read`: For contacts, flattens each row’s custom_fields entries (name, type, value) to root-level keys in Fields (custom_<normalized_name>), matching metadata keys.
- `Resilience`: If definitions fetch fails, logs a warning and returns base contact metadata only (no hard failure on the whole metadata response).
- `Errors`: Definition fetch / parse failures use common.ErrResolvingCustomFields where appropriate.

## Conventions followed

Read more: Deep Connectors Guide – Reviewer Checklist
- `Field keys`: custom_ + normalized label (lowercase, spaces → underscores, non-alphanumeric stripped) so custom labels do not collide with built-in contact fields.
- `Human-readable names`: FieldMetadata.DisplayName uses PhoneBurner display_name (definitions) / name (read payload); users select by the same logical names as in the UI-oriented API.
- `Fields vs Raw`: Custom values appear in ReadResultRow.Fields; Raw is a shallow clone of the row before flattening (Slab: do not mutate provider-shaped Raw).
- `Metadata fetch`: Definitions loaded once per metadata call (paginated list), not per contact row.

## Supported objects

- `contacts` only. Other objects in this connector (folders, members, dialsession, tags, voicemails) are not covered by the same documented custom_fields + /rest/1/customfields pattern for bulk read.

## How it works

Operation | API used | What it does
-- | -- | --
Metadata | GET /rest/1/customfields | Loads all pages of definitions, merges into contacts Fields with IsCustom: true.
Read | GET /rest/1/contacts (unchanged) | Response still includes nested custom_fields; connector clones row → Raw, flattens copy → Fields with custom_* keys.

#### Metadata
Screenshot 2026-04-22 at 10 08 05 AM

#### Read
Screenshot 2026-04-22 at 10 16 01 AM
Screenshot 2026-04-22 at 10 16 20 AM
Screenshot 2026-04-22 at 10 16 36 AM

### Integration Test
Screenshot 2026-04-22 at 12 54 28 PM
Screenshot 2026-04-22 at 12 54 59 PM
Screenshot 2026-04-22 at 12 55 16 PM

### DIFF
--- a/providers/phoneburner/custom.go
+++ b/providers/phoneburner/custom.go
@@ -51,6 +51,7 @@ func normalizeCustomFieldKey(label string) string {
 	s = strings.ReplaceAll(s, " ", "_")
 	s = nonAlphaNumUnderscore.ReplaceAllString(s, "_")
 	s = strings.Trim(s, "_")
+
 	for strings.Contains(s, "__") {
 		s = strings.ReplaceAll(s, "__", "_")
 	}
@@ -68,53 +69,14 @@ func (c *Connector) fetchMemberCustomFieldDefinitions(ctx context.Context) ([]me
 	page := 1
 
 	for {
-		url, err := urlbuilder.New(c.ProviderInfo().BaseURL, restPrefix, restVer, "customfields")
+		defs, totalPages, err := c.getMemberCustomFieldDefinitionsPage(ctx, page)
 		if err != nil {
-			return nil, fmt.Errorf("%w: %w", common.ErrResolvingCustomFields, err)
+			return nil, err
 		}
 
-		url.WithQueryParam("page_size", "100")
-		url.WithQueryParam("page", strconv.Itoa(page))
+		out = append(out, defs...)
 
-		resp, err := c.JSONHTTPClient().Get(ctx, url.String())
-		if err != nil {
-			return nil, fmt.Errorf("%w: %w", common.ErrResolvingCustomFields, err)
-		}
-
-		if err := interpretPhoneBurnerEnvelopeError(resp); err != nil {
-			return nil, fmt.Errorf("%w: %w", common.ErrResolvingCustomFields, err)
-		}
-
-		body, ok := resp.Body()
-		if !ok {
-			return nil, fmt.Errorf("%w: empty customfields body", common.ErrResolvingCustomFields)
-		}
-
-		wrapper, err := jsonquery.New(body).ObjectRequired("customfields")
-		if err != nil {
-			return nil, fmt.Errorf("%w: %w", common.ErrResolvingCustomFields, err)
-		}
-
-		arr, err := jsonquery.New(wrapper).ArrayOptional("customfields")
-		if err != nil {
-			return nil, fmt.Errorf("%w: %w", common.ErrResolvingCustomFields, err)
-		}
-
-		for _, n := range arr {
-			def, err := jsonquery.ParseNode[memberCustomFieldDefinition](n)
-			if err != nil {
-				return nil, fmt.Errorf("%w: %w", common.ErrResolvingCustomFields, err)
-			}
-
-			out = append(out, *def)
-		}
-
-		totalPages, err := jsonquery.New(wrapper).IntegerWithDefault("total_pages", 1)
-		if err != nil {
-			return nil, fmt.Errorf("%w: %w", common.ErrResolvingCustomFields, err)
-		}
-
-		if page >= int(totalPages) {
+		if page >= totalPages {
 			break
 		}
 
@@ -122,6 +84,64 @@ func (c *Connector) fetchMemberCustomFieldDefinitions(ctx context.Context) ([]me
 	}
 
 	return out, nil
+}
+
+func (c *Connector) getMemberCustomFieldDefinitionsPage(ctx context.Context, page int) (
+	[]memberCustomFieldDefinition, int, error,
+) {
+	url, err := urlbuilder.New(c.ProviderInfo().BaseURL, restPrefix, restVer, "customfields")
+	if err != nil {
+		return nil, 0, fmt.Errorf("%w: %w", common.ErrResolvingCustomFields, err)
+	}
+
+	url.WithQueryParam("page_size", "100")
+	url.WithQueryParam("page", strconv.Itoa(page))
+
+	resp, err := c.JSONHTTPClient().Get(ctx, url.String())
+	if err != nil {
+		return nil, 0, fmt.Errorf("%w: %w", common.ErrResolvingCustomFields, err)
+	}
+
+	if err := interpretPhoneBurnerEnvelopeError(resp); err != nil {
+		return nil, 0, fmt.Errorf("%w: %w", common.ErrResolvingCustomFields, err)
+	}
+
+	body, ok := resp.Body()
+	if !ok {
+		return nil, 0, fmt.Errorf("%w: empty customfields body", common.ErrResolvingCustomFields)
+	}
+
+	return parseMemberCustomFieldDefinitionsPage(body)
+}
+
+func parseMemberCustomFieldDefinitionsPage(body *ajson.Node) ([]memberCustomFieldDefinition, int, error) {
+	wrapper, err := jsonquery.New(body).ObjectRequired("customfields")
+	if err != nil {
+		return nil, 0, fmt.Errorf("%w: %w", common.ErrResolvingCustomFields, err)
+	}
+
+	arr, err := jsonquery.New(wrapper).ArrayOptional("customfields")
+	if err != nil {
+		return nil, 0, fmt.Errorf("%w: %w", common.ErrResolvingCustomFields, err)
+	}
+
+	out := make([]memberCustomFieldDefinition, 0, len(arr))
+
+	for _, n := range arr {
+		def, err := jsonquery.ParseNode[memberCustomFieldDefinition](n)
+		if err != nil {
+			return nil, 0, fmt.Errorf("%w: %w", common.ErrResolvingCustomFields, err)
+		}
+
+		out = append(out, *def)
+	}
+
+	totalPages, err := jsonquery.New(wrapper).IntegerWithDefault("total_pages", 1)
+	if err != nil {
+		return nil, 0, fmt.Errorf("%w: %w", common.ErrResolvingCustomFields, err)
+	}
+
+	return out, int(totalPages), nil
 }
 
 // flattenContactCustomFieldsInMap promotes each contacts GET "custom_fields" entry to a top-level key
--- a/providers/phoneburner/custom.go
+++ b/providers/phoneburner/custom.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"maps"
-	"regexp"
 	"strconv"
 	"strings"
 
@@ -15,11 +14,9 @@ import (
 	"github.com/spyzhov/ajson"
 )
 
-// customFieldKeyPrefix avoids collisions between PhoneBurner custom field display names
-// and built-in contact fields (e.g. "First Name" vs first_name).
+// customFieldKeyPrefix disambiguates member-defined custom field values on the contact from built-in
+// top-level contact keys when we promote custom_fields into the read record.
 const customFieldKeyPrefix = "custom_"
-
-var nonAlphaNumUnderscore = regexp.MustCompile(`[^a-z0-9_]+`) //nolint:gochecknoglobals
 
 // memberCustomFieldDefinition is a row from GET /rest/1/customfields (member-level definitions).
 // See: https://www.phoneburner.com/developer/route_list#customfields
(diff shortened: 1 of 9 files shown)


providers/phoneburner/custom.go
```diff
@@ -0,0 +1,201 @@
...
+		return common.ValueTypeSingleSelect
+	case "7":
+		return common.ValueTypeFloat
+	default:
+		return common.ValueTypeOther
+	}
+}
+
+// normalizeCustomFieldKey builds a stable, lower_snake_case key used in metadata and read flattening.
+func normalizeCustomFieldKey(label string) string {
```
Cobalt0s: I believe that naming of custom fields should be untouched. I have previously also formatted this to avoid collision.

But it seems that we should keep the field name as is and just move it up top the JSON root level.

However, I prefer "prefixes, suffixes" or any identifiable information for the field.

providers/phoneburner/read.go
```diff
@@ -166,11 +166,13 @@ func parseReadResponse(
 		}
 	}
 
-	records, err := recordsFunc(params.ObjectName)
+	baseRecords, err := recordsFunc(params.ObjectName)
 	if err != nil {
 		return nil, err
 	}
 
+	records := withContactCustomFieldFlatten(baseRecords, params.ObjectName)
```
Cobalt0s: You should do something like this: https://github.com/amp-labs/connectors/blob/main/providers/copper/handlers.go#L106

The main goal, that I am after is -- ensuring that Fields and Raw are not shared and we are operating on distinct copies. We modify and extract stuff into Fields, Raw must be left untoched!

providers/phoneburner/read_test.go
```diff
@@ -80,6 +81,39 @@ func TestRead(t *testing.T) { //nolint:funlen,gocognit,cyclop
...
+			Expected: &common.ReadResult{
+				Rows: 1,
+				Data: []common.ReadResultRow{{
+					Fields: map[string]any{
+						"contact_user_id":   "30919237",
+						"custom_lead_score": "42",
+					},
+					Raw: map[string]any{
+						"contact_user_id":   "30919237",
+						"custom_lead_score": "42",
```
Cobalt0s: This would be wrong! The Raw must be unchanged. In our case, raw shoudln't have custom properties, they are coming in the different call to enhance Fields, not Raw.

Cobalt0s (changes requested):
> no hard failure on the whole metadata response
Thats good design.

See comments below, then assign me as reviewer once addressed.

providers/phoneburner/custom.go
```diff
@@ -0,0 +1,213 @@
...
+	return record
+}
+
+// getMarshaledDataContactsWithCustomFieldsPreservingRaw builds each ReadResultRow from two separate
+// shallow copies of the contact map (same intent as copper's MakeMarshaledDataFunc(attachReadCustomFields)):
+// Raw is a clone of the row as returned by the provider (including "custom_fields" array only).
+// Fields are derived from a second clone that flattenContactCustomFieldsInMap mutates, promoting
+// custom values to top-level keys for ExtractLowercaseFieldsFromRaw. Merged custom keys exist only
+// in Fields, not in Raw, and the extracted []map from the response body is not mutated in place.
+func getMarshaledDataContactsWithCustomFieldsPreservingRaw(
```
Cobalt0s: There are existing methods that resolve this issue.

Copper uses outdated MakeMarshaledDataFunc which doesn't deal with Identifiers: https://github.com/amp-labs/connectors/blob/main/providers/copper/handlers.go#L106

So instead as per godoc use:
https://github.com/amp-labs/connectors/blob/main/providers/slack/handlers.go#L124

`common.RecordTransformer` is the only part that you should implement which will do the custom fields insertion. This influences the Fields only. The Raw is handled as well as Identifiers by `readhelpers.MakeMarshaledDataFuncWithId`.

providers/phoneburner/metadata.go
```diff
@@ -0,0 +1,68 @@
...
+
+		return result, nil
+	}
+
+	if objectMetadata.Fields == nil {
+		objectMetadata.Fields = make(common.FieldsMetadata)
+	}
+
+	for _, def := range definitions {
+		if def.DisplayName == "" {
```
Cobalt0s: This statement is a bit unusual. Does it mean that the payload was completely empty? Does it mean thtat the display name is empty and other fields are set, then what is the reason for provider to return something like this?

Should leave a comment or `fetchMemberCustomFieldDefinitions` should return non-empty values in the first place. Again it is interesting to know why the payload can be empty.

providers/phoneburner/metadata.go
```diff
@@ -0,0 +1,68 @@
...
+	if !slices.Contains(objectNames, objectContacts) {
+		return result, nil
+	}
+
+	objectMetadata, ok := result.Result[objectContacts]
+	if !ok {
+		return result, nil
+	}
+
+	definitions, err := c.fetchMemberCustomFieldDefinitions(ctx)
```
Cobalt0s, line 45: Should this always be invoked? Are there some objects that use it or custom fields can be created for all objects?

providers/phoneburner/read.go
```diff
@@ -153,17 +153,21 @@ func parseReadResponse(
...
+		return res, err
+	}
+
+	if params.ObjectName == objectContacts {
+		return common.ParseResult(
+			response,
+			common.MakeRecordsFunc(objectContacts, objectContacts),
+			nextRecordsURL(url, params.ObjectName),
+			readhelper.MakeMarshaledDataFuncWithId(
+				readContactRecordTransformer,
```
Cobalt0s, line 172: So custom fields are part of the normal response and custom fields are supported only for contacts. This should be documented in the godoc of this method.

providers/phoneburner/custom_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/amp-labs/connectors"
+	"github.com/amp-labs/connectors/common/readhelper"
+	"github.com/spyzhov/ajson"
+)
+
+func TestReadContactRecordTransformerAndMarshaledDataWithId(t *testing.T) {
+	t.Parallel()
+
+	const leadKey = "custom_Lead Score" // metadata / read request field key (prefix + display name)
```
Cobalt0s: I am a bit confused about the key name. We should avoid any emergent values via computations and should stick as close to the APIs as possible.

providers/phoneburner/custom_test.go
```diff
@@ -0,0 +1,71 @@
...
+		t.Fatal(`Raw must not contain flattened custom field key "custom_Lead Score"`)
+	}
+
+	if _, ok := row.Raw["custom_lead score"]; ok {
+		t.Fatal(`Raw must not contain Fields-style lowercased key "custom_lead score"`)
+	}
+
+	cf, ok := row.Raw["custom_fields"]
+	if !ok {
+		t.Fatal(`Raw must still include provider "custom_fields" array`)
```
Cobalt0s: This test is not needed. The read_test.go already covers everything and you already have the `custom_fields` field.

I added a new concept into Read testing: https://github.com/amp-labs/connectors/blob/main/test/utils/mockutils/readResult.go#L37. You may include `mockutils.Any` for any keys that you want to be present but dont care about the value. (In case value is very nested and pointless for the matter of testing but you care that the key is there).

providers/phoneburner/metadata_test.go
```diff
@@ -116,6 +123,42 @@ func TestListObjectMetadata(t *testing.T) { // nolint:funlen,gocognit,cyclop
...
+				Result: map[string]common.ObjectMetadata{
+					"contacts": {
+						DisplayName: "Contacts",
+						Fields: map[string]common.FieldMetadata{
+							"contact_user_id": {
+								DisplayName:  "Contact User Id",
+								ValueType:    "string",
+								ProviderType: "string",
+							},
+							"custom_Lead Score": {
```
Cobalt0s:
```
"lead score": {
  DisplayName:  "Lead Score",
  ValueType:    "float",
  ProviderType: "Numeric",
  IsCustom:     new(true),
},
```
Note: Change display name to be lower case. Also newest version of golang allows primitive values to be passed to `new` to create a pointer to that value.

providers/phoneburner/read_test.go
```diff
@@ -80,6 +81,53 @@ func TestRead(t *testing.T) { //nolint:funlen,gocognit,cyclop
...
+				},
+				Then: mockserver.Response(http.StatusOK, responseContactsWithCustomFields),
+			}.Server(),
+			Comparator: testroutines.ComparatorSubsetRead,
+			Expected: &common.ReadResult{
+				Rows: 1,
+				Data: []common.ReadResultRow{{
+					Fields: map[string]any{
+						"contact_user_id":   "30919237",
+						"custom_lead score": "42",
```
Cobalt0s: It should just be `lead score`.

From the test input files I see that fields are identified by their display name. You should just stick to it and use lower case. Spaces should be preserved.

```
 ...
 "lead score": "42",
 ...
```

providers/phoneburner/metadata.go
Fullstack-Hero: The check skips a single customfields row with a blank display_name, not an empty response; the API docs treat display_name as required on create and always show it on GET, but we still guard against bad rows because we key fields as custom_*

Fixed it as suggested.

providers/phoneburner/metadata.go
Fullstack-Hero: It is not always invoked: we only call `fetchMemberCustomFieldDefinitions` when "contacts" is in the objectNames list (see the guard at lines 28–30).

providers/phoneburner/custom_test.go
Fullstack-Hero: Fixed

providers/phoneburner/metadata.go
Cobalt0s: I see, I couldn't predict that. It makes sense. For such cases a comment is always helpful to describe the intent/reason even for something benign.

providers/phoneburner/custom.go
```diff
@@ -0,0 +1,193 @@
...
+	case "7":
+		return common.ValueTypeFloat
+	default:
+		return common.ValueTypeOther
+	}
+}
+
+// customFieldMetadataKey is the connector field name for a PhoneBurner custom field: the provider
+// display_name / custom_fields[].name lowercased with spaces preserved (e.g. "Lead Score" -> "lead score").
+func customFieldMetadataKey(displayName string) string {
```
Cobalt0s, line 54: Thanks!

providers/phoneburner/metadata.go
Cobalt0s: Oh, I see it!

providers/phoneburner/metadata.go
```diff
@@ -0,0 +1,64 @@
...
+	if len(objectNames) == 0 {
+		return nil, common.ErrMissingObjects
+	}
+
+	result, err := pbmetadata.Schemas.Select(c.ProviderContext.Module(), objectNames)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get base metadata: %w", err)
+	}
+
+	if !slices.Contains(objectNames, objectContacts) {
```
Cobalt0s: Ok after this line the method forks into 2 branches. Either return as is or attempt to invoke custom fields for contacts to enhance the metadata.

That should be expressed in some way. You can add some comments to "separate" function sections.
Or you could have a method `listObjectContactsMetadata`. Maybe some other name.

The way the funciton is layed out hints that querying customFieldDefinitions is the norm. While in fact it applies only to the contacts.

providers/phoneburner/custom_test.go
Cobalt0s: Thanks!

providers/phoneburner/read_test.go
```diff
@@ -80,6 +82,39 @@ func TestRead(t *testing.T) { //nolint:funlen,gocognit,cyclop
...
+				},
+				Then: mockserver.Response(http.StatusOK, responseContactsWithCustomFields),
+			}.Server(),
+			Comparator: testroutines.ComparatorSubsetRead,
+			Expected: &common.ReadResult{
+				Rows: 1,
+				Data: []common.ReadResultRow{{
+					Fields: map[string]any{
+						"contact_user_id": "30919237",
+						"lead score":      "42",
```
Cobalt0s, line 106: 👍

providers/phoneburner/read.go
```diff
@@ -180,6 +190,36 @@ func parseReadResponse(
 	)
 }
 
+// readWithClientSideTimeFilterIfNeeded handles members and voicemails when Since/Until require
+// client-side filtering. If it returns handled true, res and err are the outcome; otherwise the
+// caller should use the standard read path.
+func readWithClientSideTimeFilterIfNeeded(
```
Cobalt0s, line 196: 👍

providers/phoneburner/custom_test.go
```diff
@@ -0,0 +1,47 @@
+package phoneburner
+
+import (
+	"testing"
+
+	"github.com/spyzhov/ajson"
+)
+
+func TestParseMemberCustomFieldDefinitionsPage_skipsBlankDisplayName(t *testing.T) {
```
Cobalt0s, line 9: The test fits well and helps see right away what is happening in the impl. I see you declared `memberCustomFieldDefinition` much more readable. Approving.

Cobalt0s (approved):
Awesome! 🚀

PR #2902, opened 2026-04-18
[ENG-3798] feat: add google.SubscriptionEvent and history fan-out helper

anushat05:
Adds google.SubscriptionEvent + SubscriptionEventsFromHistory for Gmail push events.

### DIFF
--- a/providers/google/subscriptionEvent.go
+++ b/providers/google/subscriptionEvent.go
@@ -7,6 +7,16 @@ import (
 	"github.com/amp-labs/connectors/common"
 )
 
+// Gmail history change categories, as they appear in the history.list response.
+// Stored verbatim on SubscriptionEvent.RawName so EventType can map them to the
+// common subscription event taxonomy.
+const (
+	historyCategoryMessagesAdded   = "messagesAdded"
+	historyCategoryMessagesDeleted = "messagesDeleted"
+	historyCategoryLabelsAdded     = "labelsAdded"
+	historyCategoryLabelsRemoved   = "labelsRemoved"
+)
+
 // SubscriptionEvent is a single Gmail message-level event derived from a Gmail
 // watch push notification. Gmail pushes only carry {emailAddress, historyId};
 // the caller (typically a worker that received the push) is expected to call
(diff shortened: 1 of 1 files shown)


providers/google/subscriptionEvent.go
```diff
@@ -0,0 +1,157 @@
...
+
+var _ common.SubscriptionEvent = SubscriptionEvent{}
+
+// EventType returns the subscription event category (create/update/delete).
+// Falls back to Other if the type field is empty, matching the contract of
+// other providers' SubscriptionEvent implementations.
+func (e SubscriptionEvent) EventType() (common.SubscriptionEventType, error) {
+	if e.Type == "" {
+		return common.SubscriptionEventTypeOther, nil
+	}
```
jlimatampersand, line 58: this mapping should be done here `add(added.Message.ID, rec.ID, common.SubscriptionEventTypeCreate)`

not in the helper.

providers/google/subscriptionEvent.go
```diff
@@ -0,0 +1,157 @@
...
+			MessageID:    msgID,
+			HistoryID:    historyID,
+			EmailAddress: emailID,
+			Type:         evtType,
+		})
+	}
+
+	for _, rec := range history {
+		for _, added := range rec.MessagesAdded {
+			add(added.Message.ID, rec.ID, common.SubscriptionEventTypeCreate)
```
jlimatampersand: this conversion should be in eventname() method.

providers/google/subscriptionEvent.go
```diff
@@ -0,0 +1,157 @@
...
+// RawEventName returns the event type as a string. Gmail has no separate
+// provider-native event name (unlike HubSpot's "contact.creation"), so this
+// mirrors EventType.
+func (e SubscriptionEvent) RawEventName() (string, error) {
+	return string(e.Type), nil
+}
+
+// ObjectName returns the Gmail object name. Only "messages" is supported for
+// subscriptions today.
+func (e SubscriptionEvent) ObjectName() (string, error) {
```
jlimatampersand, line 73: I would list out all the potenial raw event, and then parse from there. 

But, for sake of time, and knowing we messages is the only object, we can create Todo ticket and comment, then move on.

jlimatampersand (approved):
Preapproved, but please resolve or respond to my comments before merging.

providers/google/subscriptionEvent.go
anushat05: Made the edits

providers/google/subscriptionEvent.go
anushat05: Added a TODO will follow up once MVP is merged

providers/google/subscriptionEvent.go
anushat05: Updated it

PR #2904, opened 2026-04-20
[CON-3195] feat: add support for quickbooks sandbox connector

Jkarage:
# Conventions
  
  Metadata Live Tests:
Screenshot 2026-04-20 at 20 54 21

Read Live Tests:
Screenshot 2026-04-20 at 20 55 49

Write Live Tests:
Screenshot 2026-04-20 at 20 56 30


laurenzlong (approved):
COMMENT REMOVED

PR #2910, opened 2026-04-21
[Customer Supprot] Fix false negative on Microsoft Token Refresh

jlimatampersand:
## Summary

Stop the Microsoft connector from classifying every Graph 401 as invalid credentials. Today every 401 maps to `common.ErrAccessToken`, which the server treats as a signal to flip the connection to `CONNECTION_STATUS_BAD_CREDS` — forcing re-auth even when the refresh token is still perfectly valid. This change inspects the response body and headers on 401 and splits the cases:

- **Genuinely transient conditions** (clock skew, AAD signing-key replication lag right after a refresh) → `common.ErrRetryable`, so Temporal retries absorb them and the connection stays healthy.
- **CAE / step-up claim challenges** (session revoked mid-flight, MFA step-up, compliant-device requirement) → a new `ClaimsChallengeError` type that wraps `common.ErrAccessToken`. The server still flips to bad_credentials (correct — the user must re-authenticate), but the error now carries structured diagnostic fields that land in existing server logs for operator visibility.
- **Everything else** on 401 → unchanged; continues to produce `common.ErrAccessToken`.

Non-401 responses are untouched.

## Problem

A Microsoft connection was observed flipping to `bad_credentials` immediately after a successful `login.microsoftonline.com` 200 OK refresh. Tracing the chain:

1. A read activity called Graph with the freshly-minted access token.
2. Graph returned 401 (transient token-validation race / CAE challenge / clock skew — all observed variants in production).
3. The connector's error handler mapped the 401 to `common.ErrAccessToken` unconditionally.
4. `shared/workflow/read/error.go:47-63` flipped the connection to `bad_credentials`, requiring manual re-auth to clear — even though the refresh token was still valid.

### DIFF
--- a/providers/microsoft/errors.go
+++ b/providers/microsoft/errors.go
@@ -1,8 +1,12 @@
 package microsoft
 
 import (
+	"encoding/json"
 	"fmt"
+	"net/http"
+	"strings"
 
+	"github.com/amp-labs/connectors/common"
 	"github.com/amp-labs/connectors/common/interpreter"
 )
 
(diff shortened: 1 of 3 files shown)


providers/microsoft/errors.go
```diff
@@ -219,3 +236,118 @@ func transientAuthErrorMessage(body []byte) (string, bool) {
...
+// The type exists so the server can still identify these events
+// specifically — via errors.As — and log or surface them differently from
+// a plain bad-credentials flip. That gives us the hook for a future
+// CONNECTION_STATUS_REAUTH_REQUIRED (or similarly-differentiated state)
+// without having to change the classification here.
+//
+// Example server-side use:
+//
+//	var chErr *microsoft.ClaimsChallengeError
+//	if errors.As(err, &chErr) {
```
Cobalt0s: Use (as per go1.26): 
```
chErr, ok := errors.AsType[*microsoft.ClaimsChallengeError](err)
```

Cobalt0s (approved):
The godocs are verbose, rather than that motivation and implementation makes sense.

PR #2911, opened 2026-04-22
chore: add company ID metadata to Procore

immdipu:


Jkarage:
Is this being added, because a user can be in several companies? A bit of context in the PR description would help. 
Also  this is for the deep connector implementation right?

immdipu:
> Is this being added, because a user can be in several companies? A bit of context in the PR description would help. Also this is for the deep connector implementation right?

yes correct. we need to add company id in the request header. this is primarily for the deep connector. For proxy user would have to attach the company id in the request header themselves.

Jkarage (approved):

PR #2917, opened 2026-04-24
Remove AssociationType from Assocation and fix HubSpot providerAssociationMetadata

laurenzlong:
**Problem:**

HubSpot associations actually an array of association types per associated record ([docs](https://developers.hubspot.com/docs/api-reference/latest/crm/associations/associate-records/guide#retrieve-associated-records)), e.g.

```
          "toObjectId": 5790939450, // one record
          "associationTypes": [ // multiple association types
            {
              "category": "HUBSPOT_DEFINED",
              "typeId": 1,
              "label": "Primary"
            },
            {
              "category": "HUBSPOT_DEFINED",
              "typeId": 279,
              "label": null
            },
```

The previous code was looping through `associationType` and creating a new association for each element of `associationType`, that means it creating multiple "associations" with the same ID. This leads to a problem in the server, because the server builds an "associationMap" keyed off of record ID. So it was essentially overwriting previous entries and only keeping the last associationType in the array. 

All of this logic seemed to be done in order to extract the appropriate `associationType` string, which we are not even reading in the server. 

Bug reported by customer ([private link](https://ampersand-company.slack.com/archives/C08TSJVEMC0/p1776877957101379?thread_ts=1775261388.668189&cid=C08TSJVEMC0))

**Solution**
- Remove `AssociationType` from `common.Association` (not used at all), and update tests accordingly
- Instead of creating multiple `common.Association` for a single associated record (one for each of its association types), just create one, and put the entire array of association types inside of `providerAssociationMetadata`


providers/hubspot/internal/crm/associations/fetch.go
```diff
@@ -157,21 +151,21 @@ func (s Strategy) fetchObjectAssociations( //nolint:cyclop,funlen
 		var assocs []common.Association
 
 		for _, assoc := range result.To {
-			for _, assocType := range assoc.AssociationTypes {
-				metadata := map[string]any{
+			assocTypes := make([]map[string]any, len(assoc.AssociationTypes))
```
laurenzlong, line 154: main fix is here

providers/hubspot/search_test.go
```diff
@@ -103,19 +103,19 @@ func TestSearch(t *testing.T) { // nolint:funlen,cyclop
 					},
 					Associations: map[string][]common.Association{
 						"companies": {{
-							ObjectId:        "29022297485",
```
laurenzlong, line 106: this is the main test that illustrates the change.

AmandaYao00 (approved):

PR #2921, opened 2026-04-24
feat: skip 404s in fetchMessage instead of failing the batch

anushat05:
### Summary
* When Gmail's GET /users/me/messages/{id} returns 404 for one id in a concurrent batch, the job was returning the error to simultaneously.DoCtx, which cancels the shared context and kills all other in-flight hydrates.
* For the subscribe flow (GetRecordsByIds) this meant one deleted/unknown message id caused the whole webhook batch to be nacked and eventually deadlettered.
* Skip 404s at fetchMessage: debug-log the missing id and return nil so the batch completes with whatever messages did resolve. The subscribe caller (fetchRecordsWithRetry) already tolerates partial/empty results. Non-404 errors are still surfaced so real failures aren't masked.

### Testing
* Added unit tests
* Ran unit tests without the fix and verified they failed and then ran with the fix and verified that passed

### DIFF
--- a/providers/google/internal/mail/messages.go
+++ b/providers/google/internal/mail/messages.go
@@ -5,8 +5,10 @@ import (
 	"errors"
 	"fmt"
 	"maps"
+	"net/http"
 
 	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/logging"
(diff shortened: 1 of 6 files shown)


eberle1080 (approved):
lgtm

PR #2923, opened 2026-04-25
feat(housecallPro): add employees webhook subscription support

Lafetz:
Adds support for subscribing to employees events.
Employee webhook events are not officially documented. Instead of having a dedicated event (employee.created), employee creation is emitted through the pro.created event.

# Test
image


AmandaYao00 (approved):
Thanks Yafet! Since this is mapping of pro.created -> employee.created is a bit unexpected, can we make sure it's documented somewhere in our customer-facing docs?

Lafetz:
> Thanks Yafet! Since this is mapping of pro.created -> employee.created is a bit unexpected, can we make sure it's documented somewhere in our customer-facing docs?

Ok, will do.

PR #2940, opened 2026-04-29
housecallPro: treat employee webhooks as passthrough events

Lafetz:
Employee webhook enrichment fails because Housecall Pro has no employee get-by-id API. Treating employee events as Other to route them through passthrough and avoid the failing fetch path.


AmandaYao00 (approved):
lgtm!

PR #2942, opened 2026-04-30
fix(gmail): cap concurrent messages.get fetches at 5

anushat05:
### Summary
* Gmail returns `429` `"Too many concurrent requests for user"` when subscribe hydration fans out one HTTP GET per message ID with no concurrency limit. The simultaneously package interprets maxConcurrent=-1 as unbounded, so a 60-message batch fired 60 simultaneous requests at Gmail.
* Cap at 5: stays under Gmail's per-user concurrency wall and uses ~half the 250-units/sec quota budget at typical latencies, leaving headroom for retries and other API traffic on the same token.

### DIFF
--- a/providers/google/internal/mail/messages.go
+++ b/providers/google/internal/mail/messages.go
@@ -18,10 +18,21 @@ import (
 
 // gmailMaxConcurrentFetches caps the number of in-flight messages.get requests per call.
 // Gmail enforces a per-user concurrent-request limit (returns 429 "Too many concurrent
-// requests for user" when exceeded) and a 250 quota-units/sec/user budget where
-// messages.get costs 5 units each. A cap of 5 keeps us under the concurrency wall and
-// uses ~half the quota-unit budget at typical latencies, leaving headroom for retries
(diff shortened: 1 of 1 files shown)


providers/google/internal/mail/messages.go
```diff
@@ -16,6 +16,14 @@ import (
 	"github.com/spyzhov/ajson"
 )
 
+// gmailMaxConcurrentFetches caps the number of in-flight messages.get requests per call.
+// Gmail enforces a per-user concurrent-request limit (returns 429 "Too many concurrent
+// requests for user" when exceeded) and a 250 quota-units/sec/user budget where
+// messages.get costs 5 units each. A cap of 5 keeps us under the concurrency wall and
```
laurenzlong: please add a relevant link to Google documentation if available

laurenzlong (approved):

providers/google/internal/mail/messages.go
anushat05: Updated the comment

PR #2880, opened 2026-04-14
[ENG-3815] feat(microsoft): Impl Delete Subscription

Cobalt0s:
# Description

Added support to remove the subscription.

# Live Test

At the end of execution the subscription will be automatically removed using this new functionality.

```
> TEST Subscribe/Write/Recieve
============== Starting Webhook Handler ==================
Starting webhook server on :4550
Run ngrok with command: `ngrok http 4550`
Local webhook server started at: "http://localhost:4550"
Env variable is missing "WEBHOOK_URL"
Please provide the public URL (e.g., from ngrok) that tunnels to this local server.
Public Webhook URL (empty string to cancel): https://f748-46-150-81-5.ngrok-free.app
Webhook URL: "https://f748-46-150-81-5.ngrok-free.app"
============== Invoking connector.Subscribe() ==================
{}
{
  "me/events": {
    "Events": [
      "create",
      "delete"
    ],
    "PassThroughEvents": [],
    "WatchFields": [],
    "WatchFieldsAll": false
  }
}
============== Invoking connector.Write/Delete() ==================
Creating object me/events:
{
  "Data": {
    "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#users('ae87552f-48fc-4dec-9322-65040bf9fdfd')/events/$entity",
    "@odata.etag": "W/\"fIY9UbXetEik54DbN4n/UQABNcwxvw==\"",
    "allowNewTimeProposals": true,
    "attendees": [],
    "bodyPreview": "",
    "categories": [],
    "changeKey": "fIY9UbXetEik54DbN4n/UQABNcwxvw==",
    "createdDateTime": "2026-05-13T20:21:15.0201475Z",
    "end": {
      "dateTime": "2026-05-13T21:00:00.0000255",
      "timeZone": "UTC"
    },
    "hasAttachments": false,
    "hideAttendees": false,
    "id": "AAMkAGY0YzAwY2ViLWQyODktNDI3NS1iNmY4LTE5YzU0MjI5ZTA4OQBGAAAAAABeMJSlO8qLToz2i2IQ1wsqBwB8hj1Rtd60SKTngNs3if9RAAB-ie_oAAB8hj1Rtd60SKTngNs3if9RAAE10fK_AAA=",
    "importance": "normal",
    "isAllDay": false,
    "isCancelled": false,
    "isDraft": false,
    "isOnlineMeeting": false,
    "isOrganizer": true,
    "isReminderOn": true,
[...]

PR #2879, opened 2026-04-14
[ENG-3815] feat(microsoft): Impl Update Subscription

Cobalt0s:
# Live Tests

The script creates subscription and then updates it to the new state.

From:
image

To:
image

Script output:
image

image

Just before the delete is invoked calling subscriptions endpoint to view subscription:
image

After the end of the script the subscriptions are removed:
image

PR #2945, opened 2026-05-01
feat(microsoft): Batch API

Cobalt0s:
# Description

Introduces the `batch` package for Microsoft Graph, providing a reusable foundation for batched requests (used by the Subscription feature. This can be extended to use cases like Salesforce-style batch read/write).

### Key points

* **Automatic batching**
  Requests are transparently chunked to comply with Microsoft’s limit of 20 requests per batch.

* **Deterministic result mapping**
  Each `RequestID` is mapped to exactly one outcome:

  * success → `Responses`
  * failure → `Errors`
    This makes it easy to correlate results with original requests.

* **Generic response handling**
  `Execute[B any]` parses successful responses into the desired type, enabling type-safe usage without additional boilerplate.
  It is implemented as a standalone function (instead of a method) due to Go’s limitations around method-level generics.

* **No top-level error**
  The function intentionally does not return an error.
  All failures (API or connector-level) are captured in `Result.Errors`, allowing callers to handle partial success scenarios and implement rollback logic more naturally.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -29,6 +29,8 @@ linters:
     - tagliatelle
     # We're not doing black-box testing, so enabling this would be counterproductive. We might want to discuss this though, it's a good practice.
     - testpackage
+    # Linter crashes when Generics are used.
+    - unparam
     # Annoying and not useful. Disabling otherwise we have to have comment everywhere
     - wrapcheck
     # Deprecated (since v2.2.0) due to: new major version. Replaced by wsl_v5.
--- a/providers/microsoft/internal/batch/operation.go
+++ b/providers/microsoft/internal/batch/operation.go
@@ -1,0 +1,115 @@
+package batch
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/amp-labs/connectors/common"
+)
+
+// ErrBatchResponse represents a failure returned by the Microsoft Graph Batch API
+// for an individual request within a batch.
+var ErrBatchResponse = errors.New("API error in batch response")
+
+// Execute performs a Microsoft Graph JSON batch request.
+//
+// It aggregates multiple logical HTTP requests into one call to the `$batch` endpoint:
+// https://learn.microsoft.com/en-us/graph/json-batching
+//
+// Core Contract:
+//   - Every RequestID provided is deterministically mapped to exactly one outcome:
+//   - success → stored in Result.Responses
+//   - failure → stored in Result.Errors
+//   - Failures may originate from:
+//   - API-level errors (non-2xx responses per request)
+//   - connector-level errors (transport, serialization, or chunk failure)
+//
+// Behavior:
+//   - Requests are automatically chunked to respect Graph API limits (max 20 per batch).
+//   - Absolute URLs are normalized to relative paths as required by Graph.
+//   - Each batch is executed independently; failures in one batch do not prevent others.
+//
+// Error Handling:
+//   - This method does NOT return a combined error.
+//   - Callers must inspect Result.Errors to handle failures.
+//   - This design ensures full visibility into partial success scenarios.
+//
+// Generics:
+//   - `B` defines the expected response body type.
+//   - The caller is responsible for ensuring compatibility with each request.
+//
+// Guarantees:
+//   - Each RequestID appears in either Responses or Errors (never both, never neither).
+//   - Order is not preserved unless explicitly reconstructed via List().
+//
+// Example:
+//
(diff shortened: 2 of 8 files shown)


Cobalt0s:
* **#2878** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2878?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2877** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2877?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2945** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2945?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2945?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.golangci.yml
```diff
@@ -29,6 +29,8 @@ linters:
     - tagliatelle
     # We're not doing black-box testing, so enabling this would be counterproductive. We might want to discuss this though, it's a good practice.
     - testpackage
+    # Linter crashes when Generics are used.
+    - unparam
```
Cobalt0s, line 33: For some reason even the newest version of linter compiled with latest golang version still crashes.

providers/microsoft/internal/batch/params.go
```diff
@@ -0,0 +1,94 @@
...
+//   - The identifier is the only mechanism used to correlate responses.
+func (p *Params) WithRequest(
+	requestIdentifier RequestID,
+	method string,
+	url *urlbuilder.URL,
+	body any,
+	headers map[string]any,
+) *Params {
+	if p.payloads == nil {
+		p.payloads = make([]*payloadRequest, 0)
```
jlimatampersand: nit: can't you just append without initializing? append(...) should work on nil array

jlimatampersand (approved):

providers/microsoft/internal/batch/params.go
Cobalt0s: You are correct, fixed.

providers/microsoft/internal/batch/operation.go
```diff
@@ -0,0 +1,115 @@
...
+func Execute[B any](ctx context.Context, strategy *Strategy, params *Params) *Result[B] {
+	bundle := &Result[B]{
+		Responses: make(map[RequestID]Envelope[B]),
+		Errors:    make(map[RequestID]Envelope[error]),
+	}
+
+	// Chunk the list of payloads into consumable sizes, otherwise API will reject large number of requests.
+	// Save the output into the bundle.Raw and bundle.Response.
+	// In case of an error the bundle.Raw and bundle.Errors are populated.
+	for _, payloads := range params.chunkPayloads() {
```
jlimatampersand: for performance use, the `simultaneous` package which will allow us to use safeguarded go routines.

jlimatampersand (changes requested):

providers/microsoft/internal/batch/operation.go
Cobalt0s: Yes, that is a natural progression. I didn't do that right away due to complexity and forgot to revisit and add parallel processing.

It is done now. I ran the live Subscribtion test. Possibly would need to create more than 20 payloads to see several routines, in theory should work.

I added this change as a dedicated commit for easier review, instead of squashing it: https://github.com/amp-labs/connectors/pull/2945/changes/661ce31eadf8022a2c945651375182ef7e140504

providers/microsoft/internal/batch/operation.go
```diff
@@ -0,0 +1,171 @@
...
+	for index, payloads := range payloadList {
+		callbacks[index] = func(ctx context.Context) error {
+			batchRoutine(ctx, strategy, payloads, responseChannel, errChannel)
+
+			return nil
+		}
+	}
+
+	// Wait for all routines.
+	if err := simultaneously.DoCtx(ctx, -1, callbacks...); err != nil {
```
jlimatampersand: let's use 3 for the max concurrency

jlimatampersand (approved):
Minor comment. Sorry about coming back on this PR late. I lost track of PRs. I fixed my PR review pipeline, so I should be responding more promptly now.

providers/microsoft/internal/batch/operation.go
Cobalt0s: Probably we would need to address this in other places in the codebase: 

image

PR #2947, opened 2026-05-01
feat(goTo): add postAuth and modules

immdipu:
- Added postAuth for GoTo. We need it to fetch the `accountKey` from `/admin/rest/v1/me`,  GoTo API endpoints require it.
- Split GoTo into two modules: `goTo` (api.getgo.com, default) and `goToConnect` (api.goto.com), since the two share auth but live on different base URLs.

### DIFF
--- a/providers/goto/authmetadata.go
+++ b/providers/goto/authmetadata.go
@@ -14,6 +14,10 @@ func (c *Connector) GetPostAuthInfo(ctx context.Context) (*common.PostAuthInfo, 
 		return nil, err
 	}
 
+	if accountKey == "" {
+		return nil, common.ErrMissingExpectedValues
+	}
+
 	c.accountKey = accountKey
 
 	catalogVars := map[string]string{
--- a/providers/goto/authmetadata.go
+++ b/providers/goto/authmetadata.go
@@ -1,0 +1,65 @@
+package gotoconn
+
+import (
+	"context"
+	"strconv"
+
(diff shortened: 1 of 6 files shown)


providers/goto/authmetadata.go
```diff
@@ -0,0 +1,69 @@
...
+
+	if accountKey == "" {
+		return nil, common.ErrMissingExpectedValues
+	}
+
+	c.accountKey = accountKey
+
+	catalogVars := map[string]string{
+		"accountKey": accountKey,
+	}
```
RajatPawar: minor: we can probably use `AuthMetadataVars.AsMap()`

providers/goto/authmetadata.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	data, err := common.UnmarshalJSON[map[string]any](resp)
+	if err != nil {
+		return "", common.ErrFailedToUnmarshalBody
+	}
+
+	if data == nil {
```
RajatPawar: From AI reviewer, worth looking at: `common.UnmarshalJSON` returns a non-nil `*T` on success, so this branch is unreachable. If the intent is to handle empty-body responses, check `resp.Body()` instead (and return `common.ErrEmptyJSONHTTPResponse`).

RajatPawar (approved):

PR #2954, opened 2026-05-05
fix(connectwise): Rename opportunities

Cobalt0s:
# Description

* Rename object `opportunities` to `sales/opportunities`.
* Remove trailing slashes at the end of object names.

# Live Tests

The old object name is no longer matching the URL endpoint:
image

However, the new name does:
image

(Note: there is the issue with the creds. For the demonstration of the object name it is sufficient because it has "resolved" and located the respective URL endpoint, which is the main goal.)

### DIFF
--- a/scripts/openapi/connectwise/metadata/main.go
+++ b/scripts/openapi/connectwise/metadata/main.go
@@ -369,20 +369,22 @@ func main() {
 
 	objects := Objects()
 	for _, object := range objects {
+		objectName := getObjectName(object)
+
 		if object.Problem != nil {
 			slog.Error("schema not extracted",
-				"objectName", object.ObjectName,
+				"objectName", objectName,
 				"error", object.Problem,
 			)
 		}
 
 		for _, field := range object.Fields {
-			schemas.Add(common.ModuleRoot, object.ObjectName, object.DisplayName, object.URLPath, object.ResponseKey,
+			schemas.Add(common.ModuleRoot, objectName, object.DisplayName, object.URLPath, object.ResponseKey,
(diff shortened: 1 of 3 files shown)


Cobalt0s:
* **#2954** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2954?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2954?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

scripts/openapi/connectwise/metadata/main.go
```diff
@@ -426,3 +428,16 @@ func collectionDisplayName(displayName string) string {
 
 	return naming.NewPluralString(collectionName).String()
 }
+
+// getObjectName hardcodes object names to make names consistent across related endpoints.
+// Fixes any trailing slashes.
+func getObjectName(object metadatadef.Schema) string {
```
laurenzlong, line 434: is this function actually necessary? Since you already fixed the schema above? It's ok for us to now reject `opportunities`, we don't have to coerce it into `sales/opportunities`

laurenzlong (approved):

scripts/openapi/connectwise/metadata/main.go
Cobalt0s: This script uses OpenAPI data and then applies set of transformers to get `schema.json`.

The updated `schema.json` is the result of this function. Therefore, this function is needed.

PS: The `schema.json` is read only file. (Some connectors that manually create it are exceptions to this rule).

PR #2960, opened 2026-05-06
Allow Salesforce Optimization to use exiting fields

jlimatampersand:


test/salesforce/subscribe/register.go
```diff
@@ -1,166 +0,0 @@
-package main
```
jlimatampersand, line 1: renamed this file to subscribe.go

PR #2975, opened 2026-05-08
[CON-3237] feat(livestorm): Read implementation

Fullstack-Hero:
## Summary

This PR delivers the Livestorm deep connector: schema-backed `ListObjectMetadata`, OAuth-based HTTP read against Livestorm’s JSON:API (application/vnd.api+json), factory wiring in connector/new.go, and provider catalog updates so Livestorm exposes Read in addition to Proxy. Reads cover list endpoints (events, people, people_attributes), paginated session chat messages (session id via ReadParams.Filter), and single job fetch by id (ReadParams.Filter). Incremental reads for events use filter[updated_since] / filter[updated_until] mapped from Since / Until. Pagination uses meta.next_page when present, otherwise meta.current_page / meta.page_count with page[number] / page[size].

## Key Features

- `Embedded metadata`: providers/livestorm/metadata/schemas.json + schema.NewOpenAPISchemaProvider for ListObjectMetadata.
- `JSON`:API read parsing: Flattens data[] / single data resources (id + attributes) into flat records; IDs surfaced via readhelper.MakeGetMarshaledDataWithId.
-` Multi-object reads`: Lists for standard objects; nested routes require Filter: session_chat_messages → session UUID, jobs → job UUID.
- `Incremental events`: Since/Until → filter[updated_since] / filter[updated_until] (Unix).
- `Pagination`: Next URL from meta.next_page or computed page index from meta.
- `Factory & catalog`: providers.Livestorm registered in connector/new.go; providers/livestorm.go sets Support.Read: true.
- `Tests`: Metadata unit tests (metadata_test.go); integration harness under test/livestorm/ (extend with read/live scripts as you finish them).

## Testing
```bash
# Unit tests
go test ./providers/livestorm -v

go test ./providers/livestorm -cover

# Integration-style scripts
go run ./test/livestorm/metadata/main.go
```

### DIFF
--- a/common/urlbuilder/url.go
+++ b/common/urlbuilder/url.go
@@ -50,12 +50,21 @@ func New(base string, path ...string) (*URL, error) {
 	return u, nil
 }
 
-// FromRawURL converts a core Go `url.URL` into `urlbuilder.URL`,
-// providing better control over query parameters and encoding.
+// FromRawURL creates an independent copy of a net/url.URL as *urlbuilder.URL.
 //
-// The input URL is not stored by pointer: a copy is made so mutating this
-// urlbuilder.URL (including String(), which updates the delegate's RawQuery)
-// never changes the original *url.URL (for example http.Request.URL).
+// Unlike direct pointer use, mutations to the returned URL—including path
+// changes via AddPath(), query updates via WithQueryParam()/RemoveQueryParam(),
+// and String() which regenerates RawQuery—do not modify the original *url.URL.
+//
+// This safety is critical when wrapping http.Request.URL or parsed endpoints,
+// as urlbuilder.URL provides immutable query keys, proper percent-encoding,
+// and builder-style modifications for robust API construction.
+//
+// Example:
+//
+//	u, _ := url.Parse("https://api.example.com?existing=1")
+//	b, _ := FromRawURL(u)
+//	b.WithQueryParam("new", "value") // Original u unchanged
 func FromRawURL(rawURL *url.URL) (*URL, error) {
 	if rawURL == nil {
 		return nil, ErrInvalidURL
--- a/common/urlbuilder/url.go
+++ b/common/urlbuilder/url.go
@@ -52,14 +52,27 @@ func New(base string, path ...string) (*URL, error) {
 
 // FromRawURL converts a core Go `url.URL` into `urlbuilder.URL`,
 // providing better control over query parameters and encoding.
+//
+// The input URL is not stored by pointer: a copy is made so mutating this
+// urlbuilder.URL (including String(), which updates the delegate's RawQuery)
+// never changes the original *url.URL (for example http.Request.URL).
 func FromRawURL(rawURL *url.URL) (*URL, error) {
-	values, err := url.ParseQuery(rawURL.RawQuery)
+	if rawURL == nil {
+		return nil, ErrInvalidURL
+	}
+
+	delegate, err := url.Parse(rawURL.String())
+	if err != nil {
+		return nil, errors.Join(err, ErrInvalidURL)
+	}
+
+	values, err := url.ParseQuery(delegate.RawQuery)
 	if err != nil {
 		return nil, errors.Join(err, ErrInvalidURL)
 	}
 
 	return &URL{
-		delegate:             rawURL,
+		delegate:             delegate,
 		queryParams:          values,
 		unencodedQueryParams: url.Values{},
 		encodingExceptions:   make(map[string]string),
--- a/common/urlbuilder/url_test.go
+++ b/common/urlbuilder/url_test.go
@@ -295,6 +295,74 @@ func TestFromRawURL(t *testing.T) { // nolint:funlen
 	}
 }
 
+func TestFromRawURLDoesNotMutateInput(t *testing.T) {
+	t.Parallel()
+
+	original, err := url.Parse("https://api.example.com/v1/items?page[number]=0&page[size]=100")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	before := original.String()
+	beforeRawQuery := original.RawQuery
+
+	wrapped, err := FromRawURL(original)
+	if err != nil {
+		t.Fatalf("FromRawURL: %v", err)
+	}
+
+	wrapped.WithQueryParam("page[number]", "5")
+	_ = wrapped.String()
+
+	if got := original.String(); got != before {
+		t.Fatalf("source URL changed after builder.String(): before %q after %q", before, got)
+	}
+
+	if original.RawQuery != beforeRawQuery {
+		t.Fatalf("source RawQuery changed: before %q after %q", beforeRawQuery, original.RawQuery)
+	}
+}
+
+func TestFromRawURLIndependentOfLaterSourceMutation(t *testing.T) {
+	t.Parallel()
+
+	original, err := url.Parse("https://example.com/path?keep=1")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	wrapped, err := FromRawURL(original)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	original.RawQuery = "tampered=1"
+
+	wrapped.WithQueryParam("keep", "2")
+	out := wrapped.String()
+
+	if strings.Contains(out, "tampered") {
+		t.Fatalf("output should not reflect post-wrap mutation of source URL: %q", out)
+	}
+
+	if !strings.Contains(out, "keep=2") {
+		t.Fatalf("expected keep=2 in %q", out)
+	}
+}
+
+func TestFromRawURLNilInput(t *testing.T) {
+	t.Parallel()
+
+	_, err := FromRawURL(nil)
+	if err == nil {
+		t.Fatal("expected error for nil *url.URL")
+	}
+
+	if !errors.Is(err, ErrInvalidURL) {
+		t.Fatalf("expected ErrInvalidURL, got %v", err)
+	}
+}
+
 func TestEquality(t *testing.T) { // nolint:funlen
 	t.Parallel()
 
--- a/common/urlbuilder/url_test.go
+++ b/common/urlbuilder/url_test.go
@@ -306,13 +306,13 @@ func TestFromRawURLDoesNotMutateInput(t *testing.T) {
 	before := original.String()
 	beforeRawQuery := original.RawQuery
 
-	wrapped, err := FromRawURL(original)
+	built, err := FromRawURL(original)
 	if err != nil {
 		t.Fatalf("FromRawURL: %v", err)
 	}
 
-	wrapped.WithQueryParam("page[number]", "5")
-	_ = wrapped.String()
+	built.WithQueryParam("page[number]", "5")
+	_ = built.String()
 
 	if got := original.String(); got != before {
 		t.Fatalf("source URL changed after builder.String(): before %q after %q", before, got)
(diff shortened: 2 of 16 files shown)


providers/livestorm/read.go
```diff
@@ -0,0 +1,241 @@
...
+
+	if params.ObjectName == objectJobs {
+		return c.buildJobReadURL(params)
+	}
+
+	return c.buildGenericReadURL(params)
+}
+
+func (c *Connector) buildSessionChatMessagesReadURL(params common.ReadParams) (*urlbuilder.URL, error) {
+	sessionID := strings.TrimSpace(params.Filter)
```
Cobalt0s, line 79: Interesting. I think it is fine to use the Filter. Then you should document this in the docs so people know that the endpoint is supported for chat messages if they specify the filter. Please, check if this option can be specified in the amp.yaml. I am not sure how it is used, although, I have added the Filter support to Klaviyo.

providers/livestorm/jsonapi.go
```diff
@@ -0,0 +1,84 @@
...
+
+// extractJSONAPIDataRecords returns flattened records from a JSON:API body.
+// The "data" property may be an array of resources or a single resource object.
+func extractJSONAPIDataRecords(root *ajson.Node) ([]map[string]any, error) {
+	items, arrErr := jsonquery.New(root).ArrayOptional("data")
+	if arrErr == nil {
+		return flattenJSONAPIResourceList(items)
+	}
+
+	obj, objErr := jsonquery.New(root).ObjectOptional("data")
```
Cobalt0s, line 19: What endpoints use it? They wouldn't be a collection of records right?

The Read focuses on collections, not single items.

providers/livestorm/read.go
```diff
@@ -0,0 +1,241 @@
...
+
+func (c *Connector) parseReadResponse(
+	_ context.Context,
+	params common.ReadParams,
+	request *http.Request,
+	resp *common.JSONHTTPResponse,
+) (*common.ReadResult, error) {
+	return common.ParseResult(
+		resp,
+		extractJSONAPIDataRecords,
```
Cobalt0s: This method has calls to flatten which should not happen.

`common.ReadResult.Data[*].Fields` -> may contain fileds that are nested and flattened, may contain custom fields. It holds the targeted content matching the query.
`common.ReadResult.Data[*].Raw` -> the raw JSON of the record which should preserve all the nesting. If there are extra calls for custom fields, or any other data to embed, -- it must not be embeded here.

The flattening, custom fields, extra embeding of side-car data can occur only via `readhelper.MakeMarshaledDataFuncWithId`.

The second argument of `common.ParseResult` method should do casting of data and no manipulation.

We have lots of similar `common.ParseResult` and cleaning all of them into consitent contract would be a difficult task.

providers/livestorm/read.go
```diff
@@ -0,0 +1,241 @@
...
+
+	return req, nil
+}
+
+func (c *Connector) buildReadURL(params common.ReadParams) (*urlbuilder.URL, error) {
+	if len(params.NextPage) != 0 {
+		return urlbuilder.New(params.NextPage.String())
+	}
+
+	if params.ObjectName == "" {
```
Cobalt0s: You could use `switch params.ObjectName`.

providers/livestorm/read.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	endpointURL.WithQueryParam("page[number]", "0")
+	endpointURL.WithQueryParam("page[size]", readhelper.PageSizeWithDefaultStr(params, defaultPageSize))
+
+	return endpointURL, nil
+}
+
+func (c *Connector) buildJobReadURL(params common.ReadParams) (*urlbuilder.URL, error) {
+	jobID := strings.TrimSpace(params.Filter)
```
Cobalt0s, line 96: I see, so many endpoints just depend on the IDs. I would consider this as a hack. We can keep it.

providers/livestorm/read.go
```diff
@@ -0,0 +1,241 @@
...
+	return endpointURL, nil
+}
+
+func buildVersionedPathURL(baseURL string, path string) (*urlbuilder.URL, error) {
+	path = strings.TrimSpace(path)
+	if path == "" {
+		return urlbuilder.New(baseURL, apiVersion)
+	}
+
+	path = strings.TrimPrefix(path, "/")
```
Cobalt0s: Triming is not needed. `urlbuilder` does it and it has unit tests supporting this functionality.

https://github.com/amp-labs/connectors/blob/main/common/urlbuilder/url_test.go#L146

providers/livestorm/read.go
```diff
@@ -0,0 +1,241 @@
...
+	if pc <= 0 || cur+1 >= pc {
+		return "", nil
+	}
+
+	cloned, err := url.Parse(previous.URL.String())
+	if err != nil {
+		return "", err
+	}
+
+	u, err := urlbuilder.FromRawURL(cloned)
```
Cobalt0s:
```
u, err := urlbuilder.FromRawURL(previous.URL)
```

Do you think we should improve `FromRawURL`: https://github.com/amp-labs/connectors/blob/main/common/urlbuilder/url.go#L56. The delegate is not fully copied. So changing query params in urlbuilder.URL would affect `url.URL` inside `http.Request`.
It does look like bad code because of the side effects.

If you want, you can improve this method and write unit tests that prove that side effects do not exist and 2 concepts are indpendent. But then we should check that by mistake nobody relies on the side effects, almost 100% certain that nobody is changing both url.URL to impact urlbuilder.URL which as of now kind of a wrapper, but was not intended that way.

providers/livestorm/read_test.go
```diff
@@ -0,0 +1,343 @@
...
+				ObjectName: "people",
+				NextPage:   testroutines.URLTestServer + "/v1/people?page[number]=2&page[size]=100",
+				Fields:     connectors.Fields("email"),
+			},
+			Server: mockserver.NewServer(func(w http.ResponseWriter, r *http.Request) {
+				w.Header().Set("Content-Type", "application/json")
+				w.WriteHeader(http.StatusOK)
+
+				body := fmt.Sprintf(
+					`{"data":[{"id":"person_1","type":"people","attributes":{"email":"a@example.com"}}],"meta":{"next_page":"http://%s/v1/people?page[number]=3&page[size]=100"}}`,
```
Cobalt0s, line 199: Interesting case, where you need `r.Host`.

What would happen if you hard code the host to be anything. Is there any particuolar reason for doing this?

providers/livestorm/read_test.go
```diff
@@ -0,0 +1,343 @@
...
+	until := time.Unix(1735776000, 0).UTC()
+
+	eventsBody := testutils.DataFromFile(t, "read-events.json")
+	eventsMultipageBody := testutils.DataFromFile(t, "read-events-multipage.json")
+	peopleFirstPageBody := testutils.DataFromFile(t, "read-people-first-page.json")
+	chatMessagesBody := testutils.DataFromFile(t, "read-session-chat-messages.json")
+	jobBody := testutils.DataFromFile(t, "read-job.json")
+	peopleAttributesBody := testutils.DataFromFile(t, "read-people-attributes.json")
+
+	tests := []testroutines.Read{
```
Cobalt0s, line 33: Very nice!

Cobalt0s (changes requested):
Looks good. One important change is needed for `Fields` <==> `Raw`.

providers/livestorm/jsonapi.go
Fullstack-Hero: Livestorm returns data as an array for lists and as a single object for get job by id (/v1/jobs/{id}), we normalize both to read rows (one row when it’s a single resource).

providers/livestorm/read_test.go
Fullstack-Hero: We use r.Host so the mocked meta.next_page URL matches the httptest server’s real host/port (the test compares that string to {{testServerURL}} after substitution). A hardcoded host would fail whenever the ephemeral port differs.

common/urlbuilder/url_test.go
```diff
@@ -295,6 +295,74 @@ func TestFromRawURL(t *testing.T) { // nolint:funlen
...
+
+	original, err := url.Parse("https://api.example.com/v1/items?page[number]=0&page[size]=100")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	before := original.String()
+	beforeRawQuery := original.RawQuery
+
+	wrapped, err := FromRawURL(original)
```
Cobalt0s: We shouldn't call it wrapped. It is its own concept "disconnected" from url.URL.

common/urlbuilder/url_test.go
```diff
@@ -295,6 +295,74 @@ func TestFromRawURL(t *testing.T) { // nolint:funlen
...
+	if got := original.String(); got != before {
+		t.Fatalf("source URL changed after builder.String(): before %q after %q", before, got)
+	}
+
+	if original.RawQuery != beforeRawQuery {
+		t.Fatalf("source RawQuery changed: before %q after %q", beforeRawQuery, original.RawQuery)
+	}
+}
+
+func TestFromRawURLIndependentOfLaterSourceMutation(t *testing.T) {
```
Cobalt0s, line 326: All tests are good, thanks!

common/urlbuilder/url.go
```diff
@@ -52,14 +52,27 @@ func New(base string, path ...string) (*URL, error) {
 
 // FromRawURL converts a core Go `url.URL` into `urlbuilder.URL`,
 // providing better control over query parameters and encoding.
+//
+// The input URL is not stored by pointer: a copy is made so mutating this
+// urlbuilder.URL (including String(), which updates the delegate's RawQuery)
+// never changes the original *url.URL (for example http.Request.URL).
 func FromRawURL(rawURL *url.URL) (*URL, error) {
```
Cobalt0s, line 68:
```
// FromRawURL creates an independent copy of a net/url.URL as *urlbuilder.URL.
//
// Unlike direct pointer use, mutations to the returned URL—including path
// changes via AddPath(), query updates via WithQueryParam()/RemoveQueryParam(),
// and String() which regenerates RawQuery—do not modify the original *url.URL.
//
// This safety is critical when wrapping http.Request.URL or parsed endpoints,
// as urlbuilder.URL provides immutable query keys, proper percent-encoding,
// and builder-style modifications for robust API construction.
//
// Example:
//  u, _ := url.Parse("https://api.example.com?existing=1")
//  b, _ := FromRawURL(u)
//  b.WithQueryParam("new", "value")  // Original u unchanged
func FromRawURL(rawURL *url.URL) (*URL, error) {
```

providers/livestorm/read.go
```diff
@@ -0,0 +1,230 @@
...
+		applyEventTimeFilters(endpointURL, params)
+	}
+
+	return endpointURL, nil
+}
+
+func buildVersionedPathURL(baseURL string, path string) (*urlbuilder.URL, error) {
+	path = strings.TrimSpace(path)
+	if path == "" {
+		return urlbuilder.New(baseURL, apiVersion)
```
Cobalt0s: Empty strings can be passed to `New()`. Under the hood it will call `JoinPath`.

I checked and it omits it:
Image

providers/livestorm/read_test.go
Cobalt0s: I guess this is because connector is building the URL using `FromRawURL`. I checked and I see this is
all because you are testing the `offset pagination`. You should state that in the test. "Offset pagination testing" or something. That would give explanation of what is hapenning in the mock server. What you are doing makes sense total sense, I didn't see what connector execution path you are testing, so it was not clear if it was avoidable.

providers/livestorm/read_test.go
```diff
@@ -0,0 +1,435 @@
...
+		Data: []common.ReadResultRow{
+			{
+				Fields: map[string]any{
+					"email":      "alpha@example.com",
+					"first_name": "Alpha",
+				},
+				Raw: map[string]any{
+					"id":   "person_a",
+					"type": "people",
+					"attributes": map[string]any{
```
Cobalt0s, line 419: Perfect! Thanks!

Cobalt0s (approved):
Great!

providers/livestorm/jsonapi.go
Cobalt0s: We should exclude "single resources", endpoint that return just one object from Read impl.

Read should be called -> ReadCollection. At least that is the intention of that method. I understand that many connectors are not utilizing the whole potential of the provider APIs. Thats not a problem yet.

PR #2984, opened 2026-05-12
[CON-3238] Feet/livestorm write delete implementation

Fullstack-Hero:


providers/livestorm/write.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+}
+
+func (c *Connector) buildWriteURL(params common.WriteParams) (*urlbuilder.URL, string, error) {
+	base := c.ProviderInfo().BaseURL
+
+	switch params.ObjectName {
+	case objectEvents:
+		if params.IsCreate() {
+			u, err := urlbuilder.New(base, "/v1/events")
```
Cobalt0s: Where possible create a constant apiVersion so all locations reference it. Later when we need to update the connector version it will be a bit easier.

providers/livestorm/write.go
```diff
@@ -0,0 +1,267 @@
...
+			"attributes": record,
+		},
+	})
+}
+
+// marshalSessionPeopleBulkBody builds POST /sessions/{id}/people/bulk JSON:API body.
+// Pass session_id alongside attributes or a full document; see
+// https://developers.livestorm.co/reference/post_sessions-id-people-bulk
+func marshalSessionPeopleBulkBody(record map[string]any) ([]byte, error) {
+	record = maps.Clone(record)
```
Cobalt0s: No need to clone. Record is already a copy.
Anytime you need a new copy you can invoke `RecordDataToMap` like you did or `WriteParams.GetRecord` which does the same thing.

providers/livestorm/write.go
```diff
@@ -0,0 +1,267 @@
...
+		u, err := urlbuilder.New(base, "/v1/users")
+
+		return u, http.MethodPost, err
+	case objectSessionPeopleBulk:
+		record, err := common.RecordDataToMap(params.RecordData)
+		if err != nil {
+			return nil, "", err
+		}
+
+		sessionID, err := stringField(record, "session_id")
```
Cobalt0s: I see so you have engineered the payload to accept session id.
In other words you have `provider-payload + provider-id-list = connector input`. 

I think we did it few times, I wouldn't recall where. That should be documented in the user facing docs.

However, the bulk operations are often part of the Bulk feature which is seperate from Write. I think lets remove it. There is no defined Bulk interface but you can find some impl in the codebase: https://github.com/amp-labs/connectors/blob/main/providers/salesforce/bulk-write.go#L20.
I think only Salesforce has it. There is even config allocated for such feature: https://github.com/amp-labs/connectors/blob/main/providers/salesforce.go#L68

providers/livestorm/write_test.go
```diff
@@ -0,0 +1,184 @@
...
+						"attributes": map[string]any{"title": "Updated"},
+					},
+				},
+			},
+		},
+		{
+			Name: "Create user",
+			Input: common.WriteParams{
+				ObjectName: objectUsers,
+				RecordData: map[string]any{"email": "new@example.com"},
```
Cobalt0s, line 106: The test demonstrating payload wrapping is good.

Cobalt0s (approved):
Write/Delete feature are looking good.

I am not sure about the Bulk object. I feel it should be removed from the Write.

PR #2986, opened 2026-05-12
[CON-3243] feat: Add google calendar subscriber connector

Jkarage:
Implements Subscribe, UpdateSubscription, DeleteSubscription, and RunScheduledMaintenance for the Google Calendar connector using the Calendar Watch API.

  Objects supported:
  - events — POST /calendars/primary/events/watch
  - calendarList — POST /users/me/calendarList/watch
  - settings — POST /users/me/settings/watch
  - acl — POST /calendars/primary/acl/watch 
 
Live Tests:
Screenshot 2026-05-12 at 16 54 47
Screenshot 2026-05-12 at 16 55 00

 Google Calendar push notifications are delivered as HTTP POST requests to the address set in WatchRequest.Address. The body is always empty. All
  information arrives in headers:

  Header: X-Goog-Resource-State
    Values: sync
    Meaning: Initial handshake on channel creation — ignore it   
    Values: exists          
    Meaning: A resource was created or updated
    Values: not_exists
    Meaning: A resource was deleted

  Header: X-Goog-Channel-ID
    Values: e.g. <base-id>-events
    Meaning: The channel ID set during subscribe

  Header: X-Goog-Resource-ID
    Values: opaque string
    Meaning: Stable ID of the watched resource (needed to stop the channel)

  Header: X-Goog-Resource-URI
    Values: URL
    Meaning: The API endpoint of the changed resource

  Header: X-Goog-Channel-Token
    Values: string
    Meaning: The token set in WatchRequest.Token, echoed back verbatim — use this to verify the notification came from Google

  Header: X-Goog-Channel-Expiration
    Values: RFC 1123 date
    Meaning: When the channel expires

### DIFF
--- a/providers/google/internal/calendar/subscribe.go
+++ b/providers/google/internal/calendar/subscribe.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"net/http"
 	"net/url"
 	"slices"
@@ -27,7 +28,7 @@ var objectWatchPaths = map[common.ObjectName]string{
 	objectNameEvents:       "calendars/primary/events/watch",
 	objectNameCalendarList: "users/me/calendarList/watch",
 	objectNameSettings:     "users/me/settings/watch",
-	objectNameAcl:          "calendars/primary/acl/watch",
+	objectNameACL:          "calendars/primary/acl/watch",
 }
 
 // WatchRequest is the caller-provided config for creating watch channels.
@@ -178,7 +179,7 @@ func (a *Adapter) Subscribe(
 // Objects present in prev but absent from params are stopped. Objects in params
 // but absent from prev are newly watched. Objects present in both are left untouched
 // (Google Calendar does not support mutating an existing channel).
-func (a *Adapter) UpdateSubscription(
+func (a *Adapter) UpdateSubscription( //nolint: cyclop,funlen
 	ctx context.Context,
 	params common.SubscribeParams,
 	previousResult *common.SubscriptionResult,
@@ -207,9 +208,7 @@ func (a *Adapter) UpdateSubscription(
 
 	// Start from a copy of the previous channels; we'll mutate it below.
 	updatedChannels := make(map[common.ObjectName]*WatchResponse, len(prevChannels))
-	for obj, ch := range prevChannels {
-		updatedChannels[obj] = ch
-	}
+	maps.Copy(updatedChannels, prevChannels)
 
 	// Stop removed channels.
 	var stopErr error
@@ -280,6 +279,7 @@ func (a *Adapter) DeleteSubscription(
 	result common.SubscriptionResult,
 ) error {
 	channels := extractChannels(&result)
+
 	return a.stopAllChannels(ctx, channels)
 }
 
@@ -388,7 +388,7 @@ func (a *Adapter) stopChannel(ctx context.Context, id, resourceID string) error 
 
 	// channels/stop returns 204 No Content on success with no body.
 	if response.Code != http.StatusNoContent && response.Code != http.StatusOK {
-		return fmt.Errorf("stopChannel: unexpected status %d", response.Code)
+		return fmt.Errorf("stopChannel: unexpected status %d", response.Code) //nolint: err113
 	}
 
 	return nil
--- a/providers/google/internal/calendar/subscribe.go
+++ b/providers/google/internal/calendar/subscribe.go
@@ -151,12 +151,14 @@ func (a *Adapter) Subscribe(
 	for _, obj := range objects {
 		resp, err := a.watchObject(ctx, obj, baseID, watchReq)
 		if err != nil {
-			rollbackErr := a.stopAllChannels(ctx, result.Channels)
-			if rollbackErr != nil {
-				return &common.SubscriptionResult{
-					Status: common.SubscriptionStatusFailedToRollback,
-					Result: result,
-				}, fmt.Errorf("subscribe: watching %q failed: %w; rollback also failed: %w", obj, err, rollbackErr)
+			if len(result.Channels) > 0 {
+				rollbackErr := a.stopAllChannels(ctx, result.Channels)
+				if rollbackErr != nil {
+					return &common.SubscriptionResult{
+						Status: common.SubscriptionStatusFailedToRollback,
+						Result: result,
+					}, fmt.Errorf("subscribe: watching %q failed: %w; rollback also failed: %w", obj, err, rollbackErr)
+				}
 			}
 
 			return &common.SubscriptionResult{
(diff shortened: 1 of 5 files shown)


providers/google/internal/calendar/subscribe.go
```diff
@@ -0,0 +1,469 @@
...
+
+	result := &CalendarSubscriptionResult{
+		Channels: make(map[common.ObjectName]*WatchResponse),
+	}
+
+	// Sort for deterministic processing order so rollback behaviour is predictable.
+	objects := sortedKeys(params.SubscriptionEvents)
+
+	for _, obj := range objects {
+		resp, err := a.watchObject(ctx, obj, baseID, watchReq)
```
jlimatampersand: you can use the `simultaneously` package to run these in goroutines.

providers/google/internal/calendar/subscribe.go
```diff
@@ -0,0 +1,469 @@
...
+		if err != nil {
+			if len(result.Channels) > 0 {
+				rollbackErr := a.stopAllChannels(ctx, result.Channels)
```
jlimatampersand: Rather than doing this within a loop, try to keep the records of the responses after exiting this loop. You can rollback if anything errors out.

providers/google/internal/calendar/subscribe.go
```diff
@@ -0,0 +1,469 @@
...
+	baseID := watchReq.ID
+	if baseID == "" {
+		baseID = uuid.New().String()
+	}
+
+	// justAdded tracks objects successfully watched in this call so they can be
+	// rolled back if a later watch in the same loop fails.
+	justAdded := make([]common.ObjectName, 0, len(toAdd))
+
+	for _, obj := range sortedKeys(toAdd) {
```
jlimatampersand: same here, rollback outside of the loop, it will be much cleaner.

providers/google/internal/calendar/subscribe.go
```diff
@@ -0,0 +1,469 @@
...
+		ObjectEvents: params.SubscriptionEvents,
+	}, nil
+}
+
+// UpdateSubscription reconciles an existing subscription with the new desired state.
+//
+// Objects present in prev but absent from params are stopped. Objects in params
+// but absent from prev are newly watched. Objects present in both are left untouched
+// (Google Calendar does not support mutating an existing channel).
+func (a *Adapter) UpdateSubscription( //nolint: cyclop,funlen
```
jlimatampersand: let's try a different logic here. 

So Google does not let you extend the expiration or renew for the calendar, so we just need to create a new one. 

Ideally, when we renew (with our own logic), we want to renew all the objects together. But if the existing ones are untouched, and we add new ones on top of those, then the update schedule clock will be consistent. So instead, we can just update all of them so the expiration can be in sync. 

So the logic can be, 

1. Create a new subscription for all of the objects we want to watch (new ones AND the existing ones)
   - If this fails, we just return an error. 
2. If creating new ones is successful. Then stop the ones we need to stop. 
   - If stopping fails, just warn and return. They will expire anyway, and we are okay with duplicates. 
   - If successful, we are good. 

This way, logic is going to be very simple.
```
func UpdateSubscription(....) (...){
   // some param prep


    var res *someResult
    var err *error
    // all objects, including new and existing
    if res, err = c.Subscribe(...); err != nil {
        return ..., fmt.Errorf("....", err) // nothing changes basically
    }

    // objects to remove: remove all old subscriptions, previously existing. 
    if err := c.DeleteSubscription(...); err != nil {
       logger.Get(ctx).Warn("failed to stop existing objects. But it will expire in OOO days")

    }

    return res, nil
   
}

File: providers/google/internal/calendar/subscribe.go [hunk 5]
```diff
@@ -0,0 +1,469 @@
...
+// Renewal requires stopping each channel and re-watching the same objects.
+// This is equivalent to DeleteSubscription followed by Subscribe with the same params.
+//
+// Callers should schedule maintenance before the earliest expiration across all channels.
+func (a *Adapter) RunScheduledMaintenance(
+	ctx context.Context,
+	params common.SubscribeParams,
+	previousResult *common.SubscriptionResult,
+) (*common.SubscriptionResult, error) {
+	if err := a.DeleteSubscription(ctx, *previousResult); err != nil {
```
jlimatampersand: if you take my suggestion in the updateSubscription, this will simplify to
```
return c.UpdateSubscription(...)
```

providers/google/internal/calendar/subscribe.go
```diff
@@ -0,0 +1,469 @@
...
+
+// VerifyWebhookMessage always returns true for Google Calendar.
+// Push notifications are delivered via Google's infrastructure and authenticated
+// at the transport layer; no application-level signature verification is needed.
+// Callers should verify the X-Goog-Channel-Token header against their stored token
+// if they set one in WatchRequest.Token.
+func (a *Adapter) VerifyWebhookMessage(
+	ctx context.Context,
+	request *common.WebhookRequest,
+	params *common.VerificationParams,
```
jlimatampersand: here in the `common.WebhookRequest`

Here you need to define this function like

```
type GoogleVerificationParam struct {
  XGooglChannelToken string
}

const tokenHeader = "X-Googl-Channel-Token"

func (a *Adapter) VerifyWebhookMessage(...)bool {
   tokenFromHeader := request.Headers.Get(tokenHeader)

   tokenFromParam, ok := params.Param.(GoogleVerificationParam)

   return tokenFromHeader == tokenFromParam
}

```

providers/google/internal/calendar/subscribe.go
```diff
@@ -0,0 +1,469 @@
...
+) (bool, error) {
+	return true, nil
+}
+
+// GetRecordsByIds is not implemented for Google Calendar.
+// Calendar push notifications deliver an empty body — only headers are sent (X-Goog-Resource-State,
+// X-Goog-Resource-ID, etc.). There is no record payload to enrich from; callers must issue
+// a follow-up API read (e.g. events.list) to fetch what actually changed.
+func (a *Adapter) GetRecordsByIds( //nolint:revive
```
jlimatampersand: @anushat05, I think this is something we need to apply the same strategy as gmail subscribe. Can you take the lead here to guide @Jkarage on how to approach the problem? 

In the meantime @Jkarage, can you split this PR into 2
- Subscribe, Update, Delete
- Verification. 

I can approve Subscribe, Update, Delete
Verification should be done differently and separately.

jlimatampersand (changes requested):
Good start here overall, but a few places to chaneg. 

1. Split the PR into 2
  - Subcribe, update, create
  - Verify

2. work on the comments

Jkarage:
@jlimatampersand I have updates this as per your suggestions.

jlimatampersand (approved):

PR #2988, opened 2026-05-12
[CON-3222] feat(GoTo): add Write

immdipu:
## Checklist

## Sample read response
```
    ~/Desktop/Ampersand/connectors    dipu/goto-write *12  go run ./test/goto/write                                                                                  ✔  system   08:58:18 PM  
time=2026-05-12T20:58:21.846+05:45 level=DEBUG msg="loading credentials file" path=./go-to-creds.json
Created meeting..
{
  "Data": {
    "conferenceCallInfo": "string",
    "joinURL": "https://meet.goto.com/285898365",
    "maxParticipants": 26,
    "meetingid": 285898365,
    "uniqueMeetingId": 285898365
  },
  "Errors": [],
  "RecordId": "",
  "Success": true
}
time=2026-05-12T20:58:23.874+05:45 level=DEBUG msg="loading credentials file" path=./go-to-creds.json
Created webhook..
{
  "Data": {
    "callbackUrl": "https://webhook.site/4930c4f8-c82c-4183-9fb0-bce3b7450cd1",
    "createTime": "2026-05-12T15:13:24.823Z",
    "eventName": "registrant.joined",
    "eventVersion": "1.0.0",
    "product": "g2w",
    "state": "INACTIVE",
    "webhookKey": "3ca2ed0b-591f-4852-ae28-dd790c9fe285"
  },
  "Errors": [],
  "RecordId": "",
  "Success": true
}
time=2026-05-12T20:58:25.021+05:45 level=DEBUG msg="loading credentials file" path=./go-to-creds.json
Created attribute..
{
  "Data": {},
  "Errors": [],
  "RecordId": "11kdy433hrsgm",
  "Success": true
}
Tests completed successfully

```


providers/goto/internal/gotocore/write.go
```diff
@@ -0,0 +1,210 @@
...
+
+func extractWriteRecordID(data map[string]any) string {
+	switch v := data["id"].(type) {
+	case string:
+		return v
+	case float64:
+		return fmt.Sprintf("%.0f", v)
+	}
+
+	return ""
```
RajatPawar, line 211: Same issue as in the other PR, ID may be empty

providers/goto/internal/gotocore/support.go
```diff
@@ -67,14 +73,19 @@ var objectRegistry = datautils.Map[string, objectConfig]{ //nolint:gochecknoglob
...
 	// SCIM API
-	"users":  {path: "identity/v1/Users", service: serviceSCIM},
-	"groups": {path: "identity/v1/Groups", service: serviceSCIM},
+	"users":  {path: "identity/v1/Users", service: serviceSCIM, writable: true},
+	"groups": {path: "identity/v1/Groups", service: serviceSCIM, writable: true},
+
+	// Only Write - these objects don't have a read endpoint,
+	//  but we want to be able to write to them via the generic write handler
+	"meetings": {path: "G2M/rest/meetings", service: serviceMeetings, writable: true},
+	"tickets":  {path: "G2M/rest/tickets", service: serviceRemoteSupport, writable: true},
```
RajatPawar: Will meetings get through the `Read` function?

providers/goto/internal/gotocore/write.go
```diff
@@ -0,0 +1,210 @@
...
+// endpoints return a plain object; a few (e.g. meetings create) return a
+// single-element array.
+func extractGenericWriteRecord(body *ajson.Node) (map[string]any, string, error) {
+	node := body
+	if body.IsArray() {
+		arr, err := body.GetArray()
+		if err != nil || len(arr) == 0 {
+			return nil, "", fmt.Errorf("empty array in write response: %w", err)
+		}
+
```
RajatPawar, line 189: would be good to mention the intention behind picking the first record here

RajatPawar (approved):
@immdipu approving this stack, but I did see only one active module, and the other one being unsupported. Are we going to follow up with an implementation for it?

immdipu:
> @immdipu approving this stack, but I did see only one active module, and the other one being unsupported. Are we going to follow up with an implementation for it?

We don’t have access to the sandbox for the other module. Once we gain access to the sandbox, we can implement it.

providers/goto/internal/gotocore/support.go
immdipu: Good catch. there's no explicit guard. meetings is in objectRegistry, so buildReadRequest would construct a request for it. The call would then fail at the API level (GoTo doesn't expose a list endpoint for `G2M/rest/meetings`), and `ListObjectMetadata` would fail too since it samples via the same path.

I'll add a readable flag.

providers/goto/internal/gotocore/utils.go
```diff
@@ -22,13 +22,6 @@ func (a *Adapter) buildObjectURL(objectName string) (*urlbuilder.URL, error) {
 		return nil, fmt.Errorf("error building URL for object %s: %w", objectName, err)
 	}
 
-	if spec.service == serviceAdmin {
-		// Admin API uses "pageSize" for pagination, while other services use "size".
-		url.WithQueryParam(queryParamPageSize, sampleSize)
-	} else {
-		url.WithQueryParam(queryParamSize, sampleSize)
-	}
-
```
immdipu, line 31: Moved pagination to the parent function so that this function can be used by other functions that don’t support pagination.

PR #2995, opened 2026-05-13
fix(hubspot): Add marketing support to ReadUsingSearchAPI

Cobalt0s:
# Live Tests

### Contacts
Since can be passed in various ways:
Peek 2026-05-14 20-40

###  Campaigns
Peek 2026-05-14 20-41

### DIFF
--- a/providers/hubspot/search.go
+++ b/providers/hubspot/search.go
@@ -34,7 +34,9 @@ func (c *Connector) Search(ctx context.Context, params *common.SearchParams) (*c
 // This endpoint paginates using paging.next.after which is to be used as an offset.
 // Archived results do not appear in search results.
 // Read more @ https://developers.hubspot.com/docs/api/crm/search
-func (c *Connector) ReadUsingSearchAPI(ctx context.Context, params SearchParams) (*common.ReadResult, error) {
+func (c *Connector) ReadUsingSearchAPI( // nolint:cyclop
+	ctx context.Context, params SearchParams,
+) (*common.ReadResult, error) {
 	ctx = logging.With(ctx, "connector", "hubspot")
 
 	if err := params.ValidateParams(); err != nil {
@@ -56,6 +58,7 @@ func (c *Connector) ReadUsingSearchAPI(ctx context.Context, params SearchParams)
 	//
 	// Read using POST endpoints intended for Search.
 	//
+
 	switch {
 	case core.CRMObjectsWithoutPropertiesAPISupport.Has(params.ObjectName):
 		// Objects outside ObjectAPI have different endpoint while both are part of CRM module.
@@ -64,14 +67,26 @@ func (c *Connector) ReadUsingSearchAPI(ctx context.Context, params SearchParams)
 			SearchParams: params,
 		})
 	case core.MarketingObjects.Has(params.ObjectName):
+		readParams, err := makeReadParamsFromSearchParams(params)
+		if err != nil {
+			return nil, err
+		}
 		// Search for marketing is not implemented, using simple read.
-		return c.readMarketing(ctx, params.ToReadParams(), core.MarketingObjects[params.ObjectName])
+		return c.readMarketing(ctx, *readParams, core.MarketingObjects[params.ObjectName])
 	case core.CommunicationObjects.Has(params.ObjectName):
+		readParams, err := makeReadParamsFromSearchParams(params)
+		if err != nil {
+			return nil, err
+		}
 		// Search for communication is not implemented, using simple read.
-		return c.readCommunications(ctx, params.ToReadParams(), core.CommunicationObjects[params.ObjectName])
+		return c.readCommunications(ctx, *readParams, core.CommunicationObjects[params.ObjectName])
 	case core.MiscellaneousObjects.Has(params.ObjectName):
+		readParams, err := makeReadParamsFromSearchParams(params)
+		if err != nil {
+			return nil, err
+		}
 		// Search for misc objects is not implemented, using simple read.
-		return c.readMiscAPI(ctx, params.ToReadParams(), core.MiscellaneousObjects[params.ObjectName])
+		return c.readMiscAPI(ctx, *readParams, core.MiscellaneousObjects[params.ObjectName])
 	default:
 		// Otherwise object belongs to Hubspot Objects API (sub-category of CRM namespace).
 		return c.searchCRMObjectsAPI(ctx, params)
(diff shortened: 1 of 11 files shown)


Cobalt0s:
* **#2995** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2995?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2995?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

providers/hubspot/search.go
```diff
@@ -64,14 +67,26 @@ func (c *Connector) ReadUsingSearchAPI(ctx context.Context, params SearchParams)
 			SearchParams: params,
 		})
 	case core.MarketingObjects.Has(params.ObjectName):
+		read, err := makeReadParamsFromSearch(params)
```
laurenzlong: call you name this `readParams`, not `read`?

providers/hubspot/search.go
```diff
@@ -64,14 +67,26 @@ func (c *Connector) ReadUsingSearchAPI(ctx context.Context, params SearchParams)
...
+		if err != nil {
+			return nil, err
+		}
 		// Search for marketing is not implemented, using simple read.
-		return c.readMarketing(ctx, params.ToReadParams(), core.MarketingObjects[params.ObjectName])
+		return c.readMarketing(ctx, *read, core.MarketingObjects[params.ObjectName])
 	case core.CommunicationObjects.Has(params.ObjectName):
+		read, err := makeReadParamsFromSearch(params)
```
laurenzlong: same here, call you name this `readParams`, not `read`?

providers/hubspot/search.go
```diff
@@ -209,6 +226,37 @@ func BuildIdFilterGroup(id string) Filter {
 	}
 }
 
+func makeReadParamsFromSearch(search SearchParams) (*common.ReadParams, error) {
```
laurenzlong: please rename to makeReadParamsFromSearchParams

providers/hubspot/types.go
```diff
@@ -96,6 +87,9 @@ type Filter struct {
 	FieldName string             `json:"propertyName,omitempty"`
 	Operator  FilterOperatorType `json:"operator,omitempty"`
 	Value     string             `json:"value,omitempty"`
+
+	isSince bool
```
laurenzlong: why do you have to modify `searchCRMParams`?

laurenzlong (changes requested):

providers/hubspot/types.go
Cobalt0s: `ReadUsingSearchAPI` is no longer operating on CRM objects (Objects API[`companies`, `contacts`] or Non-Objects API[`lists`]). This param even should be renamed to something general, so shouldn't be called `searchCRMParams`.

I added support to Marketing objects to `ReadUsingSearchAPI`.
These objects do not have search so we will have to fall back to read. To support incremental reading I would need to pass Since/Until. So they have to be recovered from FilterGroups which server constructs in the following way:
image

Essentially I would need to reverse engineer the operations of this method to recover Since/Until:
image

To simplify this I added the flags which act as markers.

laurenzlong: Ok got it! We need to actually do more careful reviews then, since this may affect existing CRM objects.

RajatPawar (changes requested):
Hey Constantin! 

* The `isSince`/`isUntil` flags - I'd really prefer we don't go this route. It feels quite brittle/invisible, and will add a quiet dependency to the entire hubspot chain. The marketing path would depend on the caller having used `BuildLastModifiedFilterGroup` / `BuildUntilTimestampFilterGroup` specifically, and anyone refactoring that could miss it very easily. Or if anyone constructs an equivalent filter by hand, we'd run a backfill. 

Given that this is high-pri, my preference would be to add `Since time.Time` and `Until time.Time` directly onto `SearchParams`. The marketing/comm/misc branches can read them directly from `SearchParams`. It needs a minor server change to the hubspot search reader. 

Longer term, the hubspot connector could expose a method that can tell us if an object supports search or not, and the server casts the connector & uses that method to route to the correct reader.

Also, can you please add a unit test in `search_test.go` for the new marketing branching logic? I don't think we have a test for `ReadUsingSearchAPI` for non-CRM objects. Basically, something that calls `ReadUsingSearchAPI` with `ObjectName: "marketing/emails"` with `Since` set

Thanks!

Cobalt0s:
> Hey Constantin!
> 
> * The `isSince`/`isUntil` flags - I'd really prefer we don't go this route. It feels quite brittle/invisible, and will add a quiet dependency to the entire hubspot chain. The marketing path would depend on the caller having used `BuildLastModifiedFilterGroup` / `BuildUntilTimestampFilterGroup` specifically, and anyone refactoring that could miss it very easily. Or if anyone constructs an equivalent filter by hand, we'd run a backfill.
> 
> Given that this is high-pri, my preference would be to add `Since time.Time` and `Until time.Time` directly onto `SearchParams`. The marketing/comm/misc branches can read them directly from `SearchParams`. It needs a minor server change to the hubspot search reader.
> 
> Longer term, the hubspot connector could expose a method that can tell us if an object supports search or not, and the server casts the connector & uses that method to route to the correct reader.
> 
> Also, can you please add a unit test in `search_test.go` for the new marketing branching logic? I don't think we have a test for `ReadUsingSearchAPI` for non-CRM objects. Basically, something that calls `ReadUsingSearchAPI` with `ObjectName: "marketing/emails"` with `Since` set
> 
> Thanks!

Makes sense.

Also for future, we should be taking into considiration `hubspot.ReadUsingSearchAPI` and `hubspot.Read` and `hubspot.Search`, where the last 2 are general contracts and `ReadUsingSearchAPI` is special case which existed before the `connectors.SearchConnector`.

For now `hubspot.Search` will return an object is not supported, because that is the case.
But for the `ReadUsingSearchAPI` everything is supported. Since/Until as you stated are the primary channels to pass that info. The only problem becomes if both Filter and Since/Until is set for the CRM objects, or if the Filter was not set -- in that case Filters are auto populated `params = params.applyTimestampFilters()`. This means that `BuildLastModifiedFilterGroup` and `BuildUntilTimestampFilterGroup` could be removed from server as long as Since/Until are passed.

Cobalt0s:
@RajatPawar the PR description was updated and it demonstrates what the changes do. Alternatively there are unit tests which you asked. For the contacts there 3 unit tests to demonstrate combinations of Since+Filter, Since, Filter.

PS: The change is backwards compatible and allows Since/Until for marketing/etc and acts smartly in case of CRM.

RajatPawar (approved):
Thanks Constantin, LGTM! And yes, search erroring out for non-CRM objects makes sense.

Can you please also make the corresponding server change? Should be a two-liner change. Once both are approved, I can coordinate the release and let you know when it's live.

laurenzlong (approved):

PR #2996, opened 2026-05-13
[ENG-3919] fix(salesforce): fix logic for flattening compound field names in subscription events

laurenzlong:
Please see Linear ticket for more details, including a link to a doc. Previous logic for handling compound fields in CDC events was wrong.

Added tests.


eberle1080 (approved):
Looks good, the logic is a bit unintuitive at first glance, but the comments explain it.

RajatPawar (approved):
I suspect that the compound <> flattening convention may not be constant enough for the way we are doing it right now, and it might be better to map these out explicitly

EDIT: Here's what a quick search highlighted. Not sure if these need handling -
- Opportunity.Fiscal 
- Location.Location 
- AlternativePaymentMethod.PaymentMethodAddress

laurenzlong:
> I suspect that the compound <> flattening convention may not be constant enough for the way we are doing it right now, and it might be better to map these out explicitly
> 
> EDIT: Here's what a quick search highlighted. Not sure if these need handling -
> 
> * Opportunity.Fiscal
> * Location.Location
> * AlternativePaymentMethod.PaymentMethodAddress

That's what I initially thought too, but according to both https://developer.salesforce.com/docs/atlas.en-us.change_data_capture.meta/change_data_capture/cdc_subscribe_compound_fields.htm and https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/compound_fields.htm, the only compound field types that exist in salesforce are:
- address fields
- name fields
- geolocation fields

This PR handles all 3, as explained in the comments above FlattenedFieldNameFromCompoundField.

The field you provided seem to be AI hallucinated fields that do not actually exist in Salesforce. And even if they did exist, they are still handled by the last line in FlattenedFieldNameFromCompoundField. (e.g. Opportunity.Fiscal -> Fiscal,  Location.Location -> Location, AlternativePaymentMethod.PaymentMethodAddress -> PaymentMethodAddress).

There are just way too many objects and fields inside of Salesforce for us to be able to do explicit mapping properly without missing some.

PR #2997, opened 2026-05-14
[Customer Support] Coverage test with real data

jlimatampersand:
Initially, our approach when testing Apex coverage was to use dummy data to cover the test. However, each different field has a different field integrity setting, e.g., eg) some fields are required and some are not, and we could not account for that fact with dummy data. 

This PR is updating the test script to use real existing data from the actual org's instance so that we don't run into failures related to dummy data. 

## Test Result
Screenshot 2026-05-14 at 12 24 09 PM
Screenshot 2026-05-14 at 12 28 37 PM

### DIFF
--- a/providers/salesforce/internal/crm/metadata/apextrigger.go
+++ b/providers/salesforce/internal/crm/metadata/apextrigger.go
@@ -278,15 +278,20 @@ func generateClassMetaXML() string {
 //     field on newRec only so (rec.<f> != oldRec.<f>) evaluates true and
 //     fieldChanged=true.
 //
-//   - `coverTriggerDelegation` does one Database.insert with allOrNone=false
-//     on a best-effort SObject. The trigger's before-insert fires before
-//     validation rules per Salesforce's order of execution, so even if the
-//     org rejects the insert, the trigger's single delegation line has
-//     already executed and counts toward coverage.
+//   - `coverTriggerDelegation` uses SeeAllData=true to find an existing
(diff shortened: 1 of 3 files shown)


providers/salesforce/internal/crm/metadata/apextrigger.go
```diff
@@ -355,16 +364,44 @@ private class %s {
         %s.process(new List<%s>{newRec}, new List<%s>{oldRec});
     }
 
-    @isTest
+    @isTest(SeeAllData=true)
```
jlimatampersand, line 367: this line allows us to use real data from org.

providers/salesforce/internal/crm/metadata/apextrigger.go
```diff
@@ -355,16 +364,44 @@ private class %s {
...
+        // step without being short-circuited by required-field validation or
+        // before-save flows that the generic makeRec() can't satisfy.
+        //
+        // All DML in test methods is rolled back automatically when the
+        // method exits, so this strategy doesn't modify the org's data —
+        // SystemModstamp, flow side effects on related records, and audit
+        // writes all roll back with the transaction. Outbound callouts /
+        // queued emails from those flows are discarded because the commit
+        // doesn't happen.
+        List<%s> existing = [SELECT Id FROM %s LIMIT 1];
```
jlimatampersand, line 381: Getting real data from the org to use for test.

laurenzlong (approved):

PR #3008, opened 2026-05-18
[CON-3211]feat(acculynx): acculynx-read 

ejoo:
# Description                                                                                                                                                                     
                                                               
  - [x] Unit tests cover read logic (placed in `/providers/acculynx`)                                                                                                               
  - [x] live tests (placed in `/test/acculynx`)                
                                                                                                                              
       
                                                                                                                                                                                    
  # Tests                                                                                                                                                                           
                                                                                                                                                                                    
  ## Unit Test                                                 

<img width="1083" height="659" alt="image" src="https://github.com/user-attachments/assets/e9c0653b-bf96-4d51-a978-3a4ba514187b" />

  ## Live Test                                                                                                                                                                      
                                                               
<img width="1381" height="888" alt="image" src="https://github.com/user-attachments/assets/4eb5890f-81bc-430b-a491-c91a198e56fc" />
<img width="1381" height="888" alt="image" src="https://github.com/user-attachments/assets/9dade030-3c82-44ab-b288-076944faedb0" />
<img width="1381" height="888" alt="image" src="https://github.com/user-attachments/assets/f924d7be-ee2d-496a-bca7-f2e462f1c37d" />
<img width="1381" height="888" alt="image" src="https://github.com/user-attachments/assets/3da31401-ae47-41ca-8fa4-1253738ce0e8" />
<img width="1381" height="888" alt="image" src="https://github.com/user-attachments/assets/59c209e1-956e-4300-8516-410c7c436325" />
<img width="1381" height="888" alt="image" src="https://github.com/user-attachments/assets/60de6034-3c54-4c88-b44b-c9917c442b5d" />

### DIFF
--- a/providers/acculynx/read.go
+++ b/providers/acculynx/read.go
@@ -1,0 +1,448 @@
+package acculynx
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+	"github.com/amp-labs/connectors/internal/datautils"
+	"github.com/amp-labs/connectors/internal/jsonquery"
+	"github.com/amp-labs/connectors/internal/simultaneously"
+	"github.com/amp-labs/connectors/providers/acculynx/metadata"
+	"github.com/spyzhov/ajson"
+)
+
+const (
+	// AccuLynx OpenAPI does not document maximum pageSize, but its API enforces
+	// server-side per-object caps: /jobs and /supplements reject pageSize > 25,
+	// other list endpoints reject > 50. 25 is the strictest cap and is safe
+	// across every object. Well within AccuLynx's 10 req/sec per-key limit.
+	defaultPageSize = "25"
+	maxPageSize     = 25
+
+	// apiVersionPrefix is prepended to every schema path. The catalog BaseURL is
+	// the smallest URL ("https://api.acculynx.com") so proxy builders can pin
+	// their own version; the deep connector targets V2 explicitly.
+	apiVersionPrefix = "/api/v2"
+
+	pageSizeParam    = "pageSize"
+	recordStartParam = "recordStartIndex"
+	pageStartParam   = "pageStartIndex"
+	pageNumberParam  = "pageNumber"
+
+	// AccuLynx 10 req/sec per API key gives plenty of headroom; 4 is a
+	// conservative cap for the per-parent fan-out.
+	maxConcurrentChildFetch = 4
+
+	// /calendars/{calendarId}/appointments requires startDate/endDate. When the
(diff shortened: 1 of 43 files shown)


providers/acculynx/read.go
```diff
@@ -0,0 +1,448 @@
...
+	// server-side per-object caps: /jobs and /supplements reject pageSize > 25,
+	// other list endpoints reject > 50. 25 is the strictest cap and is safe
+	// across every object. Well within AccuLynx's 10 req/sec per-key limit.
+	defaultPageSize = "25"
+	maxPageSize     = 25
+
+	// apiVersionPrefix is prepended to every schema path. The catalog BaseURL is
+	// the smallest URL ("https://api.acculynx.com") so proxy builders can pin
+	// their own version; the deep connector targets V2 explicitly.
+	apiVersionPrefix = "/api/v2"
```
anushat05: The /api/v2 is necessary (BaseURL and schema paths are both version-less), but hardcoding it as a const and concatenating it at each build site isn't the convention. 

The repo injects API/version prefixes via the module path in `schemas.json` e.g. `okta` /api/v1, `zendesksupport` /api/v2, `gong` /v2 — after which LookupURLPath returns the full versioned path automatically.

If the root module path were /api/v2, we could drop `apiVersionPrefix` entirely and buildInitialURL/fetchChildPage simplify to `urlbuilder.New(BaseURL, path)`. The catalog BaseURL still stays version-less, so the proxy-flexibility rationale in the comment still holds. 

It also fixes a subtle mismatch: buildReadRequest validates against the unversioned LookupURLPath but builds the request with the prefix manually added.

providers/acculynx/read.go
```diff
@@ -0,0 +1,448 @@
...
+
+	if params.ObjectName == "calendars/appointments" {
+		applyAppointmentsDateWindow(url, params)
+	}
+
+	resp, err := c.JSONHTTPClient().Get(ctx, url.String())
+	if err != nil {
+		return nil, err
+	}
+
```
anushat05, line 414: This fetches one page per parent and discards `result.NextPage` no paging loop and only sends pageSize=25 + a page param (jobs/invoices, jobs/estimates, jobs/history, etc. have real paging styles). So a job with >25 invoices silently drops the rest. Either mark these `paginationNone` if they truly return everything in one response (and fix the comment), or loop on NextPage until done. 

Untested today since all fan-out fixtures are single short pages.

anushat05 (changes requested):

providers/acculynx/read.go
ejoo: Done. `/api/v2` moved into module root.path, apiVersionPrefix dropped, generator updated. Verified with live read.

providers/acculynx/read.go
ejoo: Fixed, fetchChildPages now loops on NextPage + server-side `startDate/endDate` for `jobs/history`. Tests added. Thanks!

PR #3011, opened 2026-05-19
fix(livestorm): Remove session chat object

Fullstack-Hero:
## Summary
Remove session_chat_messages from the Livestorm deep connector. That object requires a session id in ReadParams.Filter, which scheduled reads and manifests cannot supply, so it is not a valid Read object for this connector.

Changes drop read routing, metadata, tests, and integration helpers for session_chat_messages, along with ErrSessionIDRequired. Read support for collection objects (events, people, people_attributes) and write support (events, users) are unchanged.


providers/livestorm/read.go
```diff
@@ -19,17 +19,17 @@ const (
...
+	objectJobs   = "jobs"
 )
 
 // nolint:gochecknoglobals
 var readSupportedObjects = datautils.NewStringSet(
 	"events",
 	"people",
 	"people_attributes",
-	"session_chat_messages",
+	"jobs",
```
Cobalt0s: Dont add jobs.

The endpoint you mention requires the ID: https://developers.livestorm.co/reference/get_jobs-id
There is the list/collection endpoint: https://developers.livestorm.co/reference/get_jobs-id-tasks but thats for "tasks" and also needs ID.

So there is no true "jobs" collection. The system may have many jobs but they are not requestable.

Sorry about the confusion with the `common.ReadParams.Filter`. At the moment there is no way to specify ids in `amp.yaml`. Also `Filter` has a very unique use cases.

Cobalt0s (approved):
Don't add `objectJobs`. Once you remove it, feel free to merge the PR.
The PR should be only about removing `session chat object`.

PR #3012, opened 2026-05-19
[CON-3211] feat(acculynx): add write connector

ejoo:
# Description

  - [x] Write payloads should accept what `ReadResults.Fields` is returning.
  - [x] Unit tests cover write & delete logic (placed in `/providers/acculynx`)
  - [x] live tests (placed in `/test/acculynx/write` and `/test/acculynx/delete`)

  # Tests

  ## Unit Test

### Write

image

### Delete

image

  ## Live Test — Write

image
image
image

  ## Live Test — Delete

image

### DIFF
--- a/providers/acculynx/delete.go
+++ b/providers/acculynx/delete.go
@@ -43,10 +43,10 @@ func (c *Connector) buildDeleteURL(params common.DeleteParams) (*urlbuilder.URL,
 
 	switch params.ObjectName {
 	case deletableJobsARRepresentative:
-		return urlbuilder.New(baseURL, apiVersionPrefix,
+		return urlbuilder.New(baseURL, c.modulePath(),
 			"jobs", params.RecordId, "representatives", "ar-owner")
 	case deletableJobsSalesRepresentative:
-		return urlbuilder.New(baseURL, apiVersionPrefix,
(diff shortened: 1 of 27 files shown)


anushat05 (approved):

providers/acculynx/delete.go
```diff
@@ -0,0 +1,69 @@
...
+
+// buildDeleteURL maps a deletable ObjectName + RecordId (= jobId) to its
+// fully-qualified AccuLynx URL. Returns ErrOperationNotSupportedForObject for
+// any object outside the small set AccuLynx allows deletes on.
+func (c *Connector) buildDeleteURL(params common.DeleteParams) (*urlbuilder.URL, error) {
+	baseURL := c.ProviderInfo().BaseURL
+
+	switch params.ObjectName {
+	case deletableJobsARRepresentative:
+		return urlbuilder.New(baseURL, apiVersionPrefix,
```
anushat05: You need to update delete for the apiVersionPrefix change as per comment on this PR https://github.com/amp-labs/connectors/pull/3008#discussion_r3325713878

anushat05 (changes requested):

providers/acculynx/delete.go
ejoo: Done! 🚀

PR #3021, opened 2026-05-21
[ENG-3949] feat(hubspot): Campaign assets

Cobalt0s:
# Description

* Add `readhelper.HydrateAssociations`, which initializes and populates `[]common.ReadResultRow`.
* HubSpot batch package now supports reading. The payloads are automatically grouped into the chunks.
* Marketing campaigns are read using the batch endpoint only when asset association is requested. Unlike the usual list marketing campaigns endpoint, this batch endpoint returns assets in the response.

# Live Tests

## Read 
`marketing-campaigns` with `assets` associations.

image

PR #3020, opened 2026-05-21
[ENG-3949] feat(hubspot): Marketing object renaming

Cobalt0s:
# Description

Renamed objects. Final version of the names are as follows:
* marketing-campaigns
* marketing-emails
* marketing-forms

# Live Tests

## ListObjectMetadata
Peek 2026-05-21 17-31

### Read
Peek 2026-05-21 17-36

PR #3022, opened 2026-05-21
[ENG-3949] feat(hubspot): Campaign associations to contacts

Cobalt0s:
# Description

* Add support for additional associations from `marketing-campaigns` to `contacts`.
image

## Procedure:

* Fetch contact identifdiers using `simultaneously` for each campaing id + contactType:
  * There are 3 contact types "contactFirstTouch", "contactLastTouch", "influencedContacts". The number of callabacks is therfore tripled for each campaign.
image

* Registries for the bookkeeping purposes. This is preparation for the next requests and so that we can wire everything back to produce `common.Association`
image

* Fetch all contacts
image

* Finally, each contact is linked back to the original marketing-campaign and metadata includes the contactType.
image

# Live Tests

## Read
`marketing-campaigns` with `contacts` associations. 

image

image

# Unit Tests

Connector does the following API calls:
image

Connector output:
image


Cobalt0s:
* **#3023** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3023?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3022** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3022?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3022?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3021** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3021?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3020** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3020?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

providers/hubspot/read_test.go
```diff
@@ -514,6 +516,107 @@ func TestRead(t *testing.T) { //nolint:funlen,gocognit,cyclop,maintidx
...
+									"email":            "bh@hubspot.com",
+									"firstname":        "Brian",
+									"hs_object_id":     "110338039632",
+									"lastmodifieddate": "2026-05-21T22:45:00.229Z",
+									"lastname":         "Halligan (Sample Contact)",
+									"createdAt":        "2026-05-21T22:43:55.385Z",
+									"updatedAt":        "2026-05-21T22:45:00.229Z",
+									"archived":         false,
+									"url":              "https://app-eu1.hubspot.com/contacts/148543157/record/0-1/110338039632",
+									"properties": map[string]any{
```
Cobalt0s, line 586: Properties is flattened here and also included in the final response just in case.

It seems that we need to turn associations param from `[]string` into `[]AssociationParam`. AssociationParam would have: 
* Name
* Fields -- then we will know what to extract. The same ideas should apply to associations. Otherwise, idk what to include.

Let me know if the "true" raw response should be returned.

RajatPawar: That makes sense. Field filtering for associations already happens server-side so the connector returning a verbose `Raw` is probably fine. Might be a good to have, but for now, I'd decouple it from this and return the true raw response.

RajatPawar (approved):
LGTM. Can we rename the `type` in ProviderAssociationMetadata to `associationType`? Sorry, I know that I proposed it, but now that I think about it, `type` might be too broad and might conflict in the future

providers/hubspot/read_test.go
Cobalt0s: Ok. then I will remove the flattening logic. The associated object will be the rawest of the raw.

Cobalt0s:
> LGTM. Can we rename the `type` in ProviderAssociationMetadata to `associationType`? Sorry, I know that I proposed it, but now that I think about it, `type` might be too broad and might conflict in the future

To match this idea, in the text PR instead of just calling a metadata as `type` I call it `assetType`, that is the name used in Hubspot API docs so that matches and is more descriptive than just the `type`.

PR #3026, opened 2026-05-22
fix: classify Google rate-limit responses as retryable (Gmail/Calendar/Contacts)

RajatPawar:
## Summary
- Adds `providers/google/internal/core.InterpretJSONError` that detects Google's seven rate-limit/quota reason strings (UPPER_SNAKE in `details[].reason`, camelCase in `errors[].reason`) on both 403 and 429 responses, returning `common.ErrLimitExceeded` (retryable) instead of letting the default status mapper fall through to `ErrForbidden` (unrecoverable).
- Wires the helper into Gmail (mail), Calendar, and Contacts adapters via `DirectFaultyResponder`, matching the existing HTML override pattern. Contacts' 409 → `ErrConflict` mapping is preserved through the fallback path.

## Context
Discovered while triaging an issue. The underlying HTTP response was still classified as `ErrForbidden` by the connector and killed the read activity before the limiter's backoff could matter. This change closes that gap at the connector layer.

### DIFF
--- a/providers/google/internal/calendar/adapter.go
+++ b/providers/google/internal/calendar/adapter.go
@@ -35,7 +35,7 @@ func constructor(base *components.Connector) (*Adapter, error) {
 	}
 
 	errorHandler := interpreter.ErrorHandler{
-		JSON: interpreter.NewFaultyResponder(errorFormats, nil),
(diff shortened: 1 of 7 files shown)


AmandaYao00 (approved):
very cool

PR #3028, opened 2026-05-25
fix(tests): Remove reference http.DefaultClient

Cobalt0s:
# Description

Tests should not use `http.DefaultClient` because its internal transport behaves like a singleton. In the past, closing the client could cause intermittent errors in other tests that relied on it.

## Changes

This change ensures that a new client instance is created for each test via `mockutils.NewClient()` (however, only one such case).

For most other cases, we reuse the client provided by `httptest.Server`. Since the mock server is created per unit test, and each test initializes a new connector instance, there is no risk in reusing a non-singleton client. When the test completes, the server is closed, and the associated client is cleaned up through the resulting teardown chain and connector instance is discarded.

# Note

The other files may still use `http.DefaultClient` and those are not changed. The change is about tests that run in parallel which can cause side effects via the singelton client transport instance.

### DIFF
--- a/common/http_test.go
+++ b/common/http_test.go
@@ -618,9 +618,7 @@ func TestHTTPClient_Get_Success(t *testing.T) {
 
 	client := &HTTPClient{
 		Client: &mockAuthClient{
-			doFunc: func(req *http.Request) (*http.Response, error) {
-				return http.DefaultClient.Do(req)
-			},
+			doFunc: server.Client().Do,
 		},
 	}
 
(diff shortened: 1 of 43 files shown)


Cobalt0s:
* **#3028** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3028?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3028?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

eberle1080 (approved):

PR #3032, opened 2026-05-27
[CON-3247] feat(acculynx): add subscribe connector

ejoo:
Adds the Subscribe action for the AccuLynx connector, full webhook lifecycle (Subscribe / Update / Delete), event parsing, and `GetRecordsByIds` for contacts and jobs.          
                                                                                                                                                                                    
  ## Tests                                                                                                                                                                          
                                                                                                                                                                                    
  ### Unit    
                                                   
<img width="955" height="849" alt="image" src="https://github.com/user-attachments/assets/8e3c7648-e14c-4766-a255-004d201f22af" />

  ### Live
  
<img width="1468" height="849" alt="image" src="https://github.com/user-attachments/assets/ee5d1eaa-e7be-46fc-bb81-115415ca1064" />
<img width="1468" height="849" alt="image" src="https://github.com/user-attachments/assets/5598dcca-291c-4a96-8415-141024829850" />



<img width="1705" height="957" alt="image" src="https://github.com/user-attachments/assets/7af540f4-491c-4bf9-bae9-be3613e8fd16" />
<img width="1705" height="957" alt="image" src="https://github.com/user-attachments/assets/450e5561-eb9d-4941-aeda-818a48d81c1a" />

### DIFF
--- a/providers/acculynx/subscribe.go
+++ b/providers/acculynx/subscribe.go
@@ -1,0 +1,391 @@
+package acculynx
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"slices"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+	"github.com/amp-labs/connectors/internal/datautils"
+	"github.com/go-playground/validator"
+)
+
+// AccuLynx exposes a single subscription per installation. UpdateSubscription
+// replaces the topic list in-place; consumerUrl is immutable after creation.
+// Only contact_added/changed and job_created/updated are first-class; every
+// other topic is reachable via ObjectEvents.PassThroughEvents.
+//
+// Reference: https://apidocs.acculynx.com/reference/postsubscription
+
+const (
+	webhooksVersionPrefix = "/webhooks/v2"
+	subscriptionsPath     = "subscriptions"
+)
+
+var (
+	errMissingSubscribeParams      = errors.New("acculynx: missing subscribe params")
+	errInvalidSubscribeRequestType = errors.New("acculynx: invalid subscribe request type")
+	errInvalidSubscriptionResult   = errors.New("acculynx: invalid subscription result type")
+	errNoTopicsResolved            = errors.New("acculynx: no AccuLynx topics resolved from requested events")
+	errUnsupportedSubscribeObject  = errors.New("acculynx: object does not support subscriptions")
+	errUnsupportedSubscribeEvent   = errors.New("acculynx: event type not supported by AccuLynx for this object")
+	errUnknownPassThroughTopic     = errors.New("acculynx: passThrough event is not a recognized AccuLynx topic")
+	errEmptySubscriptionID         = errors.New("acculynx: empty subscriptionId in create response")
+)
+
+// SubscriptionRequest is the provider-specific input the framework passes via
+// common.SubscribeParams.Request.
+type SubscriptionRequest struct {
+	// ConsumerURL is the HTTPS endpoint AccuLynx will POST events to. Immutable
+	// after subscription creation.
+	ConsumerURL string `json:"consumerUrl" validate:"required,url"`
+	// TechContact is the email AccuLynx contacts about subscription status.
+	TechContact string `json:"techContact" validate:"required,email"`
+}
+
+// SubscriptionResult is the persistent connector state stored between Subscribe
+// calls. The framework keeps this verbatim and passes it back on UpdateSubscription
+// / DeleteSubscription so we can identify the existing AccuLynx subscription.
+type SubscriptionResult struct {
+	SubscriptionID string   `json:"subscriptionId"`
+	ConsumerURL    string   `json:"consumerUrl"`
+	TechContact    string   `json:"techContact"`
+	TopicNames     []string `json:"topicNames"`
+	Status         string   `json:"status"`
+}
+
+// subscriptionCreateRequest mirrors the POST /webhooks/v2/subscriptions body.
+type subscriptionCreateRequest struct {
+	ConsumerURL     string   `json:"consumerUrl"`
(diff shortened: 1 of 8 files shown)


providers/acculynx/subscriptionEvent.go
```diff
@@ -0,0 +1,254 @@
...
+// ObjectName derives the target object from the topic prefix
+// (contact* → contacts, job* → jobs).
+func (e SubscriptionEvent) ObjectName() (string, error) {
+	topic, err := e.RawEventName()
+	if err != nil {
+		return "", err
+	}
+
+	switch {
+	case strings.HasPrefix(topic, "contact_"), strings.HasPrefix(topic, "contact."):
```
jlimatampersand: can't this be just strings.HasPrefix(topic, "contact")? rather than listing out? same for below?

providers/acculynx/subscriptionEvent.go
```diff
@@ -0,0 +1,254 @@
...
+	wrapper, ok := inner[key].(map[string]any)
+	if !ok {
+		return nil, fmt.Errorf("%w: expected event.%s", errMissingObjectWrapper, key)
+	}
+
+	return wrapper, nil
+}
+
+// VerifyWebhookMessage always returns true: AccuLynx does not sign webhooks
+// and exposes no verification mechanism on the wire.
```
jlimatampersand: I looked into this doc, https://apidocs.acculynx.com/docs/webhooks-listener-documentation#proposed-solution

It mentions that endpoint will return secrets. But I wasn't sure if it was actually possible.

There is testing screenshot https://apidocs.acculynx.com/docs/guidelines#testing-webhooks

If you look into the header, there is webhook secret too. So I am a little confused. Can you double check this?

providers/acculynx/subscriptionEvent.go
```diff
@@ -0,0 +1,254 @@
...
+	"errors"
+	"fmt"
+	"maps"
+	"strings"
+	"time"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/internal/datautils"
+)
+
```
jlimatampersand, line 14: Please add a sample webhook event here for better visualization.

providers/acculynx/subscribe.go
```diff
@@ -0,0 +1,391 @@
...
+
+func (c *Connector) updateSubscription(
+	ctx context.Context, subscriptionID string, payload subscriptionUpdateRequest,
+) error {
+	u, err := c.subscriptionByIDURL(subscriptionID)
+	if err != nil {
+		return err
+	}
+
+	_, err = c.JSONHTTPClient().Put(ctx, u.String(), payload)
```
jlimatampersand, line 270: does the api return the updated body? if so, we need to pase the body and return as part of subscription result. Please don't ignore this.

providers/acculynx/subscribe.go
```diff
@@ -0,0 +1,391 @@
...
+	if parsed == nil || parsed.SubscriptionID == "" {
+		return nil, errEmptySubscriptionID
+	}
+
+	return parsed, nil
+}
+
+func (c *Connector) updateSubscription(
+	ctx context.Context, subscriptionID string, payload subscriptionUpdateRequest,
+) error {
```
jlimatampersand, line 264: ideally, this will return updated result from the API

providers/acculynx/subscribe.go
jlimatampersand: nvm it seems like they return 204.

jlimatampersand (changes requested):
Looks good mostly. But can you pleaes add live test that you actually receive a webhook from Acculynx by triggering some change?

you can use https://www.svix.com/play/ and test your webhook message. 

I think they also provide test API

https://apidocs.acculynx.com/reference/postsendtestevent

ejoo:
> Looks good mostly. But can you pleaes add live test that you actually receive a webhook from Acculynx by triggering some change?
> 
> you can use https://www.svix.com/play/ and test your webhook message.
> 
> I think they also provide test API
> 
> https://apidocs.acculynx.com/reference/postsendtestevent

providers/acculynx/subscriptionEvent.go
ejoo: Good call, simplified. Thanks!

providers/acculynx/subscriptionEvent.go
ejoo: Good catch I dug into this carefully:                                                                                                                                         

> POST /webhooks/v2/subscriptions response (verified live): only {"subscriptionId": "..."}, no secret field. OpenAPI spec confirms subscriptionPostResponseOk defines           
 subscriptionId only.  
                                                 
> GET /webhooks/v2/subscriptions/{id} returns 8 fields  no secret.
                                                                                                             
> Webhook deliveries I just triggered (and ~22 captured during dev) contain only CloudFront/Datadog/Zipkin headers no x-acculynx-webhooks-secret. 
                            
> The 2023 screenshot in the docs is from a legacy implementation its payload uses Type/ResourceKind/ResourceInstance, but the live system uses topicName/event.<object>.id (different shape entirely).

providers/acculynx/subscriptionEvent.go
ejoo: Done!

providers/acculynx/subscribe.go
ejoo: Confirmed, PUT returns 200 with empty body. Current code builds the result from the request payload since AccuLynx doesn't echo it back.

providers/acculynx/subscribe.go
ejoo: Verified! PUT returns 200 + empty body. Added a comment to updateSubscription explaining why it returns only error.

jlimatampersand (approved):

PR #3036, opened 2026-05-28
fix(monday): mark write support enabled

### LINKED ISSUE #3003: Monday: catalog support.write is false but Write API succeeds end-to-end for boards
> **Scope narrowed:** This issue originally covered both `aws` and `monday`. It is now Monday-only. AWS remains untested at runtime and is not in scope here; see the comment below for context.

## Summary

Live provider metadata for `monday` reports `support.write: false`, but a Write Actions API call to create a Monday board actually succeeded end-to-end. The catalog flag appears to be under-reporting at least the tested capability.

## Runtime evidence (Monday, `boards` create)

Tested against production via the dedicated Write Actions API surface (not Proxy).

```
POST https://write.withampersand.com/v1/projects/<project>/integrations/<integration>/objects/boards
Content-Type: application/json
X-Api-Key: <redacted>

{
  "type": "create",
  "installationId": "<redacted>",
  "groupRef": "<redacted>",
  "record": {
    "name": "drift-smoke-test-board",
    "board_kind": "private"
  }
}
```

Response: **HTTP 200**

```json
{
  "operationId": "...",
  "result": {
    "success": true,
    "data": {
      "data": {"create_board": {"id": "...", "name": "drift-smoke-test-board"}}
    }
  }
}
```

A real board was created in the test Monday workspace and was deleted immediately after the test.

## Source state confirmed earlier

`providers/monday/utils.go` registers `components.WriteSupport` for `boards`, `items`, and `users`. `handlers.go` has `buildWriteRequest` and a `create_board` GraphQL mutation. `handlers.go:32` explicitly defines `ErrWriteUserNotSupported` indicating that write for `users` is intentionally not supported, but `boards` clearly is.

`providers/monday.go` catalog declaration:

```go
Support: Support{
    Proxy:     true,
    Read:      true,
    Subscribe: false,
    Write:     false,
},
```

## Ask

Given the Write Actions API actually performs a write for Monday `boards`, should `providers/monday.go` be updated to `Write: true`?

This is scoped narrowly:
- Tested: `boards` create only.
- Not tested: update / delete / bulk / batch / other Monday write objects.
- Out of scope: AWS write support (no runtime evidence collected).

## What this issue is **not**

- Not a runtime bug report. The system works; the metadata may not reflect what works.
- Not an AWS issue. AWS had similar source-vs-catalog appearance, but no runtime verification was attempted because that would require AWS Identity Center test credentials. AWS should be revisited separately if and when that evidence exists.

## References

- Docs detector: [amp-labs/docs#608](https://github.com/amp-labs/docs/pull/608)
- Docs triage: [amp-labs/docs#609](https://github.com/amp-labs/docs/issues/609)

caiopizzol:
## Summary

Marks Monday as supporting Write Actions in the provider catalog metadata.

## Why

`providers/monday` already registers write support for Monday objects, and runtime validation in amp-labs/connectors#3003 confirmed that creating a Monday board through the Write Actions API succeeds end-to-end. The generated/live catalog currently reports `support.write: false`, which causes docs drift and under-reports the capability.

## Scope

This is intentionally narrow:

- changes only `providers/monday.go`
- does not change AWS or any other provider
- runtime evidence covers Monday `boards` create only

Fixes #3003.

## Validation

- Reviewed connector source: `providers/monday/utils.go` registers `components.WriteSupport` for Monday objects.
- Runtime evidence is documented in #3003: `POST /objects/boards` returned HTTP 200 and created a real Monday board, which was cleaned up after the test.
- Local Go tests not run: `go` is not installed in this local environment.

### DIFF
--- a/providers/monday.go
+++ b/providers/monday.go
@@ -35,7 +35,7 @@ func init() {
 			Proxy:     true,
 			Read:      true,
 			Subscribe: false,
-			Write:     false,
+			Write:     true,
 		},
 	})
 }


caiopizzol:
@laurenzlong @eberle1080 lmk if that makes sense

caiopizzol:
I re-ran this through the harness and now have runtime evidence for the Monday write claim.

What passed:
- Created a Monday board through Ampersand Write Actions.
- Confirmed the board existed in Monday by reading it back by ID.
- Confirmed the Ampersand operation completed as a successful write.
- Deleted the test board after the run.
- Confirmed no `ampersand-*` test boards were left behind.

One caveat: the write succeeds, but the response does not fill in `result.recordId`. The board ID is present deeper in the raw response. That looks like a small connector response parsing bug and should be fixed or tracked separately, but it does not change the conclusion that Monday boards write works.

laurenzlong (approved):

PR #3046, opened 2026-06-02
[ENG-3917]feat: add support PKCE for salesforce

anushat05:
## Notes

Flip `Salesforce.Oauth2Opts.GrantType` from `AuthorizationCode` → `AuthorizationCodePKCE` so our Salesforce OAuth flow always sends `code_challenge`.

## Testing

  End-to-end via local server against a real SF External Client App:

  | Pass | App "Require PKCE" | Connection saved |
  |---|---|---|
  | A | ❌ unchecked | ✅ `auth_scheme=oauth2/authorizationCodePKCE`, both tokens |
  | B | ✅ checked    | ✅ same |

Screenshot 2026-06-03 at 13 54 50

Screenshot 2026-06-03 at 13 57 53


jlimatampersand (approved):

anushat05:
Tested end-to-end against a real Salesforce dev org through the full local stack (`api` + `temporal` + `token-manager`) running this branch's catalog, with `PROVIDER_CATALOG_URL` pointed at the branch so the runtime auth-URL fetch and the actual catalog agree. Used **two Salesforce users** in the dev org (independent token grants) and the toggleable `test_anusha` External Client App.

1. **Baseline (pre-change):** created a connection on `main` (non-PKCE) → tagged `oauth2/authorizationCode`, reading `contact` end-to-end to a webhook destination.
2. **Flip is live:** after applying the change, a fresh authorize URL includes `code_challenge` + `code_challenge_method=S256` (and still carries SF's `prompt=login`).
3. **Existing connection survives:** with the ECA's **Require PKCE + Refresh Token Rotation** enabled, the pre-existing `authorizationCode` connection still force-refreshes (new access token, SF introspect `active:true`) and reads
   successfully — proving the flip doesn't break already-connected customers even when they turn PKCE on.
4. **New connection (happy path):** a new connect under PKCE enforcement succeeds, is tagged `oauth2/authorizationCodePKCE`, and reads end-to-end.
5. **Parallel:** an `authorizationCode` and an `authorizationCodePKCE` connection read concurrently (2 records each, ~9ms apart) with identical behavior.
6. **Reconnect/migration:** re-authing the existing connection via `:oauth-update` succeeds under enforcement and reads fine. Note: the reconnect path preserves the original `auth_scheme` label (stays `authorizationCode`) — cosmetic, zero runtime impact (every reader treats the two schemes identically). Filed as a low-severity note.
Screenshot 2026-07-20 at 12 02 12
Screenshot 2026-07-20 at 12 02 03
Screenshot 2026-07-20 at 12 01 42

PR #3050, opened 2026-06-04
[CON-3293] feat(hubspot): document referencedObjectTypeMap and add USER mapping follow-up

ejoo:
## Summary
Follow-up to review feedback on [#3048](https://github.com/amp-labs/connectors/pull/3048).

- Document where `referencedObjectTypeMap` came from (composed by hand; HubSpot does not publish a closed enum for this field).
- Add explicit `"USER" → "users"` mapping. Without it the lowercase fallback would yield `"user"` (singular) instead of our `"users"` object name.

## Tests
- New unit test: `Property with referencedObjectType=USER becomes a reference to users`
- Full hubspot suite passes
- `make lint` clean

## Screenshot

image

### DIFF
--- a/providers/hubspot/schema.go
+++ b/providers/hubspot/schema.go
@@ -35,6 +35,16 @@ var KnownObjectTypes = map[string]string{ // nolint:gochecknoglobals
 //
 // Used by ListObjectMetadata when populating common.FieldMetadata.ReferenceTo
 // for reference-typed fields.
+//
+// Provenance: HubSpot does not publish a closed enum of valid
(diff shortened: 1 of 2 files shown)


providers/hubspot/schema.go
laurenzlong: can we actually attach a `s` at the end? As an extremely simplistic way to pluralize the word?

laurenzlong (approved):

providers/hubspot/schema.go
ejoo: Done, appended an "s" to match the plural convention.

PR #2775, opened 2026-03-19
[CON-3029] feat(connectWise): Write

Cobalt0s:
# Description

Implementations of Write and Delete.
Write chooses between using PATCH or PUT depending on the payload.

* PATCH:
image

* PUT:
image

Changed some leftover occurences of connectWise to connectwise.

# Live Tests

Create/Update/Delete contacts.

Peek 2026-06-08 20-20

PR #3053, opened 2026-06-05
feat(utils): Template for parallel fetch

Cobalt0s:
# Description

Added package parallelfetch with `Execute` to run batches of tasks concurrently using the existing `simultaneously` runner. Instead of duplicating previous code, this change centralizes parallel task execution:
* Introduces `Execute(ctx, tasks)` which runs a list of Task functions in parallel via `simultaneously.DoCtx`.
* Collects successful results into `Result.Records` keyed by task ID.
* Collects failures into `Result.Errors` keyed by task ID.
* Execute itself never returns an error; all outcomes are returned in the Result value.
* Uses buffered channels and safe channel closing to aggregate results after all callbacks complete.

### DIFF
--- a/internal/parallelfetch/fetch.go
+++ b/internal/parallelfetch/fetch.go
@@ -1,0 +1,130 @@
+package parallelfetch
+
+import (
+	"context"
+
+	"github.com/amp-labs/connectors/internal/datautils"
+	"github.com/amp-labs/connectors/internal/simultaneously"
+)
+
+// Result holds the outcome of executing multiple generic tasks, containing
+// successful records and any errors that occurred, keyed by their task ID.
+//
+// The type parameters are:
+//   - ID: The task ID type, must be comparable (usable as map key)
+//   - DATA: The data type returned by successful tasks.
+type Result[ID comparable, DATA any] struct {
+	// Records contains successfully completed task data, mapped by task ID.
+	Records datautils.Map[ID, DATA]
+	// Errors contains any errors that occurred during task execution,
+	// mapped by the task ID that failed.
+	Errors datautils.Map[ID, error]
+}
+
+// pair holds a key-value pair for channel communication during task execution.
+// This is a private helper type used internally by Execute.
+type pair[ID comparable, DATA any] struct {
+	// ID is the task identifier.
+	ID ID
+	// Data holds either the task result or error.
+	Data DATA
+}
+
+// Task is a generic function type representing an asynchronous task that
+// executes in a context and returns a task ID, data pointer, and optional error.
+//
+// The type parameters are:
+//   - T: The task ID type, must be comparable
+//   - D: The data type returned by the task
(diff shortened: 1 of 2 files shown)


Cobalt0s:
* **#3053** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3053?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#2775](https://github.com/amp-labs/connectors/pull/2775) <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2775?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#3072](https://github.com/amp-labs/connectors/pull/3072) <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3072?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3053?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

internal/parallelfetch/fetch.go
```diff
@@ -0,0 +1,130 @@
...
+				errChannel <- pair[ID, error]{
+					ID:   taskID,
+					Data: err,
+				}
+
+				return nil // nolint:nilerr
+			}
+
+			// Task succeeded with data: send to response channel
+			responseChannel <- pair[ID, DATA]{
```
jlimatampersand, line 99: we try avoid using raw channels when we can avoid. Since this we are already using simultaneously, I don't think it is necessary to use channels. We can just use map directly.

jlimatampersand (approved):
Please revise to not use the channel, but otherwise looks good.

internal/parallelfetch/fetch.go
Cobalt0s: I noticed this pattern and it was the first usage of the simultaneously package after it was added, so I assumed it was the standard approach 🤔.

However, maps aren't safe for parallel execution, so we should use channels instead.

Looking into the source code, I think we should use channels:
https://github.com/amp-labs/connectors/blob/main/internal/simultaneously/simultaneously.go#L25
https://github.com/amp-labs/connectors/blob/main/internal/simultaneously/simultaneously.go#L91

Cobalt0s: @jlimatampersand let me know if it is ok to merge this PR.

As I understand the write with the map means that the map may grow and that would be not safe with go routines. Go prohibts this with an error: https://medium.com/@diveshkumarchordia/go-maps-and-concurrency-why-you-need-synchronization-and-how-to-get-it-right-6cf5cae2f26b

Also, as I understood the goal of `simultaneously` is to control the number of routines spawned in program. Otherwise, each connector could exhaust the pool of routines.

Chris may have had more reasons for this package, but that is the main issue that I remember discussing with him.

jlimatampersand (approved):

PR #3054, opened 2026-06-05
feat(connectwise): Read by IDs

Cobalt0s:
# Description

Batch reading of records is done by using the conditions query parameter on the GET endpoint.

The challenge is the URL length limit. Very large URLs will be rejected, causing batch reads to fail. This makes batch reading a problem of grouping and partitioning identifiers across multiple URLs of acceptable size that won't return a bad request response.

Added unit tests to the procedure which groups identifiers.

# Live tests

Creating multiple contacts and then reading them by ids. At the end the test self cleans.

Peek 2026-06-08 20-20

PR #3056, opened 2026-06-05
feat(connectwise): Create Subscribtion

Cobalt0s:
# Description
The Subscription Output contains a mapping from object name to SystemCallback, the ConnectWise resource representing a webhook.

## Subscription behavior
* Create, Update, and Delete operations are all subscribed to. There is no granularity.
* The webhook URL includes a recordId query parameter at the end because ConnectWise appends the ID to this URL (treating it as a string). Ex: "https://somedomain?recordId=".
* The webhook is set to level: owner, which means:
> When set to owner, all ConnectWise PSA <contacts/tickets/projects/etc> are returned.

I created 2 username+password pairs: one used to create the webhook, and the other to trigger operations. The events arrived as expected. This validation was necessary because the memberId in the system-callback resource and the term "owner" raised some doubts.

# Live tests

### Tickets

image

image

image

### Contacts

image

image

image

image

### DIFF
--- a/providers/connectwise/internal/subscriber/create.go
+++ b/providers/connectwise/internal/subscriber/create.go
@@ -1,0 +1,262 @@
+package subscriber
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/internal/datautils"
+	"github.com/amp-labs/connectors/internal/parallelfetch"
+)
+
+// maxConcurrency is the maximum number of goroutines that will run
+// concurrently when creating subscriptions. The value is chosen arbitrarily.
+const maxConcurrency = 3
+
+func (s Strategy) Subscribe(
+	ctx context.Context,
+	params common.SubscribeParams,
+) (*common.SubscriptionResult, error) {
+	if err := params.ValidateParams(); err != nil {
+		return nil, err
+	}
+
+	input, err := s.TypedSubscriptionRequest(params)
+	if err != nil {
+		return nil, err
+	}
+
+	url, err := s.getSubscriptionURL()
+	if err != nil {
+		return nil, err
+	}
+
+	subscriptionURL := url.String()
+
+	tasks := make([]parallelfetch.Task[common.ObjectName, SubscriptionResource], len(params.SubscriptionEvents))
+
+	index := 0
+	for objectName := range params.SubscriptionEvents {
+		tasks[index], err = s.newTaskCreateSubscription(objectName, subscriptionURL, input)
+		if err != nil {
+			return nil, err
+		}
+
+		index += 1
+	}
+
+	createResult := parallelfetch.Execute(ctx, tasks, maxConcurrency)
+	state := getStateFromCreateResponse(createResult)
+
+	if len(createResult.Errors) != 0 {
+		// Some requests failed; initiate rollback.
+		return s.rollbackSubscriptionCreation(ctx, params, state, createResult)
+	}
+
+	return &common.SubscriptionResult{
+		Result: Result{
+			ObjectWebhooks: createResult.Records,
+		},
+		ObjectEvents: state,
+		Status:       common.SubscriptionStatusSuccess,
+	}, nil
+}
+
+// getStateFromCreateResponse extracts the subscription state from batch responses.
+// Successful responses set the events state; errors result in empty ObjectEvents state.
+func getStateFromCreateResponse(
+	subscriptions parallelfetch.Result[common.ObjectName, SubscriptionResource],
+) map[common.ObjectName]common.ObjectEvents {
+	result := make(map[common.ObjectName]common.ObjectEvents)
+
+	for _, subscription := range subscriptions.Records {
+		// Map successful subscription to its events.
+		objectName := webhookObjectTypeToObjectName[subscription.ObjectType]
+		result[objectName] = common.ObjectEvents{
+			Events: []common.SubscriptionEventType{
+				common.SubscriptionEventTypeCreate,
+				common.SubscriptionEventTypeUpdate,
+				common.SubscriptionEventTypeDelete,
+			},
+			WatchFields:       nil,
+			WatchFieldsAll:    false,
+			PassThroughEvents: nil,
+		}
+	}
+
+	for objectName := range subscriptions.Errors {
+		// Failed requests yield no subscription.
+		result[objectName] = common.ObjectEvents{}
+	}
+
+	return result
+}
+
+// rollbackSubscriptionCreation deletes successfully created subscriptions on partial failure.
+// It updates state to reflect remaining subscriptions after attempted rollback.
+// Returns appropriate status based on rollback success.
+func (s Strategy) rollbackSubscriptionCreation(
+	ctx context.Context,
+	params common.SubscribeParams,
+	state map[common.ObjectName]common.ObjectEvents,
+	partialCreation parallelfetch.Result[common.ObjectName, SubscriptionResource],
+) (*common.SubscriptionResult, error) {
+	requestsRegistry := make(datautils.Map[int, common.ObjectName])
+	for objectName, subscription := range partialCreation.Records {
+		requestsRegistry[subscription.ID] = objectName
+	}
+
+	removeResult := s.removeSubscriptionsByIDs(ctx, requestsRegistry.Keys())
+	if len(removeResult.Errors) == 0 {
+		// Full rollback succeeded.
+		objectNames := datautils.FromMap(params.SubscriptionEvents).Keys()
+
+		events := make(map[common.ObjectName]common.ObjectEvents)
+		for _, objectName := range objectNames {
+			events[objectName] = common.ObjectEvents{}
+		}
+
+		return &common.SubscriptionResult{
+			Result: Result{
+				ObjectWebhooks: make(map[common.ObjectName]SubscriptionResource),
+			},
+			ObjectEvents: events,
+			Status:       common.SubscriptionStatusFailed,
+		}, nil
+	}
+
+	// Partial rollback; track remaining subscriptions.
+	existingObjects := datautils.NewSet[common.ObjectName]()
+
+	for requestID := range removeResult.Errors {
+		objectName := requestsRegistry[requestID]
+		existingObjects.AddOne(objectName)
+	}
+
+	allObjects := datautils.FromMap(params.SubscriptionEvents).KeySet()
+	removedObjects := allObjects.Subtract(existingObjects)
+
+	// Clear state for successfully removed objects.
+	for _, objectName := range removedObjects {
+		state[objectName] = common.ObjectEvents{}
+	}
+
+	remainingWebhooks := make(map[common.ObjectName]SubscriptionResource)
+	for name := range existingObjects {
+		remainingWebhooks[name] = partialCreation.Records[name]
+	}
+
+	return &common.SubscriptionResult{
+		Result: Result{
+			ObjectWebhooks: remainingWebhooks,
+		},
+		ObjectEvents: state,
+		Status:       common.SubscriptionStatusFailedToRollback,
+	}, nil
+}
+
+func (s Strategy) newTaskCreateSubscription(objectName common.ObjectName,
+	url string,
+	input Request,
+) (parallelfetch.Task[common.ObjectName, SubscriptionResource], error) {
+	objectType, found := objectNameToWebhookObjectType[objectName]
+	if !found {
+		return nil, fmt.Errorf("%w: cannot subscribe to '%v' object", common.ErrObjectNotSupported, objectName)
+	}
+
+	return func(ctx context.Context) (taskID common.ObjectName, data *SubscriptionResource, err error) {
+		body := SubscriptionResource{
+			// Webhook must end with recordId without the value.
+			// The ConnectWise will append the record identifier as a raw string when sending the event.
+			WebhookURL:     input.WebhookURL + "?recordId=",
+			ObjectType:     objectType,
+			ObjectLevel:    "owner",
+			PayloadVersion: messageVersion,
+		}
+
+		resp, err := s.client.Post(ctx, url, body, s.clientIdHeader())
+		if err != nil {
+			return objectName, nil, err
+		}
+
+		subscription, err := common.UnmarshalJSON[SubscriptionResource](resp)
+		if err != nil {
+			return objectName, nil, err
+		}
+
+		return objectName, subscription, nil
+	}, nil
+}
+
+// SubscriptionResource models a ConnectWise callback (subscription) returned by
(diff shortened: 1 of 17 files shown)


Cobalt0s:
* **#3063** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3063?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3062** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3062?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3057** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3057?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3056** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3056?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3056?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3093** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3093?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3054** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3054?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2775** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2775?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3053** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3053?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#3072](https://github.com/amp-labs/connectors/pull/3072) <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3072?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

providers/connectwise/internal/subscriber/create.go
```diff
@@ -0,0 +1,261 @@
...
+	for objectName := range params.SubscriptionEvents {
+		tasks[index], err = s.newTaskCreateSubscription(objectName, subscriptionURL, input)
+		if err != nil {
+			return nil, err
+		}
+
+		index += 1
+	}
+
+	createResult := parallelfetch.Execute(ctx, tasks, -1)
```
jlimatampersand: why -1 for maxconcurrency? how does it behave?

providers/connectwise/internal/subscriber/strategy.go
```diff
@@ -0,0 +1,66 @@
...
+// Input defines the subscription request payload for the webhook provider.
+type Input struct {
+	// WebhookURL is the endpoint where webhook messages will be delivered.
+	WebhookURL string
+}
+
+// Output represents the subscription result payload.
+type Output struct {
+	// ObjectWebhooks maps object names to their associated webhook resources.
+	ObjectWebhooks datautils.Map[ObjectName, SystemCallback]
```
jlimatampersand: Outputs should always be in native Go data structures, as they should be accessible and serializable from external systems without the need to import extra packages other than the provider-specific one. ObjectName should be coming from the common package.

providers/connectwise/internal/subscriber/create.go
```diff
@@ -0,0 +1,261 @@
...
+		if err != nil {
+			return objectName, nil, err
+		}
+
+		return objectName, subscription, nil
+	}, nil
+}
+
+// SystemCallback represents a ConnectWise subscription.
+type SystemCallback struct {
```
jlimatampersand: this should be exported and accessible by external system that imports connectors package.

providers/connectwise/internal/subscriber/create.go
```diff
@@ -0,0 +1,261 @@
...
+	// Callback properties.
+	ID         SubscriptionID `json:"id"`
```
jlimatampersand: keep it as int. need to access as easily as it can be from server code.

providers/connectwise/internal/subscriber/strategy.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// ObjectName represents a Microsoft resource type used for subscriptions.
+type ObjectName = common.ObjectName
```
jlimatampersand: unnecessary, just use common.ObjectName instead.

providers/connectwise/internal/subscriber/create.go
```diff
@@ -0,0 +1,261 @@
...
+	createResult := parallelfetch.Execute(ctx, tasks, -1)
+	state := getStateFromCreateResponse(createResult)
+
+	if len(createResult.Errors) != 0 {
+		// Some requests failed; initiate rollback.
+		return s.rollbackSubscriptionCreation(ctx, params, state, createResult)
+	}
+
+	return &common.SubscriptionResult{
+		Result: Output{
```
jlimatampersand: Rename to result.

providers/connectwise/internal/subscriber/strategy.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// Input defines the subscription request payload for the webhook provider.
+type Input struct {
+	// WebhookURL is the endpoint where webhook messages will be delivered.
+	WebhookURL string
+}
+
+// Output represents the subscription result payload.
+type Output struct {
```
jlimatampersand: rename to result for consistency.

providers/connectwise/internal/subscriber/strategy.go
```diff
@@ -0,0 +1,66 @@
...
+func NewStrategy(client *common.JSONHTTPClient, providerInfo *providers.ProviderInfo, clientID string) *Strategy {
+	return &Strategy{
+		client:       client,
+		providerInfo: providerInfo,
+		clientID:     clientID,
+	}
+}
+
+// Input defines the subscription request payload for the webhook provider.
+type Input struct {
```
jlimatampersand: rename to payload or params for consistency.

jlimatampersand (changes requested):

providers/connectwise/internal/subscriber/create.go
Cobalt0s: It means no limit. Frankly, not sure what limit to put here if any.

https://github.com/amp-labs/connectors/blob/main/internal/simultaneously/simultaneously.go#L31

providers/connectwise/internal/subscriber/strategy.go
Cobalt0s: ObjectName is an alias. It is not a "wrapper" of the type. The compiler would replace it with `common.ObjectName`.

For `datautils.Map` it makes sense. The type is a decoration of basic `map[K]V` which gives methods to use. Of course if it is to be exported the datautils should not be used.

providers/connectwise/internal/subscriber/create.go
Cobalt0s: Thats a bit unusual. I am not sure why this should be used by external systems. This represents the Webhook in ConnectWise API. I will expose it, though I worry about increasing dependencies.

providers/connectwise/internal/subscriber/create.go
Cobalt0s: Ok, I didn't know that it would be used outside. I was thinking that it would be treated as a black box.

With primitive types it is easy to forget or misuse variables and makes it less readable.

Ideally I would create the SystemCallback alternative model suitable for the server. This would decouple the `Server` from `Provider API`. Otherwise the danger of making code changes in the connectors codebase rises a lot and it would require checking server code during development.

providers/connectwise/internal/subscriber/strategy.go
Cobalt0s: Its an alias, not a direct type definition. Less typing this way. I will remove it.

providers/connectwise/internal/subscriber/create.go
Cobalt0s: This name is indeed out of place. The result is used a lot, so for some reason I did this. But it is not necessary.

providers/connectwise/internal/subscriber/strategy.go
Cobalt0s: done

providers/connectwise/internal/subscriber/strategy.go
Cobalt0s: Params would be more suitable.

Package `subscriber`. The values would `subscriber.Params` and `subscriber.Result`.

Cobalt0s: image

providers/connectwise/internal/subscriber/create.go
```diff
@@ -0,0 +1,254 @@
...
+
+		return objectName, subscription, nil
+	}, nil
+}
+
+// SystemCallback represents a ConnectWise subscription.
+type SystemCallback struct {
+	// Callback properties.
+	ID         int    `json:"id"`
+	Disabled   bool   `json:"inactiveFlag"`
```
jlimatampersand: json tag mismatch

providers/connectwise/internal/subscriber/create.go
```diff
@@ -0,0 +1,254 @@
...
+		if err != nil {
+			return objectName, nil, err
+		}
+
+		return objectName, subscription, nil
+	}, nil
+}
+
+// SystemCallback represents a ConnectWise subscription.
+type SystemCallback struct {
```
jlimatampersand: is SystemCallback a term that ConnectWises actually uses? If not, lets call it ConnectWiseSubsbcribeResult. 

Please do the same for any following PRs and subscription. 

SystemCallback is very confusing, this is not a callback either.

providers/connectwise/internal/webhook/verifier.go
```diff
@@ -0,0 +1,145 @@
...
+// Metadata.key_url field.
+//
+// It parses the raw request body to extract key_url, fetches the signing key,
+// computes the expected signature from the raw body bytes, and compares the
+// computed value with the received signature.
+//
+// It returns true when the webhook signature is valid. If verification cannot
+// be performed, it returns a non-nil error. If the signature is invalid, it
+// returns (false, nil).
+func (v Verifier) VerifyWebhookMessage(
```
jlimatampersand: let's split this PR and put Verification in Separate PR.

jlimatampersand (changes requested):

providers/connectwise/internal/subscriber/create.go
Cobalt0s: The JSON flag is correct. I can change the golang property.

Cobalt0s: It is common to do it that way.
For example there are some more:
```
WebhookURL   string `json:"url"`
ObjectType string `json:"type"`
```

It makes it more readable and understandble. Otherwise the property named `SystemCallback.Type` is less clear than `SystemCallback.ObjectType`.

The `inactiveFlag` is even not used. Maybe should remove it just like `Miscellaneous properties` but I don't have a precise tactic for this.

providers/connectwise/internal/subscriber/create.go
Cobalt0s: SystemCallback is the actual name of the REST API resource returned by ConnectWise. The struct tags match the JSON returned by the API. The struct’s fields may intentionally differ slightly to abstract away some of the provider-specific details, but the underlying type is what the API uses.

If we want an exposed, more readable version of this data, your suggestion is correct: we should introduce connectwise.SubscribeResult (or similar) and maintain a conversion layer between:

* SystemCallback – the provider’s API representation (what the API returns), and
* SubscribeResult – the cleaned-up, user-facing representation.

We can also define a clearer term for the API representation layer. Some options:
* SubscriptionAPIResource
* SubscriptionResource
* SubscriptionGetResponse (if we want to emphasize it’s the GET response shape)

This layer represents the “provider’s representation of a webhook/subscription resource,” and its exact format may differ across providers and endpoints (GET vs POST vs PUT).

Also, as a side note: it’s generally bad practice (and IDEs even warn about it) to name structs, interfaces, or variables so that they start with the package name (e.g. connectwise.ConnectWiseWebhook or common.CommonWebhook). That’s another reason why SystemCallback is not ideal, but it’s still the correct ConnectWise API term. The solution is to keep SystemCallback as the internal API type and introduce a better-named, package-coherent type like SubscribeResult for external use.

providers/connectwise/internal/webhook/verifier.go
Cobalt0s: Sure!

providers/connectwise/internal/subscriber/create.go
Cobalt0s: I struggle to understand what is required by server and why we cannot treat managing "Subscription API Resources" as a black box.

It would be helpful to know what is required by the outside world to satisfy those requirements/contract. I would still prefer ro decouple provider level from any upstream logic (at least by introducing the layer to avoid unforseen integration bugs caused by server+connectors).

jlimatampersand: Okay, now that we have the Serverside work guide I shared, you should have some understanding. 

In the server repo, there subscribeInstallation function.

subscribeToProvider

function subscribes and stores the result from the Connector method to our database

Then during updateInstallation time

updateProviderSubscription is called in which 

calls prepareSubscribeParamFromResolver

Then you can see how returned results from connector method will be used. Please DM me if you have hard time understanding, I can hop on a call to explain

jlimatampersand (approved):

providers/connectwise/internal/subscriber/create.go
Cobalt0s: Yes, I do see the main loop and it makes sense how everything comes together from `subscribeInstallation`. 

I don't see any "general" way the Result is used, but I see gmail as an exception, where it makes sense that some extra info should be leaked. Gmail has special event flow via history tokens as I recall the provider docs. It doesn't look like ConnectWise would need this.

There may be some indirect or “soft” usages that are hard to find by static search, but from what I inspected I didn't spot other consumers. If you think there are hidden references I missed, point me to the likely areas and I'll dig deeper.

Having said that I will expose the data type and name it `SubscriptionResource` as something more neutral (ConnectWise docs are guarded so links to webpage is not so useful, it is also one big page). Also the `common.SubscribeParams.Request` is better to be called `subscriber.Request` over `subscriber.Params`.

providers/connectwise/internal/subscriber/create.go
Cobalt0s: Setting it to 3 like in other places.

providers/connectwise/internal/subscriber/create.go
Cobalt0s: `SystemCallback` renamed to `SubscriptionResource` and exported.

image

PR #3058, opened 2026-06-05
[CON-3243] feat: add GetRecordsById in calendar subscriber connector

Jkarage:


providers/google/internal/calendar/subscribe.go
```diff
@@ -221,6 +230,107 @@ func (a *Adapter) RunScheduledMaintenance(
...
+// Verification (comparing the X-Goog-Channel-Token header against a stored token) is
+// being delivered in a separate PR. Until then this refuses verification rather than
+// asserting authenticity it never checked — callers must not treat unverified messages
+// as trusted.
+func (a *Adapter) VerifyWebhookMessage(
+	ctx context.Context,
+	request *common.WebhookRequest,
+	params *common.VerificationParams,
+) (bool, error) {
+	return false, common.ErrNotImplemented
```
Jkarage: Should I just return true here. As I when I read the docs, if this returns false. the request is dropped.

providers/google/internal/calendar/subscribe.go
anushat05, line 35: `objectWatchPaths` supports events/calendarList/settings/acl, but `GetRecordsByIds` only handles `events` objects. Since the `updatedMin` fetch is fundamentally events-only, calendarList/settings/acl .list don't support `updatedMin`, only `syncToken`. Watching them creates channels whose notifications can't be processed. 

For MVP this scope for objectWatchPaths is only `events` (or add a guard rejecting Subscribe for objects GetRecordsByIds can't handle), and add the other three back when they get a real fetch path.

providers/google/internal/calendar/subscribe.go
anushat05: 1. Add a `CalendarVerificationParams{ ChannelToken string }` (exported, like the other calendar re-exports).
2. Implement the real `VerifyWebhookMessage` compare the `X-Goog-Channel-Token` header against `params.ChannelToken` (token will be `"amp_"+installationID`). Same idea as Zoho's verify in [providers/zoho/subscriptionEvent.go](https://github.com/amp-labs/connectors/blob/d92a1d78e3fd324976ef68fcde08f5574234ed65/providers/zoho/subscriptionEvent.go#L37), just read from the header since Calendar's body is empty.
3. Heads-up: a zero-value `&google.Connector{}` won't reach this VerifyWebhookMessage dispatches on c.Calendar != nil, so it returns `ErrNotImplemented` and the webhook gets dropped. Please expose a standalone Calendar verifier (a small exported type implementing `WebhookVerifierConnector`) the server config can point at directly.

anushat05 (changes requested):

providers/google/internal/calendar/subscribe.go
Jkarage: - CalendarVerificationParams{ ChannelToken string } — added as an exported as well.
  - VerifyWebhookMessage — compares the X-Goog-Channel-Token header against params.ChannelToken.
  - Standalone CalendarVerifier NewCalendarVerifier implements WebhookVerifierConnector and always carries the Calendar adapter, so the server config can point at it directly without hitting the c.Calendar != nil nil dispatch.

providers/google/internal/calendar/subscribe.go
Jkarage: done, only `events` left.

anushat05 (approved):
Couple of things which can be added as a follow up PR 
1.  **Event classification (separate from `GetRecordsByIds`).** The server needs each event's type (create/update/delete) to honor what the customer subscribed to (e.g. updates-only → drop creates/deletes). Like Gmail's `SubscriptionEventsFromHistory` (`providers/google/subscriptionEvent.go`), keep this a **separate** step — `GetRecordsByIds` stays a pure fetch. Add a Calendar `EventType()`/helper that derives the type from the row (no explicit type field like Gmail's history categories):     - `status == "cancelled"` → delete    - `created` within the `updatedMin` window → create else → update

2. **Unit test for `GetRecordsByIds`.** Only `VerifyWebhookMessage` is covered. The fixtures (`records-by-ids/events-page-1/2/single.json`) are already here, add a `go test` asserting pagination across pages, the empty-`recordIds` error, and the events-only guard.

PR #3067, opened 2026-06-10
fix(acculynx): startDate/endDate filter fix

ejoo:
## Summary
  - AccuLynx's `/jobs` and `/jobs/{id}/history` endpoints return HTTP 400 (`"Start Date and End Date do not have the same format"`) when only one of `startDate` / `endDate` is
  sent.
  - The Ampersand platform fires incremental reads with `Since` set and `Until` zero (the standard backfill shape), which the connector previously translated to sending `startDate`
   alone — triggering the 400.
  - Added a `pairedDateWindow()` helper that always sends both bounds. When `Until` is zero it defaults to `time.Now().UTC()`; when `Since` is zero it defaults to Unix epoch. The
  connector-side time filter still enforces precise bounds on the response, so widening upstream is safe.

  ## How it was tested
  - **Unit tests** (`providers/acculynx/incremental_test.go`) covering `pairedDateWindow` directly, plus `applyJobsIncrementalFilter` and `applyHistoryDateWindow` for the full
  Since-set/Until-zero, Until-set/Since-zero, and both-set cases.
  - **Live-tenant regression** in `test/acculynx/read/main.go` (`testBackfillShapeRead`) that exercises the exact platform shape (`Since` set, `Until` zero) against a real AccuLynx
   tenant for both `jobs` and `jobs/history`. Before the fix this reproduces the HTTP 400; after the fix both objects read cleanly.
  - **Caught during deep-integration testing** — the dashboard Operations log surfaced the exact `"Start Date and End Date do not have the same format"` 400 from AccuLynx, which is
   what motivated this fix.
  - `go test ./providers/acculynx/...` passes, `make lint` reports 0 issues

### DIFF
--- a/providers/acculynx/incremental.go
+++ b/providers/acculynx/incremental.go
@@ -116,15 +116,10 @@ func applyJobsIncrementalFilter(url *urlbuilder.URL, params common.ReadParams) {
 		return
 	}
 
+	since, until := pairedDateWindow(params.Since, params.Until)
 	url.WithQueryParam("dateFilterType", "ModifiedDate")
-
-	if !params.Since.IsZero() {
-		url.WithQueryParam("startDate", params.Since.Format(time.DateOnly))
-	}
-
-	if !params.Until.IsZero() {
-		url.WithQueryParam("endDate", params.Until.Format(time.DateOnly))
-	}
+	url.WithQueryParam("startDate", since.Format(time.DateOnly))
+	url.WithQueryParam("endDate", until.Format(time.DateOnly))
 }
 
 // applyHistoryDateWindow pushes Since/Until into AccuLynx's server-side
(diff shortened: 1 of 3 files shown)


providers/acculynx/incremental.go
```diff
@@ -143,11 +138,28 @@ func applyHistoryDateWindow(url *urlbuilder.URL, params common.ReadParams) {
...
+// the same format") on /jobs and /jobs/{id}/history when only one of
+// startDate / endDate is sent. Missing bounds default to:
+//   - Until → time.Now().UTC()
+//   - Since → Unix epoch (predates AccuLynx; safely covers all history)
+//
+// The connector-side time filter still enforces precise bounds on the
+// response, so widening upstream when a bound is missing is safe.
+func pairedDateWindow(since, until time.Time) (time.Time, time.Time) {
+	if until.IsZero() {
+		until = time.Now().UTC()
```
anushat05, line 157: One thing to confirm: endDate defaults to today at day-granularity, and the docs don't pin whether that's inclusive of the full day. Can you check a job modified today still comes back via testBackfillShapeRead?

anushat05 (approved):

providers/acculynx/incremental.go
ejoo: Good catch! verified live. I made the same proxy call the connector would generate for the backfill shape (Since = 15 days ago, Until = zero, which maps to startDate=2026-05-28, endDate=2026-06-12), modified a job today, and the just-modified job came back in the response with modifiedDate=2026-06-12T20:33:47Z (same calendar day as endDate). So AccuLynx is treating endDate at day-granularity as inclusive of the full day, today's modifications are not missed. No code change needed.

PR #3070, opened 2026-06-11
feat[ENG-3789]: catch HubSpot invalid pagination token errors

laurenzlong:
Sample error from HubSpot:  `HTTP status 400: bad request: Invalid input JSON on line 1, column 10: Cannot deserialize value of type int from String \"https://api.hubapi.com/crm/objects/v3/contacts?limit=100\u0026after=70680725625\u0026properties ....`

### DIFF
--- a/providers/hubspot/internal/core/errors.go
+++ b/providers/hubspot/internal/core/errors.go
@@ -18,7 +18,7 @@ const hubspotMigrationStatusCode = 477
 // search/list APIs reject the supplied pagination cursor as unparseable.
 // This is caused by the pagination cursor being not the appropriate type
 // for the API (because Search and List APIs use different pagination tokens).
-const invalidPaginationCursorMessageSubstring = "Cannot deserialize value of type `int` from String"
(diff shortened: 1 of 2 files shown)


providers/hubspot/internal/core/errors.go
```diff
@@ -13,6 +14,12 @@ import (
 // Migrations resolve in hours, so it is a retryable error.
 const hubspotMigrationStatusCode = 477
 
+// invalidPaginationCursorMessageSubstring is the message HubSpot returns when its
+// search/list APIs reject the supplied pagination cursor as unparseable.
+// This is caused by the pagination cursor being not the appropriate type
+// for the API (because Search and List APIs use different pagination tokens).
+const invalidPaginationCursorMessageSubstring = "Cannot deserialize value of type `int` from String"
```
RajatPawar:
```suggestion
const invalidPaginationCursorMessageSubstring = "Cannot deserialize value of type"
```
Would be good to drop this because we might misuse the cursor in either direction (search to list, list to search)

RajatPawar (approved):

PR #3072, opened 2026-06-11
[CON-3296] feat(slack): Read by IDs

Cobalt0s:
# Description

Slack has limited endpoints for reading single records. Only a handful of resources support fetching individual records instead of collections, and all of these resource names end with `.info` (e.g., `bots.info`, `conversations.info`, `users.info`).

Since Slack provides no batch endpoints for single-record reads, this implementation uses parallel requests to fetch records individually. The `parallelfetch` utility executes all requests concurrently, then combines the results into a single response.

# Live Tests
image

image

Slack App:
image

PR #3061, opened 2026-06-09
feat(utils): Add SubscriptionHelper package

Cobalt0s:
# Description

Given previous and new event objects we can categorize them into 4 categories:
* object subscriptions **to create**
* object subscriptions **to update**
* object subscriptions **to keep**
* object subscriptions **to remove**

PR #3073, opened 2026-06-11
[CON-3296] feat(slack): Subscription Event format

Cobalt0s:
# Description

This PR implements Slack event subscriptions and webhook verification for the connector.

## Event mapping

Event names are mapped to connector object names and subscription event types. The `eventNameToEventDescription` variable serves as the glue between Slack events and our objects. I've estimated the most relevant events based on Slack's API documentation.

## Webhook implementation

- **Webhook message format**: Implemented with `team_id` representing the workspace ID
- **Verification**: The `Verifier.VerifyWebhookMessage` method validates incoming webhook requests using HMAC-SHA256 signature verification, following Slack's official documentation [docs.slack.dev](https://docs.slack.dev/authentication/verifying-requests-from-slack/)
- **Test coverage**: See `message_test.go` and sample JSON files (`channel-created.json`, etc.) for webhook format examples

## Configuration

- Added new `metadata.signingSecret` property to the connector (required for subscriptions, optional for backward compatibility)
- Updated `ProviderInfo` in the provider catalog with instructions on acquiring the Signing Secret

## Subscription confirmation

The `subscriptionConfirmation` handler in `test/slack/subscription/webhook.go` implements the URL verification challenge that Slack sends when setting up the webhook URL in the dashboard.

**Note**: The server must handle this verification challenge in production.

# Live Tests

The event arrives and script invokes `VerifyWebhookMessage` as shown by the debbuger breakpoint.
Then the content is printed to the screen.

Peek 2026-06-12 02-57


Cobalt0s:
* **#3073** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3073?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3073?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3072** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3072?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3053** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3053?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#2775](https://github.com/amp-labs/connectors/pull/2775) <a href="https://app.graphite.com/github/pr/amp-labs/connectors/2775?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

internal/httpkit/body.go
```diff
@@ -0,0 +1,31 @@
...
+// ReadJSONBody parses the request body into type T and restores req.Body
+// so it can be read again by other handlers.
+func ReadJSONBody[T any](req *http.Request) (T, error) {
+	var result T
+
+	// Read entire body
+	bodyBytes, err := io.ReadAll(req.Body)
+	if err != nil {
+		return result, fmt.Errorf("failed to read body: %w", err)
+	}
```
eberle1080: You're skipping over closing the original body (after reading it). This is a resource leak.

internal/httpkit/response.go
```diff
@@ -29,3 +34,24 @@ func Status4xx(code int) bool {
...
+	}
+
+	if value := headers.Get(name); value != "" {
+		return strings.TrimSpace(value), nil
+	}
+
+	canonicalName := textproto.CanonicalMIMEHeaderKey(name)
+	if values, ok := headers[canonicalName]; ok && len(values) > 0 {
+		return strings.TrimSpace(values[0]), nil
+	}
```
eberle1080: I believe `headers.Get` already does this kind of transformation

internal/httpkit/response.go
```diff
@@ -29,3 +34,24 @@ func Status4xx(code int) bool {
 func Status5xx(code int) bool {
 	return 500 <= code && code < 600
 }
+
+func ExtractHeader(headers http.Header, name string) (string, error) {
```
eberle1080: Might want to call this `ExtractRequiredHeader` since it errors out when the header is missing.

providers/slack/internal/webhook/message.go
```diff
@@ -0,0 +1,182 @@
...
+		return "", fmt.Errorf("'event.%v.id' is not of type string", identifierHolder) // nolint:err113
+	}
+
+	return identifier, nil
+}
+
+// EventTimeStampNano
+// "event_time" property:
+// > The epoch timestamp in seconds indicating when this event was dispatched.
+func (e Event) EventTimeStampNano() (int64, error) {
```
eberle1080, line 156: Does the interface mandate this being an int64? Generally I try to avoid working with unix timestamps (unless external systems force me to) because we lose the timezone info. Better to prefer `time.Time` if that's an option. But if not, just ignore this comment.

providers/slack/internal/webhook/verifier.go
```diff
@@ -0,0 +1,100 @@
...
+//   - X-Slack-Signature header: v0=<hex-digest>
+//   - X-Slack-Request-Timestamp header: Unix timestamp
+//
+// The signature is computed as:
+//
+//	sig_basestring = "v0:" + timestamp + ":" + request_body
+//	signature = "v0=" + HMAC-SHA256(signing_secret, sig_basestring).hex()
+//
+// The request is rejected if the timestamp is more than 5 minutes old (replay attack protection).
+func (v Verifier) VerifyWebhookMessage(
```
eberle1080, line 53: I'm assuming we're not supporting verification tokens? I'm good with it, but I wonder if there are any customers that still rely on it. More of a hypothetical I guess. Would be nice though if - supposing the normal verification fails, it could check to see if the verification token header is set, and if that's *all* that's available, error out stating that "Ampersand doesn't support verification tokens, please upgrade to signing secrets"

eberle1080 (changes requested):
Looks good overall, only blocker is the resource leak in `ReadJSONBody`

internal/httpkit/body.go
Cobalt0s: It might be a mistake that I put it in the httpkit, because then it should follow up some strict rules. 

What I have read, it seems that it is fine to not close it. There is always some confusion about this in golang and I am still not sure what to do. Coming long time ago from Java I would close everything :)

This utility is called from the http server which is `http.HandlerFunc`.
https://stackoverflow.com/questions/43021058/golang-read-request-body-multiple-times

Likely, I should reshuffle some code to not have this problem in the first place. However, I would still need to read the Body but only once.

`http` package: 
image

internal/httpkit/response.go
Cobalt0s: Yes, indeed. Somehow, I am not able to replicate the problem that caused me to take this direction. I will remove it.

providers/slack/internal/webhook/message.go
Cobalt0s: Makes sense that saving the time zone can be important. This method is part of our `common.SubscriptionEvent` interface.

See: https://github.com/amp-labs/connectors/blob/7a1469332237fff17df7acaf7418a1dd428d4432/common/types.go#L789

providers/slack/internal/webhook/verifier.go
Cobalt0s: In the docs it says that it is deprecated.

image

Generating Signing Secret in the dashboard is very easy, I would just enforce it and of course better from the security perspective.

If that is not set no event shall pass :smile: .

Cobalt0s:
@eberle1080 new changes in this commit: https://github.com/amp-labs/connectors/pull/3073/changes/f2d3574ee006e3e4f4eff90b06f1d25f269810a7

eberle1080 (approved):

PR #3074, opened 2026-06-12
[CON-3252] feat: add zoominfo read connector

Jkarage:
# Conventions

Live Tests:
Screenshot 2026-06-12 at 11 30 45


providers/zoominfo/connector.go
```diff
@@ -45,5 +47,21 @@ func constructor(base *components.Connector) (*Connector, error) {
...
 
+	registry, err := components.NewEndpointRegistry(supportedOperations())
+	if err != nil {
+		return nil, err
+	}
+
+	connector.Reader = reader.NewHTTPReader(
+		connector.HTTPClient().Client,
+		registry,
+		common.ModuleRoot,
```
RajatPawar: please use connector.Module()

RajatPawar (approved):

PR #3075, opened 2026-06-12
[CON-3286] feat(BreezyHR): Read implementation

Fullstack-Hero:
## Summary
This PR implements Read for the Breezy HR connector (API v3), enabling data retrieval from Breezy’s list endpoints. Users can read records from supported Breezy objects. Company-scoped paths require metadata.company_id.

## Key Features

- HTTP Read implementation: Adds Read via the shared components.Reader (reader.NewHTTPReader), following existing deep-connector conventions.
- Company-scoped requests: Requires and uses metadata.company_id (e.g. from GET /v3/companies) to build v3 URLs under /v3/company/{company_id}/... for company-scoped objects.
- No pagination: Breezy list endpoints return full result sets in a single response; no next_page or ReadParams.PageSize handling.

## Filtering
- Provider-side: positions accepts ReadParams.Filter as the Breezy state query param (published, draft, archived, etc.). The API defaults to published when omitted.
- Connector-side incremental: positions (updated_date) and webhook_endpoints (updated_at) support ReadParams.Since / Until via connector-side time filtering (Salesfinity pattern). Breezy does not expose time-range query params on these endpoints.
- Full sync only: companies, pipelines, categories, departments, questionnaires, and templates are small lookup lists with no incremental timestamps.

## Supported objects:

- `companies`
- `positions`
- `pipelines`
- `categories`
- `departments`
- `questionnaires`
- `templates`
- `webhook_endpoints`

## Catalog metadata:
- Adds provider metadata input company_id

## Testing
```bash
# Run unit tests
go test ./providers/breezy -v

# Run with coverage
go test ./providers/breezy -cover

# Prerequisites: create breezy credentials file used by test harness
# Run integration test against real API
go run ./test/breezy/read/main.go
```

### Integration Tests
Screenshot 2026-06-18 at 6 32 35 PM
Screenshot 2026-06-18 at 6 32 49 PM
Screenshot 2026-06-18 at 6 33 04 PM
Screenshot 2026-06-18 at 6 33 32 PM
Screenshot 2026-06-18 at 6 33 40 PM
Screenshot 2026-06-18 at 6 33 51 PM
Screenshot 2026-06-18 at 6 34 01 PM
Screenshot 2026-06-18 at 6 34 09 PM
Screenshot 2026-06-18 at 6 34 30 PM
Screenshot 2026-06-18 at 6 34 46 PM
Screenshot 2026-06-18 at 6 34 59 PM


providers/breezy/metadata/README.md
```diff
@@ -1,3 +1,9 @@
 # Metadata
 
-The static file `schemas.json` was created by manually inspecting the [Breezy HR API documentation](https://developer.breezy.hr/reference/overview) and live API responses, then translating the supported objects into our static schema format.
\ No newline at end of file
+The static file `schemas.json` was created by manually inspecting the [Breezy HR API documentation](https://developer.breezy.hr/reference/overview) and live API responses, then translating the supported objects into our static schema format. All listed objects can be downloaded via read.
+
+## Read filtering
```
Cobalt0s, line 7: Ther README is helpful. This is exactly the right place for this information as it is for us not external customers and gives good overview, brief and on point.

providers/breezy/read.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	path, err := metadata.Schemas.FindURLPath(c.ProviderContext.Module(), params.ObjectName)
+	if err != nil {
+		return nil, err
+	}
+
+	if strings.Contains(path, companyIDPlaceholder) {
+		if c.CompanyID == "" {
+			return nil, ErrMissingCompanyID
```
Cobalt0s: I think we should enforce the `CompanyID` as a required field. That must be configured by marking connector struct with `common.RequireMetadata`. So then `Init` will validate the `params.Metadata`.

It should be fair to say that a connector cannot funciton without this variable, even if one endpoint like `/v3/companies` doesn't need the id.

providers/breezy/read.go
```diff
@@ -0,0 +1,158 @@
...
+
+	endpointURL, err := buildVersionedPathURL(c.ProviderInfo().BaseURL, path)
+	if err != nil {
+		return nil, err
+	}
+
+	if params.ObjectName == objectPositions && params.Filter != "" {
+		// Provider-side state filter (draft, archived, published, etc.).
+		// https://developer.breezy.hr/reference/company-positions
+		endpointURL.WithQueryParam("state", params.Filter)
```
Cobalt0s: Don't use `params.Filter`. This more of a hack rather than the norm.

Any provider-side filtering that you can construct in the connector without the need of extra information must be done. Otherwise, downgrade it to connector-side filtering or no filtering if no time fields are present.

providers/breezy/read.go
```diff
@@ -0,0 +1,158 @@
...
+		makeFilterFunc(params),
+		readhelper.MakeMarshaledDataFuncWithId(nil, idFieldForObject(params.ObjectName)),
+		params.Fields,
+	)
+}
+
+func nodeRecordsForRead(c *Connector, objectName string) common.NodeRecordsFunc {
+	switch objectName {
+	case objectPipelines:
+		return pipelineRecordNodes()
```
Cobalt0s, line 97: That is interesting. Possibly you should serve the default pipelines. Or maybe loop through each "category" and combine all of the pipelines together. There is some very particular use case for Hubspot pipelines: https://github.com/amp-labs/connectors/blob/main/providers/hubspot/metadata.go#L391

I think best approach -- return default pipelines. Other approach merge all pipelines and return them. If there are special use cases like in Hubspot that should be handled only when we get real customer. But before that the best routes are simpler.

From your test sample:
```
{
  "default": {
    "_id": "default",
    "name": "Default Pipeline",
    "type": "default",
    "pipeline": []
  },
  "default_pool": {
    "_id": "default_pool",
    "name": "Default Pool",
    "type": "pool",
    "pipeline": []
  }
}
```
I think the meat is inside "<some_key>.pipeline". But I like the approach where you preserve the type/name of the pipeline. I think it all depends what is inside the "pipeline" property. Would that be 1 pipeline or it will be many of them (my guess many of them).

providers/breezy.go
```diff
@@ -29,7 +29,8 @@ func init() {
 					DisplayName: "Company ID",
 					DocsURL:     "https://developer.breezy.hr/reference/customer-guide",
 					Prompt: "Copy the company `_id` from GET /v3/companies. " +
-						"Required for company-scoped objects such as positions, pipelines, and webhook endpoints.",
+						"Required for company-scoped objects such as positions, pipelines, categories, " +
+						"departments, questionnaires, templates, and webhook endpoints.",
```
Cobalt0s: By the way. The `webhook` is not a critical object for read.

Usually `webhook` won't be exposed by Read/Write/Metadata.
The `SubscribeConnector` would use the endpoints dealing with webhooks.
Similar how the "dicsovery/schema" endpoints are not exposed by Read but are used by ListObjectMetadata.

providers/breezy/read_test.go
```diff
@@ -0,0 +1,580 @@
...
+	})
+
+	return testroutines.ComparatorSubsetRead(serverURL, actual, expected)
+}
+
+func constructTestConnectorWithoutCompanyID(serverURL string) (*Connector, error) {
+	connector, err := NewConnector(
+		common.ConnectorParams{
+			Module:              common.ModuleRoot,
+			AuthenticatedClient: mockutils.NewClient(),
```
Cobalt0s: By the way, something that I noticed is that you could pass the `tt.Server` as func param and it has the client. So this will become `server.Client()` instead of `mockutils.NewClient()`. This comment can be disregarded as it doesn't change the essence.

providers/breezy/read_test.go
```diff
@@ -0,0 +1,580 @@
...
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	connector.SetUnitTestBaseURL(serverURL)
```
Cobalt0s: Use `SetUnitTestMockServerBaseURL`.

There are several versions of them. One was created when the first modules were introduced. Overall this one is the correct impl, others had some issues.

providers/breezy/read_test.go
```diff
@@ -0,0 +1,580 @@
...
+	"github.com/amp-labs/connectors/test/utils/mockutils"
+	"github.com/amp-labs/connectors/test/utils/mockutils/mockcond"
+	"github.com/amp-labs/connectors/test/utils/mockutils/mockserver"
+	"github.com/amp-labs/connectors/test/utils/testroutines"
+	"github.com/amp-labs/connectors/test/utils/testutils"
+)
+
+const testCompanyID = "testCompanyID"
+
+func TestRead(t *testing.T) { //nolint:funlen
```
Cobalt0s, line 20: Awesome!

test/breezy/read/main.go
```diff
@@ -0,0 +1,84 @@
...
+			continue
+		}
+
+		testscenario.ValidateMetadataContainsRead(ctx, conn, obj.name, nil)
+	}
+
+	slog.Info("Breezy read tests completed successfully!")
+}
+
+func validateMetadataSchemaOnly(ctx context.Context, conn connectors.ObjectMetadataConnector, objectName string) {
```
Cobalt0s: There are some utils that have similar goals like: `testscenario.ValidateMetadataExactlyMatchesRead` and `testscenario.ValidateMetadataContainsRead` maybe they can be helpful.

Cobalt0s (approved):
Looks good, well done!

PR #3076, opened 2026-06-12
[CON-3253] feat: add zoominfo write connector

Jkarage:
# Conventions

Live Tests:
Screenshot 2026-06-12 at 15 28 28

### DIFF
--- a/providers/zoominfo/connector.go
+++ b/providers/zoominfo/connector.go
@@ -6,6 +6,7 @@ import (
 	"github.com/amp-labs/connectors/internal/components/operations"
 	"github.com/amp-labs/connectors/internal/components/reader"
 	"github.com/amp-labs/connectors/internal/components/schema"
+	"github.com/amp-labs/connectors/internal/components/writer"
 	"github.com/amp-labs/connectors/providers"
 )
 
@@ -24,6 +25,7 @@ type Connector struct {
 	// Supported operations
 	components.SchemaProvider
 	components.Reader
+	components.Writer
 }
 
 func NewConnector(params common.ConnectorParams) (*Connector, error) {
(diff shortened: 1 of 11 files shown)


providers/zoominfo/connector.go
```diff
@@ -63,5 +65,16 @@ func constructor(base *components.Connector) (*Connector, error) {
 		},
 	)
 
+	connector.Writer = writer.NewHTTPWriter(
+		connector.HTTPClient().Client,
+		registry,
+		common.ModuleRoot,
```
RajatPawar: please use connector.Module()

providers/zoominfo/objects.go
```diff
@@ -29,6 +29,26 @@ const (
...
+const (
+	objCustomerBuyerPersonas = "customer-buyer-personas"
+	objCustomerCompetitors   = "customer-competitors"
+	objIdealCompanyProfile   = "ideal-company-profile"
+	objProducts              = "products"
+	objAudiences             = "audiences"
+	objAudienceFolders       = "audience-folders"
+	objIndustries            = "industries"
+
+	typeAudience             = "Audience"
```
RajatPawar, line 70: can you add a comment about type vs objects?

providers/zoominfo/write.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// buildWriteRequest creates or updates a writable object. The request shape
+// depends on the object's write style:
+//   - upsert (Copilot config): POST {collection}; an existing RecordId is carried
+//     in the JSON:API body as data.id.
+//   - createUpdate (Studio): POST {collection} to create; PATCH {collection}/{id}
+//     to update.
+func (c *Connector) buildWriteRequest(ctx context.Context, params common.WriteParams) (*http.Request, error) {
+	def, ok := writeObjects[params.ObjectName]
```
RajatPawar, line 47: we don't need to check for support here, the registry automatically does it

RajatPawar (approved):

providers/zoominfo/write.go
Jkarage: Yeah, Thanks I have updated this so we only capture the required details here, and not checking if its supported or not.

PR #3080, opened 2026-06-12
fix (fathom): make calls much more slowly when fetching records/summaries

laurenzlong:
See https://developers.fathom.ai/api-overview#heavy-requests-rate-limits


eberle1080 (approved):
Huh. Strange API indeed.

PR #3084, opened 2026-06-12
[ENG-3162] docs: add Subscribe onboarding + PR-process guides for new providers

jlimatampersand:
## What

Contributor docs for onboarding a new provider to **Subscribe**, in three layers:

**1. Implementation reference** — `SUBSCRIBE_ONBOARDING.md`
The full picture: interface ladder (`WebhookVerifier` → `Subscribe` → `RegisterSubscribe` → `SubscriptionMaintainer`), core types, event types, verification, provider metadata + gating, factory wiring, worked example, real-provider shapes, registration, maintenance, twin providers, testing.

**2. PR-process overview** — `SUBSCRIBE_PR_PROCESS.md`
How to ship it as a stack of small, gated-off PRs: principles, the stack diagram + merge order, why/which flags to gate on, managing the stack, and a copy-paste PR-description checklist.

**3. One focused guide per PR** — `docs/subscribe-onboarding/`
- `pr-1-metadata-and-factory.md`
- `pr-2-verification.md`
- `pr-3-subscribe-update-delete.md`
- `pr-4-registration.md` *(provider-specific, if needed)*
- `pr-5-maintenance.md` *(provider-specific, if needed)*
- `pr-6-enable.md`

Each per-PR guide has: goal, what to implement (interface + signatures), files, step-by-step, an example, a checklist, and reviewer focus — and links back to the reference and overview.

## Notes

- Docs only — no code changes.
- `RegisterSubscribeConnector` / `SubscriptionMaintainerConnector` are marked provider-specific / "if needed".
- Caller-side ("server") wiring lives outside this repo and is referenced only conceptually.


jlimatampersand:
* **#3084** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3084?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3084?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

docs/subscribe-onboarding/pr-1-provider-info.md
```diff
@@ -0,0 +1,191 @@
...
+> Part of [Contributing a Subscribe Action](../../CONTRIBUTING_SUBSCRIBE_ACTION.md). Shared concepts:
+> [`SUBSCRIBE_REFERENCES.md`](../../SUBSCRIBE_REFERENCES.md).
+
+**Required.** This is the base of the stack — everything else stacks on it.
+
+## Goal
+
+Declare the provider's subscribe metadata on its `ProviderInfo` — `Support.Subscribe` and
+`SubscribeRequirements` — with the gate (`Support.Subscribe`) **off**, and wire the connector into the
+factory if it's brand-new. This PR changes **no runtime behavior**; it's a safe no-op until the final
```
Cobalt0s, line 12: > wire the connector into the factory if it's brand-new

Are there going to be connectors that do Subscribe before ListObjectMetadata/Read/Write?
If so then mention `connector/new.go` file instead of just saying factory which is arbitrary info for newcomers.

docs/subscribe-onboarding/pr-1-provider-info.md
```diff
@@ -0,0 +1,191 @@
...
+[`Enable`](./pr-6-enable.md) PR.
+
+## What you implement
+
+Declare subscribe capability in the provider's hand-written info file, `providers/<provider>.go`,
```
Cobalt0s: hand-written vs handwritten

docs/subscribe-onboarding/pr-1-provider-info.md
```diff
@@ -0,0 +1,191 @@
...
+inside `SetInfo(...)`. Two pieces matter:
+
+- **`Support.Subscribe`** — the master "this provider supports subscribe at all" switch.
+- **`SubscribeRequirements`** — what the provider *needs* in order to subscribe. All fields are `*bool`;
+  use the package pointer helper `new(true)`. From `providers/types.gen.go`:
```
Cobalt0s: All fields are `*bool`, use new(true) or new(false).

`new` is part of golang builtin statement not a package.

docs/subscribe-onboarding/pr-1-provider-info.md
```diff
@@ -0,0 +1,191 @@
...
+
+```go
+type SubscribeRequirements struct {
+    // Maintenance: subscriptions/watches expire after a TTL and must be renewed on a schedule.
+    Maintenance *bool
+    // PostProcess: subscribing needs a third-party setup step the connector itself cannot perform
+    // (e.g. Salesforce → AWS EventBridge).
```
Cobalt0s, line 29: The Salesforce example is not intuitive. Maybe it is just me. It makes sense that it suggests to invistigate Salesforce to understand PostProcess.

docs/subscribe-onboarding/pr-1-provider-info.md
```diff
@@ -0,0 +1,191 @@
...
+    // https://developer.salesforce.com/docs/platform/pub-sub-api/guide/intro.html
+    SubscribeByAPI: new(true),
+},
+```
+
+## Gating rule
+
+> `Support.Subscribe` is the **gate** — it must be `true` for the provider to subscribe at all (via API
+> or manual/UI); `SubscribeByAPI` says whether the programmatic API approach is available. **Keep
+> `Support.Subscribe` off** for the entire stack and flip it on only in the final
```
Cobalt0s, line 126: Should it be done like this in the final PR or during development? There is another flag, which I would imagine would be the global switch, global gate under `Support`:
Image

In other words, clarify on the nature of both flags and in what order to switch them on. What matters and what doesn't?

PS: came back to a comment after reading the last 6th PR. Still wonder about both flags.

docs/subscribe-onboarding/pr-1-provider-info.md
```diff
@@ -0,0 +1,191 @@
...
+> [`Enable`](./pr-6-enable.md) PR — that's what keeps every intermediate PR a safe no-op even after it
+> merges.
+
+## Factory wiring *(brand-new providers only)*
```
Cobalt0s: I would have the docs for brand-new providers in a different document and focus only on Subscribe.

Maybe it is just me, but it is a lot of steps and information.

docs/subscribe-onboarding/pr-2-verification.md
```diff
@@ -0,0 +1,188 @@
...
+Salesloft models all three with a `map[string]any`:
+
+```go
+type SubscriptionEvent map[string]any
+type CollapsedSubscriptionEvent map[string]any
+
+var (
+    _ common.SubscriptionEvent          = SubscriptionEvent{}
+    _ common.SubscriptionUpdateEvent    = SubscriptionEvent{}
+    _ common.CollapsedSubscriptionEvent = CollapsedSubscriptionEvent{}
```
Cobalt0s, line 81: Must mention that `CollapsedSubscriptionEvent` must be exported by the said provider package.

This is important for the server to access to create an event.

It looks like we don't care about `SubscriptionEvent` and `SubscriptionUpdateEvent` being exported, as the focus is on the umbrella event: `CollapsedSubscriptionEvent`. Correct me if I am wrong. Some of it could be useful to add to the doc itself.

docs/subscribe-onboarding/pr-2-verification.md
```diff
@@ -0,0 +1,188 @@
...
+```
+
+The caller fills `Param` per installation; cast it back to your type with `common.AssertType`. The
+values inside (a signing secret, a token, an account ref, …) come from whatever your connector
+persisted — **anything your `Subscribe` returns in the `SubscriptionResult` is available to the caller
+to thread back here.** So if the provider issues a signing secret at subscribe time, return it in the
+`SubscriptionResult` (PR 3), and the caller will supply it back through `VerificationParams.Param` when
+it calls `VerifyWebhookMessage`.
+
+Implement `VerifyWebhookMessage` on your `*Connector`. The pattern (from
```
Cobalt0s, line 140: For now mention that `VerifyWebhookMessage` implementation cannot rely on the AuthorizationClient.

In general one would have an impression that connector is initialized once so each method is equal in its capabilities. But `VerifyWebhookMessage` is not declared with auth yet.

docs/subscribe-onboarding/pr-2-verification.md
```diff
@@ -0,0 +1,188 @@
...
+(still implement it if the provider does sign its webhooks).
+
+## Checklist
+
+- [ ] `var _ connectors.WebhookVerifierConnector = &Connector{}` assertion present.
+- [ ] `VerifyWebhookMessage` uses constant-time compare (`hmac.Equal`); returns `false` (not an error)
+      for untrusted requests.
+- [ ] All `SubscriptionEvent` methods implemented; `SubscriptionUpdateEvent` /
+      `CollapsedSubscriptionEvent` added where the provider needs them.
+- [ ] Unit tests cover valid / invalid / missing-signature, plus each event method against a captured
```
Cobalt0s, line 197: Connector test suites:
* `testroutines.TestCaseWebhookMessageVerification`
* `testroutines.TestCaseSubscriptionEvent` (I will create a PR to rename it from `SubscriptionEventTestCase`)

docs/subscribe-onboarding/pr-3-registration.md
```diff
@@ -0,0 +1,125 @@
...
+}
+
+type ResultData struct {
+    EventChannel     *EventChannel     `validate:"required"`
+    NamedCredential  *NamedCredential  `validate:"required"`
+    EventRelayConfig *EventRelayConfig `validate:"required"`
+}
+
+func (c *Connector) EmptyRegistrationParams() *common.SubscriptionRegistrationParams {
+    return &common.SubscriptionRegistrationParams{Request: &RegistrationParams{}}
```
Cobalt0s, line 60: It is critical to mention that Request/Result must hold a pointer not a regular value. Otherwise, the implementation won't integrate with server.

Probably we could create alternative to `components.SubscriptionInputOutput` but for registration.

docs/subscribe-onboarding/pr-4-subscribe-update-delete.md
```diff
@@ -0,0 +1,180 @@
...
+type SubscriptionRequest struct {
+    WebhookEndpoint string `json:"webhookEndpoint"`
+    Secret          string `json:"secret"`
+}
+
+type SubscriptionResult struct {
+    SubscriptionID string `json:"subscriptionId"`
+}
+
+func (c *Connector) EmptySubscriptionParams() *common.SubscribeParams {
```
Cobalt0s, line 68: No need to implement `EmptyRegistrationParams` method or `EmptyRegistrationResult`.

Embed `components.SubscriptionInputOutput[Request, Result]` to get these methods out of the box.

Request -- type in `common.SubscribeParams.Request`
Result -- type in `common.SubscriptionResult.Result`


Let me know if you have suggestions on how to better name `components.SubscriptionInputOutput`. 
Also it returns a [pointer to Request/Result](https://github.com/amp-labs/connectors/blob/main/internal/components/subscriptions.go#L53) as expected by server. Returning by value instead of by reference will cause the subsribe to not work during integration testing.

docs/subscribe-onboarding/pr-4-subscribe-update-delete.md
```diff
@@ -0,0 +1,180 @@
...
+DeleteSubscription(ctx context.Context, previousResult common.SubscriptionResult) error
+
+EmptySubscriptionParams() *common.SubscribeParams
+EmptySubscriptionResult() *common.SubscriptionResult
+```
+
+- **`Subscribe`** translates the normalized `params.SubscriptionEvents` (objects → event types) into
+  provider-specific API calls and returns the resulting state. On partial failure it should roll back
+  what it created (see Salesloft/Outreach for the parallel-create-with-rollback pattern).
+- **`UpdateSubscription`** reconciles the existing subscription (`previousResult`) with the new desired
```
Cobalt0s, line 37: Should there be any rollbacks.

docs/subscribe-onboarding/pr-4-subscribe-update-delete.md
```diff
@@ -0,0 +1,180 @@
...
+  state (`params`). The framework only calls this after it detects a change.
+- **`DeleteSubscription`** tears down everything identified by `previousResult`.
```
Cobalt0s: What is the expected previous result `SubscriptionResult.ObjectEvents`? Does it matter for the server?
Is the only important info in previousResult a custom `Result` data type?

I suppose `ObjectEvents` doesn't matter and it is a complete nuke, right?

docs/subscribe-onboarding/pr-4-subscribe-update-delete.md
```diff
@@ -0,0 +1,180 @@
...
+secret. You only define the struct and consume it.
+
+See [`providers/salesloft/subscribe.go`](../../providers/salesloft/subscribe.go) and
+[`providers/outreach/subscribe.go`](../../providers/outreach/subscribe.go) for the parallel-create-with-
+rollback pattern.
+
+## Serialization
+
+The caller **persists your `SubscriptionResult` and reads it back later** (for updates, deletes, and to
+build verification params), so it serializes/deserializes the value you place in the `Result any` field
```
Cobalt0s, line 112: I still occasionally worry that the `Result` structure may act like a stale cache. Could an outage or similar issue cause `Result` to be missing fields/records and therefore diverge from the remote provider's actual state?

Should we be concerned about this? Specifically, should `UpdateSubscription` treat responses like 400/404 (when they rely on `Result`) as a signal that `Result` might be stale and proactively recreate or refresh it?

docs/subscribe-onboarding/pr-4-subscribe-update-delete.md
```diff
@@ -0,0 +1,180 @@
...
+   [`test/utils/testroutines`](../../test/utils/testroutines/connector.go) — one per subscription
+   method. They let each method be verified independently and are what the subscription CUD / update
+   test scenarios consume (in particular `TestableSubscriptionUpdater` for the update path):
+
+   ```go
+   // <provider>_test.go  (testroutines imports "testing", so keep these in a _test.go file)
+   import "github.com/amp-labs/connectors/test/utils/testroutines"
+
+   var (
+       _ testroutines.TestableSubscriptionCreator = &Connector{} // Subscribe
```
Cobalt0s: We may omit these statements, although, they carry their own meaning acting as a "checklist".


When passing the Connector/Strategy to the `TestCase` runner the return type already enforces this. In other words, the connector participating in the unit tests signifies that it conforms to the interface:

Image

docs/subscribe-onboarding/pr-4-subscribe-update-delete.md
```diff
@@ -0,0 +1,180 @@
...
+- [ ] `Empty*` populate the provider-specific `.Request` / `.Result`.
+- [ ] `Request` / `Result` structs round-trip through serialization: exported fields with JSON tags,
+      Go native types.
+- [ ] `SubscriptionResult.ObjectEvents` reflects the actual post-operation state.
+- [ ] Manual harness runs against a real sandbox; a triggered change yields a verified webhook
+      end-to-end.
+
+## Reviewer focus
+
+- Rollback correctness on partial failure (no orphaned provider-side subscriptions).
```
Cobalt0s, line 199: Something important to note, it is a responsibility of the developer to keep the State of:
* common.SubscriptionResult.ObjectEvents
* common.SubscriptionResult.Result
with respect to remote state.

If some subscription was not removed the `ObjectEvents` must still return that object and if `Result` holds subscription records it still should be present.

In summary, both data strucutres require state managment from developer. Also one must consider how each and every `return` statement going to impact this.

A practical note from experience: batch operations (which many connectors rely on) are much easier to reason about if they avoid returning an error -- therefore such design is preferred. Instead of failing fast, prefer returning per-item results ([succeeded/failed](https://github.com/amp-labs/connectors/blob/main/providers/microsoft/internal/batch/response.go#L16)) so callers can reconcile state reliably. Early error returns often hide which items were processed and which were not, making it difficult or impossible to deduce the true remote state.

docs/subscribe-onboarding/pr-2-verification.md
```diff
@@ -0,0 +1,188 @@
+# PR 2 — Verification (`WebhookVerifierConnector`)
+
+> Part of [Contributing a Subscribe Action](../../CONTRIBUTING_SUBSCRIBE_ACTION.md). Shared concepts:
+> [`SUBSCRIBE_REFERENCES.md`](../../SUBSCRIBE_REFERENCES.md).
+
+**Required.** Stacks on [PR 1](./pr-1-provider-info.md).
```
Cobalt0s, line 6: It also requires implementation of `GetRecordsByIds`.

It would be also useful to note that the subscription Event should ignore the data of a record. Primary information is the id of affected record and other metadata only.

This would explain the requirement of the `GetRecordsByIds` in the Subscription stack, even though `GetRecordsByIds` can exist on its own.

docs/subscribe-onboarding/pr-2-verification.md
```diff
@@ -0,0 +1,188 @@
...
+}
+```
+
+The provider-specific `*VerificationParams` struct (here `SalesloftVerificationParams`) is populated
+per installation by the caller and passed through `common.VerificationParams.Param`. Hubspot and
+Outreach show HMAC-SHA256 variants; Salesloft uses HMAC-SHA1. For UI Subscription only providers whose
+events carry no provider signature, verification is bypassed by the caller rather than implemented here
+(still implement it if the provider does sign its webhooks).
+
+## Checklist
```
Cobalt0s, line 190: # Live Test

Here is an example of the script that listens on your local system and prints events to the screen:
* https://github.com/amp-labs/connectors/blob/main/test/connectwise/subscription/consumer/main.go#L19

Prerequisites:
* ngrok http running on 4550 or similar.
* Connector must implement `TestableWebhookMessageVerifier` aka must have `VerifyWebhookMessage` method.

Result:
* Valid messages will be printed on the screen. [See happy path.](https://github.com/amp-labs/connectors/blob/main/test/utils/testscenario/webhook.go#L211).


This script is useful to get started with subscribe feature. And it also demonstrates `VerifyWebhookMessage` working in action.

docs/subscribe-onboarding/pr-4-subscribe-update-delete.md
```diff
@@ -0,0 +1,180 @@
...
+- [ ] Decomposed test assertions present (`TestableSubscriptionCreator` / `TestableSubscriptionUpdater`
+      / `TestableSubscriptionRemover`) so Subscribe, Update, and Delete are each verified independently.
+- [ ] `Subscribe` rolls back partially-created subscriptions on error.
+- [ ] `UpdateSubscription` reconciles `previous` → desired state (handles additions and removals).
+- [ ] `DeleteSubscription` tears down everything in `previous.Result`.
+- [ ] `Empty*` populate the provider-specific `.Request` / `.Result`.
+- [ ] `Request` / `Result` structs round-trip through serialization: exported fields with JSON tags,
+      Go native types.
+- [ ] `SubscriptionResult.ObjectEvents` reflects the actual post-operation state.
+- [ ] Manual harness runs against a real sandbox; a triggered change yields a verified webhook
```
Cobalt0s, line 194: Unit tests:
* `TestCaseCreateSubscription`
* `TestCaseUpdateSubscription`
* `TestCaseDeleteSubscription`

# Live Test

### (1)
Script: `testscenario.ValidateSubscribeReceiveEvents`
Allows to create subscription by using connector `Subscribe` method and then triggers combinations of various `Write`/`Delete` to then await events.

Prerequisite:
* Connector must implement Read/Write/Delete
* Connector must have `VerifyWebhookMessage` and `Subscribe` methods.


Result:
* Given `map[common.ObjectName]common.ObjectEvents` and list of `testscenario.ConnectorOperation`
* Print the events expected as a consequence of subscription.

This script is useful to demonstrate Subscription is created, auto Deleted (if DeleteSubscription method is implemented) and events are displayed.

Example:
* Subscribe to Messages object Create/Delete.
* Create a message, remove it.
* Script waits for 2 events and prints them, which in our case would be Create/Delete.

### (2)
Script `testscenario.SubscriptionCreateUpdateDelete`.

Invokes `Subscribe`, `UpdateSubscription`, `DeleteSubscription`.
It prints `common.SubscriptionResult` to the screen for examination.

CONTRIBUTING_SUBSCRIBE_ACTION.md
```diff
@@ -0,0 +1,160 @@
...
+dependency: **Registration comes before Subscribe** (it creates a shared resource that all object
+subscriptions hang off of, which `Subscribe` consumes), and **Maintenance comes after Subscribe** (it
+renews what Subscribe created). `Enable` is always last.
+
+```
+  Enable the provider (PR 6)             flip Support.Subscribe on   ← merge last
+        ▲
+  Maintenance (PR 5, if needed)          SubscriptionMaintainerConnector — renews after subscribe
+        ▲
+  Subscribe / Update / Delete (PR 4)     SubscribeConnector
```
Cobalt0s, line 25: Is it ok to have
* PR4_Create, PR4_Delete, PR4_Update
or
* PR4_CreateDelete, PR4_Update

CONTRIBUTING_SUBSCRIBE_ACTION.md
```diff
@@ -0,0 +1,160 @@
...
+        ▲
+  Registration (PR 3, if needed)         RegisterSubscribeConnector — shared resource across objects
+        ▲
+  Verification (PR 2)                    WebhookVerifierConnector
+        ▲
```
Cobalt0s, line 30: Missing: `BatchRecordReaderConnector`.

CONTRIBUTING_SUBSCRIBE_ACTION.md
```diff
@@ -0,0 +1,160 @@
...
+   subscribe, maintenance). See [The big picture](./SUBSCRIBE_REFERENCES.md#the-big-picture) for how the
+   interfaces relate (note: PR order is sequenced by dependency, not by the interface ladder — see
+   [The stack](#the-stack)).
+5. **Only build what the provider needs.** `RegisterSubscribeConnector` and
+   `SubscriptionMaintainerConnector` are **provider-specific** — most providers skip them. Don't add a
+   PR for a rung the provider doesn't require.
+
+---
+
+## The stack
```
Cobalt0s, line 12: The whole section is very helpful. Should be the first thing the developer sees.

CONTRIBUTING_SUBSCRIBE_ACTION.md
```diff
@@ -0,0 +1,160 @@
...
+  just branches off whatever precedes it.
+- If you use **Graphite**, this is a normal `gt create` stack; submit with `gt submit --stack`.
+- With plain git, branch each PR off its parent and rebase the upstack when a lower PR changes.
+- When a lower PR gets review changes, restack everything above it before re-pushing.
+- You don't have to wait for PR 1 to merge before opening PR 2 — open the whole stack and let review
+  proceed in parallel; merge bottom-up as each is approved.
+
+---
+
+## PR description checklist (copy into each PR)
```
Cobalt0s: Instead of copying create `subscribe.md` file similar to [proxy_connector.md](https://github.com/amp-labs/connectors/blob/main/.github/PULL_REQUEST_TEMPLATE/proxy_connector.md#L1). Then this PR can simply reference that file.

Graphite expects one of the PR templates to be selected when the PR is first pushed, so keeping the content in a dedicated template file fits that workflow. One caveat from past investigation: supporting both Graphite-driven templates and a regular Git workflow via git hooks was problematic -- it is either or.

CONTRIBUTING_SUBSCRIBE_ACTION.md
```diff
@@ -0,0 +1,160 @@
...
+```
+
+---
+
+## Quick reference
+
+| Want to… | Go to |
+|----------|-------|
+| Understand the interfaces & types | [`SUBSCRIBE_REFERENCES.md`](./SUBSCRIBE_REFERENCES.md) |
+| See a worked example | the **Example** section in each per-PR guide (e.g. [PR 4](./docs/subscribe-onboarding/pr-4-subscribe-update-delete.md#example)) |
```
Cobalt0s: typo --> See a working example

docs/subscribe-onboarding/pr-4-subscribe-update-delete.md
```diff
@@ -0,0 +1,180 @@
...
+    if err != nil {
+        return nil, err
+    }
+    // For each object+event in params.SubscriptionEvents, call the provider API.
+    // Track successes; roll back on partial failure. Return the actual state.
+}
+
+func (c *Connector) UpdateSubscription(
+    ctx context.Context, params common.SubscribeParams, previous *common.SubscriptionResult,
+) (*common.SubscriptionResult, error) { /* reconcile previous → desired */ }
```
Cobalt0s, line 89: We may mention the `subscriptionhelper` package as shared utilities for subscribe feature.

For now it has [SegmentSubscriptionEvents](https://github.com/amp-labs/connectors/blob/main/common/subscriptionhelper/event-segments.go#L43) method.

Cobalt0s (approved):
The docs look good! 🎉  

I left comments in the spirit of thinking out loud and share perspectives from my side.

docs/subscribe-onboarding/pr-1-provider-info.md
jlimatampersand: Only `Support.Subscribe` is a gate — nothing labels the `SubscribeRequirements.*` flags (including `SubscribeByAPI`) as gates; they are indicators. The relationship and switch-on order are covered in the **Gating rule** section: `Support.Subscribe` stays off for the entire stack and flips to `true` exactly once in the final Enable (PR 6), while the requirement indicators are set to their real values in PR 1 and never toggled. This example block just shows the final enabled form, which is why all four booleans appear set together.

docs/subscribe-onboarding/pr-1-provider-info.md
jlimatampersand: Good call — I trimmed the brand-new-provider content so this doc stays Subscribe-focused. The Factory wiring section no longer teaches connector creation inline; the brand-new case now just links out to `CONTRIBUTING.md` ("Adding a Proxy Connector" / "Adding a Deep Connector"). What remains here is minimal and subscribe-specific: existing connector = nothing to do, brand-new = build it first via those links then add a constructor to `connector/new.go`, plus the twin-providers note.

docs/subscribe-onboarding/pr-4-subscribe-update-delete.md
jlimatampersand: For this doc we're intentionally keeping the hand-written `EmptySubscriptionParams` / `EmptySubscriptionResult` rather than steering newcomers to `components.SubscriptionInputOutput`, so the mechanics stay explicit. The substantive point still stands though: the example already returns pointers (`Request: &SubscriptionRequest{}`, `Result: &SubscriptionResult{}`), and I've now made the pointer requirement explicit in the doc — `.Request`/`.Result` must hold pointers or server-side deserialization fails and subscribe silently breaks in integration.

docs/subscribe-onboarding/pr-4-subscribe-update-delete.md
jlimatampersand: Good concern. `Result` is a persisted snapshot, so in principle it can drift from the provider's actual state. We're not going to bake per-provider staleness handling (e.g. treating 400/404 as "Result is stale, recreate") into this onboarding doc — that's a cross-provider design decision, not something each provider author should hand-roll. If an outage or similar causes stale data, it surfaces as a monitoring alert and we handle it there.

docs/subscribe-onboarding/pr-4-subscribe-update-delete.md
jlimatampersand: Agreed — passing the connector to the `TestCase` runner already enforces these interfaces at compile time, so the `var _ ... = &Connector{}` assertions are technically redundant. Keeping them anyway: they document intent as a checklist and surface a conformance failure right at the declaration (naming the specific interface) instead of deep in the test-runner call site. Cheap and clearer errors, so it doesn't hurt to state them here. (Also fixed the package reference in these snippets — they should be `testconn`, not `testroutines`.)

docs/subscribe-onboarding/pr-2-verification.md
jlimatampersand: If you can add a README documentation on live testing, it would be great. Once we have the doc, we can link it to this documentation.

docs/subscribe-onboarding/pr-4-subscribe-update-delete.md
jlimatampersand: Added the unit-test suites to the Testing section (`testconn.TestCaseSubscribe` for create — note the actual name is `TestCaseSubscribe`, not `TestCaseCreateSubscription` — plus `TestCaseUpdateSubscription` and `TestCaseDeleteSubscription`).

For the live-test scripts (`testscenario.ValidateSubscribeReceiveEvents` / `testscenario.SubscriptionCreateUpdateDelete`): same as the PR 2 thread — if you can add a README documenting live testing, it would be great, and we can link it from this documentation once it exists.

CONTRIBUTING_SUBSCRIBE_ACTION.md
jlimatampersand: It's up to the developer — we don't have to force any particular split (PR4_Create/Delete/Update, or PR4_CreateDelete + Update, or one combined PR are all fine) as long as the tests pass and the connector works. Note `SubscribeConnector` bundles Subscribe/Update/Delete in one interface, so the full `var _ connectors.SubscribeConnector = &Connector{}` assertion only passes once all three land — until then use the per-method `Testable*` assertions. Since the stack is gated off until the final Enable PR, intermediate states are safe no-ops either way.

CONTRIBUTING_SUBSCRIBE_ACTION.md
jlimatampersand: Not missing — `WebhookVerifierConnector` embeds `BatchRecordReaderConnector` (see `connectors.go`), so satisfying the PR 2 interface already forces the developer to implement `GetRecordsByIds`; the compiler enforces it. Since it comes along transitively, we don't need to list it separately in the stack diagram/table.

docs/subscribe-onboarding/pr-4-subscribe-update-delete.md
Cobalt0s: Makes sense, so it is only a concern for unique cases and we will look back at implementation should the issues arise if at all.

docs/subscribe-onboarding/pr-4-subscribe-update-delete.md
Cobalt0s: The checklist intent makes sense.

CONTRIBUTING_SUBSCRIBE_ACTION.md
Cobalt0s: Ok, just checking so that I am on the same page. Your motivation is clear to me. Thanks!

CONTRIBUTING_SUBSCRIBE_ACTION.md
Cobalt0s: I was thinking that it should be mentioned in the tree structure.

This could come as a surprise. Essentially batch reading is a pre-requisite for all subscribe features.

docs/subscribe-onboarding/pr-2-verification.md
Cobalt0s: @jlimatampersand 

The guide is ready: https://github.com/amp-labs/connectors/pull/3253

I refactored some code. This should make it more readable. It is a pre-req for the docs PR: https://github.com/amp-labs/connectors/pull/3252

Cobalt0s: The docs are merged.

PR #3085, opened 2026-06-13
feat: Add ReadParams.Opts to support bespoke parameters

laurenzlong:
We sometimes need a way to pass in an additional parameter to affect how a read operation works that is outside of our standard supported parameters like filters or associations. This new field is meant for bespoke parameters that only make sense for a single connector, and does not need an abstraction.

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -220,8 +220,12 @@ type ReadParams struct {
 
 	// Additional options for the read operation that the connector may support.
 	// This optional map is used for bespoke connector-specific parameters.
-	Opts map[string]any // optional
+	Opts ReadParamOpts // optional
 }
+
+// Each connector that supports ReadParams.Opts should define its own type and assert it.
+// e.g. gong.ReadParamOpts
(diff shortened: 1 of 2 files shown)


common/types.go
```diff
@@ -217,8 +217,16 @@ type ReadParams struct {
...
 	PageSize int // optional
+
+	// Additional options for the read operation that the connector may support.
+	// This optional map is used for bespoke connector-specific parameters.
+	Opts ReadParamOpts // optional
 }
 
+// Each connector that supports ReadParams.Opts should define its own type and assert it.
+// e.g. gong.ReadParamOpts
+type ReadParamOpts any
```
RajatPawar: minor: confused if this should be `ReadParamsOpts`? alternatively, `ReadOpts`

RajatPawar (approved):

common/types.go
laurenzlong: good catch, I made this `ReadParamsOpts`

PR #3078, opened 2026-06-12
chore(providers): Constructors to use components.Init

Cobalt0s:
# Description

This is a follow up to https://github.com/amp-labs/connectors/pull/2677.

Changes:
* All connectors that initialize components now use Init.
* The `components.Initialize` is removed.

Constructors that were setting the values post factum now have the natural order. Example:
image

PR #3090, opened 2026-06-16
fix(components): Init to sanitize params

Cobalt0s:
# Background

This PR continues the work discussed in https://github.com/amp-labs/connectors/pull/2775#discussion_r3416662374 and relates to https://github.com/amp-labs/connectors/pull/3078.

# Description

Previous init flow:
* Call constructor.
* Validate params; if params.Metadata is an empty map, return an error.
* Use params.Metadata to set Connector properties.
* Finish initialization.

New init flow:
* Call constructor (constructor is allowed to read/use params.Metadata).
* Validate params; if params.Metadata is an empty map, return an error.
* Finish initialization.

The constructor must be free to access and use `params.Metadata` during construction. To guarantee that, `params.Metadata` must be initialized before the constructor runs to **avoid panic**. The call `params = prepareParams(params)` ensures `params.Metadata` (and any other normalization) is set up for constructor use.

Validation remains in place: we still validate required fields. Validation can still fail if fields are empty strings or otherwise invalid (`common.ValidateParameters` uses interfaces as "tags" on the Connector struct).

### DIFF
--- a/internal/components/connector.go
+++ b/internal/components/connector.go
@@ -77,10 +77,8 @@ func Init[T any](
 		conn = nil
 	})
 
-	// Default module is always the root module
-	if params.Module == "" {
-		params.Module = common.ModuleRoot
-	}
+	// Undeclared properties are set to empty values and default values.
+	params = prepareParams(params)
 
 	transport, err := NewTransport(provider, params)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


Cobalt0s:
* **#3090** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3090?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3090?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jlimatampersand (approved):

PR #3092, opened 2026-06-16
feat:expose campaignId

anushat05:
### Notes
Customers could only map a campaign's integer `hs_object_id`, but HubSpot keys campaigns (and the campaign references on events, e.g. form-submission `hs_matched_campaign_and_assets`) by the GUID, hence there was no way to join events back to their source campaign.

This surfaces the campaign's top-level ID (the GUID) as a mappable field:
  - `schemas.json` — advertise id ("Campaign ID") on marketing-campaigns.
  - `read.go` — exclude id from the campaigns properties= query param. The Campaigns API rejects id there
 ` (Forbidden properties: [id])`. The GUID is returned at the top level of the response and surfaced into fields by the existing FlattenNestedFields transformer.

 ### Testing
* Added campaignId as a field in the integration and verified that the reads are successful and the payload returns it as a mapped field 
Screenshot 2026-06-16 at 16 09 26
Screenshot 2026-06-16 at 16 09 09

Screenshot 2026-06-16 at 16 16 40

### DIFF
--- a/providers/hubspot/internal/metadata/schemas.json
+++ b/providers/hubspot/internal/metadata/schemas.json
@@ -147,7 +147,7 @@
               "providerType": "String"
             },
             "id": {
-              "displayName": "Campaign ID (GUID)",
+              "displayName": "Campaign ID",
               "valueType": "string",
               "providerType": "String"
             },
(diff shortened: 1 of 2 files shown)


Cobalt0s (approved):
LGTM

PR #3096, opened 2026-06-17
fix(hubspot): omit empty properties param for marketing campaign reads

anushat05:
### Notes
When `id` is the only requested field on a marketing-campaigns read it is filtered out of the `properties=` query param (the Campaigns API rejects `id` there), leaving the param empty. HubSpot then rejects the request with `400 Forbidden properties: []`. Skip the param entirely when the filtered field list is empty, the response still returns the record's top-level fields, including `id`.

#### Integration 
Screenshot 2026-06-16 at 17 19 35

#### Before the change
Screenshot 2026-06-16 at 17 20 13
Screenshot 2026-06-16 at 17 19 45

#### After the change
Screenshot 2026-06-16 at 17 33 28
Screenshot 2026-06-16 at 17 33 13

### DIFF
--- a/providers/hubspot/read.go
+++ b/providers/hubspot/read.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


providers/hubspot/read.go
```diff
@@ -242,19 +243,19 @@ func (c *Connector) buildMarketingReadURL(
 			// top level of the response, not a campaign property. Drop it from the
 			// requested properties; the GUID still comes back at the top level and is
 			// surfaced into Fields by the FlattenNestedFields record transformer.
-			filtered := make([]string, 0, len(fields))
```
RajatPawar, line 245: minor: if it's not buggy, I wouldn't change it i.e. especially for hubspot, I would make the smallest possible change required

RajatPawar (approved):

providers/hubspot/read.go
anushat05: Kept it as `slices.DeleteFunc` the loop got touched only because the new empty-properties guard pushed this block over the nestif limit and failed CI. Also worth noting this loop isn't long-standing HubSpot code; I added it in the previous PR (#3092), so it's not really a risky change. Happy to revert to the loop + //nolint:nestif if you'd still prefer minimal, but this fixes the complexity for real rather than suppressing it.

PR #3103, opened 2026-06-19
[CON-3261] feat: add servicenowPKCE connector

Jkarage:
As recommended here: https://linear.app/ampersand/issue/CON-3261/research-and-implement-new-servicenow-oauth-apps

The private apps in serviceNow use the Authorizatin Code Grant, by default. For public apps its necessary to use Authorization code with PKCE grant. This adds a servicenow connector that supports Authorization code grant with PKCE.
I have tested this with the localhost proxy, works as expected.

### GET
Screenshot 2026-06-19 at 18 44 19

### POST
Screenshot 2026-06-18 at 18 34 32

### PUT
Screenshot 2026-06-18 at 18 36 06

### DIFF
--- a/providers/serviceNow.go
+++ b/providers/serviceNow.go
@@ -1,6 +1,9 @@
 package providers
 
-const ServiceNow Provider = "serviceNow"
+const (
+	ServiceNow     Provider = "serviceNow"
+	ServiceNowPKCE Provider = "serviceNowPKCE"
+)
 
 func init() {
 	// ServiceNow configuration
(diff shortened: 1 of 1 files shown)


providers/serviceNow.go
```diff
@@ -55,4 +58,57 @@ func init() {
...
+				Operators: SearchOperators{
+					Equals: false,
+				},
+			},
+		},
+		Metadata: &ProviderMetadata{
+			Input: []MetadataItemInput{
+				{
+					Name:        "workspace",
+					DisplayName: "Instance",
```
RajatPawar, line 110: Can you please use prompt & docs to indicate how to collect this?

RajatPawar (approved):

providers/serviceNow.go
Jkarage: I have added the Prompt: 

`Please provide your instance Name — for example, dev1234 if your instance base URL: https://dev1234.service-now.com.`

I couldnt find a servicenow link explaining how one can get instance name. So I have not added docs.

PR #3110, opened 2026-06-24
[CON-3317] feat(square): add proxy 

immdipu:
# Configuration
OAuth

# Conventions
- [ ] All required metadata variables are templated (`{{.var}}`) and defined in `ProviderInfo.Metadata`

### GET
URL: http://localhost:4444/v2/payments
image

### POST
URL: http://localhost:4444/v2/payments
image

### DELETE
URL: http://localhost:4444/v2/merchants/custom-attribute-definitions/oerfoiasjofas
image

## Pagination
Procore uses `cursor` and `limit` for pagination.

First Page
image

Next Page
image

# Logo & Icons

## Regular Icon
image

## Dark Icon
image

## Regular Logo

image

## Dark Logo

image

### DIFF
--- a/providers/square.go
+++ b/providers/square.go
@@ -1,0 +1,95 @@
+package providers
+
+const (
+	Square        Provider = "square"
+	SquareSandbox Provider = "squareSandbox"
+)
+
+//nolint:funlen
+func init() {
+	SetInfo(Square, ProviderInfo{
+		DisplayName: "Square",
+		AuthType:    Oauth2,
+		BaseURL:     "https://connect.squareup.com",
+		Oauth2Opts: &Oauth2Opts{
+			GrantType: AuthorizationCode,
+			AuthURL:   "https://connect.squareup.com/oauth2/authorize",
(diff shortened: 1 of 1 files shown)


providers/square.go
```diff
@@ -0,0 +1,95 @@
...
+func init() {
+	SetInfo(Square, ProviderInfo{
+		DisplayName: "Square",
+		AuthType:    Oauth2,
+		BaseURL:     "https://connect.squareup.com",
+		Oauth2Opts: &Oauth2Opts{
+			GrantType: AuthorizationCode,
+			AuthURL:   "https://connect.squareup.com/oauth2/authorize",
+			AuthURLParams: map[string]string{
+				"session": "true",
```
Cobalt0s, line 21: Could you add a comment explaining why this is required?

Cobalt0s (approved):
LGTM

providers/square.go
immdipu: sure. i've added it

PR #3111, opened 2026-06-24
[CON-3287] feat(BreezyHR): Write Delete Implementation

Fullstack-Hero:
## Summary
This PR adds Write and Delete support to the Breezy HR deep connector (on top of existing Read and metadata), using Breezy’s REST API with API key auth and required company_id input metadata. Create/update and single-record delete are implemented only where the upstream API supports them; all other objects remain read-only.

### Write support

- HTTP write handlers for positions only: create via POST /v3/company/{company_id}/positions, update via PUT /v3/company/{company_id}/position/{position_id}.
- Record ID is extracted from the response _id field via jsonquery; update uses RecordId in the URL path.
- API doc links are included in write.go for create and update payloads.

### Delete support

- **Delete (`positions` only):** Soft delete — archives the position via `PUT /v3/company/{company_id}/position/{position_id}/state` with `{"state":"archived"}`.
- Breezy does not provide an HTTP `DELETE` for positions; the connector Delete operation maps to archive.
- Archived positions no longer appear in the default published positions list.
## Integration tests

- Restructured test/breezy/ to match the FastSpring layout: per-object folders with metadata/, read/, and write-delete/ where applicable.
- Removed monolithic test/breezy/metadata/main.go and test/breezy/read/main.go.

## Non-standard provider behavior

- Path naming: Collection create uses plural positions; resource update/delete uses singular position/{id} (documented in paths.go and metadata/README.md).
- Delete is archive: Connector delete does not remove the position record; it sets state to archived so the job no longer appears in the default published positions list.
- Company scoping: Write and delete require company_id from connector metadata (same as read).
- _id in write payload: _id is stripped from the request body on write; updates rely on RecordId in the URL.

## Objects supported
Capability | Objects
-- | --
Read | companies, positions, pipelines, categories, departments, questionnaires, templates
Write | positions (create + update)
Delete | positions (archive via state update)

## Testing

- Unit tests for write and delete (mocked HTTP), plus existing read and metadata tests.
- Integration tests under test/breezy/ (per-object metadata, read, and write-delete for positions).

```bash
# Unit tests
go test ./providers/breezy -v

go test ./providers/breezy -cover

# Integration – write-delete
go run ./test/breezy/positions/write-delete
```

### Live Tests
Screenshot 2026-06-24 at 7 06 23 PM
Screenshot 2026-06-24 at 7 06 41 PM
Screenshot 2026-06-24 at 7 06 54 PM
Screenshot 2026-06-24 at 7 07 49 PM
Screenshot 2026-06-24 at 7 07 57 PM
Screenshot 2026-06-24 at 7 08 07 PM
Screenshot 2026-06-24 at 7 08 17 PM


test/breezy/positions/write-delete/main.go
```diff
@@ -0,0 +1,217 @@
...
+	defer func() {
+		if positionID == "" {
+			return
+		}
+		if err := archivePosition(ctx, conn, positionID); err != nil {
+			slog.Warn("cleanup archive failed", "position_id", positionID, "error", err)
+		}
+	}()
+
+	slog.Info("=== publish draft position (required for default positions list) ===")
```
Cobalt0s: Maybe it would be possible to use `ValidateCreateUpdateDelete` but then modify `CRUDTestSuite` to allow some custom methods to run before major steps.

For example BeforeCreate, BeforeUpdate, BeforeDelete. 
Currently `PreprocessUpdatePayload` is kind of a `BeforeUpdate` which takes `*common.WriteResult` to conditionally create the Update Payload for the Update step.

I think it would be interesting to enhance the `ValidateCreateUpdateDelete`. If you think that current `testscenario` is confusing let me know, we can work on improving it.

test/utils/testscenario/crud.go
```diff
@@ -49,6 +49,9 @@ type CRUDTestSuite struct {
 	// ValidateUpdatedFields optionally verifies that updates took effect correctly.
 	// Use this when UpdatedFields is insufficient — for example, when validating nested arrays or objects.
 	ValidateUpdatedFields func(record map[string]any)
+
+	// PostCreate runs after create succeeds and before the first read/search. Optional.
+	PostCreate func(ctx context.Context, conn ConnectorCRUD, createResult *common.WriteResult) error
```
Cobalt0s, line 54: Oh, I see you already thought of it.
Why not use it?

test/breezy/state.go
```diff
@@ -0,0 +1,59 @@
...
+
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return fmt.Errorf("failed to set position state %q: HTTP %d", state, resp.StatusCode)
+	}
+
+	return nil
+}
+
+// PublishPosition moves a draft position to published so it appears in the default positions list.
+func PublishPosition(ctx context.Context, conn *breezy.Connector, positionID string) error {
```
Cobalt0s, line 75: Ok, so this method is exposed and used by the live test. This is fine in my opinion. You should write down the documentation explaining why this connector has this method. What is important is that it is not a standard public method which implements common interfaces.

Cobalt0s (approved):
LGTM!

The PR description has a contradicotry statement:
> HTTP delete for positions only. Breezy has no DELETE endpoint for positions ...

I checked the implementation and more comments that you left across the files and everything makes sense. The Write/Delete supports only poisitions and Delete is rather a "soft" delete which marks position as archived. This would be good to inlcude in customer facing docs.

If there are no custom fields to worry about then this conenctor should be good for testing and shipping.

test/utils/testscenario/crud.go
Fullstack-Hero: Done. 
The write-delete test now uses PostCreate to call PublishPosition after create. We briefly had a manual step-by-step version for debugging live output; switched back to the shared scenario helper.

Fullstack-Hero:
> LGTM!
> 
> The PR description has a contradicotry statement:
> 
> > HTTP delete for positions only. Breezy has no DELETE endpoint for positions ...
> 
> I checked the implementation and more comments that you left across the files and everything makes sense. The Write/Delete supports only poisitions and Delete is rather a "soft" delete which marks position as archived. This would be good to inlcude in customer facing docs.
> 
> If there are no custom fields to worry about then this conenctor should be good for testing and shipping.

Done ✅.
Updated ✔️

PR #3125, opened 2026-06-30
[CON-3289] feat(breezy): Enable read and write

Fullstack-Hero:
## Description
All screenshots are attached on the ticket.


Fullstack-Hero:
@Cobalt0s what we should do for this connector now?
Everything is done it's just now how we want to manage the recurring session expires.

Cobalt0s (approved):
Indeed, it is good to merge.

PR #3131, opened 2026-07-02
[CON-3396] feat: add jobber subscribe action

Jkarage:
## Summary

Adds the Subscribe action to the Jobber connector: full `connectors.SubscribeConnector`
implementation (Subscribe / UpdateSubscription / DeleteSubscription), webhook HMAC
verification, subscription event parsing, and `GetRecordsByIds` fetch back for
hydrating Jobber's ID only webhook payloads.

Also bumps the API version to `2026-05-12` (the previous `2025-01-20` was
past end-of-life and flagged for removal the API response). 

## Testing
Screenshot 2026-07-02 at 17 08 31

Screenshot 2026-07-02 at 17 08 46


providers/jobber/graphql/query_timeSheetEntries.graphql
```diff
@@ -316,7 +316,6 @@ query TimeSheetEntryconnection{
       startAt
       ticking
       updatedAt
-      visitDurationTotal
```
jlimatampersand, line 336: what is the reason behind this?

providers/jobber/subscribe.go
```diff
@@ -0,0 +1,644 @@
...
+
+// resolveSubscriptionTopics translates the requested object events into
+// Jobber webhook topics, keyed by topic with the owning object as value.
+func resolveSubscriptionTopics(
+	events map[common.ObjectName]common.ObjectEvents,
+) (map[string]common.ObjectName, error) {
+	topics := make(map[string]common.ObjectName)
+
+	for obj, objEvents := range events {
+		root, supported := objectTopicRoot[obj]
```
jlimatampersand, line 567: Does this root object list match the `batchReadableObjects`?

Basically, the objects that we subscribe to should work end-to-end, so if any of them are going to fail, they should be on the same list. 

Can you consolidate it? Or at least the list you use here at Subscribe time should be more strict set, and ideally it should be the intersection of it. 

Can you confirm the list?

providers/jobber/subscribe.go
```diff
@@ -0,0 +1,644 @@
...
+	if len(prev.Endpoints) == 0 {
+		return nil, errMissingStoredEndpoints
+	}
+
+	return prev, nil
+}
+
+// resolveSubscriptionTopics translates the requested object events into
+// Jobber webhook topics, keyed by topic with the owning object as value.
+func resolveSubscriptionTopics(
```
jlimatampersand, line 561: Can you explain this? 

Keyed by topic value as object name. 

Can you give some example of topic names so that make it little clear?

providers/jobber/support.go
```diff
@@ -15,41 +15,41 @@ func supportedOperations() components.EndpointRegistryInput {
...
-		"vehicles",
-		"visits",
+		objectTimeSheetEntries,
+		objectUsers,
+		objectVehicles,
+		objectVisits,
 	}
 
 	writeSupport := []string{
-		"clients",
```
jlimatampersand: seems like this list changed. Can you add docs related to this?

jlimatampersand (approved):
Looks mostly good, but there are some comments I left. Please resolbe before merging.

providers/jobber/graphql/query_timeSheetEntries.graphql
Jkarage: This field used to exist when the query was first written, but Jobber has since removed it from their API entirely. It's gone from all the versions they currently serve, including the old one we were pinned to. Because GraphQL rejects the whole query if any field is unknown, this single field was breaking every timeSheetEntries read. I removed it and verified the read works against the live API now.

Jkarage: I'm adding a comment on the apiVersion, so when someone wants to update it, they should double-check that the schemas and all the queries still work as expected.

providers/jobber/subscribe.go
Jkarage: Yeah confirmed and consolidated the list. Thanks.

providers/jobber/subscribe.go
Jkarage: Topic names are Jobber's webhook identifiers in the form {OBJECT}_{EVENT}, and the function flattens the per-object request into a per topic map. For example, clients with create+update plus quotes with create and the QUOTE_SENT pass-through resolve to:

"CLIENT_CREATE": "clients"
"CLIENT_UPDATE": "clients"
"QUOTE_CREATE":  "quotes"
"QUOTE_SENT":    "quotes"

It's keyed by topic because Jobber creates one webhook endpoint per topic, so create, delete, and the update diff all operate on topic sets. The object is kept as the value so the created endpoints can be grouped back into the per-object result the framework stores.

providers/jobber/support.go
Jkarage: Not changed, I have just used the constant clientsObject in its place. to avoid the linter complaining.

PR #3135, opened 2026-07-02
fix(common): treat provider 404 as non-retryable (ENG-4099)

RajatPawar:
## What
`common.InterpretError` mapped HTTP 404 → `ErrRetryable`. For connectors whose read `ErrorHandler` is `InterpretError` (e.g. GitHub `/issues`, superSend), a permanent 404 (resource deleted / access revoked) was retried until the Temporal retry budget was exhausted, then the read schedule **paused** — and a manual unpause just re-paused it.

## Change
404 now maps to `ErrBadRequest` + `ErrNotFound` (non-retryable), matching `interpreter.DefaultStatusCodeMappingToErr`, which already treats 404 that way. The two status mappers now agree.

## Impact
All connectors using `InterpretError`: a 404 fails fast (non-retryable) instead of retry-then-pause. Sibling of the Google DWD `FAILED_PRECONDITION` classification issue (ENG-4109).

Sentry: BACKEND-WP, WN (GitHub), Y3 (superSend). Fixes ENG-4099.

### DIFF
--- a/common/error.go
+++ b/common/error.go
@@ -32,8 +32,9 @@ func InterpretError(res *http.Response, body []byte) error {
 		// Forbidden, not retryable
 		return NewHTTPError(res.StatusCode, body, headers, createError(ErrForbidden))
 	case http.StatusNotFound:
-		// Semantics are debatable (temporarily missing vs. permanently gone), but for now treat this as a retryable error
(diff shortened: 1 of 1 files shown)


RajatPawar:
Closing — team's call is that retrying 404s is acceptable (semantics are debatable, temporarily-missing vs permanently-gone, as the existing comment notes). Keeping current behavior.

PR #3136, opened 2026-07-02
[CON-3310] feat(stripe): Nested expanding of fields

Cobalt0s:
# Description

* `readhelper.SelectFields` now applies to all objects without exceptions. (allows selecting deeply nested objects/arrays/primitives)
* `customFields` are also attached to all applicable objects. This behavior is unchanged, and the unit test for custom fields still pass.
* `MakeExpandableQueryParam` can now construct complex, deeply nested `expand[]` query parameters. See the detailed docs for the algorithm.
  * `query-param_test.go` contains unit tests that show which query params are generated from a JSON path.

# Live Tests

## Checkout Sessions

Checkout Sessions behaves as before.
This does top root level expanding.
image

## Balance Transactions

Deeply nested fields are now supported:
image

Usually, when the `payment_intent` is not expanded, the raw payload looks like this, with `payment_intent` set to an ID instead of an object:
image

### DIFF
--- a/providers/stripe/metadata/query-params.go
+++ b/providers/stripe/metadata/query-params.go
@@ -1,0 +1,178 @@
+package metadata
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/amp-labs/connectors/common/naming"
+	"github.com/amp-labs/connectors/common/readhelper"
+)
+
+/*
+MakeExpandableQueryParam builds a Stripe `expand[]` query parameter for a field in JSON path format.
+Regular fields can be passed and result in empty query parameter.
+
+The connector only has partial knowledge of the nested object graph, so it can
+only infer the deepest valid expansion path from the registry of known expandable
+fields. When the requested path includes unsupported segments, the function trims
+the query at the last resolvable expandable object.
+
+Example:
+(balance_transactions, $['source']['payment_intent']['customer']['id'])
+
+	(balance_transactions, source) -> can expand
+	(source, payment_intent)       -> unknown
+	(payment_intent, customer)     -> can expand
+	(customer, id)                 -> unknown
+
+Therefore, the expand query must include everything up to customer:
+
+	data.source.payment_intent.customer
+
+Stripe expansion has a maximum nesting depth of 4 levels, so this helper stops
+as soon as the path would exceed that limit. If the requested field path cannot
+be matched to at least one expandable field, the function returns an empty string.
+
+Docs:
+https://docs.stripe.com/expand#multiple-levels
+
+Example error returned by Stripe when the limit is exceeded:
+
+	{
+	    "error": {
+	        "code": "property_expansion_max_depth",
+	        "message": "You cannot expand more than 4 levels of a property.
+						Property: data.source.payment_intent.customer.discount",
+	        "request_log_url": "",
+	        "type": "invalid_request_error"
+	    }
+	}
+*/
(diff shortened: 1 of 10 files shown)


Cobalt0s:
* **#3136** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3136?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3136?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

providers/stripe/metadata/query-params.go
```diff
@@ -0,0 +1,178 @@
...
+	    "error": {
+	        "code": "property_expansion_max_depth",
+	        "message": "You cannot expand more than 4 levels of a property.
+						Property: data.source.payment_intent.customer.discount",
+	        "request_log_url": "",
+	        "type": "invalid_request_error"
+	    }
+	}
+*/
+func MakeExpandableQueryParam(objectName, field string) string {
```
AmandaYao00, line 51: cool algorithm!

providers/stripe/metadata/query-params.go
```diff
@@ -0,0 +1,178 @@
...
+	// We construct the query from the root and stop before exceeding the 4-level limit.
+	const maxPropertyLevel = 4
+
+	queryParam := "data"
+
+	var queryParamSb108 strings.Builder
+
+	for i := 0; i <= lastExpandablePair; i++ {
+		if i+1 == maxPropertyLevel {
+			// Returning an empty string is safer than emitting an invalid expand path.
```
AmandaYao00: out of curiosity, why is it safer to return no query param than the last valid path?

providers/stripe/metadata/query-params.go
```diff
@@ -0,0 +1,178 @@
...
+		lastExpandablePair = len(pairs) - 1
+	}
+
+	// Stripe list responses require the `data.` prefix in expand paths.
+	// We construct the query from the root and stop before exceeding the 4-level limit.
+	const maxPropertyLevel = 4
+
+	queryParam := "data"
+
+	var queryParamSb108 strings.Builder
```
AmandaYao00: out of curiosity, why is this named "queryParamSb108" haha

providers/stripe/read.go
AmandaYao00: nit: nested expansion now supports all fields, not just LineItems

AmandaYao00 (approved):

providers/stripe/metadata/query-params.go
Cobalt0s: I use LLMs primarily for comments, so that could be it. I should change the phrasing. There’s no real danger, to be honest.

During development I thought the best approach would be to return as much as possible, to truncate at level 4. 
In practice, the field requested by the customer will never be found in `Raw` by definition of us truncating
So `ReadResultRow.Fields` will be empty for that field. You can treat this as connector-side field validation given provider level 4 constraint.

providers/stripe/metadata/query-params.go
Cobalt0s: I did use raw string concats and this was the result of llm fixing the comments... and I overlooked that it changed this part. Which is the right change for string building.

Cobalt0s: My bad :hot_face:

PR #3160, opened 2026-07-08
[ENG-4096] fix(salesforce): retry event relay RUN with namespaced full name

jlimatampersand:
New external client with namespace org has different convention in naming the event realy config and it returns error. 

This PR is capturing that error and retrying with the naming convention recommended from the error message. 

In a namespaced org (scratch/packaging orgs), Salesforce prepends the org
namespace to the full name of metadata it creates, so an EventRelayConfig
created as "amp_x" becomes "<ns>__amp_x". The follow-up PATCH that sets the
relay to the RUN state echoed the un-namespaced full name and was rejected
with "Full name amp_x does not match the full name <ns>__amp_x".

Detect that mismatch, extract the expected namespaced full name from the
error, and retry the PATCH once with it (persisting the corrected full name
back onto the config). Mirrors the existing isDuplicateDeveloperName recovery
pattern.

### DIFF
--- a/providers/salesforce/eventRelayNamespace_test.go
+++ b/providers/salesforce/eventRelayNamespace_test.go
@@ -1,0 +1,127 @@
+package salesforce
+
+import (
+	"errors"
+	"net/http"
+	"testing"
+
+	"github.com/amp-labs/connectors/common"
+	"gotest.tools/v3/assert"
+)
+
+func TestFullNameMismatch(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name         string
+		err          error
+		fullName     string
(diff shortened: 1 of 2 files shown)


jlimatampersand:
* **#3160** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3160?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3160?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

RajatPawar (approved):

PR #3167, opened 2026-07-09
fix(square): treat missing records key as empty page

immdipu:
Reading a paginated Square object (e.g. `payments`) errored with `key not found: [payments]` on an empty page, because Square omits the records key entirely (body is just `{}`) and we used `ArrayRequired`.

Fix: extract records with `ArrayOptional` so an empty page returns 0 rows and `Done: true`.

### DIFF
--- a/providers/square/read.go
+++ b/providers/square/read.go
@@ -60,11 +60,20 @@ func (c *Connector) parseReadResponse(
 
 	return common.ParseResult(
 		resp,
-		common.MakeRecordsFunc(cfg.responseKey),
+		makeRecordsFunc(cfg.responseKey),
 		makeNextRecordsURL(),
 		readhelper.MakeMarshaledDataFuncWithId(nil, readhelper.NewIdField("id")),
 		params.Fields,
 	)
+}
+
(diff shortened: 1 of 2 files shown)


providers/square/read.go
```diff
@@ -60,13 +60,22 @@ func (c *Connector) parseReadResponse(
...
 		makeNextRecordsURL(),
 		readhelper.MakeMarshaledDataFuncWithId(nil, readhelper.NewIdField("id")),
 		params.Fields,
 	)
 }
 
+// makeRecordsFunc extracts the records array under responseKey. Square omits the
+// key entirely on an empty last page (the body is just `{}`), so we look it up
+// optionally and treat a missing key as zero records rather than an error.
+func makeRecordsFunc(responseKey string) common.NodeRecordsFunc {
```
Cobalt0s, line 73: Probably it is worth adding `common.MakeRecordsOptionalFunc` instead of the connector local func.

I am surprised this hasn't come up before. We have something like this for `RecordsFunc` but not for `NodeRecordsFunc`.

Cobalt0s (approved):
LGTM

PR #3184, opened 2026-07-13
[ENG-4086] feat(providers): multi-step custom auth registry + Microsoft pilot

RajatPawar:
Executable side of multi-step custom auth. The catalog (openapi) only declares `CustomOpts.MultiStep`; the step handlers live here in a **compiled-in registry**, so they never cross JSON/gob (which would drop func fields).

### `customauth.go`
- `AuthContext` — pure, serializable flow data (`ConsumerInputs`, `ProviderInputs`, `Secrets`, `Metadata`, server-injected read-only `System`). `Flatten()` precedence, low→high: System, Metadata, ProviderInputs, ConsumerInputs, Secrets.
- `HTTPStep` / `RedirectStep` / `AuthStep`; handlers build the full request and return updated state by value.
- `CustomAuthFlow` + registry (`RegisterCustomAuthFlow` / `CustomAuthFlowFor`); `HasSteps` / `HasRedirect`.
- `ExtractJSONSecrets` / `ExtractQueryParamsSecrets` helpers for the simple cases.

### `microsoftClientCredentials.go`
Converted from `oauth2`/client-credentials to the custom multi-step flow: redirect to admin consent → capture `tenant` from the callback → exchange client credentials for a Graph token. Refresh re-runs the exchange. Reuses the `providers/microsoft` connector for data operations.

### Also
- Regenerated `types.gen.go` for the new catalog fields (fieldType/options, providerInputs, multiStep); matches once the openapi PR merges to main.
- Tests: `Flatten` precedence, extract helpers, Microsoft handlers, flow registration.


providers/customauth.go
```diff
@@ -0,0 +1,195 @@
...
+	// ConsumerInputs are what the consumer submitted (username, password, ...).
+	// Set once at flow start; not modified by steps.
+	ConsumerInputs map[string]string `json:"consumerInputs,omitempty"`
+
+	// ProviderInputs are what the builder configured on their provider app
+	// (clientId, clientSecret, subscription keys). Set once at flow start.
+	ProviderInputs map[string]string `json:"providerInputs,omitempty"`
+
+	// Secrets are credentials accumulated by steps (accessToken, sessionId, ...).
+	// Persisted to the connection (encrypted) when the flow completes.
```
jlimatampersand: The concept of Connection here has some Ampersand server context. So we can say 

`Persisted to the Connection table in the database in the Ampersand server.` 

So it is clear to the contributors. Same for pretty much all of the other terms and comments, if there is anything that requires server/system context, worth elaborating on them a bit more.

providers/customauth.go
```diff
@@ -0,0 +1,195 @@
...
+	if c.System == nil {
+		c.System = map[string]string{}
+	}
+
+	return c
+}
+
+// Flatten merges every sub-map into one for template resolution. Precedence,
+// lowest to highest: System, Metadata, ProviderInputs, ConsumerInputs, Secrets.
+func (c AuthContext) Flatten() map[string]string {
```
jlimatampersand, line 75: Is there a need for collision handling here?

providers/customauth.go
```diff
@@ -0,0 +1,195 @@
...
+
+	return out
+}
+
+// HTTPStep is a server-side HTTP call that acquires or refreshes credentials.
+// BuildRequest constructs the full request to send; ParseResponse extracts
+// results. State is passed by value and returned explicitly (no pointer mutation).
+type HTTPStep struct {
+	// BuildRequest builds the request to send. Required.
+	BuildRequest func(ctx context.Context, state AuthContext) (AuthContext, *http.Request, error)
```
jlimatampersand, line 91: Is AuthContext something we can expand later for each provider if they have any provider-specific context that is needed for it to be custom auth?

I can see that the fields in the AuthContext are all just maps, so I can see this will fit most of the needs, if not all. But just want to confirm if this can be extended further if there is a need for it. Otherwise, I would do below. 
```
type RequestParam struct {
   Context AuthContext

   // later more custom params
   CustomParamForProviderA *AParams
}
```

But you have better context here, so your call here. Same for. response.

providers/customauth.go
```diff
@@ -0,0 +1,195 @@
...
+
+	// ParseResponse extracts values from the response into the returned state. Required.
+	ParseResponse func(ctx context.Context, state AuthContext, resp *http.Response) (AuthContext, error)
+}
+
+// RedirectStep sends the user's browser to a URL and resumes when the provider
+// redirects back to the Ampersand callback.
+type RedirectStep struct {
+	// TimeoutSeconds bounds how long to wait for the callback. 0 = server default.
+	TimeoutSeconds int
```
jlimatampersand, line 101: Why not `time.Duration`?

providers/customauth.go
jlimatampersand, line 91: Same for the return type.

providers/customauth.go
```diff
@@ -0,0 +1,195 @@
...
+
+	// BuildURL returns the URL to redirect the browser to. Required.
+	BuildURL func(ctx context.Context, state AuthContext) (AuthContext, string, error)
+
+	// ParseCallback extracts values from the provider's callback request. Required.
+	ParseCallback func(ctx context.Context, state AuthContext, callback *http.Request) (AuthContext, error)
+}
+
+// AuthStep is one step of a flow. Exactly one of HTTP or Redirect is set.
+type AuthStep struct {
```
jlimatampersand, line 111: Do we have some enforcing or validation to have only 1? Maybe validator validation? or `.Validate` method to make sure only 1 is set.

providers/customauth.go
```diff
@@ -0,0 +1,195 @@
...
+
+// HasSteps reports whether this provider uses a multi-step custom auth flow
+// (driven via /custom-auth/connect) rather than static header/query injection.
+func (o *CustomAuthOpts) HasSteps() bool {
+	return o != nil && o.MultiStep
+}
+
+// ExtractJSONSecrets returns a ParseResponse handler that decodes the JSON body
+// and copies mapped response fields (responseKey -> secretKey) into Secrets.
+func ExtractJSONSecrets(
```
jlimatampersand: `GetSecretParser`?

ExtractJSONSecrets sounds like a function that actually extracts. But this is a function constructor

providers/customauth.go
```diff
@@ -0,0 +1,195 @@
...
+			}
+		}
+
+		return state, nil
+	}
+}
+
+// ExtractQueryParamsSecrets returns a ParseCallback handler that copies mapped
+// callback query params (paramName -> secretKey) into Secrets.
+func ExtractQueryParamsSecrets(
```
jlimatampersand: Same here.

providers/microsoftAdminConsent.go
```diff
@@ -0,0 +1,157 @@
...
+// credentials) against Microsoft Graph, reusing the providers/microsoft
+// connector. Rather than requiring the customer to enter their tenant and grant
+// admin consent out-of-band, it uses the multi-step custom auth flow: redirect
+// the admin to the consent screen, capture the tenant Microsoft returns, then
+// exchange the app's client credentials for a Graph token.
+const MicrosoftAdminConsent Provider = "microsoftAdminConsent"
+
+func init() {
+	// Token exchange is reused for both the second connect step and refresh.
+	tokenStep := HTTPStep{
```
jlimatampersand, line 22: Maybe beyond scope of this PR. 

But worth setting this custom auth initialization logic and provider info registration logic separately outside of this init function and just call here in init with fewer lines?

eg

```
func init() {
  SetInfo(MicrosoftAdminConsent, providerInfo)
  RegisterCustomAutFlow(MicrosoftAdminConsent, customAuthFlow)
}
```

jlimatampersand (approved):
Left minor comments.

providers/customauth.go
RajatPawar: The last wins, and added docs on the precedence

providers/customauth.go
RajatPawar: The maps are deliberately generic, so we should be covered!

providers/customauth.go
RajatPawar: There is serialization involved, so I'd rather keep it a simple time which is easy to deal with through various cycles of marshaling/unmarshaling

providers/customauth.go
RajatPawar: Good call, added!

providers/microsoftAdminConsent.go
RajatPawar: Good idea. I'll decouple it and leave it as a follow up. We'll discover patterns as we implement more of these, and then abstract it using a pattern like yours

PR #3191, opened 2026-07-14
[ENG-4125] feat(connectwise): Support Custom Fields

Cobalt0s:
# Description

- ConnectWise batch processing now supports paginated listing of all records. When the API returns a `next` link header, the parallel worker follows it and continues listing records until no `next` link remains.
- `ListObjectMetadata` now calls `requestCustomFields` to enrich metadata from the static file. If an object does not support custom fields, the method returns an empty field map. Otherwise, the flow is:
  1. Read a sample record, assuming at least one record exists.
  2. Use the response’s `customFields` array, which includes the full set of custom fields even when values are `null`.
  3. Collect the custom field IDs.
  4. Request **custom field definitions** using an `id in (...)` condition.
  5. Build `ObjectMetadata` with the derived value type, select options, and display name.
- Custom fields do not have a stable human-readable identifier, so the connector exposes two field names for each field:
  - the caption, which is user-facing but mutable in the ConnectWise UI;
  - `customField<id>`, which is stable and does not change.
- `ReadRecordsById` and `Read` now inherit custom fields into `Fields`. This means users can reference a custom field by either naming format, and the field will still be extracted correctly. The `Raw` value in `ReadResult` is unchanged.

# Live Tests

## Read

Read with custom fields requested. `Raw` is omitted because it is large:

image

`Raw` is returned here as well, and custom fields are still present unchanged:

image

## ListObjectMetadata

image

Field `hobby` is exposed in two formats:

image

image

## ReadRecordsById

The script creates records and then requests the custom field. Even though the field value is `null`, it is still elevated and returned under `Fields`:

image


Cobalt0s:
* **#3194** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3194?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3191** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3191?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3191?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

providers/connectwise/custom.go
```diff
@@ -0,0 +1,285 @@
...
+
+	if len(values) == 0 {
+		return nil
+	}
+
+	return values
+}
+
+func (f modelCustomField) makeFieldName1() string {
+	return fmt.Sprintf("customField%v", strconv.Itoa(f.Id))
```
Cobalt0s, line 272: First custom field `Field name` format.

providers/connectwise/delete_test.go
```diff
@@ -36,6 +36,7 @@ func TestDelete(t *testing.T) { // nolint:funlen,cyclop
 				If: mockcond.And{
 					mockcond.MethodDELETE(),
 					mockcond.Path("/v4_6_release/apis/3.0/company/contacts/57919"),
+					mockcond.Header(http.Header{"ClientId": []string{"dummy-client-id"}}),
```
Cobalt0s, line 39: Forgot to add it. The Connector must send this header for all requests.
Adding this to engrave this requirement.

providers/connectwise/metadata.go
```diff
@@ -0,0 +1,54 @@
...
+func (c *Connector) ListObjectMetadata(
+	ctx context.Context, objectNames []string,
+) (*common.ListObjectMetadataResult, error) {
+	metadataResult, err := metadata.Schemas.Select(c.Module(), objectNames)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, objectName := range objectNames {
+		fields, err := c.requestCustomFields(ctx, objectName)
```
Cobalt0s, line 19: Custom fields are fetched for objects that support such fields. The metadata for this object will be enriched.

Other objects will keep relying on static file without enrichment, aka no operational change for them.

providers/connectwise/read-by-ids.go
```diff
@@ -30,8 +31,16 @@ func (c *Connector) GetRecordsByIds(ctx context.Context,
...
 
-	marshaler := readhelper.MakeGetMarshaledDataWithId(readhelper.NewIdField("id"))
+	marshaler := readhelper.MakeMarshaledDataFuncWithId(
+		flattenCustomFields(),
+		readhelper.IdFieldQuery{Field: "id"},
+	)
 	uniqueFields := datautils.NewSetFromList(fields).List()
 
-	return marshaler(batchResult, uniqueFields)
+	list, err := jsonquery.Convertor.NodesFromArray(batchResult)
```
Cobalt0s, line 40: In order to reuse `flattenCustomFields` which operates on `ajson.Node` I am covnerting `[]map[string]any` to `[]*ajson.Node`.

As a result `ReadRecordsByIds` acts just like `Read`. See unit test for `read-by-ids.go` file.

providers/connectwise/read.go
```diff
@@ -78,9 +78,12 @@ func (c *Connector) parseReadResponse(
 ) (*common.ReadResult, error) {
 	return common.ParseResult(resp,
 		// "" is used, because root level of JSON is right away an array.
-		common.ExtractRecordsFromPath(""),
+		common.MakeRecordsFunc(""),
 		nextRecordsURL(resp),
-		readhelper.MakeGetMarshaledDataWithId(readhelper.IdFieldQuery{Field: "id"}),
+		readhelper.MakeMarshaledDataFuncWithId(
+			flattenCustomFields(),
```
Cobalt0s, line 84: This stament gives `Read` an ability to select custom fields.

providers/connectwise/custom.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	records, err := common.UnmarshalJSON[readResponses](res)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(*records) == 0 {
+		// There are no records for sampling.
+		return []string{}, nil
```
anushat05, line 146: This treats "no records" the same as "no custom fields." Custom fields are defined at the CW setup level independent of data, so an object with fields defined but zero records (new instance, unused object, permission-filtered view) returns nothing here. The fields silently never appear in metadata/mapping UI. This bites at setup time, which is exactly when users map. The per-record null guarantee doesn't help it assumes a record exists. Consider records-independent discovery via `GET /system/userDefinedFields` or at minimum, log this case distinctly so it's not silently indistinguishable from "no custom fields

providers/connectwise/custom.go
```diff
@@ -0,0 +1,285 @@
...
+	default:
+		return common.ValueTypeOther
+	}
+}
+
+// multiValueOrDefault returns MultiSelect or SingleSelect for list-style fields, and defaultValue otherwise.
+// Number, Percent, Text can all be either primitive data type or Multi/Single Select.
+func multiValueOrDefault(field modelCustomField, defaultValue common.ValueType) common.ValueType {
+	switch field.EntryTypeIdentifier {
+	case "List":
```
anushat05, line 242: List → MultiSelect looks wrong. `is_synced` is `entryTypeIdentifier: "List"` with a single value and `true/false` options, and the CW UI labels it `"List (Drop-down)" i.e. single-select. Mapping to MultiSelect misrepresents the field and could break value handling / value mappings. 

Did we confirm CW's List vs Option semantics?

anushat05 (changes requested):

providers/connectwise/custom.go
Cobalt0s: I understand this. I was going with this approach due to dozens of connectors doing this. As a general rule I never use sampling technique as it is inferior in my opinion to all other approaches. Nonetheless, it is used, so I thought this would be the most robust solution.

I don't think that there are other approaches rather than hardcoding `podId`. I am not sure if they differ among customers. Therefore, I cannot conclude if this approach will scale.

The sampling apprach that I am referring to:
* [Gitlab example](https://github.com/amp-labs/connectors/blob/main/providers/gitlab/handlers.go#L29)
* `NewObjectSchemaProvider` returns ListObjectMetadata by sampling one record. It is used by 59 connectors:
image

PS: Here is an [article on sources of Metadata](https://ampersand.slab.com/posts/sources-for-object-metadata-sl2kdcpi#h9vfg-schema-sources) for ListObjectMetadata feature.

anushat05 (approved):

providers/connectwise/custom.go
Cobalt0s: The mapping is correct. This would be the data issue. The author of this field should've used the `Checkbox` field type.

I am relying on this part of documentation:
Selection_401

Cobalt0s: I also tested the `Checkbox` and it can only accept true/false and never Yes/No as it states in the docs :sweat: 
image

PR #3194, opened 2026-07-14
[ENG-4125] feat(connectwise): Write with Custom Fields

Cobalt0s:
# Description

Custom fields are now supported as native fields for the following write cases:
* Create
image

* Update
image

* Update with operations
image

Custom Field must be of format `customField<num>`:
image

# Live Tests

### Write/Delete [Contacts](https://github.com/amp-labs/connectors/blob/3fee124b14ec6241ee6e2b0567f97ac3c98e84cc/test/connectwise/write-delete/contacts/main.go#L34)
image

### DIFF
--- a/providers/connectwise/write.go
+++ b/providers/connectwise/write.go
@@ -6,8 +6,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strconv"
+	"strings"
 
 	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/urlbuilder"
 	"github.com/amp-labs/connectors/internal/jsonquery"
 )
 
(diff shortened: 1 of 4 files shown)


Cobalt0s:
* **#3194** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3194?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3194?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3191** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3191?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

providers/connectwise/write.go
```diff
@@ -52,6 +40,119 @@ func (c *Connector) buildWriteRequest(ctx context.Context, params common.WritePa
...
+			continue
+		}
+
+		customFields = append(customFields, map[string]any{
+			"id":    fieldId,
+			"value": payload.Value,
+		})
+	}
+
+	result = append(result, patchOperationPayload{
```
anushat05, line 161: This appends replace `/customFields` on every update, even when the patch has no custom fields. So a normal update (e.g. just firstName) sends replace` /customFields []`. If CW treats a replace with an empty array as "set to empty," that wipes all custom fields on the record. Even if it's a harmless no-op, it's an unnecessary op on every write.

Suggest only appending when there's something to write:
```if len(customFields) > 0 {
    result = append(result, patchOperationPayload{
        Op: "replace", Path: "/customFields", Value: customFields,
    })
}```

(Matches the create/PUT path, which already guards with if len(customFields) != 0.)

anushat05 (changes requested):

providers/connectwise/write.go
Cobalt0s: ConnectWise is very difficult provider because of its inconsistency. I know that customFields even with empty array must be set even if your intention is not to deal with custom fields but change something different.

However, let me double check because your intuition is right. I need to validate it.

Cobalt0s: Ok, so double checked the behaviour. Here is how it works:

* [JSON Patch](https://datatracker.ietf.org/doc/html/rfc6902) for objects that have custom fields will fail if custom fields is not set.
image

* When customFields is set to empty array:
image

* Lets say the customFields is non empty array:
image

* Next time we provide different field, the `hobby` custom field is not "wipped out"
image

* Setting and array to empty does not clear `hobby` and `mobile` fields:
image

* Explicitly setting each field to null, wipes them out:
image

anushat05 (approved):

providers/connectwise/write.go
anushat05: Thank you for checking. This helps.

Cobalt0s: No problem!

PR #3196, opened 2026-07-15
[CON-3413] feat(zohoMail): add webhook payload verification and getRecordsById support

Jkarage:
# Zoho Mail webhook subscribe support (Mail + Task)

## Summary
Adds subscription support for the Zoho **Mail** module using the **manual-config webhook** pattern (`SubscribeByAPI: false`).

## Background — why this shape
Zoho Mail has **no programmatic API** to create, update, or delete webhook subscriptions. Outgoing webhooks are configured by hand in the Zoho Mail console (Settings → Integrations → Developer Space → Outgoing Webhooks), for the **Mail** or **Tasks** entity. This rules out a `RegisterSubscribeConnector`/`SubscriptionMaintainerConnector` implementation. What Zoho *does* provide is a signed data, so the connector implements the `WebhookVerifierConnector` and declares `SubscribeByAPI: false`.

## How it works
1. Customer configures the outgoing webhook in the Zoho Mail console. Zoho fires a **first handshake request** carrying the `x-hook-secret` header (once only, it is resent when some configurations are updated(only saw when we change the webhook URL)) and expects a `200`.
2. On every subsequent delivery, the connector's `VerifyWebhookMessage` recomputes `base64(HMAC-SHA256(secret, rawBody))` and compares it to `X-Hook-Signature`.
3. `SubscriptionEventList` parses the payload into a normalized event: `ObjectName` (`messages`/`tasks`), `EventType`, `RecordId`, `EventTimeStampNano`, `Workspace`.
4. Optionally, `GetRecordsByIds` fetches the full record using the composite record id.

## Expectations from the server

- **Capture & persist the `x-hook-secret`.** Zoho sends it only in the header of the *first* request. The server must read it, return `200`, persist it against the connection, and supply it back as the **`zohoMailWebhookSecret`** connection metadata value. The connector reads it from metadata; it does not capture it. Verification is therefore only possible from the second delivery onward.
- **Decode webhook bodies with `json.Decoder.UseNumber` (or otherwise preserve 64-bit integers).** `folderId` exceeds 2^53 and has no string twin; a plain `json.Unmarshal` rounds it and would corrupt the fetch URL.

## RecordIds in GetRecordsById
- **Record ids are composite:** `"<folderId>/<messageId>"` for messages, `"<groupId>/<taskId>"` (or bare `taskId` for personal tasks). Zoho's get record endpoints require the parent (folder/group) id, but `GetRecordsByIds` only receives a flat id list the parent is folded into the id so the fetch URL can be reconstructed. The webhook payload has these, so I believe the server can create these values and send them as expected.

- Messages: folderId / messageId → e.g. 7214453000000008008/1784105487965154200
- Tasks: nameSpaceId (the group id) / entityId (the task id) → e.g. 9000000002014/4000000012345

### DIFF
--- a/providers/zoho/internal/mail/verify.go
+++ b/providers/zoho/internal/mail/verify.go
@@ -1,0 +1,52 @@
+package mail
+
+import (
+	"context"
+	"crypto/hmac"
+	"crypto/sha256"
+	"encoding/base64"
+	"errors"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/internal/httpkit"
+)
+
+// Zoho Mail uses the manual-config webhook pattern (SubscribeByAPI: false in the
+// catalog): the outgoing webhook is created by hand in the Zoho Mail console
+// (Settings > Integrations > Developer Space > Outgoing Webhooks), and this
+// adapter only verifies deliveries (here), parses them into events
+// (subscriptionEvent.go), and fetches full records on demand (records.go). Zoho
+// Mail has no API for the programmatic Subscribe/Update/Delete lifecycle (unlike
+// Zoho CRM), so those operations do not exist here.
+//
+// Docs: https://www.zoho.com/mail/help/dev-platform/webhook.html
+
+// mailHookSignatureHeader is the per-request signature header Zoho Mail attaches
+// to every webhook delivery. Its value is
+// base64(HMAC-SHA256(x-hook-secret, rawRequestBody)).
+const mailHookSignatureHeader = "X-Hook-Signature"
+
+var ErrMissingWebhookSecret = errors.New("zoho mail webhook secret is not set")
+
+// VerifyWebhookMessage validates that a webhook request came from Zoho Mail by
+// recomputing the HMAC-SHA256 signature over the raw request body with the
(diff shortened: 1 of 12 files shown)


providers/zoho/internal/mail/verify.go
```diff
@@ -0,0 +1,52 @@
...
+// signing secret and comparing it (constant-time) to the X-Hook-Signature
+// header.
+func (a *Adapter) VerifyWebhookMessage(
+	_ context.Context, request *common.WebhookRequest, _ *common.VerificationParams,
+) (bool, error) {
+	if a.hookSecret == "" {
+		return false, ErrMissingWebhookSecret
+	}
+
+	got, err := httpkit.ExtractRequiredHeader(request.Headers, mailHookSignatureHeader)
```
jlimatampersand: @Jkarage Let's comment all of the logics here and return true for now. 

We don't currently have means to do the handshake for Zoho webhook secrets.

jlimatampersand (changes requested):
Sorry for the delay. Mostly good, except that verification could be bypassed due to current system limitation. Please ping me on slack, I can approve right away.

jlimatampersand (approved):

PR #3199, opened 2026-07-16
[ENG-4127] feat(connectwise): Contact communication items

Cobalt0s:
# Description

Each communication item on a contact has a category `email`, `phone`, `fax` and the fields exposed are virtual fields `AMPERSAND-defaultEmail`, `AMPERSAND-defaultPhone`, `AMPERSAND-defaultFax`.
Also there are matching fields holding communication type identifier, for example `AMPERSAND-defaultEmailId`.

# Metadata Contacts

image
image

# Read Contacts

image


Cobalt0s:
* **#3203** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3203?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3199** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3199?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3199?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

providers/connectwise/metadata.go
```diff
@@ -51,3 +54,91 @@ func (c *Connector) ListObjectMetadata(
...
+			DisplayName:  item.Description,
+			ValueType:    common.ValueTypeString,
+			ProviderType: "string",
+			ReadOnly:     new(false),
+			IsCustom:     new(false),
+			IsRequired:   new(false),
+		}
+
+		var prefix string
+		if item.EmailFlag {
```
anushat05: These are sequential ifs with no handling when none of the flags is set to a type with no `email`/`fax`/`phone` flag leaves prefix == "", so the field is added as a bare id (e.g. "1") with no `AMPERSAND- `prefix. Use a `switch`/`else-if` and skip (or handle) types with no category flag. 
(Also: if a type ever had multiple flags, the last wins a switch fixes both.)

providers/connectwise/metadata.go
```diff
@@ -51,3 +54,91 @@ func (c *Connector) ListObjectMetadata(
...
+
+		if item.FaxFlag {
+			prefix = virtualFieldContactFax
+		}
+
+		if item.PhoneFlag {
+			prefix = virtualFieldContactPhone
+		}
+
+		objectMetadata.AddFieldMetadata(prefix+item.Id.String(), field)
```
anushat05: Metadata builds the category prefix from the type's flags (EmailFlag…), but read (attachCommunicationItems) builds it from the item's CommunicationType string. 

These must always agree or the metadata key won't match the read key (field shows in the UI but returns empty). Why are we not deriving both from the same signal?

providers/connectwise/communication-types.go
```diff
@@ -0,0 +1,113 @@
...
+		if err != nil {
+			return err
+		}
+
+		identifier := item.Type.Id.String()
+		switch item.CommunicationType {
+		case "Email":
+			fields[virtualFieldContactEmail+identifier] = item.Value
+			if item.DefaultFlag {
+				fields[virtualFieldContactEmailDefault] = identifier
```
anushat05: Two things on `AMPERSAND-email-default`:
1. **Value:** it stores identifier (the type id, "1") rather than the address, so mapping -default returns "1", not the email. Should probably be `item.Value ("emma@work.com")` so the convenience field actually gives the default address.
2. **Multiple defaults:** this writes the same key for every default item, so if a category ever has more than one DefaultFlag == true, it's always last-in-array wins. If CW can return >1 default per category, we'd be asserting a wrong default. Either confirm CW guarantees one default per category, or only set -default when exactly one exists (omit otherwise) so we never send a false default.

providers/connectwise/test/write/tmp.json
```diff
@@ -0,0 +1,34 @@
+{
+  "FieldsMap": {
```
anushat05: Is this file needed?

anushat05 (changes requested):

providers/connectwise/communication-types.go
anushat05: A follow up can you verify how defaults work in ConnectWise? Specifically: is there exactly one default per communication category, i.e. can only one email item have `DefaultFlag == true` (and same for phone/fax), or can multiple be set as default at once?

Trying to confirm whether "the default email/phone/fax" is guaranteed unique.

providers/connectwise/test/write/tmp.json
Cobalt0s: Yeah, this was a mistake, removed.

providers/connectwise/metadata.go
Cobalt0s: I am not sure what you mean by the same signal. But the names returned in Metadata, do match those used by Read.

Metadata:
image

Read:
image


Metadata uses response from `/company/communicationTypes`. The response data is different from the read.
Read includes `communicationItems` as nested property. The structure of that schema differs from the metadata.

In conclusion, the sources and structure of the data differs between Read/Metadata on the Provider level. However, the Connector level bridges both and results in the same `fieldNames`. Check the PR description, live tests section. The names from Metadata script match the Read.Fields. 

NOTE: the fields are always lower cased for all connectors. It is possible that this part is confusing. To be honest, I am not sure why we make it lower case.

providers/connectwise/metadata.go
Cobalt0s: Ah, indeed. I have not thought of the default case, which is obvious with the `switch` flow. Probably should be guarding against such case.

providers/connectwise/communication-types.go
Cobalt0s: For the record, this was discussed on slack.

A TODO for me: virtual field representing a default of some category type should be the only field we are returning and must hold the value, not the id. Also, I should be more consistent with seting the default flags to true.

anushat05 (approved):
One thing:
The description is out of date. It says fields are `AMPERSAND-<type><typeId>` with separate `AMPERSAND-email-default` markers, but the code emits the static `AMPERSAND-defaultEmail` / `AMPERSAND-defaultEmailId` set instead. Worth updating so it matches, since the git history will point back to it later.

PR #3203, opened 2026-07-17
[ENG-4127] feat(connectwise): Write contact communication items

Cobalt0s:
# Description

Communication items is implemented for:
* Create (POST)
* Update (PUT)
  * Before making a PUT call a special PATCH call is maded to wipe out all the communication items, then it gurnatees that PUT will act as "expected by good provider API design".

Not implemented:
* Update (PATCH)

# Live Tests

image

### DIFF
--- a/providers/connectwise/communication-types.go
+++ b/providers/connectwise/communication-types.go
@@ -1,8 +1,13 @@
 package connectwise
 
 import (
+	"context"
+	"fmt"
 	"maps"
+	"sort"
+	"strings"
 
+	"github.com/amp-labs/connectors/common"
 	"github.com/amp-labs/connectors/common/naming"
 	"github.com/amp-labs/connectors/internal/jsonquery"
 	"github.com/spyzhov/ajson"
@@ -12,8 +17,25 @@ const (
 	communicationTypeEmail = "Email"
 	communicationTypePhone = "Phone"
 	communicationTypeFax   = "Fax"
+
+	operationAdd     = "add"
+	operationReplace = "replace"
+	operationRemove  = "remove"
 )
 
+// attachCommunicationItems attaches default communication items from a contact's
+// `communicationItems` array to the top-level `root` map as virtual fields.
+//
+// It:
+//   - Reads the optional `communicationItems` array from the provided JSON node.
+//   - Iterates over items and selects only those marked as default (DefaultFlag=true).
+//   - For each default item, writes the corresponding virtual fields for value and ID
+//     (e.g. email, emailId, phone, phoneId, fax, faxId) into `root`.
+//
+// Non-default communication items are ignored because only the default item per
+// type is exposed via the connector's Read/Metadata surfaces.
+//
+// If `communicationItems` is missing or empty, no fields are added and no error is returned.
 func attachCommunicationItems(node *ajson.Node, root map[string]any) error { // nolint:cyclop
 	communicationItems, err := jsonquery.New(node).ArrayOptional("communicationItems")
 	if err != nil {
@@ -21,8 +43,7 @@ func attachCommunicationItems(node *ajson.Node, root map[string]any) error { // 
 	}
 
 	if len(communicationItems) == 0 {
-		// This contact doesn't have communication items.
-		// Nothing to attach.
+		// Contact has no communicationItems; nothing to attach.
 		return nil
 	}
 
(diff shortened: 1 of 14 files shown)


Cobalt0s:
* **#3203** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3203?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3203?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3199** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3199?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

providers/connectwise/write.go
```diff
@@ -51,64 +77,46 @@ func makeWritePayload(params common.WriteParams, url *urlbuilder.URL) (any, stri
...
 
-func makeWriteCreatePayload(params common.WriteParams) (any, string, error) {
+func makeWriteCreatePayload(params common.WriteParams) (any, string, bool, error) {
 	record, err := params.GetRecord()
 	payloadWithCustomFields(record)
 
-	return record, http.MethodPost, err
-}
-
-func payloadWithCustomFields(record common.Record) {
```
Cobalt0s, line 61: Moved under `custom.go`.

providers/connectwise/communication-types.go
```diff
@@ -53,13 +72,594 @@ func attachCommunicationItems(node *ajson.Node, root map[string]any) error { //
...
+		intent.RemoveFax, communicationTypeFax, contact.Items, itemsRegistry)
+
+	// Execute "add/replace" before "remove" to preserve index validity.
+	// ConnectWise-specific behavior:
+	//   - Inserting at index 0 does not respect DefaultFlag and effectively appends the item anyway.
+	//   - Appending to the end of the array honors DefaultFlag.
+	output = append(output, opEmail...)
+	output = append(output, opPhone...)
+	output = append(output, opFax...)
+	output = append(output, removeEmail...)
```
anushat05: The removes go `email` →`phone` → `fax` by their original index, and JSON Patch applies them in order. So removing a lower index shifts the rest up and the next remove hits the wrong item. 
E.g. remove `email (/0)` + `phone (/1)`: after `/0` goes, phone slides to index 0, so remove` /1` deletes the fax. So instead we can sort the removes by index, highest first.

anushat05 (changes requested):

providers/connectwise/communication-types.go
Cobalt0s: So I tested if this behaviour is indeed present in ConnectWise, because they are not always true to the RFC description, and indeed the remove happens for each operation instead of looking at them as a whole.

Given all the possible communication items:
image

Then removing items at indexes:
image

Removing using the same index slides the array:
image

Cobalt0s: Good catch. I will remove items starting from the end going to beginning to avoid array sliding.

Note: when adding items to the data structure, it does not exhibit the sliding behaviour. For example adding all items at position 0 still adds them at the last index. But remove acts as a proper array method.

anushat05 (approved):
Two smaller things that can be folded into the follow-up unit-test PR rather than blocking here:
- `record["communicationItems"] = ... `silently overwrites a raw array passed alongside the AMPERSAND- fields (the // TODO). We could add a error message here, with `AMPERSAND-` kept out of the error message.
- The PUT wipe isn't atomic: if the wipe `PATCH` succeeds but the `PUT` fails, the contact is left wiped. Worth at least a comment noting it.

Approving to unblock. Let's cover both plus the `allJsonPatchOperation`s remove case in the unit tests.

Cobalt0s:
> Two smaller things that can be folded into the follow-up unit-test PR rather than blocking here:
> 
> * `record["communicationItems"] = ... `silently overwrites a raw array passed alongside the AMPERSAND- fields (the // TODO). We could add a error message here, with `AMPERSAND-` kept out of the error message.
> * The PUT wipe isn't atomic: if the wipe `PATCH` succeeds but the `PUT` fails, the contact is left wiped. Worth at least a comment noting it.
> 
> Approving to unblock. Let's cover both plus the `allJsonPatchOperation`s remove case in the unit tests.

Yes, the dedicated unit tests just for the PATCH Operation implication logic is on my todo list. Not just the usual connector operations but to test the smaller functions.

The overwrite is something that you said that makes 2 ideas collide the 2 level of abstraction, so we assumed that nobody should be passing communicationItems if they use virtual fields.

For the atomic, yes that is sort of a problem. Then end user would have to run the operation one more time for it be fixed. I don't think that it is a problem. I thought that it could be an issue, but since the intention of the PUT is already to remove what is present it is just fine. But yes, the state of the data can be left in limbo between initial and desired. We could comment somewhere, I would rather indicate this in the end user docs. I don't know any other communication channel worth mentioning this. The errors we return or logging wouldn't benefit them, and for us this information is irrelevant.

PR #3212, opened 2026-07-21
[CON-3408] feeat(Sendgrid): Deep connector Write & Delete implementation

Fullstack-Hero:
## Summary
This PR implements Write and Delete support for the SendGrid deep connector, enabling create/update and single-record delete for marketing lists, transactional templates, and ASM unsubscribe groups.

### Write support

- Implemented HTTP write handlers for supported objects.
- Create uses POST `/v3/{objectPath}`; update uses PATCH `/v3/{objectPath}/{id}.`
- JSON request bodies; response id is mapped to `WriteResult.RecordId`.

### Delete support

- Implemented HTTP delete handlers using DELETE `/v3/{objectPath}/{id}` for supported objects.
- Successful 2xx responses map to `DeleteResult{Success: true}`.

## Objects supported

- Read: contacts, lists, segments, singlesends, templates, field_definitions, verified_senders, senders, bounces, blocks, spam_reports, unsubscribes, invalid_emails, asm_groups, categories, subusers
- Write: lists, templates, asm_groups
- Delete: lists, templates, asm_groups

## Testing

- Added unit tests for write and delete
- Added integration write-delete harnesses

```bash
# Run unit tests
go test ./providers/sendgrid -v

# Run with coverage
go test ./providers/sendgrid -cover

# Integration testing
# End-to-end create -> update -> delete
go run ./test/sendgrid/lists/write-delete
go run ./test/sendgrid/templates/write-delete
go run ./test/sendgrid/asm_groups/write-delete
```

### Live Tests
Screenshot 2026-07-28 at 11 40 39 PM
Screenshot 2026-07-28 at 11 43 51 PM
Screenshot 2026-07-28 at 11 57 49 PM


providers/sendgrid/write.go
```diff
@@ -0,0 +1,119 @@
...
+	if !supportedWriteObjects.Has(params.ObjectName) {
+		return nil, common.ErrOperationNotSupportedForObject
+	}
+
+	record, err := common.RecordDataToMap(params.RecordData)
+	if err != nil {
+		return nil, err
+	}
+
+	delete(record, "id")
```
Cobalt0s, line 44: Why is this step needed?

providers/sendgrid/write.go
```diff
@@ -0,0 +1,119 @@
...
+		}, nil
+	}
+
+	recordID, err := jsonquery.New(body).TextWithDefault("id", "")
+	if err != nil {
+		return nil, err
+	}
+
+	if recordID == "" {
+		recordID = params.RecordId
```
Cobalt0s: Instead use this as a default above:
```
jsonquery.New(body).TextWithDefault("id", params.RecordId)
```

test/sendgrid/asm_groups/write-delete/main.go
```diff
@@ -0,0 +1,45 @@
...
+		},
+		map[string]any{
+			"name":        updatedName,
+			"description": description + " (updated)",
+		},
+		testscenario.CRUDTestSuite{
+			ReadFields:          datautils.NewSet("id", "name", "description"),
+			RecordIdentifierKey: "id",
+			UpdatedFields: map[string]string{
+				"name": updatedName,
```
Cobalt0s: You should also check that description is updated too.

providers/sendgrid/read_test.go
```diff
@@ -38,6 +39,22 @@ func TestRead(t *testing.T) { //nolint:funlen
 			Server:       mockserver.Dummy(),
 			ExpectedErrs: []error{common.ErrResolvingURLPathForObject},
 		},
+		{
+			Name:  "Read templates requests generations legacy,dynamic",
+			Input: common.ReadParams{ObjectName: objectTemplates, Fields: connectors.Fields("id", "name")},
+			Server: mockserver.Conditional{
+				Setup: mockserver.ContentJSON(),
```
Cobalt0s, line 46: Given the `plainTextJSONClient` could you add the unit test where provider will respond with:
```
mockserver.ContentText()
```

providers/sendgrid/http.go
```diff
@@ -0,0 +1,59 @@
...
+
+	_ = resp.Body.Close()
+	resp.Body = io.NopCloser(bytes.NewReader(body))
+
+	trimmed := bytes.TrimSpace(body)
+	if len(trimmed) == 0 {
+		return resp, nil
+	}
+
+	if trimmed[0] != '{' && trimmed[0] != '[' {
```
Cobalt0s: Is the trimming necessary, does the `json.Valid` fail?

providers/sendgrid/http.go
```diff
@@ -0,0 +1,59 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/amp-labs/connectors/common"
+)
+
+// plainTextJSONClient wraps the authenticated client so SendGrid responses that
+// return JSON with Content-Type text/plain (notably POST /v3/asm/groups) are
+// accepted by common.ParseJSONResponse.
+type plainTextJSONClient struct {
```
Cobalt0s, line 22: The change makes sense due to the internal expectation here: https://github.com/amp-labs/connectors/blob/main/internal/components/operations/http.go#L86
I think this can be noted in the comment to give more context. 

The `operation.HTTPOperation` (used by new components package) and well `common.JSONHTTPClient` (initial approach in connectors) both expect to operate on JSONs. I think the approach here is good. We have `common.XMLHTTPClient` but thats about it in terms of "data" languages.

Cobalt0s (approved):
LGTM

Please add the test for Groups object, that would be enough to test the `plainTextJSONClient`.

providers/sendgrid/write.go
Fullstack-Hero: So this is defensive sanitization: id is connector routing metadata, not a writable field for these endpoints. Same pattern exists in other connectors (e.g. Livestorm).

PR #3215, opened 2026-07-21
[Attio] Return empty updated fields without error

jlimatampersand:
Stacked on #2668.

`SubscriptionEvent.UpdatedFields` previously returned an error because Attio webhooks do not carry updated-field information. This changes it to return an empty `[]string` with no error, so callers can treat it as "no updated fields" rather than an error condition.

- Updated the unit test to expect no error and an empty slice.


jlimatampersand:
Consolidated into #3219.

PR #2387, opened 2025-11-26
[CON-2713] feat(Attio): add subscribe action (1/2)

immdipu:
## Subscribe Test Result

```
feat/attio-subscribe *11  go run ./test/attio/subscribe                                                                                ✔  system   09:07:33 PM  
time=2026-01-30T21:07:46.216+05:45 level=DEBUG msg="loading credentials file" path=./attio-creds.json
time=2026-01-30T21:07:46.220+05:45 level=DEBUG msg="HTTP request" method=GET url=https://api.attio.com/v2/objects correlationId=e61528f0-87f2-4710-947a-ffc2c6158aad headers.Accept=application/json
time=2026-01-30T21:07:46.908+05:45 level=DEBUG msg="HTTP response" details="map[correlationId:e61528f0-87f2-4710-947a-ffc2c6158aad headers:[Set-Cookie: <redacted> Strict-Transport-Security: max-age=7776000; includeSubDomains; preload Cf-Ray: 9c61f7787d913e33-BOM X-Accurate-At: 2026-01-30T15:22:46.741246000Z Access-Control-Expose-Headers: x-attio-client,x-attio-execution-id,x-attio-platform,x-attio-platform-version,x-attio-record-query-score,x-attio-app-installation-bundle-id X-Attio-Execution-Id: 39d78f2f-f8e8-407e-b7bb-c2d608988dc0 Access-Control-Allow-Origin: * Via: 1.1 google Cf-Cache-Status: DYNAMIC Server: cloudflare Date: Fri, 30 Jan 2026 15:22:46 GMT Content-Type: application/json; charset=utf-8 Vary: Origin X-Frame-Options: DENY] method:GET status:200 url:https://api.attio.com/v2/objects]"
time=2026-01-30T21:07:46.909+05:45 level=DEBUG msg="HTTP request" method=POST url=https://api.attio.com/v2/webhooks correlationId=035cac42-8fcc-4fae-b538-8df6cb7c4ecc headers.Accept=application/json headers.Content-Type=application/json
[...]

PR #3216, opened 2026-07-21
[Attio] Fix webhook event parsing for real payloads

jlimatampersand:
Stacked on #3215 (which is stacked on #2668).

Follow-up to review comments on #2387 that were marked resolved but still broke on real Attio webhook payloads.

### Bugs
- `asMap()` extracts the single event from the top-level `events` **array**, but `ObjectName`/`RawEventName`/`RecordId`/`Workspace` then called `m.Get("events")` again on the already-unwrapped event → every method failed with `key not found: events`.
- The `id` object was asserted as `map[string]string`, but `encoding/json` decodes nested objects to `map[string]any` → `RecordId`/`Workspace` always failed on real data.

Server-side, `Workspace`/`ObjectName`/`RecordId` errors are hard aborts on the webhook path, so this would have broken Attio event processing once wired in.

### Fix
- Read `event_type`/`id` directly off the unwrapped event; assert `id` as `map[string]any`.
- The old tests passed only because the helper fabricated `events` as a map and `id` as `map[string]string`. The helper now builds the real array/`any` shape, and a new `TestSubscriptionEvent_RealWebhookPayload` decodes a real JSON body end to end.

### DIFF
--- a/providers/attio/subscriptionEvent.go
+++ b/providers/attio/subscriptionEvent.go
@@ -112,21 +112,9 @@ func (evt SubscriptionEvent) EventType() (common.SubscriptionEventType, error) {
 }
 
 func (evt SubscriptionEvent) ObjectName() (string, error) {
-	m := evt.asMap()
-
-	events, err := m.Get("events")
+	name, err := evt.RawEventName()
 	if err != nil {
 		return "", err
-	}
-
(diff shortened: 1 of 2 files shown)


jlimatampersand:
Consolidated into #3219.

PR #3217, opened 2026-07-21
[Attio] Use singular "filter" key in GetRecordsByIds

jlimatampersand:
Stacked on #3216.

Follow-up to review on #2387. `GetRecordsByIds` sent the list-records query body with a `"filters"` (plural) key, but Attio expects a singular **`"filter"`** ([docs](https://docs.attio.com/rest-api/guides/filtering-and-sorting)):

```json
{ "filter": { "record_id": { "$in": ["..."] } } }
```

With the plural key the filter is ignored/rejected, so Attio returns unfiltered records instead of the requested IDs.

The existing test didn't catch this because the mock matched only on the URL path. The mock now also asserts the request body — verified by reverting the fix (the test then fails).

### DIFF
--- a/providers/attio/record.go
+++ b/providers/attio/record.go
@@ -33,8 +33,10 @@ func (c *Connector) GetRecordsByIds( //nolint:revive
 		return nil, err
 	}
 
+	// Attio's list-records query expects a singular "filter" key.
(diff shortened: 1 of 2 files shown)


jlimatampersand:
Consolidated into #3219.

PR #3220, opened 2026-07-21
[ENG-4127] feat(connectwise): Change Contacts PUT to PATCH

Cobalt0s:
# Description

To bypass validation bugs of ConnectWise API this PR replaces the use of PUT method on contacts with an equivalent PATCH.

# Live Tests

Update contacts via PUT (under the hood a PATCH equivalent)
image

Update contacts via PATCH
image

### DIFF
--- a/providers/connectwise/communication-types.go
+++ b/providers/connectwise/communication-types.go
@@ -5,10 +5,12 @@ import (
 	"fmt"
 	"maps"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/amp-labs/connectors/common"
 	"github.com/amp-labs/connectors/common/naming"
+	"github.com/amp-labs/connectors/internal/datautils"
 	"github.com/amp-labs/connectors/internal/jsonquery"
 	"github.com/spyzhov/ajson"
 )
@@ -79,7 +81,7 @@ func attachCommunicationItems(node *ajson.Node, root map[string]any) error { // 
 	return nil
 }
 
-// payloadWithCommunicationItems rewrites the write payload for a contact record
+// postPayloadWithCommunicationItems rewrites the write payload for a contact record
 // so that virtual communication-item fields are translated into a proper
 // `communicationItems` array for ConnectWise.
 //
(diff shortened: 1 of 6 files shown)


Cobalt0s:
* **#3220** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3220?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3220?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

providers/connectwise/communication-types.go
```diff
@@ -148,7 +150,224 @@ func (c *Connector) payloadWithCommunicationItems(ctx context.Context, record co
...
+			if err != nil {
+				return nil, err
+			}
+
+			continue
+		}
+
+		// Remove all other fields; they will be replaced from the new record.
+		operations = append(operations, patchOperationPayload{
+			Op:   operationRemove,
```
anushat05, line 235: `contactsFullUpdatePayload` deletes fields the caller didn't send (all unspecified top-level fields, plus non-default comm items). I think that's a regression:
* Normal updates always hit this path. The server sends the record as a plain flat map, nothing builds the patch envelope, so the surgical PATCH path is opt-in only.
* ConnectWise's own PUT never erased unspecified fields (your screenshot 1 proved it, PUT of just firstName kept everything else). So the old connector was safe, and this makes a normal write destructive.

So a customer updating just the email would now wipe the contact's phone, fax, and every unmapped field. That didn't happen before. Thoughts?

Cobalt0s: I think that is the point of the [PUT operation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/PUT).

It is suppose to be a replacement/overiding. And if the values were not set, then it means we are overiding with null values.

ConnectWise actually behaves this way (except custom fields):

* Given: 
image

* PUT with only a name:
image

Everything is wiped out. So PATCH equivalent is correct in this regard.

anushat05 (approved):

PR #3222, opened 2026-07-21
fix(connectwise): Allow partial updates

Cobalt0s:
# Bug

Partial updates were sending other default fields with empty values. This is the reason of the type/id is required error:

image

# Fix

`makeJsonPatchOperations` without id means that the field was not requested for modifications and we should omit it. (this got complex when I added the remove functionality which introduced this bug).

# Live Test

Creating contact without communication items and then updating via patch just the email:

image


Cobalt0s:
* **#3222** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3222?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3222?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

anushat05 (approved):

PR #3250, opened 2026-07-27
feat(square): enable read

immdipu:
# Installation
Mailmonkey using OAuth
# Conventions

- [x]  Read supports incremental sync

# Read and Write
- [x] Insert screenshot of metadata fields from read object installation.
image
image
image
image

- [x] Insert screenshot of Svix destination.
image
image
image
image

- [x] Screenshot of operations on dashboard
image
image
image
image


Cobalt0s (approved):
LGTM

PR #3261, opened 2026-07-29
fix(zoho): read CRM server_time under float64 and json.Number decodes

jlimatampersand:
This is blocking Zoho mail https://github.com/amp-labs/connectors/pull/3196

EventTimeStampNano read server_time via StringMap.AsInt, which only
handles the float64 that a plain json.Unmarshal produces. When the caller
decodes the webhook body with json.Decoder.UseNumber (as the server does
for Zoho to preserve 64-bit ids in the Mail module — folderId exceeds
2^53 with no string twin), server_time arrives as a json.Number, whose
reflect kind is String, so AsInt errors and the CRM event loses its
timestamp (breaking event-collapse dedup downstream).

Add serverTimeMillis: try AsInt first (the existing float64 path), then
fall back to parsing a json.Number/string. This makes the CRM event
parser tolerant of both decode modes, so enabling UseNumber for Zoho is
fully backward compatible for CRM.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/providers/zoho/serverTime_test.go
+++ b/providers/zoho/serverTime_test.go
@@ -1,0 +1,70 @@
+package zoho
+
+import (
+	"bytes"
+	"encoding/json"
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+// crmWebhookBody is a Zoho CRM notification payload. server_time is a bare JSON
+// number (epoch millis).
+const crmWebhookBody = `{
+	"server_time": 1750102639787,
+	"affected_values": [
+		{"record_id": "6756839000000575405", "values": {"Company": "Rangoni Of Test"}}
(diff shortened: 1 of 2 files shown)


jlimatampersand:
* **#3261** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3261?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3261?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

AmandaYao00 (approved):

PR #3284, opened 2026-08-04
[ENG-4159] fix(subscribe): make an explicit CDC quota-optimization opt-out reach Salesforce

AmandaYao00:
PR to support disabling CDC Quota Optimization in the self-serve CDC quota optimization stack ([proposal doc](https://ampersand.slab.com/posts/cdc-quota-optimization-self-serve-config-hwir3mvj)). 

If a resolved config's CDC Quota Optimization has all `enabled:false`, we should tear down the trigger/channels in  Salesforce. 

```
EXAMPLE: 
Account and Contact are set to enabled:false, Opportunity does not have salesforceQuotaOptimization => 
customer's expectation is that we will tear down their CDC artifacts.

{
  "provider": "salesforce",
  "subscribe": {
    "objects": {
      "account": {
        "objectName": "account",
        "destination": "webhook",
        "updateEvent": {
          "enabled": "always",
          "requiredWatchFields": ["Name"]
        },
        "salesforceQuotaOptimization": {
          "enabled": true -> false
        }
      },
      "contact": {
        "objectName": "contact",
        "destination": "webhook",
        "updateEvent": {
          "enabled": "always",
          "requiredWatchFields": ["LastName"]
        },
        "salesforceQuotaOptimization": {
          "enabled": true -> false
        }
      },
      "opportunity": {
        "objectName": "opportunity",
        "destination": "webhook",
        "updateEvent": {
          "enabled": "always",
          "requiredWatchFields": ["StageName"]
        }
      }
    }
  }
}
```


AmandaYao00:
* **#3284** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3284?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3284?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

subscribe/salesforce.go
```diff
@@ -105,11 +103,6 @@ func getSalesforceRequest(
 		return nil, nil //nolint:nilnil // documented contract: no CDC opt-in → no custom payload.
 	}
 
-	optInObjects := optInConfig.ObjectEnabled
-	if len(optInObjects) == 0 {
```
AmandaYao00, line 109: If no objects are enabled, return an empty `QuotaOptimizationObjectFields` instead of nil. This will send a request to Salesforce to redeploy the current config's triggers/channels (which are none if len(optInObjects==0). If there were previous triggers/channels set up, they will be torn down.

providers/salesforce/subscribe.go
```diff
@@ -1072,50 +1072,67 @@ func isObjectSubscribed(
...
 		return err
 	}
 
 	return c.updateExistingChannelMembers(ctx, req, diff)
 }
 
+// shouldSkipQuotaReconcile is true when updateExistingSubscriptions would only rewrite
+// empty filters over empty filters and delete no triggers. Any of the checks below means
+// there is real work (or unknown prior state), so reconcile must run.
+func shouldSkipQuotaReconcile(
```
AmandaYao00, line 1109: We skip updateExistingSubscriptions when the request has no quota fields and Ampersand’s prior state has none either (no fields, no triggers). Without that, a project with CDC config present but every object enabled:false would still enter reconcile on every UpdateSubscription and PATCH channel members with empty filter/enriched fields — a no-op that still hits Salesforce.

providers/salesforce/subscribe.go
```diff
@@ -1072,50 +1072,67 @@ func isObjectSubscribed(
...
+	// Prior fields left on prevState after prepare: tear them down (full or partial disable).
+	if len(prevState.QuotaOptimizationObjectFields) > 0 {
+		return false
+	}
+
+	// Kept-object triggers with no matching request field: delete as orphans.
+	if len(diff.apexTriggersExisting) > 0 {
+		return false
+	}
+
```
eberle1080: You might need to add this, for the case where some previous updateExistingChannelMembers failed (it's in a partially-cleaned-up state in other words).

```suggestion
  // Kept members still carrying a filter (e.g. from a failed teardown)
  for _, member := range diff.channelMembersExisting {
      if member == nil || member.Metadata == nil {
          continue
      }

      if member.Metadata.FilterExpression != "" || len(member.Metadata.EnrichedFields) > 0 {
          return false
      }
  }
```

providers/salesforce/subscribe_test.go
```diff
@@ -311,7 +315,33 @@ func TestUpsertQuotaOptimizationFieldsFiltersUnsubscribedObjects(t *testing.T) {
...
+			`"enrichedFields":[],"filterExpression":""}}`)
+	}))
+	defer server.Close()
+
+	requestCount := func() int {
+		mu.Lock()
+		defer mu.Unlock()
+
+		return requests
+	}
```
eberle1080: Very minor nit, you could probably use an atomic int for this. Saves you a mutex. Feel free to ignore though, it's just a test after all.

eberle1080 (approved):
Looks good overall. I added a suggestion for a corner case - if it makes sense, add it, if I've misunderstood the code, ignore it.

providers/salesforce/subscribe.go
AmandaYao00: makes sense

PR #3287, opened 2026-08-05
[CON-3397] feat(mailgun): add Mailgun read

ejoo:
# Description

- [x] Provider errors are mapped if non-standard (Mailgun errors are a uniform `{"message": "..."}`, surfaced via `common.InterpretError`)
- [x] Custom fields, if not human readable names, are resolved to readable names. (N/A — Mailgun has no custom fields)
- [x] Unit tests cover read logic (placed in `/providers/mailgun`)
- [x] live tests (placed in `/test/mailgun`)

# Tests

## Unit Test

image

## Live Test

image

image

image

image

image

image


providers/mailgun/connector.go
```diff
@@ -8,31 +8,67 @@ package mailgun
...
+			Connector: base,
+			workspace: params.Workspace,
+		}
+
+		// Resolve the regional base URL. US (api.mailgun.net) is the catalog
+		// default; EU is api.eu.mailgun.net. The US/EU host asymmetry prevents a
+		// clean {{.region}} template, so the base URL is set here from metadata.
+		if params.Metadata["region"] == regionEU {
+			base.SetBaseURL(euBaseURL)
+		}
```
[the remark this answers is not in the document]
ejoo: Fixed. The region check is now case-insensitive and trimmed with strings.EqualFold(strings.TrimSpace(...)).

providers/mailgun/handlers.go
```diff
@@ -0,0 +1,818 @@
...
+	if err := simultaneously.DoCtx(ctx, maxConcurrentListFetch, jobs...); err != nil {
+		return nil, err
+	}
+
+	var data []common.ReadResultRow
+	for _, rows := range perList {
+		data = append(data, rows...)
+	}
+
+	return data, nil
```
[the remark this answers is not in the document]
ejoo: Fixed. data is now allocated with make([]common.ReadResultRow, 0), so an empty result returns [] not null. Added a unit test for lists with zero members.

test/mailgun/connector.go
```diff
@@ -10,15 +10,32 @@ import (
...
+}
+
 func GetMailgunConnector(ctx context.Context) *mailgun.Connector {
 	filePath := credscanning.LoadPath(providers.Mailgun)
-	reader := utils.MustCreateProvCredJSON(filePath, false)
+	reader := utils.MustCreateProvCredJSON(filePath, false, fieldRegion)
+
+	metadata := make(map[string]string)
+	if region := reader.Get(fieldRegion); region != "" {
+		metadata["region"] = region
```
Cobalt0s: Why not allow empty string to be tracked by the key:
```
metadata["region"] = reader.Get(fieldRegion)
```

I think connector should figure out that missing region or empty string is the same thing. Maybe some part of code actually requires this, let me know.

test/mailgun/read/main.go
```diff
@@ -0,0 +1,197 @@
...
+		return fmt.Errorf("error reading %s: %w", objectName, err)
+	}
+
+	slog.Info("Read result", "object", objectName, "rows", res.Rows, "nextPage", res.NextPage, "done", res.Done)
+	utils.DumpJSON(res, os.Stdout)
+
+	return nil
+}
+
+func testPagination(ctx context.Context, conn *mailgun.Connector, objectName string) error {
```
Cobalt0s: For pagination the `testscenario.ReadThroughPages` could be helpful.

For other methods you could think about creating some `testscenario` if you find yourself creating some functions over and over across connectors.

providers/mailgun.go
```diff
@@ -10,14 +10,36 @@ func init() {
...
 		Support: Support{
 			BulkWrite: BulkWriteSupport{
 				Insert: false,
 				Update: false,
 				Upsert: false,
 				Delete: false,
 			},
-			Proxy:     true,
+			Proxy: true,
+			// Read and Write stay false until the post-merge integration-testing
```
Cobalt0s: The comment is not needed. This is understood.
The prcoess is no different for the mailgun anyway.

providers/mailgun/read_test.go
```diff
@@ -0,0 +1,723 @@
...
+	"github.com/amp-labs/connectors/test/utils/mockutils/mockcond"
+	"github.com/amp-labs/connectors/test/utils/mockutils/mockserver"
+	"github.com/amp-labs/connectors/test/utils/testconn"
+	"github.com/amp-labs/connectors/test/utils/testutils"
+)
+
+func TestRead(t *testing.T) { //nolint:funlen,maintidx
+	t.Parallel()
+
+	responseWebhooks := testutils.DataFromFile(t, "read/webhooks.json")
```
Cobalt0s, line 19: There are way too many tests. But since you created them keep them.

Also the webhooks is not a real "ampersand object". But that doesn't matter. This will be handled by the Subscribe feature later on.

providers/mailgun/connector.go
```diff
@@ -6,33 +6,73 @@
...
+	// workspace is the Mailgun sending domain (ConnectorParams.Workspace).
+	// It is substituted into domain-scoped object paths at read time and is
+	// only required for those objects (checked in buildReadRequest).
+	workspace string
 }
 
 // NewConnector creates a new Mailgun connector.
 func NewConnector(params common.ConnectorParams) (*Connector, error) {
-	return components.Initialize(providers.Mailgun, params, constructor)
+	return components.Initialize(providers.Mailgun, params, constructor(params))
```
Cobalt0s: Use `components.Init`. It has a better format of the constructor lambda out of the box.

providers/mailgun/connector.go
```diff
@@ -6,33 +6,73 @@
...
+			workspace: params.Workspace,
+		}
+
+		// Resolve the regional base URL. US (api.mailgun.net) is the catalog
+		// default; EU is api.eu.mailgun.net. The US/EU host asymmetry prevents a
+		// clean {{.region}} template, so the base URL is set here from metadata.
+		// The region value is free text, so compare case-insensitively and
+		// trimmed ("EU", " eu" must not silently fall back to the US host).
+		if strings.EqualFold(strings.TrimSpace(params.Metadata["region"]), regionEU) {
+			base.SetBaseURL(euBaseURL)
```
Cobalt0s: Can it only be EU? What about `api.mailgun.net` for `us`?

providers/mailgun/handlers.go
```diff
@@ -0,0 +1,872 @@
...
+// account's mailing lists (see parseListMembersResponse).
+const listAddressPlaceholder = "{list_address}"
+
+type paginationStyle int
+
+const (
+	// paginationCursor follows the response's paging.next URL. Mailgun always
+	// returns paging.next (even on the last page); the follow-up request returns
+	// an empty items page, which common.ParseResult treats as Done.
+	paginationCursor paginationStyle = iota
```
Cobalt0s, line 46: I guess given so much provider API behaviour variation the amount of tests is justified 😅

providers/mailgun/handlers.go
```diff
@@ -0,0 +1,872 @@
...
+	"analytics/logs": {paginationLogsToken, scopeAccount, ""},
+
+	// POST-sourced account tags (the modern Tags API, replacing the deprecated
+	// GET /v3/{domain}/tags) with body-offset pagination. last_seen filters
+	// connector-side when the API returns it as a timestamp string.
+	"analytics/tags": {paginationBodyOffset, scopeAccount, "last_seen"},
+}
+
+func (c *Connector) buildReadRequest(ctx context.Context, params common.ReadParams) (*http.Request, error) {
+	if err := params.ValidateParams(true); err != nil {
```
Cobalt0s: The validation is already done by the "template/parent" here: https://github.com/amp-labs/connectors/blob/main/internal/components/reader/http.go#L36.

providers/mailgun/handlers.go
```diff
@@ -0,0 +1,872 @@
...
+
+	for _, placeholder := range domainPlaceholders {
+		path = strings.ReplaceAll(path, placeholder, c.workspace)
+	}
+
+	return path, nil
+}
+
+func pageSize(params common.ReadParams) string {
+	if params.PageSize <= 0 || params.PageSize > maxPageSize {
```
Cobalt0s: Use `readhelper.PageSizeWithDefaultStr`

providers/mailgun/handlers.go
```diff
@@ -0,0 +1,872 @@
...
+		return jsonquery.New(node).ArrayOptional(recordsKey)
+	}
+}
+
+// timeFilter sieves a page's records by the object's timestamp field against
+// ReadParams.Since/Until. The next-page token is computed from the raw page:
+// an empty raw page ends pagination (Mailgun's cursor endpoints echo
+// paging.next even on the last page), while a page that merely filters down to
+// zero records keeps paginating.
+func (c *Connector) timeFilter(
```
Cobalt0s: I believe this method and other utils used by it can be replaced with `readhelper.MakeTimeFilterFunc`.
If you are unsure of the order you can pick `readhelper.Unordered`. This way it won't stop paginating but will filter out records outside of the time window.

providers/mailgun/handlers.go
```diff
@@ -0,0 +1,872 @@
...
+
+// listsArrayKey is the records key of the /v3/lists response.
+const listsArrayKey = "items"
+
+// parseListMembersResponse implements the lists/members read. The response in
+// hand is a page of the account's mailing lists; for each list we fetch all of
+// its members and flatten them. Pagination advances through the mailing-lists
+// list itself (offset), so one Read call returns the members of one page of
+// lists and Done fires when the lists list is exhausted.
+func (c *Connector) parseListMembersResponse(
```
Cobalt0s: I would split the `handlers.go` file. For instance, the read of special objects can be in a different file.

The file can be dedicated just for the list members.

Maybe you could split by pagination style. The file is almost 1k and it is just the read. The provider is complex.

Cobalt0s (approved):
Approving!

However, are you missing the `US` region? It is important.

providers/mailgun/handlers.go
```diff
@@ -1,33 +1,22 @@
 package mailgun
```
Cobalt0s, line 1: The files split is very nice. Much easier to follow. Easy to see clustered, related functions.

PR #3288, opened 2026-08-05
feat(common): add HTTPOption client wrapper for extra request headers/query params

jlimatampersand:
## Motivation

Some providers require an extra header or query parameter on **every** request beyond the primary auth credential — ConnectWise needs `ClientId`, Blackbaud needs `Bb-Api-Subscription-Key`. Today those are hand-rolled inside each connector's read/write handlers, so any consumer that only uses the authenticated HTTP client never sends them. Notably, the server's proxy action wraps the `generic` connector and forwards requests through the auth client directly, so proxied ConnectWise calls fail with 401 "Incorrect or disabled client ID" even though the clientId is stored in connection metadata.

## Changes

**Mechanism (common):**
- `common.HTTPOption{In, Key, Value}` — a static header (`in: header`) or URL query parameter (`in: query`) attached to every request.
- `common.NewHTTPOptionsClient(client, opts)` — decorates any `AuthenticatedHTTPClient`. Options are applied **set-if-missing**, so values already present on a request (set by the caller or by a connector's own header logic, e.g. ConnectWise's `clientIdHeader()`) win and nothing double-applies. Returns the inner client unwrapped when `opts` is empty.
- `providers.NewClientParams.HTTPOptions` — applied by `NewClient` around the auth-type-specific client, so basic/oauth2/apiKey/custom clients all inherit the behavior uniformly.

**Per-provider declaration (providers):**
- `providers.HTTPOptionSpec{In, Key, MetadataField}` + `SetHTTPOptionSpecs(provider, specs...)` — each provider declares the options it requires next to its `SetInfo` call, with values sourced from connection metadata (`ProviderMetadata.Input`).
- `providers.ResolveHTTPOptions(provider, metadata)` — consumers (the server) resolve the specs against connection metadata generically; specs with missing/empty metadata are skipped.
- Registered ConnectWise: `clientId` metadata → `ClientId` header.

## Follow-up

- Server wiring: amp-labs/server#6945 resolves the specs in `createAuthenticatedClient` so the proxy and all operations inject them automatically.
- Blackbaud registration once its metadata field is validated.
- Longer term this declaration could move into the generated catalog (amp-labs/openapi).

## Test plan

- Unit tests for the decorator (header+query application, original-request immutability for 401-retry replay, caller-value precedence, empty-options passthrough) and for spec resolution (present / missing / unregistered provider).
- `go build ./...`, `go test ./common/... ./providers/`, and `./custom-gcl run` clean.

### DIFF
--- a/common/httpoption.go
+++ b/common/httpoption.go
@@ -1,0 +1,91 @@
+// nolint:revive,godoclint
+package common
+
+import (
+	"net/http"
+)
+
+// HTTPOptionIn identifies where an HTTPOption is attached on the request.
+type HTTPOptionIn string
+
+const (
+	// HTTPOptionInHeader attaches the option as an HTTP request header.
+	HTTPOptionInHeader HTTPOptionIn = "header"
+
+	// HTTPOptionInQuery attaches the option as a URL query parameter.
+	// Note that query parameter values end up in URLs, which are commonly
+	// logged; prefer HTTPOptionInHeader for sensitive values.
+	HTTPOptionInQuery HTTPOptionIn = "query"
+)
+
+// HTTPOption is a static key/value pair attached to every request made by an
+// AuthenticatedHTTPClient, either as a header or a URL query parameter. It is
+// meant for provider requirements beyond the primary auth credential, such as
+// an extra client id or subscription key header, typically sourced from
+// connection metadata.
+type HTTPOption struct {
+	In    HTTPOptionIn `json:"in"`
+	Key   string       `json:"key"`
+	Value string       `json:"value"`
+}
+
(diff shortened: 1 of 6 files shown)


jlimatampersand:
* **#3288** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3288?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3288?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #3295, opened 2026-08-06
[CON-3397] feat(mailgun): mailgun-write/delete

ejoo:
## Conventions
- [x] Provider errors are mapped if non-standard (uniform `{"message"}` bodies surfaced via `common.InterpretError`)
- [ ] Custom fields, if not human readable names, are resolved to readable names. (N/A)
- [x] Unit tests cover read/write/delete/metadata logic (placed in /providers/mailgun)
- [x] Appropriate object names are used. Objects need to be resources, not actions (`messages` and not `messages.send`).
- [ ] Modules are only being added because: (N/A — single module)
  - [ ] Different base URLs
  - [ ] Object name collisions
  
# Tests

## Unit Test
image

## Live Test

image

image

image


test/mailgun/connector.go
```diff
@@ -18,6 +18,14 @@ var fieldRegion = credscanning.Field{ //nolint:gochecknoglobals
 	SuffixENV: "REGION",
 }
 
+// GetWorkspace returns the Mailgun sending domain from the credentials file.
+func GetWorkspace() string {
+	filePath := credscanning.LoadPath(providers.Mailgun)
+	reader := utils.MustCreateProvCredJSON(filePath, false, fieldRegion)
```
Cobalt0s: Passing fieldRegion as requirement doesn't make sense. You only request workspace and never use region.

providers/mailgun/delete.go
```diff
@@ -0,0 +1,91 @@
...
+		return nil, common.ErrOperationNotSupportedForObject
+	}
+
+	path, err := c.collectionPath(params.ObjectName)
+	if err != nil {
+		return nil, err
+	}
+
+	if scope == scopeDomain {
+		path, err = c.substituteDomain(path)
```
Cobalt0s: I think substitution shouldn't be conditional. It should always happen.
If the endpoint has `{domain_name}` it will fill it in, otherwise a no-op.

Also just noticed the schemas.json is invalid. The path `"3/{domain_name}/whitelists"` of an object is wrong. There should be no API version. However, all objects have different versions... 
Which is very odd and makes this hard to track.

Maybe instead of using the paths from the `schema.json` even if it will be sort of a duplication could you extend `collectionPathOverrides` registry that I see you have. That will be one lookup.

providers/mailgun/write_test.go
```diff
@@ -0,0 +1,479 @@
...
+			},
+			Server: mockserver.Conditional{
+				Setup: mockserver.ContentJSON(),
+				If: mockcond.And{
+					mockcond.MethodPOST(),
+					mockcond.Path("/v3/lists"),
+					mockcond.Header(http.Header{
+						"Content-Type": []string{"application/x-www-form-urlencoded"},
+					}),
+					mockcond.Body(`address=developers%40example.com&name=Developers`),
```
Cobalt0s: I think you can skip the encoding. The urlencoded is supported by [mockcond.Body](https://github.com/amp-labs/connectors/blob/main/test/utils/mockutils/mockcond/body.go#L63). It was introduced recently.

providers/mailgun/write_test.go
```diff
@@ -0,0 +1,479 @@
...
+			},
+		},
+		{
+			// Webhooks: event_types is a list on read and a repeated form field on
+			// write ("use multiple times" per the spec); the create response is a
+			// flat {webhook_id} with no envelope. Response shape is the spec's
+			// documented example (not a live capture).
+			Name: "Create webhook repeats event_types and reads the flat webhook_id response",
+			Input: common.WriteParams{
+				ObjectName: "webhooks",
```
Cobalt0s: Webhooks should be removed. It is not a valid Ampersand object for reading and writing.
This endpoint can be used by the Subscribe. If you have it, that is fine you can preserve it for now. But the unit test kind of makes it important.

Cobalt0s (approved):
LGTM

The write is as fun as a read 😄

providers/mailgun/write_test.go
```diff
@@ -275,7 +242,7 @@ func TestWrite(t *testing.T) { //nolint:funlen,maintidx
 				If: mockcond.And{
 					mockcond.MethodPOST(),
 					mockcond.Path("/v3/example.com/templates"),
-					mockcond.Body(`description=Welcome+email&name=welcome&template=%3Chtml%3EHi+%7B%7Bname%7D%7D%3C%2Fhtml%3E`),
+					mockcond.Body(`description=Welcome email&name=welcome&template=<html>Hi {{name}}</html>`),
```
Cobalt0s, line 245: By the way you can even insert the new lines after the `&` the mockcond.Body will handle it too.
There are not many properties so that may not be needed. Just so that you know about it.

Here is an example with many properties: https://github.com/amp-labs/connectors/blob/main/providers/stripe/subscribe_test.go#L102

providers/mailgun/write.go
```diff
@@ -55,7 +55,6 @@ const (
 )
 
 type writeSpec struct {
-	scope objectScope
```
Cobalt0s, line 58: It is much more readable without the scope ✔️ :tada:

PR #3304, opened 2026-08-07
[CON-3456] feat: add monaco metadata connector

Jkarage:
# Conventions

- [ ] Raw response is returned as is for **READ** & **METADATA**, no formatting or flattening is performed.

Live Tests
Screenshot 2026-08-11 at 13 58 37


buildstuff-2026 (approved):
LGTM overall. Before merge:  **Refresh `openapi.json`** from `https://api.monaco.com/openapi.json` and regenerate — `AccountResponse` gained 6 `last_*` fields since the snapshot (docs site lags the API-served spec).

PR #3316, opened 2026-08-11
[CON-3457] feat(monaco): add Read & Search action

Jkarage:
# Conventions

- [ ] Raw response is returned as is for **READ** & **METADATA**, no formatting or flattening is performed.

### DIFF
--- a/providers/monaco/objects.go
+++ b/providers/monaco/objects.go
@@ -49,7 +49,7 @@ var getListObjects = datautils.NewStringSet(
 // so `updated_at` being a filterable field with a `greater_than` operator is
 // unverified for all six. The one directly documented example is accounts
 // filtering on `created_at` with `greater_than`. Confirm against
-// GET /v1/schemas/{entity} once credentials exist; see metadata/README.md.
+// GET /v1/schemas/{entity} once credentials exist.
 //
 //nolint:gochecknoglobals
 var incrementalObjects = datautils.NewStringSet(
--- a/providers/monaco/objects.go
+++ b/providers/monaco/objects.go
@@ -17,10 +17,12 @@ const (
 	objectUsers             = "users"
 )
 
-// getListObjects are served over GET and return the whole collection in a
-// single response: the body carries `data` but no `pagination` object, so there
-// is never a next page. Everything else is listed over POST <path> with a JSON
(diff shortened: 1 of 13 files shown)


providers/monaco/objects.go
```diff
@@ -0,0 +1,62 @@
...
+
+// getListObjects are served over GET and return the whole collection in a
+// single response: the body carries `data` but no `pagination` object, so there
+// is never a next page. Everything else is listed over POST <path> with a JSON
+// body and responds with `{data, pagination, meta}`.
+//
+//nolint:gochecknoglobals
+var getListObjects = datautils.NewStringSet(
+	objectSequenceTemplates,
+	objectTags,
```
buildstuff-2026, line 30: `tags` read omits the required `object` query param** (`contacts|accounts|opportunities`, `required: true` [refer](https://docs.monaco.com/api-reference/tags/list-tags)). 

since each call returns all tags for one object type, use the `NextPage` token to step through the three types. First call (empty token) fetches `?object=contacts` and returns `NextPage: "accounts"`; next call fetches accounts and returns `"opportunities"`; last call fetches opportunities and returns done. That way Read stays stateless and the caller ends up with all tags. Please also add a test that checks the `object` param is actually sent.

providers/monaco/read.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	filters := make([]map[string]any, 0, 2) //nolint:mnd
+
+	if !params.Since.IsZero() {
+		filters = append(filters, filterRule(incrementalField, conditionGreaterThan, params.Since))
+	}
+
+	if !params.Until.IsZero() {
+		filters = append(filters, filterRule(incrementalField, conditionLessThan, params.Until))
```
buildstuff-2026: Small one: `Until` is documented as "up to and including", but `less_than` excludes the boundary. The spec mentions `less_than_or_equals` could probably the better fit here.

Jkarage: updated this to use `less_than_or_equals`.

providers/monaco/objects.go
Jkarage: Implemented as recommended.

buildstuff-2026 (approved):

PR #3318, opened 2026-08-11
[CON-3464] feat(slack): Exhaustive Unit tests

Cobalt0s:
# Background

This PR is the first in a series that introduces the **Slack for user** connector. The new connector reuses the existing **Slack for bot** implementation rather than duplicating it. The two connectors differ only in the endpoints they can access (response parsing, payload construction, query parameters, and REST methods). All other logic and structure is shared.

Before refactoring the shared code (actually a data structure/mapping), we need a comprehensive test suite to guarantee that subsequent changes do not degrade the existing Slack for bot connector's functionality.

In upcoming PRs, `support.go` will be removed and replaced with a unified registry that serves both `slack` (bot) and `slackUserScope` (user) connectors.

The business logic remains shared. *Only the data layer is extended to support both scopes*!

# Description

This PR adds exhaustive unit tests for every object and mapping currently defined in `support.go`. Specifically:

- Created unit test files covering **read**, **read batch**, **create**, and **update** operations.
- Duplicated the maps from `support.go` into the test suite to assert their contents.
- Tests invoke high-level connector functions rather than internal helpers, which are subject to change and tests must be useful in later PRs.
- Tests are fully self-sustained and do not reference any data structures. Duplication in tests makes it independent.

## Test coverage validation

To verify that the tests are exhaustive, I used the following approach:

1. Ran the new test suite.
2. Removed each entry in `support.go` one by one.
3. Confirmed that the tests fail whenever any line is removed.

This proves that the test suite fully captures the behavior provided by `support.go`.

# Future PRs

Going forward, all tests in this suite must pass to ensure no functional regression. If a future PR requires changing or removing any of these tests, the PR description will explicitly explain why and what functionality is being altered or deprecated. This ensures that:

- **Expectations** of existing functionality are **fully documented**.
- Any **outdated** or superseded **behavior** is clearly **identified**.
- Customers experience **no degradation in service**.

PR #3319, opened 2026-08-11
[CON-3464] feat(slack): Refactor as reusable connector

Cobalt0s:
# Description

This PR introduces the internal `mappings` package for the Slack connector and removes `support.go`. The external provider name remains `slack`; internally, the constructor is `slack.NewBotConnector` to make the bot scope explicit.

## The `mappings` package

The `mappings` package contains registries of object definitions. Each object definition covers:
- Collection read
- Item read
- Create
- Update
- Delete (even though connector doesn't support it, the code doesn't exist to make use of definition)

Slack API endpoints are partitioned by OAuth scopes: **user scope**, **bot scope**, **no scope**, and others. Some methods are shared between user and bot scopes. Accordingly, the registries are organized into:
- `scope-bot-user.go` — operations available to both bot and user scopes
- `scope-bot.go` — bot-only operations
- `scope-none.go` — operations requiring no scope, there are just 2 objects that connector can work with.

The Slack **bot** connector uses all three files. In a future PR, the Slack **user** connector will reuse `scope-bot-user.go` and `scope-none.go`, exclude `scope-bot.go`, and introduce its own `scope-user.go`. This is how the two connectors diverge.

### API

The package does not expose registries directly. Instead, callers use accessor functions such as:
```go
GetReadListInfo(provider, objectName)
```

The `provider` argument selects the correct data structure, and `objectName` performs the lookup. If the object is not found, an error is returned:
- `ErrObjectNotSupported` — the object is not defined at all
- `ErrOperationNotSupportedForObject` — the object exists, but the requested operation is not defined for it

All registry entries are derived from the **Slack API documentation**, with each operation linked to its corresponding docs.

## Code changes

- **URL construction**: Previously, read URLs were built via conditional string concatenation on object names. Now, URLs are resolved via mapping lookups.
- **Response parsing**: Response keys (e.g., the key holding the array, the key used for time sorting, the key holding the record ID, etc.) are now properties looked up from the config returned by the `mappings` package for each operation.
- **`support.go` removed**: All functionality previously in `support.go` is now covered by the `mappings` package.
- **Test relocation**: Live tests moved from `test/slack` to `test/slack/slackbot`.

## Future work

The `mappings` package defines `Since`/`Until` fields that are not yet used by existing functionality and some others. A follow-up PR will leverage these to improve functionality. This PR may introduce additional "data" definitions that are not immediately consumed; the goal is to preserve and port existing behavior without regression.

## Breaking/behavioral changes

Please review the following carefully:

### Response key corrections
- `conversations.requestSharedInvite`: Records are stored under the `invite_requests` key (previously assumed `invites`).
- `team.externalTeams`: Records are stored under the `organizations` key (previously assumed `external_teams`).

### Removed object
- **`reactions`**: This object has been removed. Previously, read operations supported it, but there is no consistent response format and no field that can serve as a record ID. **Question:** Is this acceptable, or should support be added back?

### New filtering support
- `conversations.listConnectInvites`: Now supports connector-side filtering via the `date_last_updated` field (unit test samples updated accordingly).
- `conversations.requestSharedInvite`: Now supports connector-side filtering via the `date_last_updated` field (unit test samples updated accordingly).

### HTTP method changes
- `auth.teams`: Read operation now uses **POST** instead of **GET**.
- `chat.scheduledMessages`: Read operation now uses **POST** instead of **GET**.

# Live Tests

## Oauth2

Given scopes:
```
files:read,remote_files:read,bookmarks:read,calls:read,calls:write,channels:read,channels:write.topic,users:read,channels:manage
```

The consent screen clearly states that scopes are for the app.
image

Below are each script that is executed using oauth2 acquired access tokens.

### authmetadata

image

### metadata

image

### read

image

### read-by-ids

image

### write

image

### subscribe/consumer

Peek 2026-08-12 23-15

Peek 2026-08-12 23-18


Cobalt0s:
* **#3322** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3322?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3366** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3366?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3319** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3319?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3319?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/amp-labs?ref=merge-queue-instructions-in-stack-comment&prId=3319).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

providers/slack/internal/webhook/verifier.go
```diff
@@ -12,25 +12,19 @@ import (
...
-	client       *common.JSONHTTPClient
-	providerInfo *providers.ProviderInfo
-
 	signingSecret string
 }
 
 // NewVerifier constructs an event message verifier.
 // The empty signingSecret won't trigger a failure to preserve backward compatibility.
 // However, no event message will be accepted.
-func NewVerifier(client *common.JSONHTTPClient, providerInfo *providers.ProviderInfo, signingSecret string) *Verifier {
```
Cobalt0s, line 30: These values are not used. Removing.

providers/slack/internal/mappings/registries.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+// nolint:gochecknoglobals
+var (
+	objectsBotConnector  = make(objectsRegistry)
+	objectsUserConnector = make(objectsRegistry)
+)
+
+func init() {
+	for name, object := range objectsForNoScopes {
```
jlimatampersand: Rather than having it in the init function, can we just define them as a registry to start with?

providers/slack/internal/mappings/registries.go
```diff
@@ -0,0 +1,80 @@
...
+	for name, object := range objectsForBotAndUserScopes {
+		objectsBotConnector.add(name, object)
+		objectsUserConnector.add(name, object)
+	}
+
+	for name, object := range objectsForBotScope {
+		objectsBotConnector.add(name, object)
+	}
+
+	objectsForNoScopes = nil
```
jlimatampersand: This seems kind of weird to set and remove. Do we ever reassign these variables?

jlimatampersand (approved):

providers/slack/handlers.go
```diff
@@ -115,20 +121,25 @@ func (c *Connector) parseReadResponse( //nolint:unparam
...
+
+	if info.TimeFilterField != "" {
 		return common.ParseResultFiltered(
 			params,
 			response,
-			nodeRecords(params.ObjectName),
-			makeTimeFilter(params.ObjectName),
+			nodeRecords(info.ResponseField),
+			makeTimeFilter(info),
 			readhelper.MakeMarshaledDataFuncWithId(nil, readhelper.NewIdField("id")),
```
jlimatampersand, line 135: Both marshal paths here (line 135 and line 144) hardcode `readhelper.NewIdField("id")`, so `ReadListInfo.ResponseIdField` / `NestedResponseIdField` are never consumed for list reads (`NestedResponseIdField` ends up as dead code — nothing reads it).

This silently produces rows with empty `Id` for two objects, per the Slack docs and this PR's own mappings:
- `team.externalTeams` — records carry `team_id`, not `id` ([team.externalTeams.list docs](https://docs.slack.dev/reference/methods/team.externalteams.list))
- `conversations.listConnectInvites` — the id is nested at `invites[i].invite.id` ([conversations.listConnectInvites docs](https://docs.slack.dev/reference/methods/conversations.listconnectinvites))

`extractIdFromRecord` returns `""` on a missing field rather than erroring, so this won't surface in tests that don't assert on `Id`.

Suggested fix (works for flat and nested since a nil zoom path behaves like the flat case):
```go
readhelper.NewNestedIdField(info.NestedResponseIdField, info.ResponseIdField)
```

providers/slack/internal/mappings/scope-bot-user.go
```diff
@@ -0,0 +1,309 @@
...
+		writeUpdateInfo: &WriteUpdateInfo{
+			Href:            "slackLists.items.update",
+			RequestIdField:  "list_id",
+			ResponseField:   "",
+			ResponseIdField: "", // None
+		},
+		// https://docs.slack.dev/reference/methods/slacklists.items.delete
+		deleteInfo: nil, // Requires query param `list_id`.
+	},
+	"team.externalTeams": {
```
jlimatampersand: Per [the team.externalTeams.list docs](https://docs.slack.dev/reference/methods/team.externalteams.list), this method supports **bot tokens only** (scopes: `conversations.connect:manage` + `team:read`) — user tokens are not listed. Since the plan is for the future user connector to reuse `scope-bot-user.go`, this entry should move to `scope-bot.go` so the user connector doesn't inherit an object it can't read.

Fix: move this whole `"team.externalTeams"` entry into `objectsForBotScope` in `scope-bot.go`.

jlimatampersand (approved):
Found some minor bugs. Left some comments. Please resolve before submitting

providers/slack/internal/mappings/scope-bot-user.go
```diff
@@ -0,0 +1,309 @@
...
+			Href:            "canvases.create",
+			ResponseField:   "",
+			ResponseIdField: "canvas_id",
+		},
+		// https://docs.slack.dev/reference/methods/canvases.edit
+		writeUpdateInfo: &WriteUpdateInfo{
+			Href:            "canvases.edit",
+			RequestIdField:  "canvas_id",
+			ResponseField:   "",
+			ResponseIdField: "canvas_id",
```
jlimatampersand: The [documented canvases.edit response](https://docs.slack.dev/reference/methods/canvases.edit) is only `{"ok": true}` — there is no `canvas_id` in it (unlike [canvases.create](https://docs.slack.dev/reference/methods/canvases.create)). As written, `WriteResult.RecordId` will always be empty for canvas updates: `write.go` does `StrWithDefault(idField, "")` and quietly returns `""`.

Either clear the field so it routes into the optimistic-success branch:

```suggestion
			ResponseIdField: "", // canvases.edit returns only {"ok": true}; no id in the response.
```

…or (nicer for callers) have the write path echo the id from the request (`RequestIdField`) when the response has no id, since the caller already supplied `canvas_id`.

providers/slack/internal/mappings/scope-bot-user.go
```diff
@@ -0,0 +1,309 @@
...
+			Href:            "slackLists.create",
+			ResponseField:   "",
+			ResponseIdField: "list_id",
+		},
+		// https://docs.slack.dev/reference/methods/slacklists.update
+		writeUpdateInfo: &WriteUpdateInfo{
+			Href:            "slackLists.update",
+			RequestIdField:  "id",
+			ResponseField:   "",
+			ResponseIdField: "list_id",
```
jlimatampersand: Same issue as `canvases.edit`: the [documented slackLists.update response](https://docs.slack.dev/reference/methods/slacklists.update) is just `{"ok": true}` — no `list_id` — so `RecordId` will always come back empty here.

```suggestion
			ResponseIdField: "", // slackLists.update returns only {"ok": true}; no id in the response.
```

…or echo the request's `id` arg in the write path, same as suggested for canvases.edit.

jlimatampersand (changes requested):
Some mappings are mapping to wrong responseID fields according to the docs. Please fix this list those operations and objects as unavailable. Please DM me or tag in in channel after resolving the issues.

providers/slack/handlers.go
```diff
@@ -115,20 +121,25 @@ func (c *Connector) parseReadResponse( //nolint:unparam
 	request *http.Request, //nolint:revive
 	response *common.JSONHTTPResponse,
 ) (*common.ReadResult, error) {
-	if objectsWithConnectorSideFilter.Has(params.ObjectName) {
+	info, err := mappings.GetReadListInfo(c.Provider(), params.ObjectName)
+	if err != nil {
+		return nil, err
+	}
+
+	if info.TimeFilterField != "" {
```
jlimatampersand, line 129: Any object with a `TimeFilterField` is now routed through `ParseResultFiltered` unconditionally — even when the caller passed no `Since`/`Until`. In that case `TimeBoundary.Contains` always returns true (`common/readhelper/time.go:67`), so the filter is a pure pass-through **except** that `MakeTimeFilterFunc` still extracts the timestamp from every record and fails the entire page with `ErrTimestampKeyNotFound` if a single record lacks the field (`common/readhelper/filter.go:190-193`).

So one record missing e.g. `updated`/`date_last_updated` turns a previously-working plain read into a hard error. Slack doc examples show these fields, but they don't guarantee presence on every record (e.g. `conversations.listConnectInvites` mixes invite shapes across direction/status).

Suggested fix: only take the filtered path when time filtering was actually requested:
```go
if info.TimeFilterField != "" && (!params.Since.IsZero() || !params.Until.IsZero()) {
```
(and/or make the filter tolerant of records missing the timestamp field — skipping them from filtering rather than failing the page — if partial presence is expected).

providers/slack/internal/mappings/registries.go
Cobalt0s: No we don't. I will change the way we define this mappings to avoid reshuffling.

providers/slack/handlers.go
Cobalt0s: Thats right. This is the current behaviour so I am preserving it, even if it is wrong. And the other PRs will fix this.

providers/slack/internal/mappings/scope-bot-user.go
Cobalt0s: Indeed, I have verified with API calls and adjusted the scraping scripts to collect scopes from the dedicated method web-pages rather than registry of scopes.

This is the only object that was inconsistent. Other objects fall into the right category.

providers/slack/internal/mappings/scope-bot-user.go
Cobalt0s: Echoing would be great. Sometimes providers have "good" create operations and update just says ok. I prefer to fallback to echoing.

providers/slack/handlers.go
Cobalt0s: @jlimatampersand 
Question: should this be true for all connectors?

We can insert the Since/Until check at the beginning of the method: https://github.com/amp-labs/connectors/blob/main/common/parse.go#L130

This method is called by most connectors that try connector side filtering. There is another method that is in use for this feature: https://github.com/amp-labs/connectors/blob/main/common/readhelper/filter.go#L60

jlimatampersand: Actually, this should be okay; I doubt that the Slack response will omit the field. The `Since` param handling is going to be handled differently for each provider, I suppose? I don't think universal handling of these filters makes sense. 

You can move on as you find it correct here.

jlimatampersand (approved):

providers/slack/handlers.go
Cobalt0s: We might actually have a consistent approach across connectors. Does filtering the records using timestamp without Since/Until make sense? I suppose not. Thats the main issue that you have found in this code. I think that it is universal.

Note: the connector side filtering had several iterations. The code changed a bit. Bugs fixed, the requirement of what it should do also became more vivid.

I decided to create the PR for this. The unit test screenshot result is in the PR definition. It shows with and without the fix. https://github.com/amp-labs/connectors/pull/3379

PR #3321, opened 2026-08-11
[CON-3464] feat(slack): Oauth2 custom query param

Cobalt0s:
# Description

This PR introduces the `SlackUserScope` provider and updates OAuth2 configuration to support Slack's custom `user_scope` query parameter.

## Provider configuration changes

- **Renamed existing provider**: `Slack` → `Slack Bot Scope` (display name updated)
- **Added new provider**: `SlackUserScope` with display name "Slack User Scope"

## Generated types

Provider types are regenerated from [amp-labs/openapi#391](https://github.com/amp-labs/openapi/pull/391). This PR introduces the `ScopeQueryParam` field on the `Oauth2Opts` struct, which allows providers to specify a custom query parameter name for passing scopes during OAuth2 authorization.

## Core change: custom scope query parameter

The `SlackUserScope` provider configuration sets `ScopeQueryParam` to `user_scope`. This is required because Slack's user-scoped OAuth flow does not accept scopes via the default `scope` query parameter—instead, they must be passed as `user_scope`.

This is the primary objective of this PR: enabling providers to pass scopes via a custom query parameter.

## Token generation changes

The `token.go` file (used by developers with `*-creds.json` files) now respects the `ScopeQueryParam` configuration:

- If `ScopeQueryParam` is empty (default): scopes are attached using the standard `scope` query parameter (existing behavior).
- If `ScopeQueryParam` is set (e.g., `user_scope`): scopes are omitted from the standard location and instead passed via `AuthOptions`, which are then included in the `oauth2.Config.AuthCodeURL` call. This results in scopes being sent using the custom query parameter.

### Slack for Bot

Given scopes:
```
files:read,remote_files:read,bookmarks:read,calls:read,calls:write,channels:read,channels:write.topic,users:read,channels:manage
```

The consent screen clearly states that scopes are for the app.
image

## Slack for User

Given the scopes:
```
bookmarks:read,calls:read,calls:write,channels:read,channels:write.topic,channels:write,files:read,remote_files:read,users:read
```

The consent screen clearly states that scopes are for the user and not the app (aka not bot).

image

PR #3320, opened 2026-08-11
[CON-3464] feat(slack): Scrap methods and scopes

Cobalt0s:
# Description

This PR introduces a Slack API documentation scraper and the resulting knowledge base used to build the `mappings` package.

## Scraper scripts

Created several scripts under `scripts/scraper/slack` that scrape the Slack API documentation to acquire a list of API methods grouped by OAuth scope:

1. **`1-scopes/main.go`**: Reads the list of scopes by their type from [https://docs.slack.dev/reference/scopes/](https://docs.slack.dev/reference/scopes/).
2. **`2-methods/main.go`**: Uses the output of the first script to query which methods are accessible via each scope.
3. **`3-views/main.go`**: Restructures the output of the second script for easier consumption.

All generated output files are committed to the repository.

## Knowledge base

The output files serve as a knowledge base for:
- Building the `mappings` package registries
- Analyzing which Slack API methods can be grouped to represent an Ampersand object
- Determining which operations (read, create, update, delete) can be supported and how (via configuration)
- Assisting developers in making informed decisions when extending or modifying connector functionality

## Future use

This knowledge base could also be leveraged for external Ampersand documentation, such as generating a "supported operations by connector" table. Or answering what scopes are needed for said object for etiher `Slack` or `SlackUserScope` connectors.

PR #3322, opened 2026-08-11
[CON-3464] feat(slack): User scope

Cobalt0s:
# Description

This PR introduces the **Slack for User** connector by exposing `slack.NewUserConnector` from the `new.go` factory.

## Connector instantiation

The `slack` package instantiates the user connector similarly to the bot connector. The only difference is the provider name:
- Bot connector: `providers.Slack`
- User connector: `providers.SlackUserScope`

Internally, the `mappings` package now includes `scope-user.go`. Based on the provider name, the mappings layer returns different configurations, enabling the two connectors to share the same codebase while operating with different API endpoints and permissions.

## Test suite

In the `test` repository, `test/slack/slackuser` mirrors the scripts in `test/slack/slackbot`. While this may appear redundant, the two connectors are functionally separate and independent. As Slack's API evolves, the number and nature of test scripts may diverge due to differences in supported endpoints between bot and user scopes.

Each test script uses `slack-user-scope-creds.json`, which is resolved universally based on the provider name. Crucially, the access token is different—it is issued on behalf of the **user**, not the app/bot. All operations (read, write, subscribe, etc.) work identically but execute with user-scoped permissions.

# Live Tests

## Oauth2

Given the scopes:
```
bookmarks:read,calls:read,calls:write,channels:read,channels:write.topic,channels:write,files:read,remote_files:read,users:read
```

The consent screen clearly states that scopes are for the user and not the app (aka not bot).

image

### authmetdata

image

### metadata

image

### read

image

### read-by-ids

Peek 2026-08-12 23-35

### write

image

### subscriber/consumer

Peek 2026-08-12 23-46


Cobalt0s:
* **#3322** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3322?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3322?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3366** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3366?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3319** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3319?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/amp-labs?ref=merge-queue-instructions-in-stack-comment&prId=3322).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

jlimatampersand (approved):

test/slack/slackuser/connector.go
```diff
@@ -0,0 +1,50 @@
...
+	Name:      "signingSecret",
+	PathJSON:  "metadata.signingSecret",
+	SuffixENV: "SIGNING_SECRET",
+}
+
+func NewConnector(ctx context.Context) *slack.Connector {
+	filePath := credscanning.LoadPath(providers.SlackUserScope)
+	reader := utils.MustCreateProvCredJSON(filePath, true, signingSecretField)
+
+	conn, err := slack.NewBotConnector(common.ConnectorParams{
```
jlimatampersand: This harness loads `SlackUserScope` credentials but constructs the connector via `NewBotConnector`, so every slackuser live test actually exercises the **bot** connector (`providers.Slack`). Reading any of the 9 new user-scope objects (`admin.*`, `stars`) through this harness returns `ErrObjectNotSupported`, and the tests that do pass (conversations, users, calls) only re-validate the bot connector — `NewUserConnector` and the `SlackUserScope` registry case are unreachable from this PR's own test suite, so the live-test evidence validated the wrong connector.

```suggestion
	conn, err := slack.NewUserConnector(common.ConnectorParams{
```

providers/slack/internal/mappings/scope-user.go
```diff
@@ -0,0 +1,145 @@
...
+
+// nolint:gochecknoglobals
+var objectsForUserScopes = map[string]Object{
+	"admin.apps.activities": {
+		// https://docs.slack.dev/reference/methods/admin.apps.activities.list
+		readListInfo: &ReadListInfo{
+			Href:            "admin.apps.activities.list",
+			Method:          http.MethodPost,
+			ResponseField:   "activities",
+			ResponseIdField: "created", // There is no ID field. Creation time is chosen as a primary key.
```
jlimatampersand: `ReadListInfo.ResponseIdField` / `NestedResponseIdField` are dead config for list reads — `parseReadResponse` in `providers/slack/handlers.go` hardcodes `readhelper.NewIdField("id")` in both branches (same issue flagged on #3319, see https://github.com/amp-labs/connectors/pull/3319#discussion_r3809122225). So this comment describes behavior that doesn't exist: 5 of the 9 new objects (`admin.apps.activities` via `created`, `admin.users.session` via `session_id`, `stars` via nested `message.permalink`, `admin.apps.approved`/`admin.apps.restricted` via nested `app.id`) have no top-level `id` and will return every row with `Id: ""` — downstream id-keyed dedup/upsert silently collapses or drops records.

Fix: wire these fields into `parseReadResponse` via `readhelper.NewNestedIdField(info.NestedResponseIdField, info.ResponseIdField)` (as `read-by-ids.go` already does for `ReadItemInfo`). Separately: per [admin.apps.activities.list](https://docs.slack.dev/reference/methods/admin.apps.activities.list), `created` is a second-resolution timestamp — multiple activities can share one value, so it's not a reliable primary key even once wired.

providers/slack/internal/mappings/scope-user.go
```diff
@@ -0,0 +1,145 @@
...
+			TimeFilterField: "",
+			SinceQP:         "min_date_created",
```
jlimatampersand: `SinceQP`/`UntilQP` have no consumers anywhere — `buildReadRequest` only sends `limit` and `cursor` — and `TimeFilterField` is `""` here, so `Since`/`Until` read params are ignored entirely for this object. An incremental sync of `admin.apps.activities` silently returns the full unbounded activity log on every run (every other new object at least gets connector-side `TimeFilterField` filtering).

Since [admin.apps.activities.list](https://docs.slack.dev/reference/methods/admin.apps.activities.list) supports `min_date_created`/`max_date_created` server-side, wiring `SinceQP`/`UntilQP` into `buildReadRequest` would fix this properly; otherwise set `TimeFilterField: "created"` as a stopgap so at least client-side filtering applies.

providers/slack/internal/mappings/scope-user.go
```diff
@@ -0,0 +1,145 @@
...
+			UntilQP:         "max_date_created",
+		},
+	},
+	"admin.apps.approved": {
+		// https://docs.slack.dev/reference/methods/admin.apps.approved.list
+		readListInfo: &ReadListInfo{
+			Href:                  "admin.apps.approved.list",
+			Method:                http.MethodGet,
+			ResponseField:         "approved_apps",
+			NestedResponseIdField: []string{"apps"},
```
jlimatampersand: Per [admin.apps.approved.list](https://docs.slack.dev/reference/methods/admin.apps.approved.list), each `approved_apps` element nests the app object under `"app"` (singular) — `{"app": {"id": "A0..."}, "scopes": [...], "date_updated": ...}` — same shape as `admin.apps.restricted` below, which correctly uses `"app"`. Latent today only because `NestedResponseIdField` is unconsumed (see comment above), but the mapping data is wrong.

```suggestion
			NestedResponseIdField: []string{"app"},
```

providers/slack/internal/mappings/scope-user.go
```diff
@@ -0,0 +1,145 @@
...
+	"stars": {
+		// https://docs.slack.dev/reference/methods/stars.list
+		// End-users can use the new Later view, but Later APIs are not currently available.
+		// As a result of this transition, the stars.list method will no longer
+		// reflect anything new that users are saving.
+		readListInfo: &ReadListInfo{
+			Href:                  "stars.list",
+			Method:                http.MethodGet,
+			ResponseField:         "items",
+			NestedResponseIdField: []string{"message"},
```
jlimatampersand: Per [stars.list](https://docs.slack.dev/reference/methods/stars.list), `items` is heterogeneous — `type` can be `message`, `file`, `channel`, `im`, or `group` — and only `message` items have a `"message"` key. A user who starred a file or channel produces records where this zoom path resolves to nothing, yielding empty ids once `NestedResponseIdField` gains a consumer. Also `permalink` isn't a stable record id. Consider deriving the id per item `type` (e.g. `file.id` / `channel` / `message.permalink`), or documenting that only message stars get ids.

test/slack/slackuser/subscription/webhook.go
```diff
@@ -0,0 +1,42 @@
...
+		if err := json.Unmarshal(data, &body); err != nil {
+			return false
+		}
+
+		if body.Challenge == "" {
+			return false
+		}
+
+		writer.WriteHeader(http.StatusOK)
+		writer.Header().Set("Content-Type", "text/plain")
```
jlimatampersand: `Header().Set` after `WriteHeader` is a no-op — headers are flushed once `WriteHeader` is called. It works today only because Go's content sniffing happens to pick plain text for the [url_verification challenge](https://docs.slack.dev/apis/events-api/#url_verification); if this ever switches to the JSON response form, the wrong Content-Type ships silently. (Same ordering bug exists in the `slackbot` copy this was cloned from.)

```suggestion
		writer.Header().Set("Content-Type", "text/plain")
		writer.WriteHeader(http.StatusOK)
```

test/slack/slackuser/subscription/webhook.go
```diff
@@ -0,0 +1,42 @@
+package subscription
```
jlimatampersand: This file is a byte-identical copy of `test/slack/slackbot/subscription/webhook.go` — the [url_verification handshake](https://docs.slack.dev/apis/events-api/#url_verification) is provider-wide, not per-scope, so any fix (e.g. the header-ordering issue below) now has to land in two places. Suggest hoisting the processor into a shared package (e.g. `test/slack/subscription`) imported by both consumer mains. The same goes for the near-verbatim `connector.go` duplication, where only the provider constant, TokenURL, and constructor differ — a shared parameterized helper would have made the `NewBotConnector` copy-paste bug impossible.

test/slack/slackuser/write/main.go
```diff
@@ -0,0 +1,93 @@
...
+
+	slog.Info("Creating calls...")
+
+	res, err := conn.Write(ctx, params)
+	if err != nil {
+		return "", err
+	}
+
+	// Print the results
+	jsonStr, err := json.MarshalIndent(res, "", "  ")
```
jlimatampersand, line 57: This marshal-and-print block (and its twin at lines 84–91) re-implements `utils.DumpJSON` (`test/utils/dump.go`), which this same PR already uses in `read/main.go` and `metadata/main.go`. Replacing both with `utils.DumpJSON(res, os.Stdout)` drops the hand-rolled error wrapping and keeps the output format consistent across the test scripts.

providers/slack/internal/mappings/registries.go
```diff
@@ -73,6 +78,13 @@ func getObject(provider providers.Provider, objectName string) (*Object, error)
 			return nil, common.ErrObjectNotSupported
 		}
 
+		return &object, nil
+	case providers.SlackUserScope:
```
jlimatampersand, line 57: The two switch cases are identical except for the registry variable, and each future scope variant adds another copied lookup block. A package-level `map[providers.Provider]map[string]Object` (the registries already exist as globals) would reduce `getObject` to a single lookup with one shared `ErrObjectNotSupported` / `ErrInvalidImplementation` path — a third variant becomes a one-line entry.

jlimatampersand:
One issue outside this diff: the `SlackUserScope` catalog entry in `providers/slack.go` still declares `Support{Proxy: false, Read: false, Subscribe: false, Write: false}`, while this PR wires a fully read/write/subscribe-capable connector into `connector/new.go`. Anything gating on `ProviderInfo().Support` (e.g. subscribe config validation) will refuse operations the connector actually supports — the bot entry sets all four to `true`. The flags should be flipped alongside this PR, or the connector shouldn't be exposed yet.

test/slack/slackuser/connector.go
Cobalt0s: Thats a bug. :cry: 
However, the tests still validated the "correct" connector. I changed the `credscanning.LoadPath` explicitly but forgot about the `slack.NewBotConnector/slack.NewBotConnector`.

The correct credentials were loaded. The scripts however, target the "shared" objects. So the difference between `slack.NewBotConnector/slack.NewBotConnector` didn't surface, because it should've complained that `mappings` is missing some object.

All the user objects are out of our reach. They require enteprise setup. All user objects are "admin" objects. Stars is something to be deprecated. However, for the sake of validating that User connector is unique and different, I created a Star record. Here is the reading with "bug" and with the "fix":

Peek 2026-08-20 23-26

providers/slack/internal/mappings/scope-user.go
Cobalt0s: The wiring is done in this PR: https://github.com/amp-labs/connectors/pull/3366.
Maybe I broke down the steps too much. The hardcoded "id" is the current behaviour of `slack` connector, so I didn't touch it here.

providers/slack/internal/mappings/scope-user.go
Cobalt0s: I see. That is a valid point. Essentially I would need PR https://github.com/amp-labs/connectors/pull/3366 to be a pre-requisite for this one.

providers/slack/internal/mappings/scope-user.go
Cobalt0s: Thank you for going over and finding all of these!

providers/slack/internal/mappings/scope-user.go
Cobalt0s: I will be removing this object. There are some other objects that didn't make it to the `mappings` for the same reason. The response format is not consistent/heterogeneous.

Stars is outdated and Slack is dropping support for this. Also stars are not visible in the UI only via API.

PS: I used the stars to demonstrate that user scope connector is different from bot in one of the replies above.

test/slack/slackuser/subscription/webhook.go
Cobalt0s: Fixed. I forget sometimes about this order rule :face_with_diagonal_mouth: 
Addressed some other places in the codebase that have this wrong order.

test/slack/slackuser/subscription/webhook.go
Cobalt0s: I will move it to the "shared" utils.

test/slack/slackuser/write/main.go
Cobalt0s: Sure, I didn't want to get into the refactoring too much.

Cobalt0s: Actually, since I am doing the Delete implementation of Slack, the script will change. Lets ignore this for this PR.
Here is the delete WIP PR: https://github.com/amp-labs/connectors/pull/3367

providers/slack/internal/mappings/registries.go
Cobalt0s: A third variant is still present with `map[providers.Provider]map[string]Object`. It is the case when the provider is not Slack or SlackUserScope. We still do the lookup:
* First by provider
* Next by object name

The reshufling won't fix this problem.
I already modified the maps to be populated with `init()` in each file. This is similar to how we load ProviderInfo from many files into 1 big map -- the Catalog. In here I do the same but group by 2 maps right away.

Result: Single object is defined in one file but may appear in both bot and user registries.

Cobalt0s:
> One issue outside this diff: the `SlackUserScope` catalog entry in `providers/slack.go` still declares `Support{Proxy: false, Read: false, Subscribe: false, Write: false}`, while this PR wires a fully read/write/subscribe-capable connector into `connector/new.go`. Anything gating on `ProviderInfo().Support` (e.g. subscribe config validation) will refuse operations the connector actually supports — the bot entry sets all four to `true`. The flags should be flipped alongside this PR, or the connector shouldn't be exposed yet.

Shouldn't we integration test first? And after the integration is complete only then switch on the flags? At least for Proxy/Read/Write.

For Subscribe as I understand I must turn it on. Please, tell me exactly what is required and I will do it. And let me know if that approach should be employed for all connectors going forward.

jlimatampersand (approved):

jlimatampersand:
oh, that provider info stuff is my ai reviewer hallucinating, I should have knocked it off. looks good. thanks!

PR #3327, opened 2026-08-12
feat(salesforce): adopt existing event relay when channel is already bound to it

jlimatampersand:
## Why

Register's create calls all recover from `DUPLICATE_DEVELOPER_NAME` by adopting the existing record — except the one case Salesforce never reports that way. Channel-binding integrity is checked **before** name uniqueness, so a surviving EventRelayConfig carrying **our own** DeveloperName (left by an earlier attempt whose rollback failed or was skipped, e.g. an activity timeout) surfaces as:

```
FIELD_INTEGRITY_EXCEPTION: We couldn't create the event relay because it contains a channel,
'amp_<hash>', that is referenced by an existing event relay, 'amp_<hash>'. ...
```

The adopt path never fires, Register fails, and its rollback then hits `DEPENDENCY_EXISTS` deleting the NamedCredential the surviving relay references — the classic wedged state, freshly re-minted by every retry (seen live today on installation `bc477872…`: channel + NC adopt succeeded, ERC create wedged).

## What

- `CreateEventRelayConfig`: on the channel-bound `FIELD_INTEGRITY_EXCEPTION`, recover via the same by-DeveloperName lookup used for duplicates — adopt the relay when it carries our name; when the channel is owned by a **foreign-named** relay, the lookup finds nothing and both errors are returned joined.
- `RunEventRelay`: skip the state PATCH when an adopted relay is already running.

With this, a re-registration against a fully-surviving orphan chain (channel + NC + ERC) self-heals end to end instead of requiring manual cleanup first.

### DIFF
--- a/providers/salesforce/events.go
+++ b/providers/salesforce/events.go
@@ -289,6 +289,21 @@ func (c *Connector) CreateEventRelayConfig(
 			return recoverDuplicateByDeveloperName[EventRelayConfig](ctx, c, "EventRelayConfig", cfg.FullName)
 		}
 
+		// Salesforce checks channel-binding integrity before name uniqueness, so
+		// a surviving relay carrying our own DeveloperName (left by an earlier
+		// partial teardown) is reported as FIELD_INTEGRITY_EXCEPTION — never as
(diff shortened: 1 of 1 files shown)


AmandaYao00 (commented):
LGTM!

PR #3328, opened 2026-08-12
[CON-39] feat: update basecamp OAuth requirements

Jkarage:
The blocker is gone. Basecamp rewrote Launchpad's OAuth to accept standard OAuth 2.0 parameters. One TokenURL now serves both grants, so a plain oauth2.Config works no per-grant URL trickery needed.

# Conventions
- [ ] DocsURLs actually link to user-friendly documentation (do not link to very technical documentation)

## GET
Screenshot 2026-08-12 at 09 26 55

## POST
Screenshot 2026-08-12 at 09 34 46

## PUT
Screenshot 2026-08-12 at 09 30 05

## DELETE
Screenshot 2026-08-12 at 09 35 26

## Invalid requests
Screenshot 2026-08-12 at 09 28 03

### DIFF
--- a/providers/basecamp.go
+++ b/providers/basecamp.go
@@ -46,7 +46,7 @@ func init() {
 				{
 					Name:        "workspace",
 					DisplayName: "Account ID",
-					Prompt:      "our account ID is the unique number located in your Basecamp URL. For example, in `https://app.basecamp.com/6258233/projects`, the account ID is 6258233.`",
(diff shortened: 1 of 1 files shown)


providers/basecamp.go
```diff
@@ -43,6 +46,8 @@ func init() {
 				{
 					Name:        "workspace",
 					DisplayName: "Account ID",
+					Prompt: "The number in your Basecamp URL. In " +
```
RajatPawar: `Your account ID is the unique number located in your Basecamp URL. For example, in `https://app.basecamp.com/6258233/projects`, the account ID is 6258233.`

Please reword this as such, and use the ` character around the link

RajatPawar (approved):

PR #3330, opened 2026-08-12
[ENG-4183] feat(acculynx): add Representative→Job and Estimate→Job associations, hydrate estimates from the detail endpoint

ejoo:
### What changed

- Representative→Job: attached in the jobs/representatives fan-out (fetchChildPages, same mechanism as Appointment→User). The representative body has no jobId (docs (https://apidocs.acculynx.com/reference/getrepresentativesforjob)), so the edge uses the fan-out parent job id. Reference shape, jobs are batch-readable, the server hydrates.
- Estimate→Job: from the embedded job stub + top-level isPrimary on every /estimates row, emitted under jobs with isPrimary in ProviderAssociationMetadata. Pure reshape, no extra fetch.
- Estimate hydration: /estimates rows are stubs (id, isPrimary, job); all substantive fields live only on GET /estimates/{estimateId}. Each row is hydrated via one detail call with ?includes=createdBy,modifiedBy,sections (concurrency-capped fan-out, custom-fields pattern). Detail populates Fields, Raw stays the list payload, a 404 mid-read keeps the thin row. job is not expanded that's the association.
- Metadata: generator merges the estimate detail schema into estimates (4 → 16 fields); api3's extractor can't read the single-object response, so properties come directly from the spec.

### Resolved open question

?includes=financials on the list can't avoid the per-record fetch: the docs support only job on the list, and live probes with financials/sections/all/details/full are silently ignored. Cost: 1+N requests per page (N ≤ 25, concurrency 4, within the 10 req/s key limit).

### Unit: go test ./providers/acculynx/...

image

### Live: go run ./test/acculynx/associations (needs acculynx-creds.json)

- 27/27 representatives carry the job edge (ObjectId matches the parent job in _link)
- 8/8 estimates carry the jobs edge with isPrimary
- 8/8 estimates hydrate financials/createdDate/estimateNumber

image
image
image
image

### DIFF
--- a/providers/acculynx/estimates.go
+++ b/providers/acculynx/estimates.go
@@ -27,6 +27,10 @@ import (
 // Like custom fields, hydration mutates only the marshalled record that feeds
 // ReadResultRow.Fields — Raw stays the untouched list payload.
 //
+// Hydration is opt-in per customer via ReadParamsOpts.HydrateEstimates (see
+// read.go) — the extra call per record is not imposed on installs that only
+// need the stub fields or the Estimate->Job association.
+//
 // References:
 //   - List: https://apidocs.acculynx.com/reference/getestimates
 //   - Detail: https://apidocs.acculynx.com/reference/getestimatebyid
--- a/providers/acculynx/estimates.go
+++ b/providers/acculynx/estimates.go
@@ -6,10 +6,10 @@ import (
 	"maps"
 
 	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/logging"
+	"github.com/amp-labs/connectors/common/readhelper"
 	"github.com/amp-labs/connectors/common/urlbuilder"
-	"github.com/amp-labs/connectors/internal/jsonquery"
 	"github.com/amp-labs/connectors/internal/simultaneously"
-	"github.com/spyzhov/ajson"
 )
 
 // AccuLynx estimate hydration:
@@ -24,8 +24,9 @@ import (
 // detail call per estimate, following the same per-record fan-out shape as
 // the custom-fields value fetch (see custom.go).
 //
-// Like custom fields, hydration mutates only the marshalled record that feeds
-// ReadResultRow.Fields — Raw stays the untouched list payload.
+// The detail is provider data, so it lands in Raw as well as Fields — raw
+// stays "the object as AccuLynx gave it", just assembled from two calls.
+// Rows whose detail fetch was skipped keep the list stub in both.
 //
 // Hydration is opt-in per customer via ReadParamsOpts.HydrateEstimates (see
 // read.go) — the extra call per record is not imposed on installs that only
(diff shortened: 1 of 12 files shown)


providers/acculynx/read.go
```diff
@@ -224,18 +224,41 @@ func (c *Connector) parseReadResponse(
...
 	var transformer common.RecordTransformer
 
-	if usesCustomFields(params.ObjectName) {
+	switch {
+	case usesCustomFields(params.ObjectName):
 		transformer, err = c.buildCustomFieldsTransformer(ctx, params, resp)
 		if err != nil {
 			return nil, err
 		}
+	case params.ObjectName == objectEstimates:
```
anushat05: This runs for every customer reading `estimates`, and it costs up to 25 extra
API calls per page. Better to let the server switch it on per customer.

There's already a way: `ReadParams.Opts` — a free-form field the server sets and the connector reads. Gong does this today: see `providers/gong/read.go` (`ReadParamsOpts` struct) and `providers/gong/flows.go` (reads it via type assertion).

Shape here:

    type ReadParamsOpts struct {
        HydrateEstimates bool
    }

    case params.ObjectName == objectEstimates && shouldHydrateEstimates(params):

`shouldHydrateEstimates` does the type assertion and returns false when `Opts`
is unset. Default off = nothing changes for existing installs.

The server change is small. This side can land first; it's a no-op until the
server sets the flag.

providers/acculynx/estimates.go
```diff
@@ -0,0 +1,168 @@
...
+		idx, id := i, estimateID
+
+		jobs[idx] = func(ctx context.Context) error {
+			detail, err := c.fetchEstimateDetail(ctx, id)
+			if err != nil {
+				if isNotFound(err) {
+					return nil
+				}
+
+				return fmt.Errorf("fetch estimate %s detail: %w", id, err)
```
anushat05: 404 is handled nicely above, but any other error fails the whole page. One estimate returning a 429 or 500 kills a read where the other 24 worked.

Real risk here: up to 25 detail calls per page at concurrency 4, and AccuLynx allows 10 requests/sec per key. The base HTTP client doesn't retry 429s.

Suggest treating transient errors like the 404 case, keep the thin row and move on, so a partial result beats no result. Worth a log line either way, since today a skipped row looks the same as an estimate with no data.

providers/acculynx/estimates.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	return *detail, nil
+}
+
+// mergeEstimateDetail returns a RecordTransformer that overlays the record's
+// detail payload onto the thin list stub. Detail wins on overlapping keys —
+// it is the authoritative superset. Records without a detail (404 mid-read)
+// pass through unchanged.
+func mergeEstimateDetail(details map[string]map[string]any) common.RecordTransformer {
```
anushat05: The detail only lands in `Fields`, so `Raw` keeps just `{id, isPrimary, job}`.

That flips what `raw` means. `raw` = the object as the provider gave it, `fields` = the subset the customer picked. Here `fields` is complete and `raw` is nearly empty. We did fetch the detail from AccuLynx, so it belongs in `raw` too.

This matters when debugging, `raw` is where people look for a missing field.

anushat05 (changes requested):

providers/acculynx/read.go
ejoo: Good catch, done! 
Gated behind `ReadParamsOpts.HydrateEstimates` (gong pattern), default off.

providers/acculynx/estimates.go
ejoo: Fair point, all failures now skip-and-log like the 404 path, only cancellation aborts. Added a 500 test.

providers/acculynx/estimates.go
ejoo: You're right, raw is where I'd look too . Detail now lands in Raw, Fields re-derived from it.

anushat05 (approved):

PR #3341, opened 2026-08-14
fix(stripe): Add object to event type mapping

Cobalt0s:
# Description

This PR improves Stripe webhook subscription support, webhook verification lifecycle handling, and test coverage.

### Explicit Stripe event mappings

- Added an explicit mapping between Ampersand object names and Stripe webhook event types.
- Replaced string-based event-name manipulation with a definition-based lookup.

### Stripe event mapping data

- Added scripts under `scripts/scraper/stripe/event-types` to extract event type names from Stripe's official documentation and transform them into the format used by the connector.
- Used the generated data to build and validate the event mapping in the connector.
- The generated data also provides an auditable list of Stripe event types and verifies coverage against the existing Ampersand object names.

### Webhook verification lifecycle

Updated the live subscription test scenario to support verification parameters that are only available after `Subscribe` completes.

The webhook server now has two lifecycle stages:

1. **Pre-subscription stage**
   - The server starts before `Subscribe` so that the generated webhook URL can be registered.
   - Provider-specific interceptors can handle webhook validation or handshake requests.
   - Connector webhook-message verification is not enabled yet.

2. **Post-subscription stage**
   - The subscription result is passed to a verification-parameter builder.
   - The builder can derive connector-specific verification parameters, such as the Stripe webhook secret.
   - The server is then configured to verify incoming webhook messages through the connector.
   - Verified events are passed to the webhook processor.

This allows interceptors to run during the entire flow while ensuring that `conn.VerifyWebhookMessage` is called only after the required subscription-dependent configuration is available.

The raw webhook runner is unchanged. It continues to load verification parameters from the provider credentials file and configures the server immediately.

### Stripe subscription result

Redesigned `stripe.SubscriptionResult`.

Previously, the result stored a mapping between object names and complete Stripe webhook API responses. This did not reflect the actual Stripe resource model because the subscription creates one webhook endpoint containing the complete event list rather than a separate webhook resource for each object.

The result now stores only the data required by the connector:

- Webhook ID.
- Webhook signing secret.
- The configured subscription events.

The complete Stripe API response is no longer retained unnecessarily. The webhook secret is also stored directly instead of being discovered by iterating over the previous object-to-response map.

### Tests and test utilities

- Updated unit tests to use the correct object names.
- Added explicit subscription-result expectations to verify the desired connector state.
- Moved reusable comparator logic into shared test utilities.
- Extended `mockcond.Body` to compare `application/x-www-form-urlencoded` request bodies semantically.
- Form body comparisons now:
  - Decode percent-encoded values.
  - Treat `+` as a space.
  - Ignore whitespace and newlines around `&`.
  - Ignore parameter order.
  - Ignore the order of repeated values.
  - Preserve duplicate values and their counts.

## Testing

- Updated unit tests for Stripe event mappings.
- Updated subscription result comparisons.
- Updated webhook and form-encoded request tests.
- Validated live subscription scenarios against the real Stripe API.
- Verified webhook interception and post-subscription message verification flows.

### Webhook

The raw webhook consumer flow remains unchanged.

Peek 2026-08-15 03-31

### Subscribe with Events

The live `ValidateSubscribeReceiveEvents` flow now requires verification parameters to be derived after `Subscribe` when the connector needs subscription-dependent data. Existing scenarios that do not require verification can leave the builder unset.
See slack for details.

### Subscribe CUD lifecycle

image

image


Cobalt0s:
* **#3341** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3341?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3341?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/amp-labs?ref=merge-queue-instructions-in-stack-comment&prId=3341).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

providers/stripe/internal/webhook/event-names.go
```diff
@@ -0,0 +1,375 @@
...
+// object-event definitions. An unsupported object returns
+// common.ErrObjectNotSupported. An unsupported operation for a known object
+// returns common.ErrSubscribeEventNotSupportedForObject.
+func resolveStripeEventName(objectName common.ObjectName, eventType common.SubscriptionEventType) (string, error) {
+	definition, ok := eventDefinitions[objectName.String()]
+	if !ok {
+		return "", fmt.Errorf("%w: object %v", common.ErrObjectNotSupported, objectName)
+	}
+
+	switch eventType { // nolint:exhaustive
```
jlimatampersand, line 335: any posibility of eventType == common.SubscriptionEventTypeOther?

providers/stripe/subscribeDelete.go
```diff
@@ -49,35 +44,3 @@ func validateSubscriptionResult(result common.SubscriptionResult) (*Subscription
 
 	return subscriptionData, nil
 }
-
-// extractEndpointID extracts the real endpoint ID shared by all composite IDs.
-func extractEndpointID(subscriptionData *SubscriptionResult) (string, error) {
```
jlimatampersand: Why did you have to strip off this logic?

jlimatampersand (approved):

providers/stripe/internal/webhook/event-names.go
Cobalt0s: I error out on that event type. I am not sure how this should be resolved.

The "other" should be mapped to something. Should we map to "everything" and return a complete list of events an object can be subscribed to? That would be more a SubscriptionEventTypeAll.

PR #3361, opened 2026-08-19
[CON-3508] feat(pipedrive): support mailboxes/mailThreads read

Jkarage:
- Adds support to read `mailboxes/mailThreads` (adding inbox as hardcoded folder)
- Adds mailThreads as an alias (instead of the whole `mailboxes/mailThreads`) 
- Failed to create/sync my mailbox so there would be a number of mailThreads to return when reading, cause the account we have requires upgrading for that. But I have printed the API response from pipedrive (showing zero records response below)
638251227-285bac52-6c83-4bb2-8f4c-d9d09470154b

Live Tests
Metadata
Screenshot 2026-08-19 at 16 37 59

Read:
Screenshot 2026-08-19 at 16 35 25

### DIFF
--- a/providers/pipedrive/internal/legacy/README.md
+++ b/providers/pipedrive/internal/legacy/README.md
@@ -26,6 +26,7 @@ Pipedrive API version : v1
 | Lead Labels | leadLabels | Read,Write |
 | Lead Sources | leadSources | Read |
 | Legacy Teams | legacyTeams | Read,Write |
+| Mail Threads | mailThreads | Read |
 | Notes | notes | Read,Write |
 | Note Fields | noteFields | Read |
 | Organizations | organizations | Read,Write |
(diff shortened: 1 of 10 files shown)


laurenzlong (changes requested):
Thanks @Jkarage! I do like the idea of using `mailThreads` and not `mailboxes/mailThreads` since there are no name collisions here. Can you actual revise ListObjectMetadata to return `mailThreads`? That way the canonical name that builder will use is `mailThreads` (but we can still accept `mailboxes/mailThreads`)

Also, I noticed that the changes were made in the `internal/legacy` folder, can you remind me why this directory is called legacy?

Jkarage:
@laurenzlong 
After these changes now `mailThreads` is the canonical name. But we still accept `mailbox/mailthreads` as well.
Metadata:
Screenshot 2026-08-22 at 09 10 59

On the legacy: 
Pipedrive has been migrating its API to v2 (slowly), and v2 now supports a few objects + incremental sync.
We currently support both v1 and v2 as separate modules. However, naming them “v1” and “v2” felt awkward, and we used `legacy` for v1 and `crm` for v2.
We only made changes to the legacy module (v1), because `mailThreads` is currently supported only in that API. 
Since legacy is the default module, builders do not need to specify it explicitly.

laurenzlong (approved):

PR #3366, opened 2026-08-19
feat(slack): Read provider side filtering

Cobalt0s:
# Description

* Fixed object identifiers. Since not all objects use an `id` field, the implementation now relies on the **Read** info to determine the correct identifier specification.
* Updated Read operations to use the **Since** and **Until** query parameters where supported. See `files` object in live tests.
* During testing, discovered that `chat.scheduledMessages.list` interprets these query parameters as the dates when messages are scheduled to be published, rather than their creation dates. As a result, the query parameters are not used for this operation; filtering is performed on the connector side using the `date_created` property.

# Live Tests

Filtering using query params (object `files`):

Peek 2026-08-21 02-06


Cobalt0s:
* **#3322** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3322?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3366** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3366?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3366?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3319** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3319?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/amp-labs?ref=merge-queue-instructions-in-stack-comment&prId=3366).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

providers/slack/handlers.go
```diff
@@ -126,13 +148,15 @@ func (c *Connector) parseReadResponse( //nolint:unparam
...
 	}
 
+	idField := info.GetIdFieldQuery()
+
 	if info.TimeFilterField != "" {
 		return common.ParseResultFiltered(
 			params,
 			response,
 			nodeRecords(info.ResponseField),
 			makeTimeFilter(info),
```
jlimatampersand, line 176: **[bug]** `makeTimeFilter` hardcodes `readhelper.TimestampFormatUnixSec` for every object with a `TimeFilterField`, but Slack's `conversations.list` channel `updated` field is epoch **milliseconds**. Parsing ms as seconds yields year ~55148 timestamps: with `Until` set every record fails the boundary check (read returns 0 rows); with only `Since` set every record passes (filtering is a no-op). This PR added per-object `RangeTimestampFormat` for query params — the connector-side filter should use a per-object format the same way instead of the hardcoded constant. (Pre-existing on main, but this PR's new field is the natural fix; the unit test masks it by feeding seconds-scale values for a field Slack returns in ms.)

providers/slack/internal/mappings/scope-none.go
```diff
@@ -11,9 +13,7 @@ func init() {
 				Method:          http.MethodPost,
 				ResponseField:   "scheduled_messages",
 				ResponseIdField: "id",
-				TimeFilterField: "",
-				SinceQP:         "oldest",
-				UntilQP:         "latest",
+				TimeFilterField: "date_created",
```
jlimatampersand, line 18: **[bug risk]** Adding `TimeFilterField: "date_created"` means a single `scheduled_messages` record with a missing/null `date_created` now aborts the entire read: `extractUnixSecTimestamp` returns `ErrTimestampKeyNotFound` when `IntegerOptional` yields nil, and the error propagates out of `ParseResultFiltered`, failing the whole page. Pre-PR this object had no connector-side filter, so such a page synced fine. Consider tolerating records with a missing timestamp field (skip or include) rather than hard-failing the read.

providers/slack/support-read_test.go
```diff
@@ -29,16 +29,21 @@ func TestObjectResponseField(t *testing.T) {
...
 		"conversations":                     "updated",
 		"conversations.listConnectInvites":  "date_last_updated",
 		"conversations.requestSharedInvite": "date_last_updated",
-		"files":                             "created", // TODO must use Since/Until instead of connector side.
 		"usergroups":                        "date_update",
 		"users":                             "updated",
 	}
 
 	readRequestMap := map[string]mockcond.Condition{
+		"admin.apps.activities": mockcond.And{
```
jlimatampersand: **[test gap]** This `admin.apps.activities` entry is dead: the object is absent from `responseFields`, and `createTestsForReadMappings` iterates `responseFields` keys only, so no test case is generated and this `mockcond.Body` assertion never runs. Since no mapping defines `SinceQP` for any POST object either, the new POST-body Since/Until block in `buildReadRequest` ships with zero coverage and zero live users — either add `admin.apps.activities` to `responseFields` (with a real mapping) or drop this entry.

common/readhelper/time.go
```diff
@@ -121,3 +122,24 @@ func (b TimeBoundary) After(params common.ReadParams, timestamp time.Time) bool
...
+	case TimestampFormatUnixSec:
+		return strconv.FormatInt(input.Unix(), 10)
+	case TimestampFormatUnixMs:
+		return strconv.FormatInt(input.UnixMilli(), 10)
+	case TimestampFormatUnixUs:
+		return strconv.FormatInt(input.UnixMicro(), 10)
+	default:
+		// Treat as a standard Go layout string. Example: "time.RFC3339".
+		// Normalize to UTC so query params are always in UTC.
+		return input.UTC().Format(formatKind)
```
jlimatampersand, line 140: **[robustness]** The default branch treats any unrecognized `formatKind` — including the empty string — as a Go layout. A mapping that sets `SinceQP`/`UntilQP` but forgets `RangeTimestampFormat` (the exact shape of the pre-PR mappings) makes `buildReadRequest` call `TimeFormat(since, "")` → `input.UTC().Format("")` → `""`, silently sending an empty `ts_from=` to Slack and quietly breaking incremental sync. Nothing validates the `SinceQP`↔`RangeTimestampFormat` pairing at init time; consider validating there or returning an error for an empty kind.

providers/slack/internal/mappings/scope-none.go
jlimatampersand, line 18: **[efficiency]** Dropping `SinceQP: "oldest"`/`UntilQP: "latest"` in favor of connector-side filtering forces full-history pagination on every incremental read: `makeTimeFilter` passes `readhelper.Unordered`, and `MakeTimeFilterFuncWithZoom` never sets `stopPaging` for `Unordered`, so a 1-hour `Since` window still fetches and parses every page of every scheduled message in the workspace. If `oldest`/`latest` were dropped because they filter on scheduled-post time rather than `date_created` (per the PR description), consider layering the connector-side `date_created` filter on top of a provider-narrowed window instead of an exhaustive scan.

common/readhelper/time.go
```diff
@@ -121,3 +122,24 @@ func (b TimeBoundary) After(params common.ReadParams, timestamp time.Time) bool
...
+// TimeFormat formats input according to the given timestamp format kind.
+// Supported kinds:
+//   - TimestampFormatUnixSec
+//   - TimestampFormatUnixMs
+//   - TimestampFormatUnixUs
+//   - time.RFC3339, time.RFC3339Nano, or any Go layout string.
+func TimeFormat(input time.Time, formatKind string) string {
+	switch formatKind {
+	case TimestampFormatUnixSec:
+		return strconv.FormatInt(input.Unix(), 10)
```
jlimatampersand: **[boundary semantics]** `strconv.FormatInt(input.Unix(), 10)` floors sub-second `Since` values. For `files`, the old connector-side filter compared `time.Unix(created, 0)` against the full-precision `Since` (so `created=1050` with `Since=1050.5s` was excluded), while the new path sends `ts_from=1050`, which Slack treats inclusively — already-synced boundary records get re-returned on every incremental sync whenever `Since` carries sub-second precision. Worth documenting or rounding up (`ceil`) for the since side.

common/readhelper/filter.go
```diff
@@ -277,9 +291,9 @@ func extractUnixMsTimestamp(nodeRecord *ajson.Node, timestampKey string, zoom ..
 	return &t, nil
 }
 
-// extractUnixSecTimestamp reads an integer seconds epoch field and converts it
-// to a time.Time using time.Unix.
-func extractUnixSecTimestamp(nodeRecord *ajson.Node, timestampKey string, zoom ...string) (*time.Time, error) {
+// extractUnixMsTimestamp reads an integer microsecond epoch field and converts it
```
jlimatampersand: **[cleanup]** Doc comment copy-paste: this says `extractUnixMsTimestamp` but the function is `extractUnixUsTimestamp`. More broadly, `extractUnixSecTimestamp`/`extractUnixMsTimestamp`/`extractUnixUsTimestamp` are three copy-pasted bodies differing only in the `int64 → time.Time` conversion — one helper taking a `func(int64) time.Time` converter (`time.UnixMilli`, `time.UnixMicro`, sec wrapper) collapses them and keeps the nil-check/error-wrap logic in one place.

providers/slack/handlers.go
```diff
@@ -103,6 +105,16 @@ func (c *Connector) buildReadRequest(ctx context.Context, params common.ReadPara
 			body["cursor"] = params.NextPage.String()
 		}
 
+		if !params.Since.IsZero() && info.SinceQP != "" {
```
jlimatampersand: **[cleanup]** The Since/Until wiring is copy-pasted four times (Since+Until in both the POST-body and GET-query branches), which is what forces the new `nolint:cyclop`. A small helper yielding the `(param, value)` pairs — or a shared readhelper function, since other declarative-mapping providers will need identical wiring — removes the duplication and the lint suppression, and gives a single edit site for fixes like the empty-`RangeTimestampFormat` guard.

common/readhelper/time.go
```diff
@@ -121,3 +122,24 @@ func (b TimeBoundary) After(params common.ReadParams, timestamp time.Time) bool
...
+
+// TimeFormat formats input according to the given timestamp format kind.
+// Supported kinds:
+//   - TimestampFormatUnixSec
+//   - TimestampFormatUnixMs
+//   - TimestampFormatUnixUs
+//   - time.RFC3339, time.RFC3339Nano, or any Go layout string.
+func TimeFormat(input time.Time, formatKind string) string {
+	switch formatKind {
+	case TimestampFormatUnixSec:
```
jlimatampersand, line 133: **[reuse]** The `unix_sec` branch re-implements the existing `datautils.Time.Unix` helper (`internal/datautils/timing.go:32`, `strconv.FormatInt(input.Unix(), 10)`). Two homes for epoch-string formatting can drift — consider calling the existing helper or consolidating.

common/readhelper/filter.go
```diff
@@ -13,17 +13,24 @@ import (
...
-// holds a Unix epoch value in seconds (int64), rather than a formatted string.
-// Internally this maps to time.Unix, allowing reuse of MakeTimeFilterFunc
-// for providers that store timestamps as integer seconds.
-const TimestampFormatUnixSec = "unix_sec"
+// TimestampFormatUnixUs is a special format string that signals the timestamp field
+// holds a Unix epoch value in microseconds (int64), rather than a formatted string.
+// Internally this maps to time.Unix(sec, nsec), allowing reuse of MakeTimeFilterFunc
+// for providers that store timestamps as integer microseconds.
+// Microsecond is referred to as `us`. See: https://en.wikipedia.org/wiki/Microsecond.
+const TimestampFormatUnixUs = "unix_us"
```
jlimatampersand: **[dead code]** `TimestampFormatUnixUs` (and its extract/format branches) has no consumer anywhere in the repo — `unix_us` is referenced only inside `common/readhelper` itself. Untested speculative branches in shared infrastructure add review surface (this one's doc comment is already wrong); consider adding it when a provider actually needs microsecond epochs.

jlimatampersand (changes requested):

providers/slack/handlers.go
Cobalt0s: Makes sense. I will fix this bug and introduce the per `TimeFilterField` time format. I went along with `TimestampFormatUnixSec` and didn't question this field. I will check every applicable object.

Cobalt0s: Makes sense. I will fix this bug and introduce the per `TimeFilterField` time format. I went along with `TimestampFormatUnixSec` and didn't question this field. I will check every applicable object.

Turns out only `conversations.list` was special. I explicitly added a comment to each `info` spec with converted time and `unit of time`.

common/readhelper/filter.go
Cobalt0s: During dev I added this and then figured out it is not needed. I thought of keeping it. I will remove it, you are right it is not justified if not used and no explicit plans for it.

common/readhelper/time.go
Cobalt0s: I was thinking that maybe `datautils.Time` should belong to `readhelper`. The latter package is more recent and it looks like the time formatting is only important for the read. At the time of development I thought about it but was not sure about the ideal path for it. I will wire up the `readhelper` with `datautils` for the time being.

common/readhelper/time.go
Cobalt0s: This statement is true globally for all connector cases. I am not sure on the best approach here given the divercity of time formats and de factor differences in API behaviour within even single provider.

Essentially, both Since and Until are floored. For the Since it may mean duplicates which in general is always better than missing data. For Until that could be a problem. I think going slightly before Since and after the Until points in time would be better (floor the Since and ceil the Until).

Is there any action I should take on this?

common/readhelper/time.go
Cobalt0s: It is assumed that developer using `readhelper.TimeFormat` will pass valid time format. This is not really a "run time" value, but a hardcoded value.

The time format should never be empty if the developer decided to use the utility.

providers/slack/internal/mappings/scope-none.go
Cobalt0s: The code embodies the contract, expectations on the name, location of the field and its format.
Connector failing means the underlying provider API has changed and connector needs to be reimplemented to fix the breaking change by provider.

The issue raised should be looked at comprehensively. Currently, it is a hard failure. Alternatively, we could log the issue and "silently" switch off the connector-side filtering. But I imagine that is not so easy as that abruptly breaks the contract/expacation between Ampersand and clients. Ex: one day the records are filtered and the other everything is returned. It still might be better than Read returning errors.

I think we would greatly benefit from error enrichment at several ciritcal func stages. So instead of just logging that timestamp cannot be extracted the error should include provider, object, operation, feature maybe other metadata to indicate precisly what happened.

Cobalt0s: This object doesn't have the query params that we would classify as Since/Until.
It is similar in some sense with Gong provider. The time qurying here is to be done in future. I would treat it as "general" fields which just happen to be time based.

providers/slack/handlers.go
Cobalt0s: I would rather preserve the explicit nature of conditionals then pulling them into the function.
As a small improvement just pulling the POST branch into a function instead of focusing on Since/Until.

Image

providers/slack/support-read_test.go
Cobalt0s: Removed and added it to the last PR which relies on it.

jlimatampersand (approved):

PR #3387, opened 2026-08-25
[CON-3512] fix(acculynx): paginate with pageStartIndex and terminate on the response envelope

ejoo:
## Problem

`Read` never terminated on wide sweeps: AccuLynx returns a full first page at any `recordStartIndex`, so the partial-page check — the connector's only exit — never fired. Production climbed to `recordStartIndex=565050` on an account of 163 jobs, re-emitting page 1 as duplicates.

## Offset parameter findings (live, per ticket step 1)

| Request | Records returned | Envelope echo |
|---|---|---|
| `?pageSize=25&recordStartIndex=25` | same as page 1 | `"pageStartIndex": 0` |
| `?pageSize=25&recordStartIndex=565050` | same as page 1 | `"pageStartIndex": 0` |
| `?pageSize=25&pageStartIndex=25` | **advances** | `"pageStartIndex": 25` |

Same result with `dateFilterType=ModifiedDate`. `recordStartIndex` is silently ignored on all 13 endpoints tested, not just `/jobs`; `pageStartIndex` is the only working offset. The spec documents `recordStartIndex` and is wrong.

## Fix

- Moved all paginated objects (19 entries + the default) to `pageStartIndex`; deleted `paginationOffsetRecord`/`paginationPageNumber` so nothing can be wired to an ignored param again.
- `makeNextPage` now stops when the echoed `pageStartIndex` ≠ the requested offset, when consumed records reach `count`, or on a short page. Deviation from the ticket: the echo is compared to the *requested* offset, not the previous page's — stateless, catches the ignore one page earlier.
- Added `maxTopLevelPages = 400` (mirrors `maxChildPagesPerParent`), erroring with the object name. It applies only when the response reports no `count`, so accounts over 10k records aren't capped.
- Applied the same exit rule to the two hand-rolled loops in `custom.go`, which had the same bug and spun inside a single `Read` call.

## Verification

- `TestReadPaginationTermination`: runaway, count-exit, non-advancing echo, page cap. The runaway guard fails against pre-fix code as required — pre-fix, a single `Read` never returns (`panic: test timed out` in `fetchCustomFieldDefinitions`).
- Live before/after, same account: jobs went from looping indefinitely / 25 unique / 25 duplicates to 7 pages / 165 unique / 0 duplicates, matching the API's reported count=165 exactly. All 33 read objects swept: looping=0, no regressions.
- Existing tests green, `make lint` 0 issues.

### pre-fix test timeout
image 

### suite passing 
image
 
### before/after audit 
image

 image

### 33-object sweep 
image
duplicates=2 are rows sharing a field-definition id by design (the _link differs per record); errors=3 are pre-existing bare-array objects that fail key not found both before and after this change. looping=0 is the number this PR moves.


providers/acculynx/incremental.go
Cobalt0s: Could you rename the timeKey to `TimeFilterField` or alike, this is to indicating that it is used by connector side filtering.

providers/acculynx/read.go
```diff
@@ -47,6 +52,12 @@ const (
 	// with a clear error pointing the caller at Since/Until — see fetchChildPages.
 	maxChildPagesPerParent = 200
 
+	// maxTopLevelPages bounds pagination for a single top-level object, the
+	// counterpart to maxChildPagesPerParent for nested fetches. At 25 records
+	// per page it allows 10,000 records before erroring. It applies only when
+	// the response reports no count — see makeNextPage.
+	maxTopLevelPages = 400
```
Cobalt0s: So this field is the field which means how many pages can be requested before it terminates. Maybe the name could be changed to `maxPagesForUnknownTotal` or something similar. The "top level" is a bit not clear at least to me.

providers/acculynx/read_test.go
```diff
@@ -1167,3 +1167,185 @@ func constructTestReadConnector(serverURL string) (*Connector, error) {
...
+			// answering every offset with a full page and a count it never
+			// reaches. Read must still stop.
+			Name: "Runaway server that ignores the offset terminates",
+			Input: common.ReadParams{
+				ObjectName: "jobs",
+				Fields:     connectors.Fields("id"),
+				PageSize:   25,
+				NextPage:   testconn.URLTestServer + "/api/v2/jobs?pageSize=25&pageStartIndex=25",
+			},
+			Server: mockserver.Fixed{
```
Cobalt0s: Please, use conditional mockserver. You should assert the `mockcond.Path` and `mockcond.QueryParam`.

Also, you don't really care about the data in these tests. We care about `Rows`, `NextPage`, `Done` -- all is achieved by using `testconn.ComparatorPagination` instead of verbose `testconn.ComparatorSubsetRead`.

The tests will become easier. You are checking the request building and the pagination side of the connector output.

providers/acculynx/read_test.go
```diff
@@ -1167,3 +1167,185 @@ func constructTestReadConnector(serverURL string) (*Connector, error) {
...
+				Setup:  mockserver.ContentJSON(),
+				Always: mockserver.Response(http.StatusOK, paginationRunawayResponse),
+			}.Server(),
+			Comparator: testconn.ComparatorSubsetRead,
```
Cobalt0s: You don't really care about the data in these tests. We care about `Rows`, `NextPage`, `Done` -- all is achieved by using `testconn.ComparatorPagination` instead of verbose `testconn.ComparatorSubsetRead`.

The tests will become easier. You are checking the request building and the pagination side of the connector output.

Also the server can be shortened to one liner as utility for all tests. The only difference is the response body.
Maybe even instead of using the files stored under `test` repo inlining data here would make it more readable.

Ex: 
```
buildServer(`{
  "count": 163,
  "pageSize": 25,
  "pageStartIndex": 0}`, 25)
```
And signature:
```
func buildServer(envelope map[string]int, numItemsToInclude int)
```

Then when reading a test one would see the connector input, envelope and connector output in one place. The items itself is dummy data.

Cobalt0s (approved):
Approving to unblock.

I wonder what objects would utilize `maxTopLevelPages`?
In general I try to avoid relying on the pagination envelope data, because it can be not consistent across the endpoints. I think I understood the algorithm, you are trying to validate that `pageStartIndex` matches the expecation -- aka what we have previously requested. All that is built on "count" envelope property without one the equation breaks so you rely on `maxTopLevelPages` as a backup.

### Suggestion
Since, such a possibility still exists, I think we should use [AggregateNextPage](https://github.com/amp-labs/connectors/blob/main/common/readhelper/aggregation.go#L12) to avoid relying on provider all together. I think the next page that connector uses can carry additional info. For instance the next page would become: `readhelper.AggregateNextPage[RecordSummary]`, where:
```
type RecordSummary struct {
  LastRecordId string
}
```
So this next page token will have `Context.LastRecordId, Context.FirstRecordId` and then for the current page you will check if the ids from last page are present. If they are then we are getting "duplciates". 

Ex:
* first page {1,2,3,4,5} then next page token: {Value: http://....., Context: {First: 1, Last: 5}}
* second page {6,7,8,9,10} --> all ids are new so continue
* second page {1,2,3,4,5} --> previous result had 1 and 5 and we are seeing it again, therefore we cannot proceed return next page as empty.

# Provider side filtering
I see [jobs](https://apidocs.acculynx.com/reference/getjobs) can use Since/Until param but currently doesn't. You should go over each object in the docs and determine what query params, what format can be used by each object.

cc @RajatPawar

providers/acculynx/incremental.go
ejoo: Done! 🚀

providers/acculynx/read.go
ejoo: Done! 🚀

providers/acculynx/read_test.go
ejoo: Done!! switched to ComparatorPagination, added buildPaginationServer(envelope, numItems), fixtures deleted and inlined. 👍

providers/acculynx/read_test.go
```diff
@@ -1167,3 +1170,159 @@ func constructTestReadConnector(serverURL string) (*Connector, error) {
...
+		},
+		{
+			// Offset honoured: the first page must hand back a cursor.
+			Name: "Count-based exit continues while records remain",
+			Input: common.ReadParams{
+				ObjectName: "jobs",
+				Fields:     connectors.Fields("id"),
+				PageSize:   25,
+			},
+			Server:     buildPaginationServer(t, map[string]int{"count": 50, "pageSize": 25, "pageStartIndex": 0}, 25),
```
Cobalt0s, line 1239: Thanks, the test is much more readable.

ejoo: Thanks!

PR #3400, opened 2026-08-28
feat(callRail): add support to read sms-threads

Jkarage:
Adds support to read sms-threads (found this while testing)

Live Tests:
Screenshot 2026-09-01 at 12 39 12
Screenshot 2026-09-01 at 12 39 25

### DIFF
--- a/providers/callrail/handlers.go
+++ b/providers/callrail/handlers.go
@@ -4,16 +4,13 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"net/http"
-	"strconv"
 
 	"github.com/amp-labs/connectors/common"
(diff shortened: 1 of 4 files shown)


providers/callrail/handlers.go
```diff
@@ -152,30 +149,6 @@ func (c *Connector) buildWriteRequest(ctx context.Context, params common.WritePa
 	return http.NewRequestWithContext(ctx, method, url.String(), bytes.NewReader(jsonData))
 }
 
-func retrieveRecordId(body *ajson.Node) (string, error) {
```
jlimatampersand, line 155: unused?

jlimatampersand (approved):

providers/callrail/handlers.go
Jkarage: Yes, its no longer used.

PR #3432, opened 2026-09-03
fix(salesforce): interpret object-shaped error bodies, pass metadata to catalog

RajatPawar:
Two independent Salesforce fixes, both surfaced while scoping [ENG-4237](https://linear.app/ampersand/issue/ENG-4237). Neither depends on that work — they are live bugs today.

## 1. `interpretJSONError` rejected object-shaped bodies

It unmarshalled only into `[]jsonError`, so any error body that was a bare JSON object failed to decode and we returned `json.Unmarshal failed: ...` — discarding the provider's actual message and leaving the error unclassified.

It now accepts a single object as a one-element array. A body matching neither shape falls through to `common.InterpretError` for status-based handling, rather than surfacing the decode failure as the error itself.

## 2. The connector passed only `workspace` into `ReadInfo`

`oldConstructor` called `providers.ReadInfo(provider, &params.Workspace)`, while `components.NewProviderContext` supplies the **full metadata map** as catalog variables to the module adapters.

A catalog entry referencing any metadata key therefore resolved differently for the two halves of the same connector — the outer `Connector` would see an unsubstituted template while `crmAdapter` and its sub-adapters resolved correctly. Both now receive the same variables, with `workspace` appended last so it wins a name collision.

## Testing

- `TestInterpretJSONError_AcceptsSingleObjectBody` — object body maps to the right sentinel and preserves the raw body
- `TestInterpretJSONError_FallsBackOnUnrecognizedBody` — non-JSON, object-without-message, and empty bodies all fall through to status handling instead of a decode error
- `TestCatalogVariables` — metadata keys are available for substitution, and workspace wins a collision

### DIFF
--- a/providers/salesforce/connector.go
+++ b/providers/salesforce/connector.go
@@ -136,17 +136,19 @@ func oldConstructor(params *parameters) (*Connector, error) {
 // catalogVariables returns the substitution variables used to resolve this
 // connector's ProviderInfo. Connection metadata is included alongside the
 // workspace so that catalog entries may reference any metadata key, matching
-// what components.NewProviderContext supplies to the module adapters. Workspace
-// is appended last so it wins over a metadata entry of the same name.
+// what components.NewProviderContext supplies to the module adapters.
+//
+// The metadata map is copied because the workspace entry is written into it,
+// and callers should not observe that.
 func catalogVariables(params *parameters) []catalogreplacer.CatalogVariable {
-	metadataVars := params.Metadata.GetCatalogVars()
-	vars := make([]catalogreplacer.CatalogVariable, 0, len(metadataVars)+1)
-
-	for _, variable := range metadataVars {
-		vars = append(vars, variable)
+	metadata := maps.Clone(params.Metadata.Map)
+	if metadata == nil {
+		metadata = make(map[string]string)
 	}
 
-	return append(vars, &params.Workspace)
+	metadata[catalogreplacer.VariableWorkspace] = params.Workspace.Name
+
+	return paramsbuilder.NewCatalogVariables(metadata)
 }
 
 // Provider returns the connector provider. For twin providers such as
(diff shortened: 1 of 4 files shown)


providers/salesforce/internal/crm/core/errors.go
```diff
@@ -99,7 +99,17 @@ func (e *fieldNotFoundError) Unwrap() []error {
...
 	if err := json.Unmarshal(body, &errs); err != nil {
-		return fmt.Errorf("json.Unmarshal failed: %w", err)
+		// Salesforce reports errors as a JSON array, but a single JSON object is
+		// also accepted so that payloads shaped that way — such as those authored
+		// by an intermediary in front of Salesforce — are still interpreted.
+		var single jsonError
+		if json.Unmarshal(body, &single) != nil || single.Message == "" {
+			// The body matches neither shape, so fall back to status-based handling
+			// rather than reporting the decode failure as the error itself.
+			return common.InterpretError(res, body)
```
Cobalt0s, line 109: Previously it returned `fmt.Errorf`, I think it is fine to fallback to `common.InterpretError`.

This introduces 2 formats, which is fair: error as array or single item.

providers/salesforce/connector.go
```diff
@@ -116,7 +117,7 @@ func oldConstructor(params *parameters) (*Connector, error) {
 
 	var err error
 
-	conn.providerInfo, err = providers.ReadInfo(conn.Provider(), &params.Workspace)
+	conn.providerInfo, err = providers.ReadInfo(conn.Provider(), catalogVariables(params)...)
```
Cobalt0s, line 120: This can be done via `paramsbuilder.NewCatalogVariables`.

Ideally, the best solution would be to migrate Salesforce to the use of `components`.
Internally it handles ProviderInfo: https://github.com/amp-labs/connectors/blob/main/internal/components/provider.go#L39

Cobalt0s (approved):

providers/salesforce/connector.go
RajatPawar: Good catch, done! I am not sure if our components abstraction supports subscription related functionality well enough to warrant a migration as of this time - we'll end up with a lot of custom code anyway. What do you think?

PR #3439, opened 2026-09-03
[ENG-4220] fix(hubspot): continue GET pagination when search receives a URL next-page token

jlimatampersand:
## Problem

[ENG-4220](https://linear.app/ampersand/issue/ENG-4220/provider-400-bad-request): HubSpot reads fail with `400 VALIDATION_ERROR: unable to deserialize field "after"` when a full next-page URL is sent as the search cursor.

HubSpot's two CRM pagination modes return incompatible next-page tokens:

- Plain GET reads paginate with `paging.next.link` — a **full URL**
- Search (`POST /search`) paginates with `paging.next.after` — a **numeric offset cursor**

A read that starts without `Since`/`Until`/filters runs on the GET path and produces URL tokens. When such a read is interrupted, the server persists the token in the read state and resumes with a freshly computed non-zero `Since` (`PrepareParams` in `server/shared/workflow/read/state.go`). With `Since` set, the connector routes to the search path and drops the URL into the `"after"` body field, which the provider rejects. Note this is not an escaping issue: the GET cursor is an object-id-style value, never a valid search offset (which must stay under 10,000).

## Fix

In `searchCRMObjectsAPI`, detect a URL-shaped `NextPage` token and honor it as the GET-pagination continuation it is — fetch that URL and parse with the GET parser (same response shape, same associations marshaller). The interrupted unfiltered read finishes on the GET path; once exhausted, subsequent runs start filtered search reads normally.

Caveat: pages served through this recovery path don't apply the `Since` filter — same behavior as if the original run had finished uninterrupted, so a resumed backfill may deliver records outside the delta window.

## Testing

- Added regression test `Since read with URL next page token continues GET pagination`; verified it fails without the fix and passes with it
- `go test ./providers/hubspot/...` and `make fix` clean

### DIFF
--- a/providers/hubspot/internal/core/parse.go
+++ b/providers/hubspot/internal/core/parse.go
@@ -24,14 +24,31 @@ Pagination format:
     }
   }
 }
+
+The two extractors below produce next-page tokens of different styles, and the
+values are NOT interchangeable:
+
+  - GET read responses carry both fields: "after" is the record ID of the next
+    record (e.g. "22222625904") and "link" is a ready-made URL embedding it.
+    The GET read path stores the "link" URL as the token (GetNextRecordsURL).
+  - Search responses carry only "after", an integer position into the result
+    set capped at 10,000 (e.g. "200"). The search path stores it as the token
+    (GetNextRecordsAfter) and sends it back in the "after" request body field.
+
+A URL token resumed into the search path is rejected by the provider with a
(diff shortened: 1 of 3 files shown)


RajatPawar (changes requested):
I don't think this is the right place to solve it. This seems to be an edge-case with our workflow state management which ensures that reads without `Since`/`Until` do not interfere with reads that have `Since`/`Until`. So we should solve this in the server.

PR #3445, opened 2026-09-04
refactor(readhelper): return time.Time by value from timestamp helpers

Jkarage:
`extractTimestamp`, `convertIntToTime` and `convertStrToTime` returned `*time.Time`. Using pointer semantics for a small value type like `time.Time` is not idiomatic Go.

This PR changes them to return `time.Time` by value. That matches the standard library convention for timestamps and removes the dereferences at the two call sites in `FilterSortedRecords` and `MakeTimeFilterFuncWithZoom`.

### DIFF
--- a/common/readhelper/filter.go
+++ b/common/readhelper/filter.go
@@ -87,7 +87,7 @@ func FilterSortedRecords(data *ajson.Node, recordsKey string, since time.Time, /
 		}
 
 		// Check if this record is newer than our reference time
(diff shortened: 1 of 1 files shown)


Cobalt0s (approved):
Sorry for being late, I missed it.

The change is great! Approving.
You are right on using the time by value, I should make it a habit. Please rebase, in case any packages need to sync up.

PR #3463, opened 2026-09-09
chore: upgrade to Go 1.27 and golangci-lint v2.13.2

eberle1080:
Upgrades the toolchain to Go 1.27 and brings the linter forward to match. Companion to amp-labs/amp-common#263.

| | Before | After |
|---|---|---|
| `go.mod` | 1.26.5 | **1.27.1** |
| CI Go — `linter.yml` | 1.26.1 | **1.27.1** |
| CI Go — `build.yml` | 1.26.5 | **1.27.1** |
| golangci-lint (3 pins) | v2.7.1 | **v2.13.2** |

## The linter bump is mandatory, not cosmetic

golangci-lint v2.7.1 **cannot analyze Go 1.27 code at all**:

```
cannot decode "internal/goarch", export data version 4 is
greater than maximum supported version 2
```

Go 1.27 advanced the export-data format past what v2.7.1's bundled `x/tools` can read, so every package fails to typecheck. The version is pinned in **three** places that must agree — `Makefile`, `.custom-gcl.yml`, and `.github/workflows/linter.yml` — and all three are updated. The `nogoroutine` and `modulelinter` plugins recompile against v2.13.2 with no dependency changes.

## Verification

- `go build ./...`, `go vet ./...` — clean
- **`make lint` — exit 0, and idempotent** (run twice, both pass)
- `go test ./...` — **163 packages pass, 0 failures**

## 🐛 A real bug this surfaced

`TestSingularMarshal` in `common/naming` passed on Go 1.26 and failed on 1.27. This is a genuine defect, not a lint nit:

`SingularString` and `PluralString` are used as **map keys**, and `encoding/json` uses `TextUnmarshaler` for map keys — handing it the *unquoted* key. But `UnmarshalText` delegated to `UnmarshalJSON`, which calls `json.Unmarshal` and expects a *quoted* JSON string. So decoding `{"data":{"potatoes":...}}` failed with `invalid character 'p' looking for beginning of value`. It was also asymmetric with `MarshalText`, which correctly writes the raw text.

Both types now format the raw text directly. **Verified under Go 1.26 as well**, so this is not a 1.27-only patch.

## `make lint` could not converge

gofumpt and gofmt disagree about alignment groups in a map literal containing one outlier-length key: one reformats the file, the other reverts it, so the verification run always reported it as unformatted — an unwinnable loop.

I first assumed the fix was removing redundant formatters from `.golangci.yml`. **I tested that, it did not work, and I reverted that config churn** rather than leave an unjustified change. The real fix is at the source, and both files read better for it:

- `providers/acculynx/incremental.go` — a blank line separating the long `company-settings/...` keys into their own alignment group
- `providers/jobber/subscribe.go` — hoisting a multi-line composite literal out of a multi-value `return`

## Config changes

**Renames.** Version bumps rename linters, which silently breaks `disable:` entries — `exhaustruct` → `exhaustruct_v5` (2850 issues reappeared) and `gomodguard` → `gomodguard_v2`. Both disabled under their new names; `gomodguard_v2` remains enabled via `default: all`, so no coverage is lost.

**Two scoped exclusions**, each documented inline:

| Rule | Scope | Why |
|---|---|---|
| `goconst` | `providers/`, `scripts/` | Each file is an independent declarative description of one SaaS API. Two providers sharing a literal (a header name, a content type) is a coincidence, not a shared concept — hoisting constants would couple unrelated provider definitions. 828 of 891 findings. |
| `gosec` G101 | `providers/` | Declares the *names* of auth headers and fields, not secrets. All 134 findings. G101 stays active everywhere else, where a real hardcoded credential could appear. |

## Everything else is fixed, not suppressed

16 `prealloc`, 7 `goconst` in `common/` extracted as real constants, and a stale `//nolint:stylecheck` (stylecheck was merged into staticcheck in v2, so the directive had become a no-op that the runner warns about).

## Eight justified nolints

Each carries a written reason. The four gosec findings were investigated individually; none is a real vulnerability:

| Site | Rule | Why |
|---|---|---|
| `providers/utils.go`, `salesforce/jwt/jwt.go` | G704 SSRF | Replays a request this connector already sent, at the same URL; refreshing credentials does not change the destination |
| `snowflake/postAuth.go` | G202 | A query body cannot be bound as a parameter, so prefixing `EXPLAIN` is the only way to validate a user-supplied Dynamic Table definition. Same rationale as the already-excluded G201 |
| `memstore/hashing.go` | G115 | `int64`→`uint64` is a same-width reinterpretation for a deterministic encoding |
| netsuite ×2, quickbooks ×2 | unqueryvet | `SELECT *` during metadata discovery — the column set is precisely what is being discovered, so it cannot be named |

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ install/linters:
 	@echo "Installing gci..."
 	go install github.com/daixiang0/gci@latest
 	@echo "Installing golangci-lint $(GOLANGCI_LINT_VERSION)..."
-	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b $$(go env GOPATH)/bin $(GOLANGCI_LINT_VERSION)
+	go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@$(GOLANGCI_LINT_VERSION)
 	@echo "Installing typos..."
 	cargo install typos-cli
 	@echo "Linters installed successfully!"
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # ====================
 
 # Linter versions - keep in sync with .github/workflows/linter.yml
-GOLANGCI_LINT_VERSION=v2.7.1
+GOLANGCI_LINT_VERSION=v2.13.2
 
 # Install all linters required by make fix
 .PHONY: install/linters
--- a/.custom-gcl.yml
+++ b/.custom-gcl.yml
@@ -1,5 +1,5 @@
 # Ensure that the version matches linter version in CI pipeline step at ".github/workflows/linter.yml".
-version: v2.7.1
+version: v2.13.2
 plugins:
   - module: 'github.com/amp-labs/connectors/tools/linters/nogoroutine'
     path: './tools/linters/nogoroutine'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
         # git ls-remote https://github.com/actions/setup-go.git refs/tags/v7
         uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e
         with:
-          go-version: "1.26.5"
+          go-version: "1.27.1"
           cache: false
 
       - name: Build
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -9,7 +9,7 @@ jobs:
 
       - uses: actions/setup-go@v4
         with:
-          go-version: "1.26.1"
+          go-version: "1.27.1"
           cache: false
 
       - name: Generate go code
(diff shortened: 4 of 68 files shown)


Makefile
Cobalt0s: Change to:
```
go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@$(GOLANGCI_LINT_VERSION)
```

Cobalt0s (approved):
LGTM

I tried it locally, updated golang and everything works fine.
However, before running `make lint`, I had to run the following:
```
rustup update stable # required for custom-gcl
make install/linters
make linter-rebuild
```

These actions should taken by everybody, after this PR is merged.

eberle1080:
> LGTM
> 
> I tried it locally, updated golang and everything works fine. However, before running `make lint`, I had to run the following:
> 
> ```
> rustup update stable # required for custom-gcl
> make install/linters
> make linter-rebuild
> ```
> 
> These actions should taken by everybody, after this PR is merged.

@Cobalt0s thanks for the tip! I'll post about it once this merges.

PR #3465, opened 2026-09-09
[CON-3479] feat(reply): add reply-read

ejoo:
# Description

- [x] Read supports pagination and incremental sync (incremental: `sequences` only — see above)
- [x] Provider errors are mapped if non-standard
- [x] Unit tests cover read logic (placed in `/providers/reply`)
- [x] live tests (placed in `/test/reply`)

# Tests

## Unit Test

image

## Live Test

image

image

image

image

image

### DIFF
--- a/providers/reply/read.go
+++ b/providers/reply/read.go
@@ -1,0 +1,203 @@
+package reply
+
+import (
+	"context"
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/amp-labs/connectors/common"
+	"github.com/amp-labs/connectors/common/readhelper"
+	"github.com/amp-labs/connectors/common/urlbuilder"
+	"github.com/amp-labs/connectors/internal/datautils"
+	"github.com/amp-labs/connectors/internal/jsonquery"
+	"github.com/amp-labs/connectors/providers/reply/metadata"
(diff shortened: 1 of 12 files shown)


providers/reply/read.go
```diff
@@ -0,0 +1,203 @@
...
+
+	// Sequences is the only object with a provider-side time filter, and it
+	// scopes by creation time; Until has no provider-side parameter, so the
+	// full window is enforced connector-side on the created field during
+	// parsing. Contacts declare createdAt/lastModifiedAt in their schema, but
+	// the live API always returns null for both and the filter endpoint
+	// silently ignores time rules, so no other object can be read
+	// incrementally (verified against the live API).
+	// https://docs.reply.io/api-reference/sequences/list-all-sequences
+	if params.ObjectName == objectSequences && !params.Since.IsZero() {
```
jlimatampersand, line 70: I think we should just do a full read because we want to capture updates. 

So, we need to do a full read, and then do client-side filtering by `lastModifiedAt` or equivalent for all objects

If no such field is available, then it should be a full read and mention it in our provider guide documentation that it is full read.

jlimatampersand (changes requested):

PR #3478, opened 2026-09-11
[CON-3512] fix(acculynx): tolerate capitalised event wrapper casing in webhooks

ejoo:
`job.contacts.primary_changed` webhooks cannot be resolved to a record: AccuLynx documents the event wrapper lowercase (`event.job`) and sends it that way for every topic except this one, which sends `event.Job`. The exact-match lookup dropped the wrapper, so `RecordId()` failed.

Fix: `lookupObjectWrapper` tries the exact key first and falls back to a case-insensitive match. All working topics are unchanged. Unit test uses the real payload captured from repeated live deliveries.

# Tests

## Unit Test

image
image
image

### DIFF
--- a/providers/acculynx/subscriptionEvent.go
+++ b/providers/acculynx/subscriptionEvent.go
@@ -314,12 +314,43 @@ func (e SubscriptionEvent) objectWrapper() (map[string]any, error) {
 		key = objectWrapperJob
 	}
 
-	wrapper, ok := inner[key].(map[string]any)
+	wrapper, ok := lookupObjectWrapper(inner, key)
 	if !ok {
 		return nil, fmt.Errorf("%w: expected event.%s", errMissingObjectWrapper, key)
 	}
 
 	return wrapper, nil
+}
+
+// lookupObjectWrapper reads the object wrapper out of the inner event,
+// tolerating the key casing AccuLynx actually sends.
+//
+// Their schema and examples document lowercase ("event.job", "event.contact")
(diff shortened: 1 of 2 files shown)


providers/acculynx/subscriptionEvent.go
```diff
@@ -314,14 +314,45 @@ func (e SubscriptionEvent) objectWrapper() (map[string]any, error) {
...
+// and every topic serialises it that way except job.contacts.primary_changed,
+// which sends "event.Job" — confirmed across repeated live deliveries against
+// a payload otherwise identical to their documented example. An exact-match
+// lookup drops the wrapper for that topic, so RecordId fails and the event
+// cannot be resolved to a record.
+//
+// Falling back to a case-insensitive match keeps that topic working and covers
+// the topics that can only be triggered from the AccuLynx UI, where the same
+// defect would otherwise surface as an unexplained webhook failure.
+func lookupObjectWrapper(inner map[string]any, key string) (map[string]any, bool) {
```
RajatPawar, line 338: I believe we have normalized maps as utils for this, and we should reuse those cc @Cobalt0s may have more context here

RajatPawar (approved):
LGTM - please address the feedback before merging this in!

providers/acculynx/subscriptionEvent.go
ejoo: Done! replaced the hand-rolled lookup with `common.ToStringMap(inner).GetCaseInsensitive(key)`

PR #3488, opened 2026-09-15
[CON-3536] feat(acculynx): read all jobs and job financials

ejoo:
# Description

`GET /jobs` without an `assignment` parameter returns only the jobs assigned to users; unassigned jobs are a disjoint population reachable only via `assignment=unassigned`, and no "all" value exists (probed live). The jobs read now sweeps both listings: when the default (assigned) sweep exhausts, the read continues into the unassigned listing, with the phase carried in the next-page URL — no new pagination state. Incremental (`dateFilterType=ModifiedDate`) rides both sweeps. Nested `jobs/*` objects inherit the fix through the parent fan-out. (`readhelper.AggregateNextPage` was considered and doesn't fit: it composes parallel multi-context tokens as JSON, while this is a sequential sweep over URL tokens.)

Adds `jobs/financials` (`GET /jobs/{jobId}/financials`): a one-to-one child of jobs read through the standard fan-out. The endpoint answers with a single object, so the response root is the record (one row per job); fields come from the spec's `financials` component (the endpoint has no list schema for the explorer). Leaf requests send `includes=worksheet,amendments` — an expanded worksheet arrives with its sections and items, and an empty worksheet omits the key entirely (verified live).

Unassigned jobs answer 404 on the representatives, estimates and milestone-history leaves (assignment-driven — an assigned Lead answers 200 on all of them; the API reference documents only a generic 404). The fan-out now treats a not-found leaf as "no child records", the same way non-user calendars are skipped in GetRecordsByIds.

Live verification: jobs and jobs/financials both sweep 286/286 unique — the assigned listing alone carries 170, so 116 previously-invisible unassigned jobs are now read; jobs/representatives completes across both populations.

# Tests

## Unit Test

image
image
image
image
image

## Live Test

image
image
image
image


providers/acculynx/read.go
```diff
@@ -477,17 +507,43 @@ func (c *Connector) fetchChildPages(
...
 		resp, err := c.JSONHTTPClient().Get(ctx, reqURL.String())
 		if err != nil {
+			// Some leaves do not exist for every job: unassigned jobs answer
+			// 404 on representatives, estimates and milestone-history. The
+			// trigger is the assignment itself, not the milestone — an
+			// assigned job in the Lead milestone answers 200 on all leaves
+			// (verified live; the API reference documents only a generic
+			// 404). A missing leaf means "no child records", not a failed
+			// read — the same way non-user calendars 404 in GetRecordsByIds.
+			if isNotFound(err) {
```
laurenzlong, line 549: doesn't this short circuit the `for range` loop and cause us to not read the rest of the child pages when we encounter one that is missing? I think we would want to skip this child and then move onto the next child.

laurenzlong (changes requested):

providers/acculynx/read.go
ejoo: Good catch to check but this loop iterates the pages of a single parent's leaf, the per-child iteration is one level up in fetchChildrenForParents, which runs an independent fetchChildPages per parent. So a 404 ends only that one parent's walk and the rest continue exactly as you describe. Covered by the "Fan-out skips a parent whose leaf answers 404" unit test (job-002 404s, job-001's rows still come back) and the live sweep (286 jobs, ~116 unassigned 404s, completes). Added a sentence to the comment making the scope explicit.

laurenzlong (approved):

PR #3492, opened 2026-09-15
[ENG-4265] feat(slack): support message events in subscribe (thread replies, edits, deletes)

jlimatampersand:
## Why

Slack has no API for finding threads with new replies. Replies arrive as `message` events with a `thread_ts`, making event-driven ingestion the only efficient approach—especially under Slack’s 2025 rate limits.

The Slack webhook package currently drops message events because they are unmapped.

## What changed

- Added message event mappings:
  - `message` → create
  - `message_changed` → update
  - `message_deleted` → delete
  - `thread_broadcast` → create
- Dispatch message events by `subtype`; unknown subtypes remain ignored.
- Identify messages by `<channel>:<timestamp>`, using the subtype-specific timestamp field.
- Added `MessageEvent`, which provides the record directly from the webhook payload because Slack cannot fetch a single message by ID.
- Added `common.ErrRecordsOnlyInline`, returned per object when records exist solely in the webhook payload, and return it for `messages`. It says more than "fetch unavailable": it tells the caller where the record does live, so the caller proceeds with no fetched rows and reads each event's inline record instead of dropping the batch or retrying a fetch that can never succeed.
- Kept `common.ErrGetRecordNotSupportedForObject` for objects that are merely unfetchable—readable but with no `.info` endpoint, or absent from the mappings—matching the other mapping-driven connectors.
- Kept non-message events on the existing fetch-by-ID path.

## Notes

- Apps must subscribe to the relevant `message.*` events, have matching `*:history` scopes, and be members of the conversations they monitor.
- A `messages` read object is still needed to satisfy server configuration validation.
- Companion PR `amp-labs/server#7385` handles `ErrRecordsOnlyInline` in the messenger. It must merge with this bump, or message events are dropped.
- The capability is derived from the same code path that performs the fetch, so it cannot drift from what `GetRecordsByIds` actually does. It asks a different question from `SubscriptionEventWithRecord`: ConnectWise events carry inline records *and* are fetchable, so neither implies the other.
- `message_changed` does not include changed fields, so it requires `watchFieldsAuto: all`.

## Testing

- Added golden and inline-record tests covering reply, changed, and deleted message events.
- Added coverage confirming non-message events still use the existing path.


Cobalt0s (approved):
LGTM 🚀

Cobalt0s:
@jlimatampersand, you need to rebase and push again. The PR you did for CI was merged. This is why the "ci expected" step is stuck in waiting.

providers/acculynx/connector.go
Cobalt0s, line 92: The inhereted methods are here:
https://github.com/amp-labs/connectors/blob/main/internal/components/provider.go#L50
https://github.com/amp-labs/connectors/blob/main/internal/components/provider.go#L54

They "display methods" use the provider name and **module name**. Both of which are dependent on loading and resolving the `ProviderInfo`.

Maybe we should rethink the design of components. cc @RajatPawar.
I would prefer the connector to be initialized via some constructors to avoid the boilerplate and any panics due to tight coupling. Connectors already don't have much state besides the loading of ProviderInfo.

It looks like, with the way it is used -- the goal is to treat connector methods as a "namespace" of static functions kind of like utils. In that case we should not use `type Connector struct` in any package.

Also the changes in here are not unique to this Slack PR per say and touche all connectors with subscribe feature.

PR #3502, opened 2026-09-17
[ENG-4236] feat(salesforce): carry the outbound message record inline

AmandaYao00:
Implements `Record()`, which is called during subscription event enrichment. In contrast to most providers, for which we do a `GetRecordsById()` call to fetch the record from the provider, the OM already has all the fields we need. So we bypass the `GetRecordsById()` call and call `Record()` instead, which converts the SOAP payload into the standard `ReadResultRow` format used downstream. 

**Testing**
Did E2E testing to verify delivered payloads:
Screenshot 2026-09-17 at 5 02 05 PM

PR #3499, opened 2026-09-16
[ENG-4236] feat(subscribe): resolve Salesforce flow subscribe through the deps seam

AmandaYao00:
Connectors will build subscription requests for both Salesforce Flow and CDC modes. This PR adds a `SalesforceFlowResolver` dependency that the server implements and injects. The resolver provides the connectors request builder with the `SelectedFields` needed to build the flow request.

PR #3506, opened 2026-09-18
feat(salesforce): support updating flow subscriptions

AmandaYao00:
UpdateSubscription only ever ran the CDC path. A flow subscription reaching it was rejected before
doing anything — flow installations have no registration, and the first thing the function did was
require one — and had it got past that, it would have run quota-field upserts and channel-member
reconciliation against a subscription that has neither.

The flow branch now diverts before the registration checks, mirroring Subscribe, which already
does this for the same reason. It covers both cases that involve flow:

- flow to flow, where a field change means redeploying the outbound message. The flow and its
  outbound message deploy as one atomic package keyed by object, so there is no in-place edit.
- switching between modes, where every artifact is replaced regardless.

Both reduce to the same two steps, and DeleteSubscription and Subscribe already branch on the mode
they are handed, so replaceSubscription reuses them rather than reimplementing either.

Teardown runs first, deliberately. Provisioning both stacks at once would deliver duplicate events
for every record change in the window, and the existing dedupe would not collapse them: CDC and
outbound messages carry different event shapes and identifiers.

There is no cross-mode transaction. A failure after teardown leaves the subscription with no
artifacts, recoverable by retrying the update — the teardown half then finds nothing to remove and
the create half provisions from scratch.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


AmandaYao00:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3506?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3506** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3506?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3506?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3502** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3502?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3499** <a href="https://app.graphite.com/github/pr/amp-labs/connectors/3499?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/amp-labs?ref=merge-queue-instructions-in-stack-comment&prId=3506).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

AmandaYao00:
Closing — superseded by `amanda/sf-flow-update` (dd955868), which already does this and does it better.

I built delete-then-recreate for flow-to-flow. That branch reconciles in place, which is the correct behavior: Salesforce replaces the outbound message in place and activates a new flow version within the same deploy, so tearing down first causes an unnecessary event gap. It also handles two things mine did not — deleting deactivated flow versions to stay under the 50-versions-per-flow cap, and removing artifacts for objects that dropped out of the subscription.

The cross-mode half is equivalent, except that branch returns a Failed-status result carrying prevState when teardown fails, where mine returned a bare error.

I missed it because I only looked at connectors main, where the series stops at (5/8). The branch name collision when I tried to create `amanda/sf-flow-update` was the signal I should have followed.

PR #3507, opened 2026-09-18
fix(subscribe): nil-safe Provider/String on webhook verifier connectors

jlimatampersand:
## Problem

Slack message webhooks were panicking in the messenger before any verification ran:

```
panic: runtime error: invalid memory address or nil pointer dereference
slack.(*Connector).Provider(...)
	<autogenerated>:1
metrics/generated.WrapConnector(...)
	shared/metrics/generated/utils.go:100
main.verifyWebhook(...)
	messenger/subscribeEventProcessor_verify.go:216
```

The server wraps a provider's verifier connector for metrics, and the wrapper's first act is to call `Provider()` on it. The subscribe registry registers **bare** connectors as verifiers: Slack's is built with only a webhook `Verifier` (`NewWebhookVerifierConnector`), and most others are zero-value (`&jobber.Connector{}`). In both cases the embedded `*components.Connector` is nil. `Provider` and `String` are declared on `components.Connector` with **value receivers**, so the promoted call dereferences that nil pointer and panics.

The blast radius is worse than a single dropped event. The pubsub receiver recovers the panic and then nacks, so the message is redelivered and panics again, forever.

## Fix

Each affected connector declares its own `Provider` and `String` that fall back to a static identity when the base is absent, and delegate to the base otherwise. This matches what `attio`, `salesforce`, `outreach`, `salesloft` and `zoho` already do, which is why they were never affected. Delegation means a real connector is unchanged, so the Slack user-scope connector still reports `slackUserScope` rather than a hardcoded `slack`.

Connectors fixed: `slack`, `jobber`, `hubspot`, `stripe`, `microsoft`, `acculynx`, `connectwise`, `housecallPro`.

Only `slack` and `jobber` are reachable today, since the server bypasses verification for the rest. That bypass is a server-side policy with open TODOs to remove it, so the others were fixed too rather than left as landmines.

`stripe` reaches the same defect through a different path: it embeds `*core.Base`, which embeds the base connector.

## Test

`TestVerifierConnectorsAnswerProviderAndString` walks every entry in `providerConfigs` and calls both methods on each registered verifier connector, rather than naming providers. A new provider that registers a bare connector fails here instead of in production.

It earns its keep: I wrote it after fixing `slack` and `jobber`, and it immediately found the other six.

## Verification

- `go test ./subscribe/...` passes, 25 provider/module subtests.
- Tests pass for all eight touched provider packages.
- `make fix` reports 0 issues.
- Confirmed the test fails without the fix, panicking in the same promoted-method wrapper as production.


eberle1080 (approved):

providers/acculynx/connector.go
```diff
@@ -83,3 +83,27 @@ func constructor(base *components.Connector) (*Connector, error) {
...
+
+// Provider returns the provider this connector talks to.
+//
+// It is declared here rather than inherited from the embedded *components.Connector because the
+// subscribe registry uses a zero-value Connector as its webhook verifier, which has no base. The
+// base's Provider has a value receiver, so the promoted call would dereference that nil pointer
+// and panic before any verification ran.
+func (c *Connector) Provider() providers.Provider {
+	if c == nil || c.Connector == nil {
+		return providers.AccuLynx
```
laurenzlong: why wouldn't this always return `providers.AccuLynx` (one line function)? same for all the other files here. When would c.Connector.Provider() ever be something different?

laurenzlong (changes requested):

providers/acculynx/connector.go
jlimatampersand: You're right for six of the eight, and I've simplified those to a one-line constant (acculynx, connectwise, housecallPro, hubspot, jobber, stripe). Each is only ever built under one provider, so delegating bought nothing.

Two have to keep delegating, because they're built under twin providers and a constant would mislabel them:

- **slack** — `NewBotConnector` initializes under `providers.Slack`, `NewUserConnector` under `providers.SlackUserScope`. Only the base knows which one a given instance is. `Slack` stays as the fallback for the baseless verifier connector, which the registry shares between both entries.
- **microsoft** — `NewConnectorForProvider` exists for exactly this, and `connector/new.go:589` calls it with `providers.MicrosoftAdminConsent`.

I kept `String` delegating everywhere. The base returns `<provider>.Connector[<module>]`, so collapsing it to a constant would drop the module from existing log output; the fallback only kicks in when there's no base to ask.

laurenzlong (approved):

providers/acculynx/connector.go
laurenzlong: ah ok makes sense! can you add a comment to those 2 providers, so no one makes the assumption I made?

PR #3521, opened 2026-09-21
migrate simultaneously and future to amp-common

eberle1080:
The motivation for doing this is 2-fold:
* Move re-usable stuff to amp-common (this qualifies)
* The one in amp-common lets us have better control over total system-level parallelism (avoid goroutine spikes by setting a global ceiling via an injectable Executor)

Both packages now live in github.com/amp-labs/amp-common and are shared across Ampersand services, so drop the copies here and point every caller at the upstream ones.

Deleting internal/simultaneously and internal/future leaves the callers as pure import swaps: amp-common's simultaneously.Job is a type alias for func(ctx context.Context) error, so the 34 []Job slices compile unchanged. No call site needed a Do -> DoCtx conversion; every one already used the context-aware form.

The nogoroutine linter excluded internal/future and internal/simultaneously from its bare-'go' check, since those two packages were the only ones that legitimately spawned goroutines. Both are gone, so the exclusions are dead config and settings drops to {}, with a comment on how to re-add a path. Its diagnostic and the README now recommend future.GoContext and simultaneously.DoCtx over the context-free variants, since those let the caller cancel the work.


Cobalt0s (approved):
LGTM 👍

